PR #129, opened 2013-07-27
support balance resetting and make assertion semantics same with ledger

xiaoruoruo:
To support balance resetting, we need to keep track of "current" account balances. The algorithm for balance assertion/assignment/resetting is basically the same: derive the amount from the assertion, and compare it to the actual amount, if any.

If there are multiple assertion postings for the same account in one transaction, it's not always possible to derive&update posting by posting. So when this happens, I don't update the balances mapping until all postings are checked for a transaction. This seems to be the semantics of ledger 3.0 also. So the tests are also updated/modified to reflect these situation.


johannesgerer:
# Update

`ledger` seems to have updates its semantics to be more consistent:

@simonmichael your example from above now is not considered balanced. The transaction accepted by both `hledger` and `ledger` now looks like this:

```
2013/1/1 after
      a      $2 = $2
      b      $-1 = $-1
      b      $-1 = $-2
```
This makes more sense.

# The semantics

For a given transaction:

1. First handle all postings with assertions and/or amounts:
    1.1. Posting with amount: Add the amount to the account balance.
        1.1.1       ... and assertion: Check Assertion == Account Balance
    1.2. Posting without amount: infer Amount = Assertion - Account Balance
2. Then handle a posting without amount or assertion: infer Amount = - Transaction Balance

# Why this feature is SO IMPORTANT:

I use hledger to track my personal finances including my wallet and other cash accounts that are frequently changing. Tracking all cash expenses up to a certain point of time can be hard due to forgetfulness etc, and because things like cash withdrawals have some delay until they find their way into my ledger through my bank statements. This is why at any given point I can easliy count my wallet at hand, but the ledger balance might not yet be up to date. With an "assigment" posting this situation can be handled extremely easy and fail safe. Not having this feature means constant mundane and error prone manual work. 
Does this make any sense?

# The pull request

I will send an updated pull request. As the parsing and data types already support assertions with missing amounts, there should be only little change needed.

What do I need to provide to help/guarantee a fast merge on your end?

ony:
@johannesgerer I'm happy to hear that this functionality may be merged into `hledger`.
But please consider next example

```
2016/12/05 open
    a  = $10
    b

2016/12/06 multipost
    a  = $2
    a  -$2  ; May add assertion like: = $0
    c  ; <-- $8 or $6?
```

I'm not sure that you can re-order postings. Either you should reject such kind of transactions or you should process in order of writing, I believe.

simonmichael:
@johannesgerer thanks for the update. I'm glad to hear about the fix in Ledger. I'll very happy to review and test your patch.

simonmichael:
(Or your update of @xiaoruoruo's patch, if that's what it is).

FYI #157 is a semi-duplicate issue with some more recent design discussion.

simonmichael:
Also related: #288. This feature seems to be quite popular.

PR #246, opened 2015-04-23
hledger-lib: Update for base-compat-0.8.0

sol:


sol:
@simonmichael This is shows how `base-compat` is meant to be used.  I only did it for `hledger-lib`, the other components would need similar changes.

simonmichael:
Thanks! More on the way.

PR #251, opened 2015-05-02
hledger-web: Set default javascript parameter values in the functions addformReset, addFormShow

kmels:
http://demo.hledger.org/journal gives some javascript errors, these prevent showing the nice chart and kb shortcuts from working

### DIFF
--- a/hledger-web/static/hledger.js
+++ b/hledger-web/static/hledger.js
@@ -126,7 +126,8 @@ function registerChartClick(ev, pos, item) {
 //----------------------------------------------------------------------
 // ADD FORM
 
-function addformShow(showmsg=false) {
+function addformShow(showmsg) {
+  showmsg = typeof showmsg !== 'undefined' ? a : false;
   addformReset(showmsg);
   $('#addmodal')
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks!

I don't see the error in firefox. What's your browser version ?

kmels:
Hey, my version is chrome 41.0.2272.118
I can see the chart fine in Firefox too..  [it seems like this feature is only supported by Firefox at the moment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Browser_compatibility)

simonmichael:
Ack - thanks!

simonmichael:
Hi Carlos, this breaks the add form in my firefox: ReferenceError: a is not defined 141:12. Can you fix ?

simonmichael:
PS only when you press the `a` key. It works when you click add a transaction. (Which is odd, why doesn't the latter go through the same code path).

kmels:
Hi Simon! Omg, I don't know why I put there this letter. I am fixing it.

PR #254, opened 2015-05-17
Changes to make `make test` run

ricklupton:
I had some trouble getting `make test` to run the tests, but got it working with some changes to the makefile:
 - test/ADDONS -> test/addons
 - add bin/ to path when running shelltest so hledgerdev is found
 - use `cabal exec` to find shelltest in cabal sandbox

I think it makes sense why this is needed, if 
 - you normally use a case-insensitive filesystem,
 - you have hledgerdev in your path, and
 - you don't install shalltest in a sandbox?

But I'm new to using Haskell so feel free to suggest better ways of fixing it!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,8 @@ GHCPKG=ghc-pkg
 HADDOCK=haddock
 CABAL=cabal
 CABALINSTALL=cabal install -w $(GHC)
-SHELLTEST=shelltest -j16 --hide-successes
+TESTFRAMEWORKOPTS=-j16 --hide-successes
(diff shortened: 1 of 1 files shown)


simonmichael:
Let's not require that tools like shelltest be installed in a local sandbox - I'd normally install them once in ~/.cabal/bin and use them across multiple projects. Your use of cabal exec is interesting, hopefully it works whether the tool installed in a sandbox or elsewhere.

Thanks Rick!

ricklupton:
cabal exec seems to work whether it's installed in a sandbox or not. Thanks!

PR #263, opened 2015-05-26
hledger-web: Fixes keyboard shortcut for adding a transaction. 

kmels:
This error happens in Chrome and in Firefox too.

See pull request #251. 

The link 'Add a transaction' is working because it is triggered by bootstrap `data-toggle=modal data-target="#addmodal`. The keyboard follows another path because it resets the form before calling `modal` on `$("#addmodal")`.. that helped discover that the link 'Add a transaction' does not clean the form in fact. :)

Should I leave it like this? Should I clean the form after clicking 'Add transaction' ? I think the best would be:
 - Put a link on the form (below the 'X') to clean the fields.
 - (Possibly) do not clean the form when pressing 'a'.

### DIFF
--- a/hledger-web/Handler/JournalR.hs
+++ b/hledger-web/Handler/JournalR.hs
@@ -34,7 +34,7 @@ getJournalR = do
   hledgerLayout vd "journal" [hamlet|
        <h2#contenttitle>#{title}
        <!-- p>Journal entries record movements of commodities between accounts. -->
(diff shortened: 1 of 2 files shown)


simonmichael:
I think we want to always reset the form before opening it, and there's no need for UI for this. Right now if you open the form, enter some data, cancel it, and click add transaction again you'll see the old data is still there, which we don't want.

kmels:
Ok I will commit again to clear it just as the shortcut does.

kmels:
Updated to clear the form before.

simonmichael:
Just noticed this, thanks!

simonmichael:
In fact, it seems not just the last commit but this whole pull request did not make it into hledger 0.26. Sorry about that.

Issue #165, opened 2014-03-06
Allow multiple files in -f parameter


markokocic:
It would be nice to be able to combine multiple files using -f parameters. Hledger could just processed them continually in order of specification.

I know it is currently possible to do something like `cat file1 file2 | hledger inc -f -` but I would still find it more convenient to be able to just use `hledger inc -f file1 file2`

simonmichael:
Thank you imuli! Merged.

PR #264, opened 2015-05-27
handle multiple -f parameters sensibly

imuli:
Fixes bug #165, though instead of `hledger inc -f file1 file2` it is `hledger inc -f file1 -f file2`: the former would interfere with specifying patterns on the command line.

Notes:

This did make the API slightly redundant, exporting both readJournalFiles and readJournalFile. (I have a version without readJournalFile, but the web frontend isn't tested.)

The first file is added to journalFilePaths with the contents of all the files and I haven't looked exhaustively for bugs that may cause, but I suspect it might only be parser error messages with the wrong file name. That looks like a problem to be solved along side the include file parsing one (hledger-lib/Hledger/Read/CsvReader.hs:812).

The important possible bug that I came up with was fine though: `hledger add -f file1 -f file2` behaves sensibly, matching on all input files and appending to the first.

### DIFF
--- a/data/alias.journal
+++ b/data/alias.journal
@@ -1,0 +1,2 @@
+alias expenses	= equity:draw:personal
+alias assets = assets:personal
--- a/data/business.journal
+++ b/data/business.journal
@@ -1,0 +1,4 @@
+2014/1/1
+    expenses:office supplies  $1
+    assets:business checking
+
--- a/data/personal.journal
+++ b/data/personal.journal
@@ -1,0 +1,4 @@
+2014/1/2
+    expenses:food  $1
+    assets:cash
+
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -1060,6 +1060,12 @@ With the `--depth N` option, commands like [account](#account), [balance](#balan
 and [register](#register) will show only the uppermost accounts in the account
 tree, down to level N. Use this when you want a summary with less detail.
 
+### Multiple files
+
+One may specify the `--file FILE` option multiple times. This is equivalent to
+concatenating the files to standard input and passing `--file -`, except that
+the add command functions normally and adds entries to the first specified file.
+
 ## Queries
 
 One of hledger's strengths is being able to quickly report on precise subsets of your data.\
--- a/hledger-lib/Hledger/Read.hs
+++ b/hledger-lib/Hledger/Read.hs
@@ -15,6 +15,7 @@ module Hledger.Read (
        readJournal,
        readJournal',
        readJournalFile,
+       readJournalFiles,
        requireJournalFileExists,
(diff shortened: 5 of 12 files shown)


simonmichael:
That's great! Thank you!

One of the things I imagine this being used for is to pull in a bunch of alias directives, without having to write them all as `--alias` options, eg `hledger -f aliases.journal -f a.journal`.

Could you clarify the semantics of aliases now ? Eg say I have a file defining some aliases for accounts in journals a and b. If I do `hledger print -f a.journal -f b.journal -f aliases.journal -f a.journal`, I'm slightly confused about what's going on. I might expect the aliases to affect only the last -f a.journal, but they seem to be affecting both instances of a.journal (though not b.journal). If we can explain this in a reasonable way, it might be ok.

Also, could you add a simple shell test (eg in `tests/cli/multiple-files.test`).

And where/how should we update the manual  (
[Usage](http://hledger.org/manual.html#usage) ? 
[Account aliases](http://hledger.org/manual.html#account-aliases) ?
[Options](http://hledger.org/manual.html#options) ?
)
and command line help ?

imuli:
Regarding aliases, I can't get the [example](http://hledger.org/how-to-use-account-aliases.html) to work with regexps, only fixed patterns (but that looks like a separate bug).

The semantics of `hledger (command) -f a -f b -f c -f a` should be identical to `cat a b c a | hledger (command) -f -`  except that the anything that modifies the journal will modify the first file instead of creating '-' (is that a bug?). With the slightly modified alias example files:

`personal.journal`:

	2014/1/2
	    expenses:food  $1
	    assets:cash

`business.journal`:

	2014/1/1
	    expenses:office supplies  $1
	    assets:business checking
	
`alias.journal`:

	alias expenses	= equity:draw:personal
	alias assets = assets:personal

Everything appears to work out as one would expect - could you provide sample files that don't?

	$ hledger print -f personal.journal -f business.journal -f alias.journal -f personal.journal
	2014/01/01
	    expenses:office supplies            $1
	    assets:business checking           $-1
	
	2014/01/02
	    expenses:food            $1
	    assets:cash             $-1
	
	2014/01/02
	    equity:draw:personal:food            $1
	    assets:personal:cash                $-1

simonmichael:
Note that hledger.org docs correspond to the latest release, the docs in master describe the latest aliases behaviour. And, I just fixed an aliases bug (aliases matching the full account name weren't being applied). Now everything is working just as you say - excellent. I'm ready to merge your commits (rebased) if you are.

imuli:
Oh! That makes sense. I'm happy with things as they are. The only other change on my mind that I might have made is renaming journalFileFromOpts to journalFile*s*FromOpts, but that's pretty minor.

PR #309, opened 2015-12-16
Small improvements to make the web output valid HTML

TomK32:
I'm sending the HTML output through the validator and see what I can fix. That's the first few things with plently more to come. My register page went from 960 to 280 errors :)


simonmichael:
I just noticed this (seems I need better notifications settings). Thank you very much!

PR #347, opened 2016-05-22
lib: use all 76 doctests

ony:
Add some work-around for preprocessor into `doctest.hs` and `.ghci`.


ony:
Hmm. I though that `-hida-package` tolerant to absence of package.

PR #357, opened 2016-06-13
Add h/j/k/l movement to hledger-ui

mitchellwrosen:
Is this wanted? I had to remap help's `h` to just `?`.

### DIFF
--- a/hledger-ui/Hledger/UI/AccountsScreen.hs
+++ b/hledger-ui/Hledger/UI/AccountsScreen.hs
@@ -5,7 +5,7 @@
 
 module Hledger.UI.AccountsScreen
  (accountsScreen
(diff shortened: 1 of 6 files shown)


simonmichael:
I'd like to keep a simple approachable feel for non-techies, so I'm not super keen on this, though I can see how it would make good sense for vi users. If this catches on we need rebindable keys.

simonmichael:
Might, *might* need. 

Options and configuration are more headache, but we'll probably go there in the end.

simonmichael:
My fingers went looking for this today, I figured I'd try it for a while.. do you still have this commit ?

simonmichael:
Not to worry, got it working.

Issue #358, opened 2016-06-18
Partial record accessors in hledger-ui


mitchellwrosen:
Screen is defined as a sum type with record accessors that differ in each data constructor. So, e.g.

```haskell
tsTransaction :: Screen -> NumberedTransaction
```

is a partial function that only works for `TransactionScreen`.

It would probably be smart to refactor this data type.

simonmichael:
I'm not happy with it either, but I haven't found a better way to structure the types yet.

mitchellwrosen:
How about this?

```haskell
class Screen s where
  ... some stuff, probably sInit, sDraw, sHandle ...

data AccountsScreen = ...
instance Screen AccountsScreen

data TransactionsScreen = ...
instance Screen TransactionsScreen

data SomeScreen = forall s. Screen s => SomeScreen s

data UIState = ...
  { screens = [SomeScreen], ... }
```

simonmichael:
I like the sound of it! If you get some time, please give it a try and see how it works out.

mitchellwrosen:
Hm, I looked into this a bit. I think because of the tangling between `Screen` and `UIState` it won't be possible to do what I outlined above - specifically, you need to be able to peek inside a `UIState` and know what `Screen` is active, because `sInit`, `sDraw`, and `sHandle` all take a `UIState` as input, where the active screen is assumed to be the type that `sInit` is called on (errors otherwise).

So, I first tried adding a type parameter to `UIState` to track the current screen, then ended up tracking the entire stack of previous screens, and ended up bringing in way too much type-level programming, hitting weird hard-to-type functions like `regenerateScreens`.

Anyways, back to the original problem of partial record selectors, I think it could be solved rather simply. Instead of

```
data Screen
  = AccountScreen { sInit :: ... }
  | TransactionScreen { sInit :: ... }
```

have:

```
data Screen 
  = AScreen AccountScreen
  | TScreen TransactionScreen

data AccountScreen = AccountScreen { asInit :: ... }

data TransactionScreen = TransactionScreen { tsInit :: ... }
```

Then, if you want one vocab for init, handle, and draw, just use

```
class ScreenClass s where
  sInit :: s -> ...

instance ScreenClass AccountScreen
instance ScreenClass TransactionScreen
```

Basically, it's the same solution as above, but without the existential `SomeScreen` that `UIState` carries around, because it needs to know what screen it actually is.

simonmichael:
[...]

PR #359, opened 2016-06-18
Remove partial record accessors in hledger-ui

mitchellwrosen:
#358

### DIFF
--- a/hledger-ui/Hledger/UI/AccountsScreen.hs
+++ b/hledger-ui/Hledger/UI/AccountsScreen.hs
@@ -5,7 +5,7 @@
 
 module Hledger.UI.AccountsScreen
  (accountsScreen
- ,asInit
+ ,asInit_
  ,asSetSelectedAccount
  )
 where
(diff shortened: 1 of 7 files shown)


simonmichael:
Interesting! It feels like we are now handling screens with tongs. And no more partial functions. Does it prevent construction of nonsensical values ?

I imagine we can find ways to clean it up further, eg sInit/sDraw/sHandle don't use their first argument.

And Screen is still a closed/monolithic type of course. I wonder how we could reduce the amount of screen-specific knowledge there, confining that more to the screen modules. Do you think there's some other sane solution based on typeclasses that we haven't found yet ?

mitchellwrosen:
I do think there are other ways of cleaning this interface up, yeah, but I think the biggest wins would involve untangling `Screen` from `UIState`. So instead of

```haskell
sInit :: Screen -> Day -> Bool -> UIState -> UIState
```
it might be
```haskell
sInit :: Day -> Bool -> Screen
```
(passing in whatever pieces of `UIState` are necessary, like maybe `UIOpts`).

I'm a bit confused about why it's called `init` if there is already an existing screen inside the `UIState` passed into `sInit`, though.

Then, it may make sense to parameterize `UIState` by the current screen, so as to avoid the partial pattern matches in `sInit`, `sDraw`, `sHandle`.

But, I'm not totally familiar with the code, and am just taking baby steps at the moment :)

This particular patch just gets rid of the partial functions at the cost of some wrapping/unwrapping.

mitchellwrosen:
Ah I see what you mean about `sInit`/`sDraw`/`sHandle` not using their first argument. I didn't even notice. It's because before they were record accessors, but they always operated on the screen in the given UIState.

simonmichael:
"init" is not a great name. It's called to update the screen's data model state (eg the list of items to show) on entering the screen but also whenever needed, eg when the filter is changed or the file is reloaded. (It is not called for the smallest UI state changes, like moving the selection.) I didn't call it "update" because in game loops that is called on every tick, but perhaps update or [re]generate or something else would be better than init.

simonmichael:
@mitchellwrosen: looking at this PR today, it doesn't seem to remove the "init function called with wrong screen type, should not happen" error calls. I'm confused, wasn't that the purpose of removing partial record accessors ?

mitchellwrosen:
@simonmichael That would require a more significant refactoring - I believe you would need to parameterize `UIState` by the screen it's showing. This is just restructuring the monolithic ADT with partial record accessors into separate types

simonmichael:
Ok, one year is long enough! I'm rejecting this one. Let's hold out for another cleanup with more compelling benefits.

PR #361, opened 2016-06-25
Adding "--value"/"-V" option for balancesheet/incomestatement/cashflow and register

mstksg:
Let me know if this is in the spirit of the library/utility! :)

I tried my best to adhere to the style that I saw used across the library, but feel free to tell me if there are any stylistic inconsistencies!

I see some potential for refactoring out balancesheet/incomestatement/cashflow commands, too, and would be willing to do it if you think that it'd be helpful to do so.

### DIFF
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -9,6 +9,7 @@ module Hledger.Reports.PostingsReport (
   PostingsReport,
   PostingsReportItem,
   postingsReport,
+  postingsReportValue,
   mkpostingsReportItem,
 
   -- * Tests
@@ -28,6 +29,7 @@ import Test.HUnit
 import Hledger.Data
 import Hledger.Query
 import Hledger.Utils
+import Hledger.Reports.BalanceReport
 import Hledger.Reports.ReportOptions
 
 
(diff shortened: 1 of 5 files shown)


mstksg:
hm, just realized that it'd make a lot more sense to just "market value-ize" the entire journal itself before anything happens, instead of the reports.

simonmichael:
Great! Can you also update the manual for these commands - cf "Market value" section in hledger/doc/commands-balance.m4.md.

You could be right about the ordering. I found it harder to convince myself it would be correct and better that way. I'm not totally clear on how to decide what order things should be done in. If you try it, be sure to run the tests.

By all means do refactor these commands. As you know they're just a thin layer above the balance command, and arguably should be some kind of alias rather than a full command. Do you find them useful with -V ? One way or another, it would be nice to bring them to feature parity with the balance command, eg by basing them on multiBalanceReport.

mstksg:
I find -V useful because I track assets with lots of different currencies and commodities, and the summaries can be a bit unweildy because each balance and transaction can take several lines :)  -V lets me see things as one-line/one-amount aggregates

simonmichael:
Sometimes `-B` is enough for that, but it's not the same as -V I know. There's also `cur:CURRENCY` to see them one at a time.

simonmichael:
Also, the balance command can be configured to show multi-commodity amounts on one line.. just throwing out ideas.

mstksg:
@simonmichael oh, what is that configuration option? :O

simonmichael:
http://hledger.org/hledger.html#custom-balance-output

simonmichael:
@mstksg were you going to refactor this, and add docs ? If it's done similarly to the balance command, I don't mind if it's not optimally factored. Releasing 1.1 soon, I can update the docs if you prefer.

simonmichael:
It seems I may not have time to update this for HEAD today - please do it if you can.

simonmichael:
This didn't make 1.1 as I have no more hacking time today. I regret this, my apologies.

mstksg:
No problem!  I've been out of town for the holidays.  I'll probably get some time to finish this up for the next release when I get back, thank you for your patience :)

Issue #362, opened 2016-06-28
possibility to set default month


LukasGasior1:
I like to keep my journal files split by month (eg. 2016-06.journal) and include them in main file. It would be nice to be able to specify default month (similar way as default year) at the beginning of a file and only specify a day for transaction. If that makes sense I can volunteer (or at least try) to implement such feature as I'm currently learning Haskell.

simonmichael:
I guess it makes sense. It would be Ledger-incompatible. You might or might not want to be able to combine it with a year. Would be a nice experiment, if you make it I would try it out.

LukasGasior1:
I gave it some thought. First approach is to use separate directive, like `M06`, and another one is to combine it with a kind of "date prefix" directive, like `D2016/06` (which IMO looks better). Are there any plans to add support for different date formats in the future?

simonmichael:
Not at present. 

The simplest thing might be to extend the existing Y directive to allow an optional month as well as year.

```
Y2016/7
```

PR #364, opened 2016-07-03
Update default year directive to allow optional month

LukasGasior1:
This adds possibility to specify default month in `Y` directive.
Example usage:

```
Y2016/07

01 a transaction
  ...
```

Transaction date here is 2016/07/01.

Related issue: #362

### DIFF
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -224,6 +224,8 @@ instance NFData MarketPrice
 
 type Year = Integer
 
+type Month = Int
+
 -- | A Journal, containing transactions and various other things.
 -- The basic data model for hledger.
 --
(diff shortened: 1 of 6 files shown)


hledger-lib/Hledger/Data/Types.hs
```diff
@@ -237,6 +239,7 @@ type Year = Integer
 data Journal = Journal {
   -- parsing-related data
    jparsedefaultyear      :: (Maybe Year)                          -- ^ the current default year, specified by the most recent Y directive (or current date)
+  ,jparsedefaultmonth     :: (Maybe Month)                         -- ^ the current default month, specified by the most recent Y directive containing month part (or current date)
```
simonmichael: Could we disallow having a default month without a default year, perhaps jparsedefaultyear :: Maybe (Year, Maybe Month) ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -191,14 +198,16 @@ datep = do
...
-  [y,m,d] <- case (dateparts,currentyear) of
-              ([m,d],Just y)  -> return [show y,m,d]
-              ([_,_],Nothing) -> fail $ "partial date "++datestr++" found, but the current year is unknown"
-              ([y,m,d],_)     -> return [y,m,d]
-              _               -> fail $ "bad date: " ++ datestr
+  currentmonth <- getMonth
+  [y,m,d] <- case (dateparts,currentyear,currentmonth) of
+              ([d],Just y,Just m) -> return [show y,show m,d]
+              ([m,d],Just y,_)    -> return [show y,m,d]
+              ([_,_],Nothing,_)   -> fail $ "partial date "++datestr++" found, but the current year is unknown"
```
simonmichael: Could you add an error message for the case when the month is unknown

simonmichael:
Thanks! Comments added to the commit, could you also update the journal manual (hledger-lib/doc/hledger_journal.5.m4.md).

LukasGasior1:
@simonmichael Thanks for your comments, I've updated the code accordingly

simonmichael:
Hmm, I don't remember why I didn't merge this. I think the mail list was unenthusiastic and there didn't seem to be quite enough value to justify another Ledger incompatibility ?

LukasGasior1:
I think that was the case

PR #365, opened 2016-07-05
Support vty 0.7

cocreature:
The explicit exports are necessary to prevent ambiguity errors with `Mode` and `setMode` in `vty`. I have to admit that I didn’t get around to testing vty 0.6 (which was previously excluded by the bounds) but it works with vty 0.7.


simonmichael:
Thanks!

PR #382, opened 2016-08-03
fix the haddock for amounts

samdoshi:
The haddocks for `amounts` are a bit broken, see [here](https://hackage.haskell.org/package/hledger-lib-0.27.1/docs/Hledger-Data-Amount.html#v:amounts).


simonmichael:
Thanks!

Issue #396, opened 2016-09-06
hledger-rewrite doesn't compile


ascjones:
Looks like because of change to Megaparsec

simonmichael:
Thank you!

PR #397, opened 2016-09-06
Fix up hledger-rewrite to work with Megaparsec

ascjones:
Compiles and works now. However the --help doesn't seem to work - it shows the whole hledger help rather than the help for this specific command, not sure how that works.


ascjones:
Fix for #396

PR #407, opened 2016-09-24
Fix csv rules parsing

hpdeifel:
A single line containing `account1 ` (note the space at the end) should
parse as assignment of the empty string to account1. At least it did
until commit 4141067.

The problem is that megaparsec's `space` parses multiple space
characters as opposed to parsec. So in the example above it would
incorrectly consume the newline.


simonmichael:
Nice, thank you!

PR #409, opened 2016-10-03
Improve the performance of post summation

cocreature:
This improves the performance on my example query from `300ms` on average to `290ms` (where the performance is bound by parsing and checking for balanced transactions). This is sadly a bit less than I hoped for but if we find a few of these small improvements it can definitely help a lot.


cocreature:
Thinking about this a bit more, I think I am creating a space leak here. I’m closing it for now and I’ll see if I can come up with a better way to do this.

PR #410, opened 2016-10-04
Small performance improvements

cocreature:
This shaves another 10 to 20ms of the total runtime of 300ms in my testcase. Again this is a small improvement but they add up :)

### DIFF
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -195,6 +195,7 @@ postingsDateSpan' wd ps = DateSpan (Just $ postingdate $ head ps') (Just $ addDa
 
 -- AccountName stuff that depends on PostingType
 
+{-# INLINABLe accountNamePostingType #-}
 accountNamePostingType :: AccountName -> PostingType
 accountNamePostingType a
     | T.null a = RegularPosting
@@ -221,17 +222,20 @@ joinAccountNames a b = concatAccountNames $ filter (not . T.null) [a,b]
 -- | Join account names into one. If any of them has () or [] posting type
 -- indicators, these (the first type encountered) will also be applied to
 -- the resulting account name.
+{-# INLINABLE concatAccountNames #-}
 concatAccountNames :: [AccountName] -> AccountName
 concatAccountNames as = accountNameWithPostingType t $ T.intercalate ":" $ map accountNameWithoutPostingType as
     where t = headDef RegularPosting $ filter (/= RegularPosting) $ map accountNamePostingType as
 
 -- | Rewrite an account name using all matching aliases from the given list, in sequence.
 -- Each alias sees the result of applying the previous aliases.
+{-# INLINABLE accountNameApplyAliases #-}
 accountNameApplyAliases :: [AccountAlias] -> AccountName -> AccountName
+accountNameApplyAliases [] a = a
 accountNameApplyAliases aliases a = accountNameWithPostingType atype aname'
   where
     (aname,atype) = (accountNameWithoutPostingType a, accountNamePostingType a)
-    aname' = foldl
+    aname' = foldl'
              (\acct alias -> dbg6 "result" $ aliasReplace (dbg6 "alias" alias) (dbg6 "account" acct))
              aname
              aliases
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -195,7 +195,7 @@ postingsDateSpan' wd ps = DateSpan (Just $ postingdate $ head ps') (Just $ addDa
 
 -- AccountName stuff that depends on PostingType
 
-{-# INLINABLe accountNamePostingType #-}
+{-# INLINABLE accountNamePostingType #-}
 accountNamePostingType :: AccountName -> PostingType
 accountNamePostingType a
     | T.null a = RegularPosting
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -170,7 +170,7 @@ statusp =
   try
     (do many spacenonewline
         (char '*' >> return Cleared) <|> (char '!' >> return Pending)) <|>
-  return Uncleared <?> "statusp"
+  return Uncleared <?> "cleared status"
 
 {-# INLINABLE codep #-}
 codep :: TextParser m String
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -95,6 +95,7 @@ parseAndFinaliseJournal' parser assrt f txt = do
 setYear :: Year -> JournalStateParser m ()
 setYear y = modify' (\j -> j{jparsedefaultyear=Just y})
 
+{-# INLINABLE getYear #-}
 getYear :: JournalStateParser m (Maybe Year)
 getYear = fmap jparsedefaultyear get
 
@@ -117,12 +118,14 @@ popParentAccount = do
     []       -> unexpected (Tokens ('E' :| "nd of apply account block with no beginning"))
     (_:rest) -> put j{jparseparentaccounts=rest}
 
+{-# INLINABLE getParentAccount #-}
 getParentAccount :: ErroringJournalParser AccountName
 getParentAccount = fmap (concatAccountNames . reverse . jparseparentaccounts) get
 
 addAccountAlias :: MonadState Journal m => AccountAlias -> m ()
 addAccountAlias a = modify' (\(j@Journal{..}) -> j{jparsealiases=a:jparsealiases})
 
+{-# INLINABLE getAccountAliases #-}
 getAccountAliases :: MonadState Journal m => m [AccountAlias]
 getAccountAliases = fmap jparsealiases get
 
@@ -161,18 +164,23 @@ parserErrorAt pos s = throwError $ sourcePosPretty pos ++ ":\n" ++ s
 --- * parsers
 --- ** transaction bits
 
+{-# INLINABLe statusp #-}
 statusp :: TextParser m ClearedStatus
 statusp =
-  choice'
-    [ many spacenonewline >> char '*' >> return Cleared
-    , many spacenonewline >> char '!' >> return Pending
-    , return Uncleared
-    ]
-    <?> "cleared status"
+  try
+    (do many spacenonewline
+        (char '*' >> return Cleared) <|> (char '!' >> return Pending)) <|>
+  return Uncleared <?> "statusp"
 
+{-# INLINABLE codep #-}
 codep :: TextParser m String
-codep = try (do { some spacenonewline; char '(' <?> "codep"; anyChar `manyTill` char ')' } ) <|> return ""
+codep =
+  try
+    (do some spacenonewline
+        between (char '(') (char ')') (many (noneOf [')']))) <|>
+  return "" <?> "codep"
 
+{-# INLINABLE descriptionp #-}
 descriptionp :: ErroringJournalParser String
 descriptionp = many (noneOf (";\n" :: [Char]))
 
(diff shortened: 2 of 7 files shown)


hledger-lib/Hledger/Data/Posting.hs
```diff
@@ -195,6 +195,7 @@ postingsDateSpan' wd ps = DateSpan (Just $ postingdate $ head ps') (Just $ addDa
 
 -- AccountName stuff that depends on PostingType
 
+{-# INLINABLe accountNamePostingType #-}
```
simonmichael: Typo (small e)

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -161,18 +164,23 @@ parserErrorAt pos s = throwError $ sourcePosPretty pos ++ ":\n" ++ s
 --- * parsers
 --- ** transaction bits
 
+{-# INLINABLe statusp #-}
```
simonmichael: small e

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -161,18 +164,23 @@ parserErrorAt pos s = throwError $ sourcePosPretty pos ++ ":\n" ++ s
...
-  choice'
-    [ many spacenonewline >> char '*' >> return Cleared
-    , many spacenonewline >> char '!' >> return Pending
-    , return Uncleared
-    ]
-    <?> "cleared status"
+  try
+    (do many spacenonewline
+        (char '*' >> return Cleared) <|> (char '!' >> return Pending)) <|>
+  return Uncleared <?> "statusp"
```
simonmichael: In general we should be trying to make parser errors more meaningful to users, this goes in the other direction

hledger-lib/Hledger/Read/JournalReader.hs
```diff
@@ -145,8 +145,8 @@ addJournalItemP =
   -- all journal line types can be distinguished by the first
   -- character, can use choice without backtracking
   choice [
-      directivep
-    , transactionp          >>= modify' . addTransaction
+      transactionp          >>= modify' . addTransaction
```
simonmichael: Why this change in precedence ?

hledger-lib/Hledger/Utils/Debug.hs
```diff
@@ -147,6 +156,7 @@ dbg9IO = tracePrettyAtIO 9
 
 -- | Pretty-print a message and a showable value to the console if the debug level is at or above the specified level.
 -- dbtAt 0 always prints. Otherwise, uses unsafePerformIO.
+{-# INLINABLe tracePrettyAt #-}
```
simonmichael: small e

cocreature:
> You've probably remembered to check functional tests too ? ("make test")

I think so, I’ll run them again just to be sure. Could we enable them on Travis? Tests that have to be run manually have the tendency to get broken easily.

> What determines which functions get the INLINABLE pragma ? What would be the policy going forward ?

I mostly looked at the [flamegraph](https://github.com/fpco/ghc-prof-flamegraph.git) of a profile and checked which functions seem worth inlining and put INLINABLE pragmas on those. There are three criterions that I generally use

1. Is this function called only once? In this case inlining doesn’t increase code size and often enables further optimizations.

2. Is this function polymorphic? Inlinable exposes the unfolding so functions can be specialized.

3. Is this function small and called from a lot of different modules?

None of these are hard criterions and in the end I just guess and see what seems to make an effect. There are probably some unnecessary inlinables (as in the performance doesn’t improve significantly. They all seemed to inline after adding the pragma) and I’m sure there are other places where we could gain something. I don’t think a policy is useful, the decision has to be made on a case by case basis.

> Could you comment on SO answers like http://stackoverflow.com/questions/21356897/do-i-need-to-bother-with-inline-inlinable-pragmas-for-small-exported-functions and http://stackoverflow.com/questions/15440804/disadvantages-of-using-inlinable-pragma ? Are things like build time, build memory usage, executable size affected much ?

I don’t feel qualified enough to answer these especially given that people actually working on GHC have already answered. I haven’t done any measures on the build and the executable size because tbh I’m more than willing to accept longer build times for better runtime performance. If you are looking for better build time performance there are bigger fish to fry, e.g. isolating the TH in `hledger` to avoid building almost every module twice.

hledger-lib/Hledger/Read/Common.hs
cocreature: I was opting for consistency with things like `codep`. I’ll happily change it back.

cocreature:
Here are the benchmarks:

#### before
```
benchmarking read bench/10000x1000x10.journal
time                 2.001 s    (1.982 s .. 2.023 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.995 s    (1.991 s .. 1.998 s)
std dev              4.042 ms   (0.0 s .. 4.658 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking print
time                 1.651 s    (1.566 s .. 1.718 s)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 1.637 s    (1.623 s .. 1.647 s)
std dev              16.23 ms   (0.0 s .. 18.25 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking register
time                 2.088 s    (1.993 s .. 2.157 s)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 2.088 s    (2.077 s .. 2.097 s)
std dev              15.57 ms   (0.0 s .. 16.29 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking balance
time                 211.3 ms   (207.2 ms .. 213.2 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 208.0 ms   (206.5 ms .. 209.3 ms)
std dev              1.708 ms   (981.7 μs .. 2.203 ms)
variance introduced by outliers: 14% (moderately inflated)

benchmarking stats
time                 194.8 ms   (186.3 ms .. 201.7 ms)
                     0.998 R²   (0.993 R² .. 1.000 R²)
mean                 195.8 ms   (191.0 ms .. 198.6 ms)
std dev              4.466 ms   (2.133 ms .. 6.387 ms)
variance introduced by outliers: 14% (moderately inflated)
```

#### after
```
Benchmarking hledger in /home/moritz/code/haskell/hledger/hledger with criterion
benchmarking read bench/10000x1000x10.journal
time                 1.842 s    (1.824 s .. 1.862 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.829 s    (1.820 s .. 1.833 s)
std dev              7.411 ms   (0.0 s .. 8.004 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking print
time                 1.605 s    (1.541 s .. 1.688 s)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 1.607 s    (1.594 s .. 1.618 s)
std dev              16.00 ms   (0.0 s .. 18.12 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking register
time                 2.181 s    (2.175 s .. 2.191 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 2.148 s    (2.137 s .. 2.154 s)
std dev              10.03 ms   (0.0 s .. 11.33 ms)
variance introduced by outliers: 19% (moderately inflated)

benchmarking balance
time                 212.9 ms   (207.0 ms .. 217.8 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 214.1 ms   (212.3 ms .. 216.3 ms)
std dev              2.491 ms   (1.684 ms .. 3.335 ms)
variance introduced by outliers: 14% (moderately inflated)

benchmarking stats
time                 190.2 ms   (162.4 ms .. 199.7 ms)
                     0.991 R²   (0.971 R² .. 1.000 R²)
mean                 204.2 ms   (195.0 ms .. 229.3 ms)
std dev              18.65 ms   (1.565 ms .. 24.39 ms)
variance introduced by outliers: 16% (moderately inflated)

Benchmark bench: FINISH
```

I ran the tests a few times and I’ve always seen a similar improvement on parsing time while the other measurements seem meaningless (sometimes one won sometimes the other).

simonmichael:
Thanks for your comments and results, this is valuable exploration. I feel cautious about merging clever GHC management like this because I think we can quickly get into diminishing returns, with improvements that are fragile, requiring ongoing testing. 

I am doing some testing also. So far I see these INLINABLE pragmas causing no change in executable size, no change in runtime memory usage, about an 8% compilation slowdown, and a possible slight speedup in parsing.

I have to say "possible" because I'm having trouble measuring definitively. I'm looking at ways to make this easier, which we need generally. (I'm getting good value from the quick, low detail, comparative reports of tools/simplebench.hs, which I think I'll improve and package separately as quickbench.)

simonmichael:
Eg here are my current numbers. It's expected that some of these are inflated since simplebench doesn't do perfect isolated measurements, but multiple measurements on a generally quiet machine give pretty good results in my experience. So the apparent variance between yesterday and today is a bit unexpected, I need to double-check things and test more.

Yesterday:
```
***** run time, quickbench
****** 1
+-----------------------------------------++--------------------+---------------------+
|                                         || hledger-410-before | hledger-410-8bde75c |
+=========================================++====================+=====================+
| -f bench/10000x1000x10.journal print    ||               3.00 |                2.83 |
| -f bench/10000x1000x10.journal register ||               3.46 |                3.46 |
| -f bench/10000x1000x10.journal balance  ||               1.70 |                1.64 |
+-----------------------------------------++--------------------+---------------------+
****** 2
+-----------------------------------------++--------------------+---------------------+
|                                         || hledger-410-before | hledger-410-8bde75c |
+=========================================++====================+=====================+
| -f bench/10000x1000x10.journal print    ||               3.01 |                2.85 |
| -f bench/10000x1000x10.journal register ||               3.45 |                3.41 |
| -f bench/10000x1000x10.journal balance  ||               1.83 |                1.72 |
+-----------------------------------------++--------------------+---------------------+
```

Today:
```
****** $ for n in 1 2 3; do quickbench -p2 -n2 -f 410-run-time.bench hledger-410-before hledger-410-8bde75c; done
******* Using 410-run-time.bench
 Running 3 tests 2 times with 2 executables at 2016-10-11 23:35:17.435554 UTC:

 Summary (best iteration):

 +-------------------------------------------------++--------------------+---------------------+
 |                                                 || hledger-410-before | hledger-410-8bde75c |
 +=================================================++====================+=====================+
 | -f hledger/bench/10000x1000x10.journal print    ||               2.94 |                2.97 |
 | -f hledger/bench/10000x1000x10.journal register ||               3.49 |                3.43 |
 | -f hledger/bench/10000x1000x10.journal balance  ||               1.74 |                1.74 |
 +-------------------------------------------------++--------------------+---------------------+
******* Using 410-run-time.bench
 Running 3 tests 2 times with 2 executables at 2016-10-11 23:35:50.965074 UTC:

 Summary (best iteration):

 +-------------------------------------------------++--------------------+---------------------+
 |                                                 || hledger-410-before | hledger-410-8bde75c |
 +=================================================++====================+=====================+
 | -f hledger/bench/10000x1000x10.journal print    ||               3.04 |                3.12 |
 | -f hledger/bench/10000x1000x10.journal register ||               4.01 |                4.01 |
 | -f hledger/bench/10000x1000x10.journal balance  ||               2.07 |                1.91 |
 +-------------------------------------------------++--------------------+---------------------+
******* Using 410-run-time.bench
 Running 3 tests 2 times with 2 executables at 2016-10-11 23:36:28.226218 UTC:

 Summary (best iteration):

 +-------------------------------------------------++--------------------+---------------------+
 |                                                 || hledger-410-before | hledger-410-8bde75c |
 +=================================================++====================+=====================+
 | -f hledger/bench/10000x1000x10.journal print    ||               2.95 |                3.00 |
 | -f hledger/bench/10000x1000x10.journal register ||               3.95 |                3.70 |
 | -f hledger/bench/10000x1000x10.journal balance  ||               1.81 |                1.94 |
 +-------------------------------------------------++--------------------+---------------------+
```

cocreature:
Closing since I think there are easier changes that should be done first.

PR #411, opened 2016-10-05
Add a cabal.project file

cocreature:
I also snuck some ignores in there. `*~` is for backup files created by emacs and `cabal.project.local` contains *temporary* changes to things like ghc-options so it should never be commited.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ test.hs
 */doc/*.[0-9].md
 */doc/*.info
 dist-newstyle/
+cabal.project.local
+*~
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks! You might have ideas about updates needed to http://hledger.org/download.html (without complicating it further) or to cabal-install.sh.

cocreature:
Not sure if it makes sense to add it atm since `cabal new-build` is still in alpha or something. ftr you can just build it using `cabal new-build hledger hledger-lib hledger-ui hledger-web hledger-api` or if you are using cabal from git master `cabal new-build` which defaults to all targets.

PR #412, opened 2016-10-07
Anon feature

### LINKED ISSUE #265: anonymise data for easier sharing
This is mostly a thought that occurred with it's value quite debatable. 

It seems to me that one of the reasons we always talk about it but never show how we organize our ledger files in it's entirety because the data contained in it's contents are obviously quite private. It would be nice if at any moment I could create a complete copy of my files with account numbers and descriptions changed, and amounts normalized (e.g. from 0-1000 instead of the real ones). I.e. a fully functional ledger database but without any real data behind it.

Maybe that could trigger the generation of some pre-defined databases for different kinds of use cases (business, consultant, trading, etc).

<bountysource-plugin>
---
There is a **[$30 open bounty](https://www.bountysource.com/issues/18410180-anonymise-data-for-easier-sharing?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

bascott:
Resolve #265

### DIFF
--- a/doc/lib.m4
+++ b/doc/lib.m4
@@ -147,6 +147,9 @@ m4_define({{_reportingoptions_}}, {{
 : will transform the journal before any other processing by replacing the account name of every posting having the tag TAG with content VALUE by the  account name "TAG:VALUE".
 : The TAG will only match if it is a full-length match. The pivot will only happen if the TAG is on a posting, not if it is on the transaction. If the tag value is a multi:level:account:name the new account name will be "TAG:multi:level:account:name".
 
+`--anon`
+: show anonymized accounts and payees
+
 }} )m4_dnl
 m4_dnl
 m4_define({{_hledgerdescription_}}, {{
--- a/hledger/Hledger/Cli/CliOptions.hs
+++ b/hledger/Hledger/Cli/CliOptions.hs
@@ -143,6 +143,7 @@ reportflags = [
  ,flagReq  ["depth"]         (\s opts -> Right $ setopt "depth" s opts) "N" "hide accounts/postings deeper than N"
  ,flagNone ["empty","E"]     (setboolopt "empty") "show items with zero amount, normally hidden"
  ,flagNone ["cost","B"]      (setboolopt "cost") "show amounts in their cost price's commodity"
+ ,flagNone ["anon"]              (setboolopt "anon") "output ledger with anonymized accounts and payees."
  ]
 
 -- | Common output-related flags: --output-file, --output-format...
--- a/hledger/Hledger/Cli/Utils.hs
+++ b/hledger/Hledger/Cli/Utils.hs
@@ -23,12 +23,15 @@ module Hledger.Cli.Utils
     )
 where
 import Control.Exception as C
+import Data.Hashable (hash)
 import Data.List
 import Data.Maybe
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import Data.Time (Day)
+import Data.Word
+import Numeric
 import Safe (readMay)
 import System.Console.CmdArgs
 import System.Directory (getModificationTime, getDirectoryContents, copyFile)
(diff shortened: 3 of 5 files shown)


simonmichael:
Sorry @bascott, I meant to process this PR next and lost track.

Looks like a great start, thank you! Would you mind documenting it as well. I think probably a manual section similar to "output destination", and linked anywhere that is linked (hledger/doc/commands[-balance].m4.md I think). Or, under OPTIONS ? I guess it applies to all commands not just balance/register/print.

simonmichael:
Anonymisation of commodities would also be nice, perhaps ?

Any thoughts on anonymising amounts ? Without that you can still figure out a lot about someone's financial life. But of course, things still have to balance, pass balance assertions etc.

simonmichael:
There's a problem with query matching, eg `hledger print --anon SOMEACCT` matches nothing (using the original or hashed form of SOMEACCT). Oh, I see why, the hash is different on each run. Is that deliberate ?

simonmichael:
Forgot to say: we should have a few tests (functional ones probably) for this.

bascott:
Regarding the hash being different on each run, that was deliberate to match the functionality on ledger. It should be easily changed if you'd prefer to have the same hash on each run.

simonmichael:
Yes, let's make it deterministic - I don't think having it change it buys us anything.

bascott:
Gonna work on anonymising amounts. Are withdrawals from accounts represented internally as negative numbers?

simonmichael:
Ideally we'd keep as much working as possible. Eg for when reporting some complex bug with prices and balance assertions. But try the easiest things first I guess.

Sign of a "withdrawal" depends on the type of account.

AaronM04:
What is the threat model? My understanding was that we can assume only the intended user has access to ledger files and/or hledger web.

On October 20, 2016 12:52:43 PM PDT, Simon Michael <notifications@github.com> wrote:
>Anonymisation of commodities would also be nice, perhaps ?
>
>Any thoughts on anonymising amounts ? Without that you can still figure
>out a lot about someone's financial life. But of course, things still
>have to balance, pass balance assertions etc.
>
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/simonmichael/hledger/pull/412#issuecomment-255210122

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

simonmichael:
For me the vision for this is making it easy to share non-trivial examples drawn from our real-world data and usage. Creating realistic synthetic examples is a lot of work. Generally we don't like to reveal our actual data for privacy and security reasons. If we could remove those concerns just by adding a flag, it would open up a lot more troubleshooting, support and sharing of tips and examples.

The threat model is that we share some "anonymized" data thinking it gives away no private information, when actually it's quite easy for a motivated person or bot to "decrypt" it and learn a lot about your finances and life.

After thinking about it slightly, it's a rather interesting challenge, since crypto is hard and also because of the goal of generating useful shareable examples. For that, we'd like to preserve constraints such as balanced transactions, balance assertions passing, balance zero points, prices etc. Also we'd like to generate easy human-readable names instead of random-looking data. I don't know how far we will go, and it may need a lot of experimentation. What we have so far is good, and when we are able to obfuscate amounts in some non-reversible way it should help a lot.

simonmichael:
@bascott, do you have some docs ready to go ? If not I'll do it - getting close to release time.

bascott:
@simonmichael I've documented the anon option in the reportingoptions macro in hledger/doc/lib.m4. I originally included the regenerated documents in the commit, but took them out because the version of m4 on my system made a lot of changes to the documents in various places other than including the new option and I'm not familiar enough with man page or info page syntax to manually edit them.

simonmichael:
Thank you @bascott, good call. I seem to be using GNU M4 1.4.6.

Merging this, thank you!

bascott:
@simonmichael Glad to have helped.

About anonymising the amounts and adding tests, should I raise separate issues for those?

simonmichael:
For tests related to the work so far, I would just reference the original issue. For next steps, you could either open a new issue, or reopen #265 after claiming the bounty. Either is fine with me.

simonmichael:
This feature appears questionable in 2023 (including the determinism that I pushed for): input welcome at #2133.

Issue #340, opened 2016-05-04
hledger-web UI should be usable on mobile


simonmichael:
Split off from #324. http://demo.hledger.org does not fit well on small screens, eg a 320px wide iPhone 4.

nikhiljha:
It's basically impossible to have the table look good in portrait on a mobile device. I think a solution would be something like...

```md
# 8/20/3020
Lorem ipsum some description here.
savings           +1000
equity            -1000

# 8/20/3020
Lorem ipsum some cooler description here.
savings           +1000
equity            -1000

# 8/18/3020
Lorem ipsum I like to eat pancakes.
savings           +1000
equity            -1000
```

That way the amount is still right justified (so you can look to the right easily) and everything else is more or less in a place that makes sense. Still working on spacing for the actual web app, but I'll send a screenshot soon(tm).

PR #418, opened 2016-10-11
Added .btn-default to unify buttons across browsers

theSuess:
By overriding the default brower-styles using the class `btn-default`, the buttons on hldeger-web are now rendered the same on all browsers. It also gives the interface a cleaner look
before_after

### DIFF
--- a/hledger-web/Foundation.hs
+++ b/hledger-web/Foundation.hs
@@ -382,7 +382,7 @@ addform _ vd@VD{..} = [hamlet|
          |]
        | otherwise = [hamlet|
           <td #addbtncell style="text-align:right;">
-           <button type=submit .btn .btn-lg name=submit>add
+           <button type=submit .btn .btn-default .btn-lg name=submit>add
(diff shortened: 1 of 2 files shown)


theSuess:
The default bootstrap distribution offers seven kind of buttons.
btnclasses

For this usecase, making the search button primary should probably be enough. I would leave the "?" button as is though because the first one being a button should be enough of a hint to suggest that it is clickable as well

simonmichael:
My habit is to stick with minimal, low-colour styling (but enough to indicate buttons, etc.)  until a better UI theme arrives. Won't styling just this button differently from the overall UI might seem odd ? 

But I'll probably merge whatever you do anyway, to refresh the eyes.

theSuess:
You're right, for fancy buttons the ui needs to be overhauled.
I'll probably tackle issue #340 next so we'll see how everything plays togeher. Thanks for merging!

simonmichael:
Thanks @theSuess!

PR #429, opened 2016-11-13
Docker compatibility 🤖🙏

glushchenko:


simonmichael:
Thanks. If I understand rightly, this makes hledger-web also respond to incoming requests to any IP address configured for this machine, rather than just local requests from this machine ? Any comments on https://github.com/simonmichael/hledger/pull/432 which goes the other way ?

Is it cross-platform ? http://serverfault.com/a/300410 says "ping 0.0.0.0 on a Windows system results in an error message".

simonmichael:
For hledger-web and hledger-api, I am thinking it's best to be consistent, to listen only on localhost by default, and to add a --host option which would allow 0.0.0.0 or other address to be configured.

simonmichael:
OR (devil's advocate): we can prioritise just-works over secure-by-default, keep things simple, and assume that people starting a web service should expect that it will serve web requests.

glushchenko:
--host argument is a good idea

mhansen:
+1 to a --host option (secure-by-default) now that there's a use case for the option :-)

simonmichael:
hledger-web --host IPADDR should work now. (I feel it should accept hostnames too, +1 to anyone who fixes that.)

PR #432, opened 2016-11-20
Default to serving API only on loopback interface

mhansen:
Before, hledger-api was binding on *4 (any IPv4 address), so anyone on the local network could access the API and the files it was serving.

Switch this to only listening on the loopback interface, so only the local machine can request the data (more secure-by-default). You can still put it behind a reverse proxy if you want to expose it.

We can discuss whether there should be another command line argument for host, if you like? I'm easy - just lean away from command line arguments unless we have a use case for them, and I'm not sure we do here yet?

### DIFF
--- a/hledger-api/hledger-api.hs
+++ b/hledger-api/hledger-api.hs
@@ -97,7 +97,8 @@ serveApi p d f j = do
   printf "Starting web api http://localhost:%d/api/v1 for %s\n" p f
   printf "and file server  http://localhost:%d        for %s/\n" p d
   printf "Press ctrl-c to quit\n"
-  Warp.run p $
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks. Any comments on https://github.com/simonmichael/hledger/pull/429 which goes the other way (to make Docker work) ? I guess -web and -api should be consistent.

simonmichael:
Adding a `--host` option would be easy and I think it's probably a good idea.

simonmichael:
Let's discuss more on #429..

simonmichael:
That should take care of it, thanks for the report.

mhansen:
Thanks very much Simon, I appreciate this (and all your work on hledger!).

Issue #436, opened 2016-12-01
Does `instance Monoid Journal` drop too much information?


johannesgerer:
Whenever a journal is read with `readJournalFiles` the `mconcat` in `Read.hs:141` will call `mappend _ mempty`, which means that things like `jparsedefaultcommodity` are set to `Nothing`.

One unfortunate consequence is: calling `readJournalFile f` will give a different result from `readJournalFiles [f]`.

Is this intended or would you accept a pull request to change this?

simonmichael:
Thanks for the report. That sounds not ideal, I would gladly receive a pull request.

simonmichael:
I remember the docs say that parse state is reset for each file when reading multiple files, so I'm not absolutely sure it's a bug. Your PR might clarify things.

simonmichael:
Great - thanks!

PR #437, opened 2016-12-01
Changed behavior of `readJournalFiles` to be identical to `readJournalFile` for singleton lists

johannesgerer:
This solves  #436.

### DIFF
--- a/hledger-lib/Hledger/Read.hs
+++ b/hledger-lib/Hledger/Read.hs
@@ -36,6 +36,7 @@ module Hledger.Read (
 ) where
 
 import Control.Applicative ((<|>))
+import Control.Arrow (right)
 import qualified Control.Exception as C
 import Control.Monad.Except
 import Data.List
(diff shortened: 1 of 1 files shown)


hledger-lib/Hledger/Read.hs
```diff
@@ -126,8 +127,11 @@ defaultJournalPath = do
 --
 readJournalFiles :: Maybe StorageFormat -> Maybe FilePath -> Bool -> [PrefixedFilePath] -> IO (Either String Journal)
 readJournalFiles mformat mrulesfile assrt prefixedfiles = do
-  (either Left (Right . mconcat) . sequence)
+  (right mconcat1 . sequence)
     <$> mapM (readJournalFile mformat mrulesfile assrt) prefixedfiles
+  where mconcat1 :: Monoid t => [t] -> t
+        mconcat1 [] = mempty
+        mconcat1 x = foldr1 mappend x
```
simonmichael: I often hear foldr is the wrong fold to use.. is it the right one here, do you think ?

simonmichael: Is the haddock for readJournalFiles still correct (about the parse state) ?

johannesgerer: the old `mconcat` also used it, so no change here. At least it is explicitly visible now.

johannesgerer: Yes the docs are correct. Actually, only the new version of the function does what the docs say, as the parse state from the last file now lives on, due to the use of `mconcat1`, which only uses `mempty` for empty lists.

johannesgerer:
This is marked as 'merged', yet i cannot find the commit in branch master. Why?

simonmichael:
I don't know! I just clicked the merge button on github. I've redone it manually, thanks.

Issue #288, opened 2015-10-11
Support balance resetting


mathstuf:
Ledger support is here:

http://ledger-cli.org/3.0/doc/ledger3.html#Resetting-a-balance

I would like to use this for other things (will post to the list).

Related to #258.

simonmichael:
#129 may have some code to salvage

simonmichael:
Duplicate of #157.

Issue #157, opened 2014-01-12
postings with just a balance assertion but no amount should work [$10]


simonmichael:
Ledger allows assertions to be made without writing the amount explicitly before the =, eg:

```
  acct        = 1
```

but hledger requires the amount, which can be tedious:

```
  acct     0 = 1
```
##

simonmichael:
I have some work in progress for this. It's rather tricky to implement generally, eg if you want to allow for out of order transactions and postings with arbitrary posting dates. (Inferring amounts from assertions requires knowing the account balance at this point in time, which requires that all previous transactions *and* any previously-dated postings or transactions we've not yet parsed have been fully inferred.. which might require inferring asserted amounts for postings on other accounts, starting a whole new round of processing which won't be solvable in all cases..)

ony:
I don't think you need to have all previously-dated postings fully inferred. Actual numbers only required when you make a report.
Back in past I considered `Journal` to be a `Monoid` that can be combined. I.e. `Transaction` can be wrapped as a `Journal` and be combined with other one to form a new value. `Journal` in one of aggregated form/projection represents a changes in balances of multiple accounts. Part of accounts will get their values independently from input balances and thus may be calculated during `mappend` of `Journal`'s.
I.e. in balance reports some accounts will have exact values, some may be represented as an offsets and all others will be unknown (equivalent of zero offset but without any associated transaction).

P.S. I do actively use approach with balance reset. Though there is only one use case that have no work around.
When I track my budget I often do it not so precisely. Periodically I simply check balance for cash or write down bank account balance notification. To do that I use something like this:

```
2016/12/04
    Assets:Cash:Wallet  =$40
    Equity:Unbalanced
```

If  `Equity:Unbalanced` is pretty high that means that I need to change my behavior to write down my expenses more precisely to be able to understand them. But sometimes I may find out where did my money gone. And I simply put transaction in the middle of journal.

```

2016/12/01
    Assets:Cash:Wallet  -$120
    Expenses:Clothes

2016/12/04 balance reset
    Assets:Cash:Wallet  =$40
    Equity:Unbalanced
```

That's pretty useful that I don't need to change "balance reset" transaction up to the moment when it isn't required anymore. I.e. when I figure out all my expenses that resulted in particular balance `ledger` reports about transaction with no balancing.

[...]

PR #438, opened 2016-12-07
Balance Assignments and accounts resetting

johannesgerer:
This is my first version (untested). I will be testing it tomorrow.

Be invited to start reviewing it now. 

See #129 #288 #157


johannesgerer:
It compiles now and passes all tests (including new ones).

Notes: It mimics `ledger`'s behavior in ignoring temporal order and strictly using parse order for both the transactions and the postings.

Core functionality in concisely implemented in `Hledger.Data.Journal.inferOrCheckPosting` taking only 7 lines.

simonmichael:
Great!

First comment, I'm sorry but abandoning temporal order won't work. Ledger's parse-order behaviour is easier to implement but more confusing for users, hledger and Beancount have it right here.

johannesgerer:
I understand. Temporal order is nice to have. I implemented it, with the restriction, that assignments (i.e. postings with assertions but no amount) are only allowed in transactions where postings have no `pdate` defined. The more general problem is much more complex.

Time complexity should not worse than before.

simonmichael:
Nice! Quick thoughts: 

Would you mind squashing these into one or more logical patches for easy review. 

There's a typo you probably noticed already (cf travis). 

Your restriction sounds quite reasonable. 

How necessary is unsafeFreeze ? Someone (me) may overlook it and cause a bug in future.

Could you add haddocks for a few more of the new top-level functions ? One sentence is fine. It may seem redundant but I find it quite helpful.

johannesgerer:
Ok. `unsafeFreeze` is gone.

I have added haddock to everything.

And cleaned it up in 5 separate commits.

hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -463,8 +470,8 @@ journalApplyAliases aliases j@Journal{jtxns=ts} =
 -- check balance assertions.
 journalFinalise :: ClockTime -> FilePath -> Text -> Bool -> ParsedJournal -> Either String Journal
 journalFinalise t path txt assrt j@Journal{jfiles=fs} = do
-  (journalNumberAndTieTransactions <$>
-    (journalBalanceTransactions $
+  (journalTieTransactions <$>
```
simonmichael: Can you explain more why numbering transactions is no longer needed ? Is it not used for eg the transaction ids in `print -O csv` ?

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -198,7 +198,7 @@ data Posting = Posting {
       pcomment          :: Text,              -- ^ this posting's comment lines, as a single non-indented multi-line string
       ptype             :: PostingType,
       ptags             :: [Tag],             -- ^ tag names and values, extracted from the comment
-      pbalanceassertion :: Maybe MixedAmount, -- ^ optional: the expected balance in the account after this posting
+      pbalanceassertion :: Maybe Amount, -- ^ optional: the expected balance in the account after this posting
```
simonmichael: IIRC the doc should be changed to "the expected balance in this commodity in the account after this posting". I can fix it later.

tests/journal/balance-assertions.test
```diff
@@ -136,3 +136,140 @@ hledger -f - stats
 # >>> /Transactions/
 # >>>2
 # >>>=0
+
```
simonmichael: Tests! These make me very happy :)

simonmichael: Oops, submitted before I was finished.. still learning github's review flow. 

Overall I like this a lot. I think it is not easy code for future hledger hackers to understand, eg at handleObject I gave up (for now). But I'm not sure much can be done about it; the feature is inherently complex. It seems to me you've done a good job encapsulating everything in a principled and compact way, and perhaps we will find small improvements in future making it more transparent. And your tests give confidence. I'll do some additional testing of my own next. Great work @johannesgerer.

tests/cli/multiple-files.test
```diff
@@ -43,7 +43,7 @@ hledger print -f personal.journal -f business.journal -f alias.journal -f person
 
 
 # 3. files can be of different formats
-hledger print -f personal.journal -f a.timeclock -f b.timedot
+hledger print -f personal.journal -f ../journal/a.timeclock -f ../journal/b.timedot
```
johannesgerer: They failed with `make test`. This change is required, because these journal are in a different directory. (And the way, they were relative even before the change.)

hledger-lib/Hledger/Data/Journal.hs
johannesgerer: The numbering now is performed in  `journalBalanceTransactions`, which uses it to get back the parsing order after iterating over time ordered postings.

hledger-lib/Hledger/Data/Types.hs
johannesgerer: i fixed it.

tests/journal/balance-assertions.test
johannesgerer: I think it would be beneficial, I you changed the docs, because you see the whole picture and this also helps check again, if everything works as expected.

tests/cli/multiple-files.test
simonmichael: Doh! I have uncommited symlinks. Thank.

simonmichael:
This assertion fails with the new code:
```
1/1
    a      $1
    b      -1 zorkmids

1/2
    a     $-1 = $0
    b
```

johannesgerer:
Ok, thanks for spotting this! I did not handle prices correctly.

It now removes all prices before checking any balances. It this ok?

 The test case is now included in the tests and works!

johannesgerer:
What can I do to make it more accessible?

`handleObject` receives different objects ordered by date. It can handle either a single posting (from an already balanced transaction without assigments) or a whole transaction with assignments (which is required to no posting with `pdate` set.).

For a single posting, there is not much to be done. Only add its amount to its account and check the assertion, if there is one. This functionality is provided by `addAmountAndCheckBalance`.

For a whole transaction, it loops over all postings, and performs `addAmountAndCheckBalance`, if there is an amount. If there is no amount, the amount is inferred by the assertion or left empty if there is no assertion.

simonmichael:
Thanks for the fix. Re code clarity, I would look at names first, I often find more informative names for things after a little time passes. Here's a few top-of-the-head ideas:

journalBalanceTransactions' -> journalBalanceTransactionsStatefully, journalProcessAssignmentsAndAssertions, ... ? By the way the presence of the boolean assrt parameter makes me wonder if checking assertions can be separated out and done as an outer step after processing assignments, or not ?

BalanceState -> BalanceStateM, BalancingComputation or something, hinting that it's a monad not a state value ?

toDated, "This converts a transaction into a list of objects" - actually it returns a computation generating that list, right ? toDated -> findAssertionDependencies ?

handleObject -> ? I can't come up with a clear name for this, which is usually a bad sign. It seems to do two fairly different jobs, doesn't it ? Maybe it can be refactored. And/or add your helpful comment above to it.

After names feel good, I would look at code, to see if the hardest functions could have an easier to understand implementation.

But you have good haddocks in place and the code doesn't affect other modules, so none of this will delay merging.

simonmichael:
Here's some benchmarking, of the balance command with the 10k sample file (10k transactions, 0 assertions, 0 assignments):
```
$ quickbench -N 3 -w hledger,hledger-assignments 'hledger -f data/10000x1000x10.journal balance' 'hledger -f data/10000x1000x10.journal balance --ignore-assertions'
Running 2 tests 1 times with 2 executables at 2016-12-09 13:31:00 PST:

Best times 1:
+-----------------------------------------------------------++---------+---------------------+
|                                                           || hledger | hledger-assignments |
+===========================================================++=========+=====================+
| -f data/10000x1000x10.journal balance                     ||    1.82 |                1.90 |
| -f data/10000x1000x10.journal balance --ignore-assertions ||    1.69 |                1.90 |
+-----------------------------------------------------------++---------+---------------------+

Best times 2:
+-----------------------------------------------------------++---------+---------------------+
|                                                           || hledger | hledger-assignments |
+===========================================================++=========+=====================+
| -f data/10000x1000x10.journal balance                     ||    1.84 |                1.90 |
| -f data/10000x1000x10.journal balance --ignore-assertions ||    1.74 |                1.89 |
+-----------------------------------------------------------++---------+---------------------+

Best times 3:
+-----------------------------------------------------------++---------+---------------------+
|                                                           || hledger | hledger-assignments |
+===========================================================++=========+=====================+
| -f data/10000x1000x10.journal balance                     ||    1.81 |                2.00 |
| -f data/10000x1000x10.journal balance --ignore-assertions ||    1.75 |                1.91 |
+-----------------------------------------------------------++---------+---------------------+
```

And of my personal journal (1k transactions, 140 assertions, 0 assignments):
```
$ quickbench -N 3 -w hledger,hledger-assignments 'hledger balance' 'hledger balance --ignore-assertions'Running 2 tests 1 times with 2 executables at 2016-12-09 13:32:26 PST:

Best times 1:
+-----------------------------++---------+---------------------+
|                             || hledger | hledger-assignments |
+=============================++=========+=====================+
| balance                     ||    0.49 |                0.47 |
| balance --ignore-assertions ||    0.43 |                0.48 |
+-----------------------------++---------+---------------------+

Best times 2:
+-----------------------------++---------+---------------------+
|                             || hledger | hledger-assignments |
+=============================++=========+=====================+
| balance                     ||    0.42 |                0.48 |
| balance --ignore-assertions ||    0.40 |                0.42 |
+-----------------------------++---------+---------------------+

Best times 3:
+-----------------------------++---------+---------------------+
|                             || hledger | hledger-assignments |
+=============================++=========+=====================+
| balance                     ||    0.41 |                0.45 |
| balance --ignore-assertions ||    0.38 |                0.44 |
+-----------------------------++---------+---------------------+
```

Here's another run including the 100k sample file:
```
Best times:
+--------------------------------------------++---------+---------------------+
|                                            || hledger | hledger-assignments |
+============================================++=========+=====================+
| -f data/100x100x10.journal print           ||    0.06 |                0.06 |
| -f data/1000x1000x10.journal print         ||    0.31 |                0.32 |
| -f data/10000x1000x10.journal print        ||    2.84 |                3.04 |
| -f data/100000x1000x10.journal print       ||   29.63 |               33.51 |
| -f data/100000x1000x10.journal print ff    ||   17.48 |               18.22 |
| -f data/100x100x10.journal register        ||    0.07 |                0.06 |
| -f data/1000x1000x10.journal register      ||    0.36 |                0.36 |
| -f data/10000x1000x10.journal register     ||    3.41 |                3.49 |
| -f data/100000x1000x10.journal register    ||   35.97 |               35.96 |
| -f data/100000x1000x10.journal register ff ||   17.88 |               20.36 |
| -f data/100x100x10.journal balance         ||    0.06 |                0.07 |
| -f data/1000x1000x10.journal balance       ||    0.31 |                0.28 |
| -f data/10000x1000x10.journal balance      ||    1.97 |                2.05 |
| -f data/100000x1000x10.journal balance     ||   19.96 |               20.69 |
| -f data/100000x1000x10.journal balance ff  ||   18.39 |               20.37 |
+--------------------------------------------++---------+---------------------+
```

I think these numbers are saying the new code brings a slight performance penalty, whether you use assertions or not. With luck it can be optimised further, if not I think it's not a blocker.

simonmichael:
I should say, I slightly mistrust that last run. I'll post another one shortly.

simonmichael:
```
$ quickbench -w hledger,hledger-assignments
Running 15 tests 1 times with 2 executables at 2016-12-09 13:44:16 PST:

Best times:
+--------------------------------------------++---------+---------------------+
|                                            || hledger | hledger-assignments |
+============================================++=========+=====================+
| -f data/100x100x10.journal print           ||    0.07 |                0.05 |
| -f data/1000x1000x10.journal print         ||    0.33 |                0.32 |
| -f data/10000x1000x10.journal print        ||    2.88 |                3.06 |
| -f data/100000x1000x10.journal print       ||   30.72 |               32.19 |
| -f data/100000x1000x10.journal print ff    ||   17.47 |               18.48 |
| -f data/100x100x10.journal register        ||    0.08 |                0.06 |
| -f data/1000x1000x10.journal register      ||    0.38 |                0.34 |
| -f data/10000x1000x10.journal register     ||    3.51 |                3.64 |
| -f data/100000x1000x10.journal register    ||   36.02 |               36.35 |
| -f data/100000x1000x10.journal register ff ||   18.00 |               19.01 |
| -f data/100x100x10.journal balance         ||    0.05 |                0.05 |
| -f data/1000x1000x10.journal balance       ||    0.25 |                0.29 |
| -f data/10000x1000x10.journal balance      ||    1.83 |                2.07 |
| -f data/100000x1000x10.journal balance     ||   18.15 |               18.76 |
| -f data/100000x1000x10.journal balance ff  ||   16.50 |               18.00 |
+--------------------------------------------++---------+---------------------+
```

johannesgerer:
Peformance: I am a bit surprised that the penalty is that is that small, considering that other than in the original version, the whole list of transactions needs to be created in an array and then copied. But I would need to see some profiling to say more.

Assertion check in extra step: The assignments require basically the same loop over temporally sorted objects, which implies that not doing checks there also, would be suboptimal.

Names: I changed some names. Feel free to change anything you want.

Easier implementation: The current implementation looks to me as a direct translation of the verbal description of the code, where every sentence is one line/statement. Plus, my Haskell style might be different from yours. I would mind if you changed it, I do not hang on to code I have written, I am only interested in the feature.

Most important: Do you think you could push the merged version to hackage? I need for another project of mine I want to push (https://github.com/johannesgerer/buchhaltung).

johannesgerer:
PS: Performance: Other than the use of ST there has not been any performance optimization at the expense of simplicity.

The point is, that even without assertions checks, the balances need to be built, in order to be able to infer the assigned amounts. You could add a check in `journalBalanceTransactions` that simply returns `fmap balanceTransaction $ jtxns j` if there are no assertions in the whole journal.

simonmichael:
I'll merge this in the next day or two, and it can be improved if needed.

Unfortunately for buchhaltung (which I'm looking forward to see) I think it will be a little longer until it's on hackage. I aim to release 1.1 this month, but I'd like to get a few more things in. I'll try to make it soon.

simonmichael:
PS this only works for installing from source, but you could provide a stack.yaml which can pull dependencies from a branch on github as well as hackage.

johannesgerer:
What about 1.0.3, as the new functionality is a superset of the existing one?

simonmichael:
I prefer to use those for bugfixes, and save the features for major releases.

hledger-lib/Hledger/Read.hs
```diff
@@ -126,8 +127,11 @@ defaultJournalPath = do
 --
 readJournalFiles :: Maybe StorageFormat -> Maybe FilePath -> Bool -> [PrefixedFilePath] -> IO (Either String Journal)
 readJournalFiles mformat mrulesfile assrt prefixedfiles = do
-  (either Left (Right . mconcat) . sequence)
+  (right mconcat1 . sequence)
```
ony: Why can't we use `mconcat` here?

As I understand the only difference between `mconcat1` and `mconcat` is that custom implementation avoids doing `mappend (last x) mempty` and starts with `mappend (last (init x)) (last x)`.
If there is some issues with that maybe we should re-consider type `Journal`.

johannesgerer:
@ony Yes, this is already merged:  #437

simonmichael:
[Documented](http://hledger.org/journal.html#balance-assignments) and merged. Thanks @johannesgerer !

PR #439, opened 2016-12-08
more general parser types enabling reuse outside of IO

johannesgerer:
As a user of the library I was often unable to use the provided parsers outside of IO or ExceptT. This commit should not change to values, only generalize their types.


johannesgerer:
Any plans to merge this?

simonmichael:
I was going to do it after #438 to avoid complicating that, but here goes. Thanks!

johannesgerer:
Thanks, I merged it into #438

PR #441, opened 2016-12-19
use isSpace in lstrip

ssbl:


simonmichael:
isSpace allows newlines, which seems like a big change. But tests seem to pass, so I'm ok with it. Thanks!

ssbl:
@simonmichael Thanks!

PR #456, opened 2016-12-31
Fix hledger-equity

ony:
- Make it compatible with current `BalanceReportItem`
- Make multi-commodity postings to be a valid journal entries


simonmichael:
Thanks!

PR #460, opened 2017-01-05
Feature/pivot implicit tags

ony:
As suggested @simonmichael use of `--pivot` with transaction-level data may be reasonable when additional filtering applied.
```
2016/02/16 (1138) Auchan
    assets:bank account
    expense:snacks                      2 EUR
    expense:grocery                    20 EUR

2016/02/16 StarBars
    assets:bank account                               
    expense:coffee                     5 EUR
```
With `hledger --pivot payee bal --no-total --flat ^expense` produces nice report by payees:
```
              22 EUR  payee:Auchan
               5 EUR  payee:StarBars
```

### DIFF
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -25,6 +25,7 @@ module Hledger.Data.Posting (
   hasAmount,
   postingAllTags,
   transactionAllTags,
+  postingAllImplicitTags,
   relatedPostings,
   removePrices,
   -- * date operations
@@ -167,6 +168,18 @@ postingStatus Posting{pstatus=s, ptransaction=mt}
   | s == Uncleared = case mt of Just t  -> tstatus t
                                 Nothing -> Uncleared
   | otherwise = s
+
+-- | Implicit tags for this transaction.
+transactionImplicitTags :: Transaction -> [Tag]
+transactionImplicitTags t = filter (not . T.null . snd) [("code", tcode t)
+                                                        ,("desc", tdescription t)
+                                                        ,("payee", tdescription t)
+                                                        ]
+
+-- | Tags for this posting including implicit and any inherited from its parent transaction.
+postingAllImplicitTags :: Posting -> [Tag]
+postingAllImplicitTags p = ptags p ++ maybe [] transactionTags (ptransaction p)
+    where transactionTags t = ttags t ++ transactionImplicitTags t
 
 -- | Tags for this posting including any inherited from its parent transaction.
 postingAllTags :: Posting -> [Tag]
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -659,6 +659,8 @@ compareAmount ord q Amount{aquantity=aq} = case ord of Lt      -> aq <  q
                                                        AbsEq   -> abs aq == abs q
 
 -- | Does the match expression match this posting ?
+--
+-- Note that for account match we try both original and effective account
 matchesPosting :: Query -> Posting -> Bool
 matchesPosting (Not q) p = not $ q `matchesPosting` p
 matchesPosting (Any) _ = True
@@ -667,7 +669,8 @@ matchesPosting (Or qs) p = any (`matchesPosting` p) qs
 matchesPosting (And qs) p = all (`matchesPosting` p) qs
 matchesPosting (Code r) p = regexMatchesCI r $ maybe "" (T.unpack . tcode) $ ptransaction p
 matchesPosting (Desc r) p = regexMatchesCI r $ maybe "" (T.unpack . tdescription) $ ptransaction p
-matchesPosting (Acct r) p = regexMatchesCI r $ T.unpack $ paccount p -- XXX pack
+matchesPosting (Acct r) p = matchesPosting p || matchesPosting (originalPosting p)
+    where matchesPosting p = regexMatchesCI r $ T.unpack $ paccount p -- XXX pack
 matchesPosting (Date span) p = span `spanContainsDate` postingDate p
 matchesPosting (Date2 span) p = span `spanContainsDate` postingDate2 p
 matchesPosting (Status Uncleared) p = postingStatus p /= Cleared
--- a/hledger/Hledger/Cli/Utils.hs
+++ b/hledger/Hledger/Cli/Utils.hs
@@ -93,7 +93,7 @@ pivot tag j = j{jtxns = map pivotTrans . jtxns $ j}
   pivotPosting p
     | Just (_ , value) <- tagTuple = p{paccount = joinAccountNames tag value, porigin = Just $ originalPosting p}
     | _                <- tagTuple = p
-   where tagTuple = find ((tag ==) . fst) . ptags $ p
+   where tagTuple = find ((tag ==) . fst) . postingAllImplicitTags $ p
 
 -- | Apply the anonymisation transformation on a journal, if option is present
 anonymiseByOpts :: CliOpts -> Journal -> Journal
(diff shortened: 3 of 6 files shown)


simonmichael:
Very nice. We'll need to document it also. I think a new subsection of http://hledger.org/manual.html#options would be a good place. I can help expand it.

My vote would be to allow only the canonical field names for now: `code` and `description`. I like to avoid multiple spellings if possible. If it's too tedious we could revisit later.

It might be out of scope, but here are some commands I ran just now, some of which might be considered confusing. I'll follow up on this later.
```
$ hledger -f data/pivot-460.journal bal --pivot x
             -27 EUR  assets:bank account
              27 EUR  expense
               5 EUR    coffee
              20 EUR    grocery
               2 EUR    snacks
--------------------
                   0
$ hledger -f data/pivot-460.journal bal --pivot desc
--------------------
                   0
$ hledger -f data/pivot-460.journal --pivot payee bal --no-total --flat ^expense
              22 EUR  payee:Auchan
               5 EUR  payee:StarBars
$ hledger -f data/pivot-460.journal bal --pivot desc expense
              27 EUR  desc
              22 EUR    Auchan
               5 EUR    StarBars
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal --pivot desc expense --drop 1
              27 EUR  
              22 EUR    
               5 EUR    
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal --pivot desc expense --drop 1 --flat
              22 EUR  Auchan
               5 EUR  StarBars
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal --pivot desc expense  --flat
              22 EUR  desc:Auchan
               5 EUR  desc:StarBars
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal --pivot desc expense --drop 1 --flat
              22 EUR  Auchan
               5 EUR  StarBars
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal expense
              27 EUR  expense
               5 EUR    coffee
              20 EUR    grocery
               2 EUR    snacks
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal expense  --pivot desc 
              27 EUR  desc
              22 EUR    Auchan
               5 EUR    StarBars
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal   --pivot desc 
--------------------
                   0
$ hledger -f data/pivot-460.journal bal   --pivot desc  -E
                   0  desc
                   0    Auchan
                   0    StarBars
--------------------
                   0
$ hledger -f data/pivot-460.journal bal expense  --pivot desc 
              27 EUR  desc
              22 EUR    Auchan
               5 EUR    StarBars
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal expense  --pivot payee
              27 EUR  payee
              22 EUR    Auchan
               5 EUR    StarBars
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal expense  --pivot code
              22 EUR  code:1138
               5 EUR  expense:coffee
--------------------
              27 EUR
$ hledger -f data/pivot-460.journal bal   --pivot code
              -5 EUR  assets:bank account
               5 EUR  expense:coffee
--------------------
                   0
$ hledger -f data/pivot-460.journal bal   --pivot code -E
              -5 EUR  assets:bank account
                   0  code:1138
               5 EUR  expense:coffee
--------------------
                   0
$ hledger -f data/pivot-460.journal bal expense  --pivot code
              22 EUR  code:1138
               5 EUR  expense:coffee
--------------------
              27 EUR
```

ony:
Ideally, I'd like to have directives that can configure `hledger` in `ledger-cli` compatible way to use with old journals. For example some of my postings looks like:
```
2016/01/03 *
    food  $20  ; Payee: McDonalds
    grocery  $80  ; Payee: SevenEleven
    wallet

2016/01/03 StarBucks
    coffee  $5
    wallet
```

Thus I'd like to keep alias `payee` for `desc`. Though I agree that documentation should use only `desc` in examples.

Regarding examples. I remember something like `ledger reg -p 2017 -P --sort date '^expense'`.
Now can achieve the similar output with: `hledger reg -p 2017 --pivot desc '^expense' -D`

simonmichael:
I've annotated https://github.com/simonmichael/hledger/pull/460#issuecomment-270759957 , with one idea for improvement.

ony:
> Perhaps postings without the tag should be excluded or assigned a blank/default value ?

You can exclude transactions without tag by using `tag:...` in match query.
We can introduce `effective:...` query expression which will result in applying a match to an "effective" posting by slicing off `porigin` (replacing with `Nothing`) and thus rendering it as the very original posting.
Probably we'll need to inject implicit tags and form "transformed" posting on early stages right after parsing the whole transaction.
I.e. after such changes you'll be able to exclude postings without tag by:
```shell
$ hledger bal --pivot desc 'effective:^desc:'   # effective account names should match /^desc:/
$ hledger bal --pivot desc 'effective:tag:^desc$'  # should have effective tag matching /^desc$/
```

Having an empty or `(missing)`  might make sense to identify untagged amount. But this can be done as part of a separate feature. Just update code of `hledger/Hledger/Cli/Utils.hs`:
```haskell
  pivotPosting p
    | Just (_ , value) <- tagTuple = p{paccount = joinAccountNames tag value}
    | _                <- tagTuple = p{paccount = joinAccountNames tag T.empty}  -- instead of just `p`
```

Note that right now "untagged amount" is seen as a split between accounts. Having top-level "account" prefix with a name of tag that allows distinguishing normal accounts and ones introduced by `--pivot`. This may help to identify transactions which needs

simonmichael:
Thoughts: 

I see some use of "effective" in comments here and there (or maybe in the other PR) - I would use "original" everywhere, eg "original account name", for consistency, and because "effective date" is an well known (and confusing) term in *ledger land.

Is always matching on both the pre- and post-pivot, or both the pre- and post-alias, data is the best way ? It seems a bit imprecise, I wonder if it will cause confusing match behaviour somewhere down the road.

I vote to change "desc" to "description", because that's how it's spelled everywhere else. Keeping "payee" is fine, it can be a synonym for now and we might support some convention for separating payee name from other descriptive text in future. (Beancount separates them with |, I sometimes use two spaces).

I've installed this too. Let's test a bit more and try to discover unforeseen interactions.

simonmichael:
I think ordinary (non pivot) tag queries have broken:
```
~/src/hledger$ hledger-1.1 -f examples/sample.journal reg tag:d
~/src/hledger$ hledger -f examples/sample.journal reg tag:d
2008/01/01 income                                  assets:bank:checking                              $1            $1
                                                   income:salary                                    $-1             0
2008/06/01 gift                                    assets:bank:checking                              $1            $1
                                                   income:gifts                                     $-1             0
2008/06/02 save                                    assets:bank:saving                                $1            $1
                                                   assets:bank:checking                             $-1             0
2008/06/03 eat & shop                              expenses:food                                     $1            $1
                                                   expenses:supplies                                 $1            $2
                                                   assets:cash                                      $-2             0
2008/12/31 pay off                                 liabilities:debts                                 $1            $1
                                                   assets:bank:checking                             $-1             0
```

simonmichael:
[Related comments](https://github.com/simonmichael/hledger/pull/472#pullrequestreview-16543370_)

ony:
@simonmichael, true. Change [`73385f2`](https://github.com/simonmichael/hledger/pull/460/commits/73385f2087bd1f11428de82d505b24e258e77594) introduced that. As result, you were able to use implicit tags `desc`, `payee` and `code` in `tag:...` query expression.
Should be fixed in [`cb8426d5`](https://github.com/simonmichael/hledger/pull/460/commits/cb8426d5).

simonmichael:
Nice! Thank you. We can tweak further in trunk.

Issue #99, opened 2013-04-08
generate automated transactions (postings), like ledger


simonmichael:
_Original author: simon@joyful.com (March 04, 2013 21:06:26)_

We'd like to have automated transactions or postings supported in hledger, eg for automatically generating interest postings based on another posting's amount. We have ledger's design (http://ledger-cli.org/3.0/doc/ledger3.html#Automated-Transactions) but other ideas could be considered.

_Original issue: http://code.google.com/p/hledger/issues/detail?id=99_

simonmichael:
I question this. We have hledger-interest which generates explicit transactions, which is more manual but allows much more control and predictability. Needs discussion/developers, closing.

singpolyma:
This would be very useful for compatibility with ledger.  the complexity mostly would come from interpreting the expression language, yes?

simonmichael:
Yes interpreting the expression language, for a start. That alone seems a really big task. Also there's: finding out what that language is (from the C++ code, as neither the docs nor any person knows all of it). Troubleshooting/fixing/documenting the inevitable incompatibilities in the features that have been implemented. Tracking the language as Ledger evolves over time. Troubleshooting/testing/fixing/documenting all the interesting interactions we'll find between automated transactions and existing features.

Arguably it would be useful to at least specify and implement a compatible subset of it. It sounds like it would be useful for some folks, and it would make us seem more Ledger compatible - right ? But most Ledger users that try it will grumble that it doesn't support essential-thing-they-rely on; hledger users will be constantly asking for just one more feature; and it will add complexity and drag to the project which I'm not sure it will justify, even if we manage to attract a dedicated maintainer for it.

Soo.. that's my reaction today, but I'd love to be proven wrong. How interested are you really ?  :-)

simonmichael:
PS, I said "not sure it will justify" partly because I have philosophical/design doubts too, namely that generating "invisible transactions" adds a level of complexity and error potential for users that is completely absent from the current hledger, where everything is visible and explicit, the journal is readable by humans and accountants even without software, and we do everything we can to make things intuitive and minimise avoidable complexity.

All that said, if someone made it, I'd be first in line to try it and figure out how I can best use it. An add-on script might be a good way to start, for preprocessing journals with automated transactions.

simonmichael:
[...]

Issue #442, opened 2016-12-19
print should preserve implicits amounts the way the user wrote them


peti:
When I `print` a journal file with hledger, then the output gives the amount for both sides of the (trivial) transaction like this:

~~~
2016/12/19 Jasmin
    Ausgaben:Restaurant        EUR 13.00
    Vermögen:Bargeld:Peter    EUR -13.00
~~~

If I `print` that same journal with the C++ `ledger`, however, then it omits the amount from the second part of the transaction and prints just:

~~~
2016/12/19 Jasmin
    Ausgaben:Restaurant                    EUR 13.00
    Vermögen:Bargeld:Peter
~~~

Personally, I feel that the second for is neater because it omits redundant information, so now I wonder whether there's a way to make hledger `print` output in the same way?

simonmichael:
I seem to remember some reason for changing it to the current behaviour - making conversion prices explicit or something ? There's no option to change it at the moment.

ony:
I don't think this is a good behavior. I like ledger because it preserves what user write and allows to get report on it.
This feature results in making transaction written by user with explicit amounts more implicit.
I don't mind if `ledger print` will preserve original content. I.e. if user omits number it also omits it in output.

simonmichael:
This I agree with: we shouldn't remove explicitness that the user may have put there intentionally, eg for error checking and readability.

peti:
There is no one true way for `hledger` to behave in this case, IMHO. As far as I am concerned, the behavior should be configurable. My personal preference is for `hledger` to keep the journal the way I wrote it -- i.e. don't insert or remove any amounts. Others may have other preference (i.e. make all posting amounts explicit).

simonmichael:
Following helpful discussion here and on #465 and #hledger, I think the priorities for print are now:

1. don't remove explicitness the user has added
2. print valid journal format by default, adding whatever explicitness needed to do that easily
3. preserve the user's original implicit/explicit style when it does not conflict with 2

simonmichael:
With #466, 1 and 2 should now be covered, and 3 remains open for PRs and design decisions, like when/whether to activate various modes such as:

- add sufficient explicitness for correct output (current default)
- preserve original implicit/explicit entries as far as possible
- make all entries as implicit as possible
- make all entries as explicit as possible

peti:
[...]

PR #471, opened 2017-01-12
Preserve implicit amounts and prices the way user wrote them in output of print command

ony:
See discussion in #442 .
Summary:
- By default `hledger print` do not inject inferred amounts into an output. I.e. keeps balance assignments and transaction balancing postings as is. Also, it doesn't put inferred prices to balance transaction.
- With `hledger print --explicit` we can get the same be haviour we had before this change.


simonmichael:
Nice. I'm going to test this one locally for a bit too. We'll need to mention this in the print docs in the manual.

peti:
One remaining issue is that `EUR 0` gets re-written to `0`. It's probably not a big deal, but IMHO there's an undesirable loss of information. `USD 0` and `EUR 0` are two different things and I'd prefer for `hledger` to distinguish between the two.

ony:
@peti C++ `ledger` also shows only `0`. Looks like both works the same. I'll look into that probably after playing with `hledger rewrite` in hope to resolve #99 some day.

peti:
> C++ `ledger` also shows only 0.

Yes, it does. Still, I think that behavior sucks, especially when the currency is dropped from balance assertions.

peti:
OK, I played around with the new code quite a bit and it works great. I didn't encounter any problems (other than that `EUR 0` thing).

simonmichael:
> Also, it doesn't put inferred prices to balance transaction.

Is this change required for this issue ? Does it not change some reports ?

ony:
@simonmichael , only `print` report affected. I've moved related changes to `Print.hs`. I.e. `showTransaction` is not affected. You can bring back inferred prices with `print --explicit`.
My intention was to keep changes of original input as minimal as possible as long as it doesn't affect interpretation. In the case of inferred prices all needed information already there.

simonmichael:
Thanks for the testing @peti. 0, EUR 0, USD 0 etc. are different internally (this is used by hledger-web's register chart, eg). I don't recall if the cli commands expose that though. Can you paste a minimal example ?

peti:
~~~~
$ cat test.ledger 
2017/01/13 Test
    Foo         EUR 0 = EUR 0
    Bar         USD 0 = USD 0
$ hledger -f test.ledger print
2017/01/13 Test
    Foo             0 = 0
    Bar             0 = 0
~~~~

peti:
This is what it looks like from within `ghci`:

~~~
λ> nulltransaction { tpostings = [nullposting { pamount = mixed [eur 0]}] } 
0000/01/01
                 0
~~~

simonmichael:
Thanks, yes I see now. Commodified zero was added for hledger-web, but all cli commands and probably other tools still treat zero as commodityless, following the original Ledger style. This isn't print-specific and may be a lot of work, let's not tackle it in this issue.

ony:
```diff
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -258,7 +258,7 @@ showPriceDebug (TotalPrice pa) = " @@ " ++ showAmountDebug pa
 -- zero are converted to just \"0\". The special "missing" amount is
 -- displayed as the empty string.
 showAmount :: Amount -> String
-showAmount = showAmountHelper False
+showAmount = showAmountHelper True
 
 showAmountHelper :: Bool -> Amount -> String
 showAmountHelper _ Amount{acommodity="AUTO"} = ""
```

But 19 tests fails (not only `print` related).
Maybe someday there will be a `PrintStyle` that is passed through all `showTransaction` and down.

PR #472, opened 2017-01-13
Pivot tweaks (extends #460, #442)

simonmichael:


hledger-lib/Hledger/Data/Posting.hs
```diff
@@ -171,7 +171,7 @@ postingStatus Posting{pstatus=s, ptransaction=mt}
 -- | Implicit tags for this transaction.
 transactionImplicitTags :: Transaction -> [Tag]
 transactionImplicitTags t = filter (not . T.null . snd) [("code", tcode t)
-                                                        ,("desc", tdescription t)
+                                                        ,("description", tdescription t)
```
ony: The name "desc" were used for consistency with `Query.hs` where you use `desc:SomePayeeRegex`.

simonmichael: Oh! You're right, that's an exception. Arguably a valid one, since that is a query prefix rather than a field name. I think "payee" is the one that will get used far more often, so I'm inclined to not to bother adding "desc" as a special synonym here (or to explore other options like accepting any unique prefix) unless we experience a usability problem. 

We might as well add "note" along with "payee".

PR #477, opened 2017-01-13
Apply ModifierTransaction's from journal during hledger-rewrite.

ony:
Probably the most natural usage would be `hledger rewrite -f rewrites.hledger`. See tests for examples.

Currently no compatibility with C++ `ledger` expressions. Same for posting values. Commodity `*` is used to identify relative amounts to posting.

*Note*: users who migrated from C++ `ledger` probably want to move their automated transactions into a separate file to avoid unintentional rewrites. Maybe we should add option to control application of automated transactions from journal.


simonmichael:
Merging without excessive brain work since this is an add-on. Thanks a lot!

simonmichael:
PS if you can, add any new usage details at least to the file comment, and I can make sure it shows up in docs somewhere.

PR #478, opened 2017-01-13
Prettify option parsing errors

hpdeifel:
Megaparsec's show instance for ParseError doesn't produce digestible
strings. parseErrorPretty has to be used instead.


simonmichael:
Thanks!

PR #484, opened 2017-01-17
rewrite: inherit dates, change application order

ony:
For budgeting it is important to inherit actual date of posting if it
differs from date of transaction. These dates will be added
as a separate line of comment.

More natural order of rewrites is when result of first defined one is
available for all next rewrites.

### DIFF
--- a/bin/hledger-rewrite.hs
+++ b/bin/hledger-rewrite.hs
@@ -5,7 +5,7 @@
   --package megaparsec
   --package text
 -}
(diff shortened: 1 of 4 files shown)


simonmichael:
I guess this one has one commit not yet in master, and needs rebasing ?

ony:
It is in `master` as part of #490 . This can be closed.

simonmichael:
Thanks, closing.

PR #485, opened 2017-01-17
Budgeting addon to simulate behavior of C++ ledger with option --budget

ony:
Until we'll get a better budgeting solution let's borrow one from C++ ledger approach.
- Modification transactions are active like in `hledger-rewrite`. If you like to fill budget buckets/envelopes manually.
- Periodic transactions inject `Budget transaction`s on a range of journal prepending any transaction in original journal. Can be disabled with `--no-offset` when you just need bucketing feature.
- Shrink tree of account to a smaller one according to account names used postings of periodic transactions.

Differences from C++ ledger:
- Posting match expression in modification transaction uses the syntax of queries of command-line.
- Period expression uses the syntax of `--period` option. I had to change keywords to lowercase.
- Order of postings in periodic transactions doesn't matter. Bucketing is done in a bit more natural way in accordance with accounts tree. 

Note that first argument of this addon should always be some subset of a `hledger` commands. Currently supported ones:
- `balance` - obvious.
- `register` - still useful with options like `-p 'monthly in 2017'`.
- `print` - mostly for debugging purposes.
- `balancesheet`, `incomestatement`, `cashflow` - doubtful.

Also be careful with automated transactions because they may result in balance assertion failures. Rules of thumb:
- Modifier transaction should follow same rules that usual transactions follow (i.e. keep balance for real and balanced virtual postings).
- Don't change the balance of real asset and liability accounts for which you usually put assertions. Keep in mind that `hledger` do not apply modification transactions.
- In periodic transactions to offset your budget use either top-level account like `Assets` or introduce a "virtual" one like `Assets:Bank:Budget` that will be a child to the one you want to offset.

P.S. With `--no-offset` and `--no-buckets` it might be considered as a resolution of #99 .

### DIFF
--- a/bin/.gitignore
+++ b/bin/.gitignore
@@ -1,3 +1,4 @@
+hledger-budget
 hledger-chart
 hledger-check-dates
 hledger-dupes
--- a/bin/hledger-budget.hs
+++ b/bin/hledger-budget.hs
@@ -40,11 +40,21 @@ journalBalanceTransactions' opts j = do
 
 withJournalDo' :: CliOpts -> (CliOpts -> Journal -> IO ()) -> IO ()
 withJournalDo' opts = withJournalDo opts . wrapper where
-    wrapper f opts' j = f opts' =<< journalBalanceTransactions' opts' j{ jtxns = ts' } where
+    wrapper f opts' j = do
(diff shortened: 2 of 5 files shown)


ony:
Rebased and added documentation section for `rewrite` addon.

simonmichael:
Great. Thank you!

PR #486, opened 2017-01-17
prices: new addon 

ony:
I've created this one when I had to use pipeline of `hledger-rewrite` and `hledger` to resurrect prices from original journal. I tried this before I've got #485.
Still might be useful to report all know prices. Additionally option `--costs` gives ability to build prices from postings costs.


simonmichael:
Looks good, could you add a usage docs comment as in the other scripts.

PR #490, opened 2017-01-19
Make hledger-rewrite tool suitable for re-factoring original journals

ony:
### What is being proposed, in one sentence ?
New option to `hledger-rewrite` to generate patch suitable for applying on original files.

### What real problem is being solved ?
Right now output of `hledger-rewrite` is just a one big journal. Everything except transaction is stripped. Transactions re-ordered. I guess most people wouldn't use it to rewrite their hand edited ledger journal for these reasons.
New output `hledger-rewrite --diff` produce only diff between an original text and the re-written transactions ignoring everything else. Benefits:
- Diff is suitable for feeding as input for `patch` tool. As result this allows updating some chunks of original files (multiple allowed).
- You can pass output through `colordiff` to see eye appealing demonstration of what re-write is going to perform.

### What is the proposed UI and help, if applicable ?
```
Output:
     --diff                         generate diff suitable as an input for
                                    patch tool
```

### What is a rough draft of the reference documentation, or other details ?
People familiar with `patch` and `diff` unix tools may want to read example from [`tests/bin/rewrite.test`](https://github.com/ony/hledger/blob/feature/rewrite-as-refactor-tool/tests/bin/rewrite.test#L150):
```
# Rewrite using diff output
runghc ../../bin/hledger-rewrite.hs --diff -f- assets:bank and 'amt:<0' --add-posting 'expenses:fee  $5'  --add-posting 'assets:bank  $-5'
<<<
2016/01/01 withdraw
    assets:cash            $20
    assets:bank

2016/01/02 withdraw
    assets:cash
    assets:bank           $-30
>>>
--- -
+++ -
@@ -1,3 +1,5 @@
 2016/01/01 withdraw
     assets:cash            $20
     assets:bank
+    expenses:fee            $5
+    assets:bank            $-5
@@ -5,3 +7,5 @@
 2016/01/02 withdraw
     assets:cash
     assets:bank           $-30
+    expenses:fee            $5
+    assets:bank            $-5
>>>2
>>>=0
```

### What is in and out of scope for this ?
[Allowing modification of original files (multiple) with preserving rest of content through patch tool](https://sketchboard.me/LAiRw203cSmQ#/)

### How will this interact with existing features and systems ?
Output expected to be used for `patch` utility. Thus output should be something like unified diff.
```sh
% hledger-rewrite --diff Agency --add-posting 'Expenses:Taxes  *0.17' | patch    
patching file ledger-2017.dat
% hledger-rewrite | colordiff
...
```

### How will we test or ensure this remains effective ?
- shelltests

### What is an estimate of hours required ?
A prototype is ready already.

### Which of our current Projects does this relate to, if any ?
~~`hledger-rewrite`~~
**Update**: [2017 2. improve real-world usefulness](https://github.com/simonmichael/hledger/projects/3)

### DIFF
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -212,8 +212,8 @@ instance Eq Posting where
     (==) (Posting a1 b1 c1 d1 e1 f1 g1 h1 i1 _ _) (Posting a2 b2 c2 d2 e2 f2 g2 h2 i2 _ _) =  a1==a2 && b1==b2 && c1==c2 && d1==d2 && e1==e2 && f1==f2 && g1==g2 && h1==h2 && i1==i2
 
 -- | The position of parse errors (eg), like parsec's SourcePos but generic.
--- File name, 1-based line number and 1-based column number.
-data GenericSourcePos = GenericSourcePos FilePath Int Int
+data GenericSourcePos = GenericSourcePos FilePath Int Int    -- ^ name, 1-based line number and 1-based column number.
+                      | JournalSourcePos FilePath (Int, Int) -- ^ file name, inclusive range of 1-based line numbers (first, last).
   deriving (Eq, Read, Show, Ord, Data, Generic, Typeable)
 
 instance NFData GenericSourcePos
--- a/bin/hledger-rewrite.hs
+++ b/bin/hledger-rewrite.hs
@@ -4,8 +4,9 @@
   --package hledger
   --package megaparsec
   --package text
+  --package Diff
 -}
-{-# LANGUAGE OverloadedStrings, LambdaCase #-}
+{-# LANGUAGE OverloadedStrings, LambdaCase, DeriveTraversable, ViewPatterns #-}
 {-
 
 hledger-rewrite [PATTERNS] --add-posting "ACCT  AMTEXPR" ...
@@ -40,15 +41,18 @@ TODO:
 
 -}
 
-import Data.Monoid
+import Control.Monad.Writer
+import Data.List (sortOn, foldl')
 import qualified Data.Text as T
 -- hledger lib, cli and cmdargs utils
-import Hledger.Cli
+import Hledger.Cli hiding (outputflags)
 -- more utils for parsing
 -- #if !MIN_VERSION_base(4,8,0)
 -- import Control.Applicative.Compat ((<*))
 -- #endif
+import Text.Printf
 import Text.Megaparsec
+import qualified Data.Algorithm.Diff as D
 import Hledger.Data.AutoTransaction (runModifierTransaction)
 
 cmdmode :: Mode RawOpts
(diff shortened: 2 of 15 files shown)


hledger-lib/Hledger/Data/Types.hs
```diff
@@ -214,6 +214,7 @@ instance Eq Posting where
 -- | The position of parse errors (eg), like parsec's SourcePos but generic.
 -- File name, 1-based line number and 1-based column number.
 data GenericSourcePos = GenericSourcePos FilePath Int Int
+                      | LedgerSourcePos FilePath (Int, Int) -- ^ range of lines in ledger journal
```
simonmichael: Am I understanding this commit rightly: it gives us the option to report a range of lines, rather than a single line and column position, in parse errors ?

I introduced GenericSourcePos in 2015 saying "abstract parsec's SourcePos so as to derive NFData. The NFData instance helps us time things with criterion." Now we use megaparsec, that reason may no longer be valid. Your new usage seems a reason to keep it around though, perhaps with a better name ?

The haddock needs an update.

I'd prefer to use "Journal" instead of "Ledger"; that's the term used elsewhere.

simonmichael:
PS that proposal format worked really well, thanks. Re the last question, I don't know if you were aware, by Projects I meant the high-level project priorities at https://github.com/simonmichael/hledger/projects (I've added a link on the HEP page). So I guess one this PR relates to is "improve real-world usefulness" (yay).

hledger-lib/Hledger/Data/Types.hs
ony: > Am I understanding this commit rightly: it gives us the option to report a range of lines, rather than a single line and column position, in parse errors ?

You right. But referring the range is only the part of the story. Note that I intentionally ignore columns since they'll refer full lines for ledger/journal entries.
It makes a special case of `GenericSourcePos` that valid only for "Ledger" journal (both `hledger` and C++ alternative `ledger`). The reason to distinguish is that we know only how to generate such kind of transaction entries. I.e. we cannot make a good in-line patch for "time log" journals simply because only some subset of all possible `Transaction` entries may be represented in a "time log" entry. Adding an extra posting probably will kick out `Transaction` read from "time log" from that subset.

> Your new usage seems a reason to keep it around though, perhaps with a better name ?

Yes. I don't really like this name. Especially after adding a new ctor.

> The haddock needs an update.

True. I'll update this.

> I'd prefer to use "Journal" instead of "Ledger"; that's the term used elsewhere.

Agree. Will update.

simonmichael: Oh I see. I think "Journal" still works; in this context I think of hledger, ledger and even beancount files as "journal formats", but not time logs or CSV files.

ony:
@simonmichael, thank you for clarification regarding projects. I've added a link to that project.

ony:
Made an attempt to document new features of `hledger-rewrite` command in [09a71e7](https://github.com/simonmichael/hledger/pull/490/commits/09a71e7)

simonmichael:
Thank you!

PR #491, opened 2017-01-19
cli: fix bug in pivot for postings without tag

ony:
Without this fix for postings without tag query checked effective account which is always empty text ("").


simonmichael:
Oops. Thanks!

PR #494, opened 2017-01-25
hledger-ui: add support for brick 0.17 and vty 0.15

peti:
I hope I didn't break backwards compatibility with older versions ... I tried not to. :-)


simonmichael:
Sweet! Thanks.

PR #495, opened 2017-01-25
equity: generate assertions

ony:
Helps to ensure that expectations for these transactions are met. I.e. there is no other postings in between or that an opening transaction is the first one in a journal.


simonmichael:
Seems like a good idea. Thanks!

PR #499, opened 2017-01-28
Escape account names

### LINKED ISSUE #498: Accounts with '?' confound web/ui registers
Accounts with question marks show up in balances, but their transactions are not listed in `hledger ui -- --register` nor `hledger web`.

    1/1
        First?  $1
        Second

    1/2
        Second  $1
        Third

    1/3
        Third  $2
        First?

anim

I suspect the problem is the question mark is not escaped in `accountNameToAccountRegex`, but I'm still compiling the universe and haven't tried it out yet. Also not sure what your preferred solution would be in this case.

(I just noticed this while working on cleaning up some imported data... I don't intend to keep the question marks for long.)

chreekat:
Fixes #498.

### DIFF
--- a/hledger-lib/Hledger/Data/AccountName.hs
+++ b/hledger-lib/Hledger/Data/AccountName.hs
@@ -149,15 +149,22 @@ clipOrEllipsifyAccountName :: Int -> AccountName -> AccountName
 clipOrEllipsifyAccountName 0 = const "..."
 clipOrEllipsifyAccountName n = accountNameFromComponents . take n . accountNameComponents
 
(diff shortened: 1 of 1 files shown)


simonmichael:
Applying the quick fix. Thanks! How robust would you estimate this is.. are there other regex constructs that might break it in future ?

chreekat:
I added all the magic characters I'm aware of! I do think it's going to be
robust enough to let it stand, and it's a fairly benign bug anyway. Perhaps
I should have added a comment to that effect.

PR #501, opened 2017-01-30
rewrite: support for replace and comment

ony:
With `--replace` it is possible to drop original posting from
transaction.

With `--txn-comment` it is possible to add transaction-level comment.

### DIFF
--- a/bin/hledger-rewrite.hs
+++ b/bin/hledger-rewrite.hs
@@ -155,6 +155,10 @@ https://github.com/simonmichael/hledger/issues/99
   |]
   [flagReq ["add-posting"] (\s opts -> Right $ setopt "add-posting" s opts) "'ACCT  AMTEXPR'"
            "add a posting to ACCT, which may be parenthesised. AMTEXPR is either a literal amount, or *N which means the transaction's first matched amount multiplied by N (a decimal number). Two spaces separate ACCT and AMTEXPR."
+  ,flagReq ["txn-comment"] (\s opts -> Right $ setopt "txn-comment" s opts) "'COMMENT'"
+           "add transaction comment."
+  ,flagNone ["replace"] (setboolopt "replace-posting")
+           "replace/delete original posting from transaction."
   ,flagNone ["diff"] (setboolopt "diff") "generate diff suitable as an input for patch tool"
   ]
   [generalflagsgroup1]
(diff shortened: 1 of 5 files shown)


simonmichael:
Cf [rewrite rule ideas](https://gist.github.com/simonmichael/dde46960d7464f655515319a787b264d)

varac:
Nice! What's the current status of this ?

simonmichael:
I'm not sure. Since this PR there was work done on rewriting to support https://github.com/simonmichael/hledger/blob/master/bin/hledger-rewrite.hs and hledger-budget. @ony any idea ?

ony:
@varac, @simonmichael, I can re-base this change. But comments still applicable. I.e. `AutoTransaction.hs` adds complexity to library while being used only in two plugins `hledger-rewrite` and `hledger-budget`.

ony:
Hmm... According to #557 support for automated transaction spills out of `AutoTransaction.hs` already.

@simonmichael, do we still looking for the way to avoid this functionality to be part of `hledger-lib`?

P.S. Locally I get errors regarding empty output in `tests/csv/read.test`. And `bin/compile.sh` that is required for `make functest` doesn't make life of developer easier...

simonmichael:
@ony: I'll think more on this when I have resolved some in-progress issues. For now, let's resolve these test issues you've mentioned.

> P.S. Locally I get errors regarding empty output in tests/csv/read.test. And bin/compile.sh that is required for make functest doesn't make life of developer easier...

Suggestions welcome. Maybe you should open a discussion on the mail list about what to do with these addons generally ?

simonmichael:
> @simonmichael, do we still looking for the way to avoid this functionality to be part of hledger-lib?

IIRC no. I think I was getting used to the idea of having in hledger-lib, with some cleanup/relanguaging at some point maybe.

simonmichael:
I don't see the csv/read.test errors with current master. Any idea on why it's failing for you ?

The experimental addons can now also be built with  "make addons", and that is done by "make functest". Let's see if that helps. It might hurt travis, but I think travis is not running functional tests at the moment.

simonmichael:
@ony please update this PR if you think it is still valuable (and separate the prices: commits which seem to be a different issue).

ony:
I suggest to close this. I didn't use this much. If someone will want this we always can re-open this pull request.

simonmichael:
I see you suggested closing this @ony, will do. If I understand correctly, this added some more features to the hledger-rewrite addon, which is now a built in command. Feel free to reopen if those features become valuable again.

Issue #475, opened 2017-01-13
print should not strip commodity from zero amounts when it is specified in the journal


peti:
`hledger print` (and other commands) should not reduce `EUR 0` to `0`.

[context: [#471](#https://github.com/simonmichael/hledger/pull/471#issuecomment-272439142)]

[note: some discussion below is from @simonmichael thinking this issue was about making all commands show commodified zero]

ony:
I guess this is a big deal and needs fixing...
```ledger
; bug.journal
2017/1/1
    assets
    expenses  $5

2017/1/1
    assets  = $0
    expenses  = $0
```
```sh
% COLUMNS=80 hledger -f bug.journal reg 
2017/01/01                      assets                         $-5           $-5
                                expenses                        $5             0
2017/01/01                      assets                          $5            $5
                                expenses                       $-5             0
```
```sh
% hledger -f bug.journal print | COLUMNS=80 hledger -f - reg           
2017/01/01                      assets                         $-5           $-5
                                expenses                        $5             0
2017/01/01                      assets                           0             0
                                expenses                         0             0
```

simonmichael:
Thanks for the example. Hmm, what exactly is going on here. Your file (1):
```
$ cat examples/475.journal
; bug.journal
2017/1/1
    assets
    expenses  $5

2017/1/1
    assets  = $0
    expenses  = $0
```
The expected balance report (2):
```
~/src/hledger$ hledger -f examples/475.journal bal -E
                   0  assets
                   0  expenses
--------------------
                   0
```
Output of print --explicit (3):
```
~/src/hledger$ hledger -f examples/475.journal print --explicit
2017/01/01
    assets             $-5
    expenses            $5

2017/01/01
    assets              $5 = 0
    expenses           $-5 = 0

```
The balance report is unchanged, as expected (4):
```
~/src/hledger$ hledger -f examples/475.journal print --explicit | hledger -f- bal -E
                   0  assets
                   0  expenses
--------------------
                   0
```
Ordinary style-preserving print (5):
```
~/src/hledger$ hledger -f examples/475.journal print
2017/01/01
    assets
    expenses            $5

2017/01/01
    assets     = 0
    expenses   = 0

```
That changes the balance report to something different (6):
```
~/src/hledger$ hledger -f examples/475.journal print | hledger -f- bal -E
                 $-5  assets
                  $5  expenses
--------------------
                   0
```
[...]

PR #505, opened 2017-02-03
Added a simple 'showMarketPrice' function to show market price directives in a journal-compatible way

mstksg:
Not too big a deal!  Not sure if it's worth an entire module.  Tried writing tests (parse and re-show), but couldn't get it to work cleanly without cyclic imports.

By the way, I noticed that hledger parses market price directives with "full" datetimes, but throws away the times.   Would it be worthwhile to modify the `MarketPrice` type to also include a full datetime, potentially, as well, if it was given one?

### DIFF
--- a/hledger-lib/Hledger/Data/MarketPrice.hs
+++ b/hledger-lib/Hledger/Data/MarketPrice.hs
@@ -12,18 +12,22 @@ are thousands separated by comma, significant decimal places and so on.
 
 -}
 
-{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings, LambdaCase #-}
 
 module Hledger.Data.MarketPrice
 where
 import qualified Data.Text as T
--- import Test.HUnit
+import Test.HUnit
 
 import Hledger.Data.Amount
-import Hledger.Data.Types
 import Hledger.Data.Dates
+import Hledger.Data.Types
+-- import Hledger.Read.JournalReader
 import Hledger.Utils
+import Hledger.Utils.Parse
 
+-- | Get the string representation of an market price, based on its
+-- commodity's display settings.
 showMarketPrice :: MarketPrice -> String
 showMarketPrice mp = unwords
     [ "P"
@@ -31,3 +35,12 @@ showMarketPrice mp = unwords
     , T.unpack (mpcommodity mp)
     , (showAmount . setAmountPrecision maxprecision) (mpamount mp)
     ]
+
+tests_Hledger_Data_MarketPrice = TestList $
+  [
+    -- this test needs Hledger.Read.JournalReader, which causes cyclic imports?
+    -- "showParsedMarketPrice" ~: do
+    --   let mp = parseWithState mempty marketpricedirectivep "P 2017/01/30 BTC $922.83"
+    --       mpString = (fmap . fmap) showMarketPrice mp
+    --   mpString `is` (Just (Right "P 2017/01/30 BTC $922.83"))
+  ]
--- a/hledger-lib/Hledger/Data/MarketPrice.hs
+++ b/hledger-lib/Hledger/Data/MarketPrice.hs
@@ -16,18 +16,18 @@ are thousands separated by comma, significant decimal places and so on.
 
 module Hledger.Data.MarketPrice
 where
-import Data.Time.Format
 import qualified Data.Text as T
 -- import Test.HUnit
 
 import Hledger.Data.Amount
 import Hledger.Data.Types
+import Hledger.Data.Dates
 import Hledger.Utils
 
 showMarketPrice :: MarketPrice -> String
 showMarketPrice mp = unwords
     [ "P"
-    , formatTime defaultTimeLocale "%0Y/%m/%d" (mpdate mp)
+    , showDate (mpdate mp)
     , T.unpack (mpcommodity mp)
     , (showAmount . setAmountPrecision maxprecision) (mpamount mp)
     ]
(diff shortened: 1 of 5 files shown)


hledger-lib/Hledger/Data/MarketPrice.hs
```diff
@@ -0,0 +1,33 @@
+{-|
+
+A 'MarketPrice' represents a historical exchange rate between two
+commodities. (Ledger calls them historical prices.) For example, prices
+published by a stock exchange or the foreign exchange market.  Some
+commands (balance, currently) can use this information to show the market
+value of things at a given date.
+
+thing we are tracking, and some display preferences that tell how to
```
simonmichael: comment looks a little broken here ?

hledger-lib/hledger-lib.cabal
```diff
@@ -110,6 +110,7 @@ library
       Hledger.Data.Dates
       Hledger.Data.Journal
       Hledger.Data.Ledger
+      Hledger.Data.MarketPrice
```
simonmichael: As this is generated from package.yaml, add the module there too. This will probably fix the Travis error.

hledger-lib/Hledger/Data/MarketPrice.hs
```diff
@@ -12,22 +12,35 @@ are thousands separated by comma, significant decimal places and so on.
...
 showMarketPrice mp = unwords
     [ "P"
     , showDate (mpdate mp)
     , T.unpack (mpcommodity mp)
     , (showAmount . setAmountPrecision maxprecision) (mpamount mp)
     ]
+
+tests_Hledger_Data_MarketPrice = TestList $
+  [
+    -- this test needs Hledger.Read.JournalReader, which causes cyclic imports?
```
simonmichael: Could move this test to JournalReader or Hledger.Read perhaps ? I've done that with others

hledger-lib/hledger-lib.cabal
simonmichael: PS, I think the new show function is not used yet ? Maybe you plan to use it for print [--explicit] ? Ultimately we want the ability to print any and all directives. There are some [questions to resolve](https://github.com/simonmichael/hledger/issues/475#issuecomment-274340537).

mstksg:
new fixes seem to pass on travis :)  and the new test in `Hledger.Read.JournalReader` seems to pass as well.

hledger-lib/hledger-lib.cabal
mstksg: new show function is indeed not yet used within `hledger` itself.  But I am using it directly in some of my own utility scripts (https://github.com/mstksg/jle-utils/blob/master/app/UpdatePrices.hs#L75).  Looking into being able to have parse+print being a non-lossy process is a good next step that this show function could hopefully be used for!

simonmichael:
Ok looks good. And UpdatePrices looks useful, hope you make it a first-class addon . Thanks!

hledger-lib/Hledger/Data/MarketPrice.hs
```diff
@@ -12,22 +12,35 @@ are thousands separated by comma, significant decimal places and so on.
...
+import Hledger.Utils.Parse
 
+-- | Get the string representation of an market price, based on its
+-- commodity's display settings.
 showMarketPrice :: MarketPrice -> String
 showMarketPrice mp = unwords
     [ "P"
     , showDate (mpdate mp)
     , T.unpack (mpcommodity mp)
     , (showAmount . setAmountPrecision maxprecision) (mpamount mp)
```
ony: Why do you need to extend precision? Each `Amount` remembers original precision with which it was parsed.

If you calculate `Amount` from somewhere you should be responsible for setting proper `AmountStyle`. I don't think it is good idea to force precision.

hledger-lib/Hledger/Data/MarketPrice.hs
```diff
@@ -12,22 +12,35 @@ are thousands separated by comma, significant decimal places and so on.
...
     ]
+
+tests_Hledger_Data_MarketPrice = TestList $
+  [
+    -- this test needs Hledger.Read.JournalReader, which causes cyclic imports?
+    -- "showParsedMarketPrice" ~: do
+    --   let mp = parseWithState mempty marketpricedirectivep "P 2017/01/30 BTC $922.83"
+    --       mpString = (fmap . fmap) showMarketPrice mp
+    --   mpString `is` (Just (Right "P 2017/01/30 BTC $922.83"))
```
ony: Any reason for commenting out test?

P.S. For such a simple case you may want to look at `doctest` as an alternative.

hledger-lib/hledger-lib.cabal
ony: `bin/hledger-prices.hs` may benefit from this module.

hledger-lib/Hledger/Data/MarketPrice.hs
mstksg: It might be a bug with amount parsing then; I had a file with a currency given in small fractions of USD:

~~~
P 2016/07/01 VND $0.000045
~~~

And, without the precision forcing, it got printed out as

~~~
P 2016/07/01 VND $0.00
~~~

Might investigate this to see if the problem is actually the precision parsing.  I am also bothered by the precision forcing, which feels like an ugly hack that could break a lot of situations.

ony:
```
zsh% hledger prices -f -
P 2016/07/01 VND $0.000045
P 2016-07-01 VND $0.000045
```
But with:

```
zsh% hledger prices -f -
commodity $
    format $1,000,000,000.00
P 2016/07/01 VND $0.000045
P 2016-07-01 VND $0.00
```

That's a bug.`commodity` directive shouldn't affect prices. This may result in truncating prices in output of `hledger print`

mstksg: I see the issue now.  Not sure how to best resolve this.  Would this be a bug in `showAmount`, which truncates prices when it shouldn't be?

PR #506, opened 2017-02-04
Refactoring balancesheet/incomestatement/cashflow, bringing to feature parity with 'balance'

mstksg:
**in progress**

As mentioned in https://github.com/simonmichael/hledger/pull/361, but never followed up.

I've successfully refactored/abstracted out `balancesheet`, `incomestatement`, and `cashflow`  under a unified wrapper.  From there I'm looking to bring things to some sort of feature parity with `balance` to the extent of what is relevant -- looks like it's just missing `-V`, `-N`, `--no-elide`, and `--format`, anyways.

Briefly thought about even including `balance` as a part of the abstraction, but not sure how to meaningfully handle the "table" views for `balancesheet` and `incomestatement`.  Maybe just print out two tables, one after the other?

### DIFF
--- a/hledger/Hledger/Cli/BalanceView.hs
+++ b/hledger/Hledger/Cli/BalanceView.hs
@@ -6,6 +6,7 @@ module Hledger.Cli.BalanceView (
  ,balanceviewReport
 ) where
 
+import Control.Monad (unless)
 import Data.Time.Calendar (Day)
 import Data.List (intercalate)
 import Data.Monoid (Sum(..), (<>))
@@ -29,6 +30,10 @@ balanceviewmode bv@BV{..} = (defCommandMode $ bvmode : bvaliases) {
      groupUnnamed = [
       flagNone ["flat"] (\opts -> setboolopt "flat" opts) "show accounts as a list"
      ,flagReq  ["drop"] (\s opts -> Right $ setopt "drop" s opts) "N" "flat mode: omit N leading account name parts"
+     ,flagNone ["value","V"] (setboolopt "value") "convert amounts to their market value on the report end date (using the most recent applicable market price, if any)"
+     ,flagNone ["no-total","N"] (\opts -> setboolopt "no-total" opts) "omit the final total row"
+     ,flagNone ["no-elide"] (\opts -> setboolopt "no-elide" opts) "don't squash boring parent accounts (in tree mode)"
+     ,flagReq  ["format"] (\s opts -> Right $ setopt "format" s opts) "FORMATSTR" "use this custom line format (in simple reports)"
      ]
     ,groupHidden = []
     ,groupNamed = [generalflagsgroup1]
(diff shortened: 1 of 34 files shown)


mstksg:
Alright, took a little shorter than I thought.  But as discussed in #361, I've turned "-V"/"--value" into a global report option, so it should work for `balance`, `register`, `print`, `balancesheet`, `cashflow`, `incomestatement`, etc.  All tests in the test suite pass.

Might have to rearrange a bit in the documentation to reflect this.  But I think this is a consistent way to do this, considering that `-B`/`--cost` is already a global option ... however, not sure if this deviates from the official `ledger` spec in undesired ways.

mstksg:
I've edited the relevant documentation that I can see, editing the `.m4` and `m4.md` files.  I assume that some build system is used to compile and render them, but I'm not sure how to test it and see the effects myself?  Is there a way for me to build the manpages/info/help files locally?

hledger/Hledger/Cli/BalanceView.hs
```diff
@@ -0,0 +1,75 @@
...
+import Data.Time.Calendar (Day)
+import Data.List (intercalate)
+import Data.Monoid (Sum(..), (<>))
+import System.Console.CmdArgs.Explicit
+
+import Hledger
+import Hledger.Cli.Balance
+import Hledger.Cli.CliOptions
+
+data BalanceView = BV { bvmode    :: String
```
simonmichael: Let's add explanatory haddocks for this type and functions.

I wouldn't mind using the full BalanceView name as constructor, in the interest of keeping the code clear and obvious for contributors. It's not used that often.

simonmichael:
Sounds great! To view docs, I do `./Shake.hs && ./Shake all && make site-preview`.

hledger/Hledger/Cli/BalanceView.hs
simonmichael: A few compile warnings remain, I think

mstksg:
Added haddocks for `BalanceView` type and accessors.   I actually am not using `MutliBalanceReport`; I pretty much actually just pulled out the repeated parts of the original commands and abstracted out their parameters, so they don't render any tables.  Will look into moving things into `MultiBalanceReport`, which looks like it would bring things into feature completeness with `balance`.

simonmichael:
Great, thank you.

mstksg:
By the way, running `./Shake.hs`, i get some errors with stack saying 'invalid command: exec'; after removing the shebang pattern and running directly with stack, I'm getting some hakyll build errors?  hakyll is failing to build `site/hakyll-std/hakyll-std`.

simonmichael:
Is your stack really old ? 1.3+ recommended.

I guess hakyll-std deps are not getting fully installed, try running "site/hakyll-std/hakyll-std.hs" once to fix that. Feel free to update the resolver in hakyll-std.hs, using a nightly may no longer be needed.

simonmichael:
> Feel free to update the resolver in hakyll-std.hs, using a nightly may no longer be needed.

Correction, it is, as hakyll and pandoc are still not in lts 7.

mstksg:
looks like the problem was just me using an old version of *stack*!  Everything seems to work now.  Pushed the re-rendered documentation.

simonmichael:
I got a little trigger-happy.. wanted to see if I could push my earlier separate fix for MarketPrice in hledger-lib.cabal and then rebase a cleaner version of this. Seems to have worked. Thank you! Adding multi-column support will be most welcome if you get around to it.

mstksg:
No problem!  Good to hear that everything works!

simonmichael:
print -V looks a little weird.
```
2017/1/1 opening balances
  (assets:depot)  $3000

2017/1/2 buy shares at $200
  ; let's assume no fees
  assets:shares   10 TSLA @ $200
  assets:depot

; market price, has jumped since yesterday's purchase
P 2017/1/3 TSLA $250
```

Old print behaviour:
```
$ hledger -f t.j print
2017/01/01 opening balances
    (assets:depot)         $3000

2017/01/02 buy shares at $200
    ; let's assume no fees
    assets:shares  10 TSLA @ $200
    assets:depot

$ hledger -f t.j print -B
2017/01/01 opening balances
    (assets:depot)         $3000

2017/01/02 buy shares at $200
    ; let's assume no fees
    assets:shares         $2000
    assets:depot
```

New -V support:
```
$ hledger -f t.j print -V
2017/01/01 opening balances
    (assets:depot)         $3000

2017/01/02 buy shares at $200
    ; let's assume no fees
    assets:shares  $2500 @ $200    ; <- should be... just $2500 without the price I guess
    assets:depot
```

mstksg:
Hm, yes, that's a little weird.  Are you convinced that

        assets:shares        $2500

is the desired behavior here?  If so we can probably go about to make that happen.

simonmichael:
It will probably need a bit more thought. I'm realising that the current -V behaviour is too simple for reports which show a span of time - print, register, multicolumn balance - since it converts everything using the price as of a single date, the report end date.

simonmichael:
Here's what Ledger does (notes the price change, does not change transaction amounts):
```
$ ledger print -f examples/investment2.journal  -V
2017/01/01 opening balances
    (assets:depot)                             $3000

2017/01/02 buy shares at $200
    ; let's assume no fees
    assets:shares                            10 TSLA @ $200
    assets:depot

2017/01/03 Commodities revalued
```

simonmichael:
And register, just for completeness:
```
$ ledger reg -f examples/investment2.journal  -V
17-Jan-01 opening balances       (assets:depot)                     $3000         $3000
17-Jan-02 buy shares at $200     assets:shares                      $2000         $5000
                                 assets:depot                      $-2000         $3000
17-Jan-03 Commodities revalued   <Revalued>                          $500         $3500
```

PR #511, opened 2017-02-11
Add missing assrt param to documentation.

srbaker:
I was working through the documentation, and this wouldn't work as-is. I noticed that there is an additional argument `assrt` which is not used here. Passing `True` causes this example to work.


simonmichael:
Thanks! That is a rather old doc comment. It should perhaps be moved somewhere better, or made into a doctest that we can automatically test.

PR #512, opened 2017-03-08
Use more accurate license tag in Cabal file.

peti:


peti:
Those changes come from recent versions of `hpack`. I just ran that tool; I didn't edit the Cabal files myself.

simonmichael:
I see, `./.` seems to be a workaround for https://github.com/sol/hpack/issues/119. 

Next, are we sure `GPL3` is valid ? I remember looking for it in the past, and it's not mentioned in the [Cabal docs](https://www.haskell.org/cabal/users-guide/developing-packages.html#package-properties) -> license -> [License type](https://www.haskell.org/cabal/release/cabal-latest/doc/API/Cabal/Distribution-License.html#t:License).

simonmichael:
The `./.` change happens with hpack 0.16 or greater. My stack binary has hpack 0.14 built in (and building a new stack doesn't seem to change that). This can cause hpack warnings and thrashing between the `.` and `./.` formats. Are you not using stack ?

peti:
> Next, are we sure GPL-3 is valid? 

Yes.

> I remember looking for it in the past, and it's not mentioned in the Cabal docs -> license -> License type.

Note that the GPL constructor of `License` has a `Version` argument. That is where the `-3` bit ends up.

> Building a new stack doesn't seem to change [the hpack version].

`stack` can be built with recent versions of `hpack`. You might be able to force that by passing `--constraint=hpack==0.17.*` on the command-line to `cabal` (or whatever it is, exactly, that you use to build `stack`). It probably doesn't happen by default because your package database contains version 0.14 of the tool already.

> Are you not using stack?

No. I used it for a while, but then I went back to plain `cabal-install` (with Nix behind it to provide pre-built binaries if possible).

simonmichael:
> Note that the GPL constructor of License has a Version argument. That is where the -3 bit ends up.
> 
Ah! Thank you.
> 
> stack can be built with recent versions of hpack. You might be able to force that by passing --constraint=hpack==0.17.* on the command-line to cabal (or whatever it is, exactly, that you use to build stack). It probably doesn't happen by default because your package database contains version 0.14 of the tool already.
> 
No, I've removed hpack and/or installed 0.17.0 manually, but it's not that easy to build a stack that uses it - seems to require checking out stack HEAD. I don't want hledger installers to have to do that to avoid hpack warnings, so let's postpone/deal with that separately. Would you mind resubmitting with just the GPL-3 change.

peti:
According to http://hackage.haskell.org/package/stack, `stack` requires `hpack >=0.14.0 && <0.18`, so I don't see why building it with a recent version should be a problem. We certainly do build `stack` with a recent `hpack` in NixOS and it works fine.

simonmichael:
For non-cabal users, installing a stack that uses hpack 0.17 is not yet that easy, because of other dependency issues. I'd prefer to separate these issues.

peti:
I'm not sure what real-world action you associate with "separating these issues"? How do we move forward from here?

simonmichael:
I meant, can you (force) push a new commit to this PR which just changes the license field. That makes it easy to accept.

simonmichael:
PS I'm sorry for the tedious push back on what you probably thought was a trivial fix. :) I can do it if you prefer. Also I clarified my earlier comment a bit at https://github.com/simonmichael/hledger/pull/512#issuecomment-285597496

simonmichael:
Merged https://github.com/simonmichael/hledger/commit/e2c8a6ae42d1f8f35b631757e2207bd183cea3e6. Thanks!

Issue #514, opened 2017-03-15
incomestatement and cashflow reports incorrectly include historical amounts


joshkehn:
Pull #506 appears to have caused a regression when rendering `is` and `cf` reports as they now include historical amounts outside of the reporting period.

***

From [mailing list thread](https://groups.google.com/forum/#!topic/hledger/34ZRe4j5xUM). 

Given a ledger [is.txt](https://github.com/simonmichael/hledger/files/845754/is.txt) and version `1.1.98` of hledger (built from 7b4e11ba), the `is` and `cf` reports incorrectly include historical data outside of the reporting period. This bug is not present on version `1.1` of hledger (from homebrew).

Balance correctly calculates for each year:

```
: hledger -f is.l bal -e 2017 -Y -T -N
Balance changes in 2015/01/01-2016/12/31:


                   ||        2015    2016       Total
===================++=================================
 assets:checking   ||  $10,000.00 $-40.00   $9,960.00
 expense:hosting   ||           0  $50.00      $50.00
 revenue:clients:A || $-10,000.00       0 $-10,000.00
 revenue:clients:B ||           0 $-10.00     $-10.00
```

`incomestatement` is incorrect:

```
: hledger -f is.l is -b 2016 -e 2017
Income Statement


Revenues:
         $-10,010.00  revenue:clients
         $-10,000.00    A
             $-10.00    B
--------------------
         $-10,010.00


Expenses:
              $50.00  expense:hosting
--------------------
              $50.00


Total:
--------------------
          $-9,960.00
```

simonmichael:
That's not good! Great catch! @mstksg would you be able to take a look ?

mstksg:
I'll take a look!  I wonder if the bug exists in balancesheet too, and if it was a problem in my initial generalization process or a problem in applying the abstraction to each individual command.

joshkehn:
@mstksg That's a good question. Since `balancesheet` is a historical point-in-time report I don't think this affects it.

mstksg:
Ah I see.  Should the behavior for bs, is, and cf be different then in regards to usage with -b, -e ?

simonmichael:
> Ah I see. Should the behavior for bs, is, and cf be different then in regards to usage with -b, -e ?

I think so. incomestatement and cashflow are period reports, but balancesheet is a snapshot report and should probably ignore a start date (like balance --historical).

joshkehn:
[...]

PR #515, opened 2017-03-18
incomestatement / cashflow behavior bugfix

mstksg:
for #514: is/cf are now properly period reports, and bs is a snapshot report.  period/snapshot mode is now a configuration flag for BalanceView

### DIFF
--- a/hledger/Hledger/Cli/BalanceView.hs
+++ b/hledger/Hledger/Cli/BalanceView.hs
@@ -15,7 +15,6 @@ module Hledger.Cli.BalanceView (
 ) where
 
 import Control.Monad (unless)
-import Data.Time.Calendar (Day)
 import Data.List (intercalate)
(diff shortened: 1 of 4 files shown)


mstksg:
In the issue, there was some talk about emitting a warning for `-b` when called with `balancesheet` -- is this something worth pursuing, do you think?  `-b` is handled as a global argument, so we'd have to fiddle with the command argument library to be able to get special behavior for this I think?

simonmichael:
Great, thanks for the fix.

I'm not too concerned about silently ignoring -b for snapshot reports - I think it's pretty harmless and intuitive. I agree warning about it would be a nice touch.

simonmichael:
PS minor, but if you do any more on this there's a warning in BalanceView.hs

mstksg:
Ah, didn't notice it before because 'stack install' seemed to eat it for some reason.  Looks like the latest commit should fix it?

PR #516, opened 2017-03-20
Add balance/cashflow/incomestatement shelltest

joshkehn:
These pass the current version of hledger available on homebrew and
cover the bug reported in #514 and fixed in #515.


simonmichael:
Thank you!

PR #518, opened 2017-03-21
Multicolumn reports for bs/cf/is, and -T/-A support

mstksg:
This format seemed the most natural to me, but let me know if this makes sense.

~~~
2017/01/01 * Paycheck
    assets:bank:checking             $6000
    assets:bank:savings              $4000
    income:work

2017/01/02 * Ice Cream
    expenses:dessert        $5.50
    liabilities:visa

2017/02/01 * Paycheck 2
    assets:bank:checking             $3000
    assets:bank:savings              $2000
    income:work

2017/02/02 * Fancy Dinner
    expenses:food                 $20.00
    liabilities:american express
~~~

And:

~~~
$ hledger balancesheet -M
Balance Sheet

                              ||  2017/01/31  2017/02/28
==============================++=========================
 Assets                       ||
------------------------------++-------------------------
 assets:bank:checking         ||    $6000.00    $9000.00
 assets:bank:savings          ||    $4000.00    $6000.00
------------------------------++-------------------------
                              ||   $10000.00   $15000.00
==============================++=========================
 Liabilities                  ||
------------------------------++-------------------------
 liabilities:american express ||           0     $-20.00
 liabilities:visa             ||      $-5.50      $-5.50
------------------------------++-------------------------
                              ||      $-5.50     $-25.50
==============================++=========================
 Total                        ||    $9994.50   $14974.50
~~~

with -T and -A

~~~
$ hledger balancesheet -MTA
Balance Sheet

                              ||  2017/01/31  2017/02/28     Total   Average
==============================++=============================================
 Assets                       ||
------------------------------++---------------------------------------------
 assets:bank:checking         ||    $6000.00    $9000.00 $15000.00  $7500.00
 assets:bank:savings          ||    $4000.00    $6000.00 $10000.00  $5000.00
------------------------------++---------------------------------------------
                              ||   $10000.00   $15000.00 $25000.00 $12500.00
==============================++=============================================
 Liabilities                  ||
------------------------------++---------------------------------------------
 liabilities:american express ||           0     $-20.00   $-20.00   $-10.00
 liabilities:visa             ||      $-5.50      $-5.50   $-11.00    $-5.50
------------------------------++---------------------------------------------
                              ||      $-5.50     $-25.50   $-31.00   $-15.50
==============================++=============================================
 Total                        ||    $9994.50   $14974.50 $24969.00
~~~

(Hopefully) meaningful behavior for -N

~~~
$ hledger balancesheet -MN
Balance Sheet

                              ||  2017/01/31  2017/02/28
==============================++=========================
 Assets                       ||
------------------------------++-------------------------
 assets:bank:checking         ||    $6000.00    $9000.00
 assets:bank:savings          ||    $4000.00    $6000.00
------------------------------++-------------------------
                              ||   $10000.00   $15000.00
==============================++=========================
 Liabilities                  ||
------------------------------++-------------------------
 liabilities:american express ||           0     $-20.00
 liabilities:visa             ||      $-5.50      $-5.50
------------------------------++-------------------------
                              ||      $-5.50     $-25.50
~~~

### DIFF
--- a/hledger/Hledger/Cli/Balance.hs
+++ b/hledger/Hledger/Cli/Balance.hs
@@ -242,7 +242,6 @@ module Hledger.Cli.Balance (
  ,multiBalanceReportAsText
  ,renderBalanceReportTable
  ,balanceReportAsTable
- ,alignBalanceReportTable
  ,tests_Hledger_Cli_Balance
 ) where
 
@@ -492,14 +491,19 @@ multiBalanceReportAsText opts r =
 renderBalanceReportTable :: Table String String MixedAmount -> String
 renderBalanceReportTable = unlines . trimborder . lines
                          . render id (" " ++) showMixedAmountOneLineWithoutPrice
+                         . align
   where
     trimborder = ("":) . (++[""]) . drop 1 . init . map (drop 1 . init)
+    align (Table l t d) = Table l' t d
+      where
+        acctswidth = maximum' $ map (strWidth . unwords . words) (headerContents l)
+        l'         = padRightWide acctswidth . unwords . words <$> l
 
 -- | Build a 'Table' from a multi-column balance report.
 balanceReportAsTable :: ReportOpts -> MultiBalanceReport -> Table String String MixedAmount
 balanceReportAsTable opts r@(MultiBalanceReport (colspans, items, (coltotals,tot,avg))) =
    addtotalrow $ Table
-     (T.Group NoLine $ map (Header . padRightWide acctswidth) accts)
+     (T.Group NoLine $ map Header accts)
      (T.Group NoLine $ map Header colheadings)
      (map rowvals items)
   where
@@ -513,7 +517,6 @@ balanceReportAsTable opts r@(MultiBalanceReport (colspans, items, (coltotals,tot
     renderacct (a,a',i,_,_,_)
       | tree_ opts = replicate ((i-1)*2) ' ' ++ T.unpack a'
       | otherwise  = T.unpack $ maybeAccountNameDrop opts a
-    acctswidth = maximum' $ map strWidth accts
     rowvals (_,_,_,as,rowtot,rowavg) = as
                              ++ (if row_total_ opts then [rowtot] else [])
                              ++ (if average_ opts then [rowavg] else [])
@@ -523,12 +526,6 @@ balanceReportAsTable opts r@(MultiBalanceReport (colspans, items, (coltotals,tot
                                     ++ (if row_total_ opts then [tot] else [])
                                     ++ (if average_ opts then [avg] else [])
                                     ))
-
-alignBalanceReportTable :: Table String a b -> Table String a b
-alignBalanceReportTable (Table l t d) = Table l' t d
-  where
-    acctswidth = maximum' $ map (strWidth . unwords . words) (headerContents l)
-    l'         = padRightWide acctswidth . unwords . words <$> l
 
 -- | Figure out the overall date span of a multicolumn balance report.
 multiBalanceReportSpan :: MultiBalanceReport -> DateSpan
--- a/hledger/Hledger/Cli/Balance.hs
+++ b/hledger/Hledger/Cli/Balance.hs
@@ -474,9 +474,9 @@ multiBalanceReportAsCsv opts (MultiBalanceReport (colspans, items, (coltotals,to
 -- | Render a multi-column balance report as plain text suitable for console output.
 multiBalanceReportAsText :: ReportOpts -> MultiBalanceReport -> String
 multiBalanceReportAsText opts r =
-    unlines [ printf "%s in %s:" typeStr (showDateSpan $ multiBalanceReportSpan r)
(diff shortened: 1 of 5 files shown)


mstksg:
Ah, so at the moment, `incomestatement`/`cashflow` force the mode to be `PeriodChange`, and `balancestatement` forces the mode to be `HistoricalBalance`, giving sort of interesting behavior.

Maybe I should have these be configurable as a flag, and having those current modes as being defaults instead.

mstksg:
New update should allow manual user override of cumulative/period/change

simonmichael:
Looking good!

> Ah, so at the moment, incomestatement/cashflow force the mode to be PeriodChange, and balancestatement forces the mode to be HistoricalBalance, giving sort of interesting behavior.

I think it's kind of the point of these commands to be more conventional and less flexible than the balance command. 

Do you think it's useful to be able to switch them between period/cumulative/historical modes ? What do you mean by interesting behaviour ? Also I think -T and -A aren't really meaningful for balancesheet/snapshot reports.

mstksg:
Sorry, interesting was a bad word choice; I meant potentially unexpected.

I've found it useful to switch between `--historical` and `--changes -T` for balancesheet when I want to see how much things have changed per-week.  I can infer it from `--historical`, but `--changes -T` makes it a little more easy to see at a glance.

I agree that `-A` and `-T` aren't meaningful for snapshot reports, but the same issue also arises for `bal --historical`.  The only "improvement" I can imagine would be to give an error on or silently ignore `-A`/`-T` only when used with `--historical`?

simonmichael:
With the two reports we are using, there can be a difference in which columns have data, eg `hledger -f examples/sample.journal bs -M`. Enforcing -E seems to fix it, though I don't know if that is the right fix in all cases.

There are a few warnings.

simonmichael:
Lots of functional tests (`make functest`) are failing due to extra whitespace, I think.

mstksg:
Ah yes I see the undesired behavior with periods that have empty columns for certain periods at the beginning or end.  I'll work on a fix for that in the coming days.  `-E` works, though; perhaps that's the simplest solution?  Is there any reason where `-E` would be undesirable behavior for a multi column balance view?

simonmichael:
Forcing -E (as we do for register -A) sounds reasonable until we find out otherwise. I am thinking of some case where it extends the report dates, to specified start/end date or journal start/end date or something.

simonmichael:
Thanks!

mstksg:
heads up, running without -E still shows empty rows.  in the near future I might submit a patch fixing this.

simonmichael:
Also, in my haste I forgot to ask about docs. We should update those.

simonmichael:
More backlog items: 
- with -A, no value is shown in the Average column for the Total row
- the Total row is redundant for the cashflow report

simonmichael:
- `hledger incomestatement -M [--cumulative|--historical]` shows the end balances of income statement accounts, but is not an actual income statement, so the report title is misleading. Should we show titles like Income Statement,  Income Statement Cumulative Ending Balance, Income Statement Historical Ending Balance ?
- `hledger -f examples/sample.journal bal checking date:2008/12 --historical` shows checking's correct historical ending balance of $0, but `hledger -f examples/sample.journal bs date:2008/12 --historical` does not. The multicolumn version (eg with `-M`) gets it right.

simonmichael:
- a nitpicky thing, for the record: --change for is and cf, and --historical for bs, are the same as the default report except they add the "Balance Changes"/"Historical Ending Balances" subtitles, which are maybe redundant and better omitted.

simonmichael:
- also, I can't quite get my head around bs --cumulative, I think that one will never be meaningful or non-confusing - am I wrong ? Not sure if there's anything to be done about it, even so.

mstksg:
We can probably get rid of the extra subtitle when the flags are redundant; should be straightforward.

I suppose regarding --cumulative, we can parameterize each command by default and allowed balance types, as well.  Would it be safe to say that all three balance types make sense is/cf, or are there any ones that are never meaningful?

simonmichael:
Re meaningless report/mode combinations - I feared the modes combined in a bad way, but they don't. `bs -M --change` is similar to `cf -M`, `bs -M --historical` is the same as `bs -M`, etc. It's fine.

@mstksg do you have any tests/changes coming for 1.2 ? It's getting close.

mstksg:
It looks like in the course of writing tests I found some unexpected behavior in the case where there are no accounts in a given category.  Trying to figure out a nice way to resolve this.

simonmichael:
Consolidating some notes on this first PR where I'll be able to find them.

- cashflow totals are still a little weird: single-column cashflow report shows 2 or 0 total rows depending on -N (should show 1 or 0 I guess); multi-column cashflow report shows 1 total row regardless of -N (should show 1 or 0).

- --flat has no effect on a single-column balancesheet report

simonmichael:
- boring parent accounts aren't elided properly in balancesheet report (#534)

PR #525, opened 2017-03-26
Ongoing fixes to bs/cf/is reports

mstksg:
Ongoing fixes (and documentation) to issues mentioned in the final messages of PR #518

### DIFF
--- a/hledger/Hledger/Cli/BalanceView.hs
+++ b/hledger/Hledger/Cli/BalanceView.hs
@@ -121,15 +121,21 @@ balanceviewReport BalanceView{..} CliOpts{reportopts_=ropts, rawopts_=raw} j = d
             sumAmts = case amts of
               a1:as -> foldl' (zipWith (+)) a1 as
               []    -> []
+            totavg  = totsum `divideMixedAmount`
+                        fromIntegral (length sumAmts)
             mergedTabl = case tabls of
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks for this. I thought it would add the missing average in the grand totals row, but after merging I noticed it doesn't seem to (see below). Also all tests are passing, so maybe one of them needs to be updated:

```
$ stack build hledger && stack exec -- hledger -f examples/sample.journal bs -YAT
Balance Sheet

                      ||  2008/12/31    Total  Average 
======================++===============================
 Assets               ||                               
----------------------++-------------------------------
 assets:bank:checking ||           0        0        0 
 assets:bank:saving   ||          $1       $1       $1 
 assets:cash          ||         $-2      $-2      $-2 
----------------------++-------------------------------
                      ||         $-1      $-1      $-1 
======================++===============================
 Liabilities          ||                               
----------------------++-------------------------------
 liabilities:debts    ||          $1       $1       $1 
----------------------++-------------------------------
                      ||          $1       $1       $1 
======================++===============================
 Total                ||           0        0          
```

mstksg:
It looks like the bug was due to me forgetting that if's are greedy; adding parentheses around my if statements fixed this it looks like!  And yeah, should probably be good to get tests for this :)

PR #527, opened 2017-03-27
Fixing ongoing issues in balance statement/income statement/cashflow

mstksg:
Unless I'm mistaken, the PR in its current statement fixes the issues mentioned in #518 and #525 .

1. `--average` for final total row is rendered properly, even in the absence of `--total`.
2. No final total column for multicolumn cashflow.  This means that cashflow essentially ignores `-N` in multicolumn mode.
3. Extra clarification is shown for titles in the case where an "unusual" balance report type is used for a given command.  For example, `--historical` for incomestatement properly indicates the type of statement being shown.  Only shows up when the mode is explicitly overridden.
4. `--historical` now works properly in simple single-column mode
5. documentation for bs/cf/is updated to reflect new changes in functionality.

### DIFF
--- a/hledger/Hledger/Cli/BalanceView.hs
+++ b/hledger/Hledger/Cli/BalanceView.hs
@@ -79,7 +79,14 @@ balanceviewQueryReport
 balanceviewQueryReport ropts q0 j t q = ([view], Sum amt)
     where
       q' = And [q0, q j]
-      rep@(_ , amt) = balanceReport ropts q' j
+      rep@(_ , amt)
+        -- For --historical/--cumulative, we must use multiBalanceReport.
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks! Accidentally merged a little sooner than planned, never mind.

I see the overall average working now.

cashflow totals are still a little weird: single-column cashflow report shows 2 or 0 total rows depending on -N, multi-column cashflow report shows 1 total row regardless of -N. Not your fault.

The report title refinements are nice. I might have put them on the same line, so there's a single clear title (easier for docs etc.). In single-column mode though, the subtitle appears in the wrong place, above Total:.

The other changes sound good too.

PR #528, opened 2017-03-28
Add an option to use unicode in balance tables

### LINKED ISSUE #522: Add an option to use unicode in tables
Unicode has a bunch of [box drawing characters](https://en.wikipedia.org/wiki/Box-drawing_character) which enable relatively pretty tables even in terminals. It would be nice if `hledger` provided an option to make use of those instead of the pure ASCII tables since I personally find them easier to read.

I’ll probably give this a shot myself but I’m not sure when I’ll get to that so if someone else wants to pick this up be my guest.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43347685-add-an-option-to-use-unicode-in-tables?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cocreature:
fixes #522 

Currently this is just enabled for the balance report but I think that’s also the primary user of tables. It also needs some tests but I haven’t yet figured out where to add them. I’ll try to get to that in the next few days.


simonmichael:
Nice! tests/balance/pretty.test could be a good place. hledger/doc/balance.m4.md is the place to document the new flag. This will be great for bs/is/cf as well, rendered by Hledger/Cli/BalanceView.hs I think.

simonmichael:
I notice the top and left lines are double and the bottom line is single. Would it be better to flip these, so the double line emphasises the grand total ?

cocreature:
Alright, tests & docs added and I also added support for this option in `BalanceView`.

cocreature:
> I notice the top and left lines are double and the bottom line is single. Would it be better to flip these, so the double line emphasises the grand total ?

I don’t think so. The bottom line separates the totals from the individual values while the top line separates titles from the individual values. IMHO totals and individual values are more closely related (e.g. they are both numbers) than the titles and the values.

simonmichael:
In accounting reports I think it's traditional to use a thick/double line below columns of numbers, highlighting the grand totals. What do others think ?

simonmichael:
(NB this visual difference in line weights is more or less visible depending on the font, for anyone wondering..)

simonmichael:
Thanks!

PR #532, opened 2017-03-30
basic functional tests for multicolumn modes for is/bs/cf

mstksg:
Just some basic tests; not sure exactly how much coverage is required, but hopefully this is a start :)


simonmichael:
Great! That's probably more than needed. The failure is my fault, I'll fix.

PR #533, opened 2017-03-31
Disabling --cumulative for balancesheet, and also removed subtitle headings for 'default modes'

mstksg:
`balancesheet --historical`, `incomestatement --change`, and `cashflow --change` no longer show the subheading qualifying their balance types.  In effect, `--historical` becomes a no-op for `balancesheet`, etc.

`balancesheet --cumulative` is an error because `--cumulative` is not a flag in the flag group.  It doesn't show up on `balancesheet -h`.


simonmichael:
@mstksg is there still value in this PR, or should we close it ?

mstksg:
I think that this behavior was what was desired, but I'm not sure the behavior was ever integrated into HEAD in some other way.  Let me check what is the current behavior and get back to you.

simonmichael:
Closing this, feel free to reopen/clarify.

PR #543, opened 2017-04-28
Add Arch Linux to download page

funilrys:
I just adopted and updated the [Arch User Repository (AUR) package](https://aur.archlinux.org/packages/hledger/) so here we go :smile_cat:


simonmichael:
Wonderful, thank you!

PR #548, opened 2017-05-13
CsvReader: simplify '-' signs if there exist more than one

### LINKED ISSUE #524: Negating amount breaks parser for negative amounts
### Reproducing

Specific trace:

```
hledger: error: could not parse "$--5000.00" as an amount
the CSV record is:       "2017/01/04", "Merchant Credit", "-5000.00"
the amount rule is:      -%payment
the currency rule is:    $
the default-currency is: unspecified
the parse error is:      ParseError {errorPos = SourcePos {sourceName = "", sourceLine = Pos 1, sourceColumn = Pos 3} :| [], errorUnexpected = fromList [Tokens ('-' :| "")], errorExpected = fromList [Tokens (',' :| ""),Tokens ('.' :| ""),Label ('d' :| "igit")], errorCustom = fromList []}
you may need to change your amount or currency rules, or change your skip rule
```

Attached the rules [bank.csv.rules.txt](https://github.com/simonmichael/hledger/files/867967/bank.csv.rules.txt) and the CSV [bank.csv.txt](https://github.com/simonmichael/hledger/files/867968/bank.csv.txt) which cause this.

### Why this matters

First, it's in an example [rules](https://github.com/simonmichael/hledger/blob/master/examples/csv-rules/amazon.csv.rules#L22). 

Second, currently American Express outputs amounts as “inflow” (positive) amounts in—some—CSV exports. It's a liability account, it has a credit balance, these are credits to the liability and a debit to the expense account. 

A possible user solution is to flip `account1` and `account2` and let some other logic in hledger figure out the balance to credit the liability. This means a common set of rules can't be used since another CSV might have “outflow” (negative) amounts. In that case, `account1` and `account2` for common rules would be reversed.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43398775-negating-amount-breaks-parser-for-negative-amounts?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Sebatyne:
It prevents hledger to crash later, as it fails to read
amount strings containing more than one '-'

Fix #524

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -35,6 +35,7 @@ import Control.Monad.Except
 import Control.Monad.State.Strict (StateT, get, modify', evalStateT)
 -- import Test.HUnit
 import Data.Char (toLower, isDigit, isSpace)
+import Data.List (findIndices)
 import Data.List.Compat
 import Data.Maybe
 import Data.Ord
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks!

PR #553, opened 2017-05-24
Add --not-pending (-X) Option

cwarden:
Add --not-pending (-X) option to exclude pending transactions/postings.

Update documentation so --uncleared and --not-pending are congruent,
meaning all transactions not explicitly cleared or pending.


simonmichael:
Nice, thanks for the PR! There is plenty of confusion around the status field and I'm not sure yet if this change is the right solution. I will update docs soon to make more clear what we currently have. -X has an existing meaning in Ledger and we probably won't use it for this.

simonmichael:
I believe http://hledger.org/manual.html#status now describes what we currently have. AFAIK we still match Ledger pretty closely (with status: added).

I don't use pending (!) myself and I'm not sure how you are using it. Arguably we have too many features and options, and I am mindful of feature creep and user overwhelm; hledger is supposed to be simpler and easier than Ledger. I wouldn't mind removing pending, unless there are compelling uses for it. Though that might hurt Ledger interop. How many Ledger users use it ? If we keep it, I understand that consistency is nice. I think this is a case where we should first work out the docs we want to end up with. Going to try summoning reviewers.

cwarden:
What do you think about the changes in 937c9800c1d0892e71c4db7d742ab139454abc28, which remove Uncleared as a status for transactions (and postings) altogether?  Uncleared only becomes a filter for transactions.  Note that this has no impact on the functional tests, which suggests that Uncleared really isn't a transaction status.

I think removes the ambiguity around Uncleared as a status by making Pending and Cleared the only statuses that a transaction can have, or it can have no status.

I use Pending to flag transactions I need to look at again, such as receivables and payables.  The `!` flag is aesthetically well suited for these.  Then I mark them Cleared when recording a transaction that settles the Pending transaction, e.g. I've received a check for an invoice.

simonmichael:
@cwarden I feel a bit dim and find your changes hard to fully understand. 

> I think removes the ambiguity around Uncleared as a status by making Pending and Cleared the only statuses that a transaction can have, or it can have no status.

So there's still effectively three statuses, right ? I see that you want more complete/convenient UI for matching pending, but what's the reason for changing the internals this way ?

cwarden:
On Thu, May 25, 2017 at 08:45:08AM -0700, Simon Michael wrote:
> @cwarden I feel a bit dim and find your changes hard to fully
> understand.

This is probably my fault.  I'm still a haskell newbie so I may not
have approached the changes in the best way.

> > I think removes the ambiguity around Uncleared as a status by
> > making Pending and Cleared the only statuses that a transaction
> > can have, or it can have no status.
> 
> So there's still effectively three statuses, right ? I see that you
> want more complete/convenient UI for matching pending, but what's
> the reason for changing the internals this way ?

Transactions can still have three statuses, Cleared, Pending, or a
default of no status.

Uncleared is no longer an overloaded term.  When filtering,
--uncleared simply means not Cleared, rather than Uncleared or
Pending.

This change also makes it easy to fix filtering with status: (see
be0a4b482fbabbaefb69301e3f6a74d843fe55a5) because UnclearedFilter can
be different from NoStatusFilter.

So I think these changes to separate ClearedStatus and
ClearedStatusFilter are useful apart from introducing a new
--not-pending flag (with the caveat that an experienced haskell
development might suggest a better implementation).

cwarden:
I rebased against dca7ffc1 and squashed my commits so you can review the changes against the latest documentation.  I also removed the changes to the 1.2 docs based on the discussion in #hledger.

simonmichael:
Thanks, this is helping.

I see you replaced Uncleared|Pending|Cleared with Maybe (Pending|Cleared) in Transaction and Posting. Ok, sounds reasonable.

And you've added a separate ClearedStatusFilter type. Presumably this was needed to make the semantics clearer. Ok. PendingFilter | ClearedFilter | UnclearedFilter | NotPendingFilter | NoStatusFilter. It looks a little odd to me, are the Un and Not values necessary given that we already have Not in the query dsl ? Instead of Status NotPending, we could use Not (Status Pending).

I'd like to reserve X for other things (`-X/--exchange`). 

Let's defer the hledger-ui changes for a little. 

Re the hledger CLI. You've added --pending, which Ledger also has - good. 

I'm less sure about --not-pending. How often will this be needed in practice ? Is typing not:status:! good enough ? Sometimes I think we should be reducing the overlap between flags and query arguments. The latter are more powerful and better supported across all the hledger tools.

cwarden:
On Thu, May 25, 2017 at 12:08:14PM -0700, Simon Michael wrote:
> Thanks, this is helping.
> 
> I see you replaced Uncleared|Pending|Cleared with Maybe
> (Pending|Cleared) in Transaction and Posting. Ok, sounds reasonable.
> 
> And you've added a separate ClearedStatusFilter type. Presumably
> this was needed to make the semantics clearer. Ok. PendingFilter |
> ClearedFilter | UnclearedFilter | NotPendingFilter | NoStatusFilter.
> It looks a little odd to me, are the Un and Not values necessary
> given that we already have Not in the query dsl ? Instead of Status
> NotPending, we could use Not (Status Pending).

This is an example of where being a haskell novice is probably failing
me.  What do you suggest as the signature for
`clearedStatusFromRawOpts`?

Or should I change `ClearedStatusFilter` to something like:

```
data ClearedStatusFilter = Filter ClearedStatus | NoStatus | IsNot ClearedStatusFilter`
```

If this is an OK approach, I could use some guidance on how to avoid
lots of parentheses (see 69c9b5d33b63fe4e5154a39cb0752c2f9d37217c).  I
also had to use `IsNot` to avoid `Ambiguous occurrence ‘Not’` errors.
Tips?

> I'd like to reserve X for other things (`-X/--exchange`). 

No objection.  Choosing `-X` was arbitrary.

> Let's defer the hledger-ui changes for a little. 
> 
> Re the hledger CLI. You've added --pending, which Ledger also has - good. 

You implemented `--pending` in
d1f63334eeeb0b99cc0887d4ffa62cd3313f2e49.

> I'm less sure about --not-pending. How often will this be needed in
> practice ? Is typing not:status:! good enough ? Sometimes I think we
> should be reducing the overlap between flags and query arguments.
> The latter are more powerful and better supported across all the
> hledger tools.

I haven't been using hledger long enough to say how frequently I use
`not:status:!`.

cwarden:
@simonmichael, I think 0e749dd5ff955907006b22aa4d4484836f4bd655 is more in line with what you were suggesting about using the existing query dsl.

Is there a way of defining `clearedstatus_` as a `Query` that can only be constructed with `Status` or `Not Status`?

simonmichael:
Sorry for the delay @cwarden (hledger hacking time has been a bit scarce). Those ideas are not quite what I was thinking. Let me step back a minute. Am I correct - ignoring hledger-ui, the user-visible changes you have introduced are two:

1. fix `status:` to match things with empty status
2. add a `--not-pending` flag

1 is needed but I don't think 2 is worthwhile for the moment.

I think we can fix 1 with a smaller change that doesn't add complexity, eg the new type seems unnecessary.

cwarden:
@simonmichael, I'll take a stab at fixing `status:` without introducing `ClearedStatusFilter`.

(I'll try to put aside my unease about the inconsistency between `--uncleared`/`not:status:!`.)

simonmichael:
We can keep discussing that. There's some inconsistency in this area as you've pointed out. As I see it, the Ledger-style `--uncleared` flag brings confusion, muddying the terminology for users and developers. I don't want to complicate the situation by adding another flag. I think we could give a reasonable explanation of why --uncleared is the way it is (ledger compatibility, pending being transient state..) and how using status: is the more general alternative. We could also consider changing how --cleared/--pending/--uncleared work. Eg making them symmetrical so --uncleared no longer matches pending things, and perhaps allowing them to be combined.

cwarden:
What do you think of 3945bcafa104f580cad1399fe94f1c5274ec5fd0?  It's still a rather intrusive change to fix `status:`, but I think it removes the ambiguity around uncleared transactions.

simonmichael:
Related [mail list discussion](https://groups.google.com/d/msg/hledger/6rG7kvyCQck/nQu7kPMdBQAJ).

simonmichael:
Re 3945bca: I still don't really see the benefit of replacing the three-value type with a Maybe (two-value) type. It doesn't seem to make code clearer. Semantically, I think three values is more correct.

cwarden:
Representing the absence of a value as simply another value doesn't feel quite right, but perhaps `Maybe` isn't the best way to represent it.  287033e31f13daec12e0577ee9e92dd2bdd0161d returns to a three-value `ClearedStatus`, replacing `Uncleared` with `Void` (aka Unmarked aka NoStatus).

It updates parsing of command-line parameters to parse `status:0` and `--uncleared` as `Not $ Status Cleared` and `status:` as `Status Void`.

simonmichael:
Trying #572 for now, based on your work here. Thanks!

PR #557, opened 2017-05-27
Add Support for Rewriting Multipler Postings Into Different Commodities

cwarden:
When generating a new posting as a multiple of an existing posting,
support conversion to a different commodity.  For example, postings in
hours can be used to generate postings in USD.

Automatic transactions generated from rewrite rules use the commodity,
amount style, and transaction price if the rewrite defines a commodity.


simonmichael:
Neat, thanks!

ony:
@simonmichael it looks like support of real expressions like `(amount * 5 JPY / 100 USD)` and may worth of planing. Having parser support may at least make this case-specific formats generalized.

PR #558, opened 2017-05-28
Use pandoc From Stackage Snapshot

cwarden:
Use pandoc from stackage snapshot when building documentation to ensure
that pandoc and the pandoc filters are built against the same version of
pandoc-types; otherwise, the filters can fail with an obtuse error (see
https://github.com/jgm/pandoc/issues/3217).


simonmichael:
Thanks!

PR #560, opened 2017-05-31
Fix typo in help docs

afarrow:
Documentation listed `---man` as the command to bring up the man pages instead of the correct `--man`

### DIFF
--- a/doc/mockups/help.txt
+++ b/doc/mockups/help.txt
@@ -516,7 +516,7 @@ Get help:
  help|man|info        show any of the hledger manuals in text/info/man format
                       see also -h, CMD -h, --help|--man|--info
 *** d
-Get help: (see also -h, CMD -h, --help|---man|--info)
+Get help: (see also -h, CMD -h, --help|--man|--info)
  help|man|info        show any of the hledger manuals in text/info/man format
 
 ** hledger (commands list)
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks!

simonmichael:
Although, it was only a mockup and nobody looks at those. But still.

afarrow:
I found it when I downloaded version 1.2 via Homebrew and was looking through the docs to figure out how to use hledger. So I guess 1 person looks there

simonmichael:
Nifty. May I ask how you found your way to the mockups, exactly ? Did you clone the hledger source ?

afarrow:
If you're asking how I originally found the typo, I typed `hledger -h` and one of the options listed was to simply run `hledger` to list commands. So I ran `hledger` and down at the bottom of that output, it printed:
```
Help: (see also -h, CMD -h, --help|---man|--info)
 help|man|info        show any of the hledger manuals in text/man/info format
```
which contains the typo. The was all via the version of hledger I downloaded via Homebrew.

If you're talking about the fix that I just committed, I cloned the source and (since I had already seen it and knew what I was looking for) just searched all the files in the repo for `---man`.

simonmichael:
Aha. The mockups are just copy-pasted from actual output, no need to fix them. The bad output actually comes from the help text in hledger/Hledger/Cli/Main.hs - would you mind fixing there ?

simonmichael:
Doh, I see you took care of that too in the same commit. Thanks for explaining.

PR #562, opened 2017-06-04
prices: improve precision and add inverted costs

ony:
- Precision is extended for inferred prices from division.
- Support for reverse price inference to support use cases when people bring money into transaction from account with non-main currency.
- Sort prices by dates.

### DIFF
--- a/bin/hledger-prices.hs
+++ b/bin/hledger-prices.hs
@@ -13,7 +13,7 @@ import Data.Maybe
 import Data.String.Here
 import Data.Time
 import qualified Data.Text as T
-import Control.Monad
+import Control.Monad (when)
 import Hledger.Cli
 
 ------------------------------------------------------------------------------
@@ -30,6 +30,13 @@ Print all prices from the journal.
 showPrice :: MarketPrice -> String
 showPrice mp = unwords ["P", show $ mpdate mp, T.unpack . quoteCommoditySymbolIfNeeded $ mpcommodity mp, showAmountWithZeroCommodity $ mpamount mp]
 
+divideAmount' :: Amount -> Quantity -> Amount
+divideAmount' a d = a' where
+    a' = (a `divideAmount` d) { astyle = style' }
+    style' = (astyle a) { asprecision = precision' }
+    extPrecision = (1+) . floor . logBase 10 $ (realToFrac d :: Double)
+    precision' = extPrecision + asprecision (astyle a)
+
 amountCost :: Day -> Amount -> Maybe MarketPrice
 amountCost d a =
     case aprice a of
@@ -37,7 +44,7 @@ amountCost d a =
         UnitPrice pa -> Just
             MarketPrice { mpdate = d, mpcommodity = acommodity a, mpamount = pa }
         TotalPrice pa -> Just
-            MarketPrice { mpdate = d, mpcommodity = acommodity a, mpamount = pa `divideAmount` abs (aquantity a) }
+            MarketPrice { mpdate = d, mpcommodity = acommodity a, mpamount = pa `divideAmount'` abs (aquantity a) }
 
 postingCosts :: Posting -> [MarketPrice]
 postingCosts p = mapMaybe (amountCost date) . amounts $ pamount p  where
--- a/bin/hledger-prices.hs
+++ b/bin/hledger-prices.hs
@@ -10,18 +10,19 @@
 {-# LANGUAGE QuasiQuotes #-}
 
 import Data.Maybe
+import Data.List
 import Data.String.Here
 import Data.Time
 import qualified Data.Text as T
-import Control.Monad (when)
 import Hledger.Cli
 
 ------------------------------------------------------------------------------
 cmdmode = hledgerCommandMode
   [here| prices
-Print all prices from the journal.
+Print all market prices from the journal.
   |]
-  [flagNone ["costs"] (setboolopt "costs") "print transaction prices from postings instead of market prices"]
+  [flagNone ["costs"] (setboolopt "costs") "print transaction prices from postings"
+  ,flagNone ["inverted-costs"] (setboolopt "inverted-costs") "print transaction inverted prices from postings also"]
   [generalflagsgroup1]
   []
   ([], Nothing)
@@ -37,6 +38,18 @@ divideAmount' a d = a' where
     extPrecision = (1+) . floor . logBase 10 $ (realToFrac d :: Double)
     precision' = extPrecision + asprecision (astyle a)
 
+invertPrice :: Amount -> Amount
+invertPrice a =
+    case aprice a of
+        NoPrice -> a
+        UnitPrice pa -> invertPrice
+            -- normalize to TotalPrice
+            a { aprice = TotalPrice pa' } where
+                pa' = (pa `divideAmount` (1 / aquantity a)) { aprice = NoPrice }
+        TotalPrice pa ->
+            a { aquantity = aquantity pa * signum (aquantity a), acommodity = acommodity pa, aprice = TotalPrice pa' } where
+                pa' = pa { aquantity = abs $ aquantity a, acommodity = acommodity a, aprice = NoPrice, astyle = astyle a }
+
 amountCost :: Day -> Amount -> Maybe MarketPrice
 amountCost d a =
     case aprice a of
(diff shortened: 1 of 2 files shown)


tests/bin/prices.test
```diff
@@ -1,7 +1,7 @@
 # Test prices addon
 
 # by default only market prices are reported
-../../bin/hledger-prices -f-
+../../bin/hledger-prices.hs -f-
```
simonmichael: 5a7f74c6ffbe7cd39404302aa5ab37ec2e1629ed made all the bin/* tests use compiled executables, in quest of speedier tests. It probably would be best to discuss this separately, and not to do it for just one file.

bin/hledger-prices.hs
```diff
@@ -19,9 +20,10 @@ import Hledger.Cli
 ------------------------------------------------------------------------------
 cmdmode = hledgerCommandMode
   [here| prices
-Print all prices from the journal.
+Print all market prices from the journal.
   |]
-  [flagNone ["costs"] (setboolopt "costs") "print transaction prices from postings instead of market prices"]
+  [flagNone ["costs"] (setboolopt "costs") "print transaction prices from postings"
+  ,flagNone ["inverted-costs"] (setboolopt "inverted-costs") "print transaction inverted prices from postings also"]
   [generalflagsgroup1]
```
simonmichael: I found it not obvious what "inverted" or "cost" meant here, and had to study the tests. Using "cost" is understandable since it appears in the -B/--cost-basis flag, it's famiiar, short, and it often makes sense - but less so when you're selling rather than buying things. P directives can represent costs, too. 

The docs prefer "transaction price", which is directionless and also hints at the similarity with and difference from P "market prices". Worth using that term instead ? --transaction-price is long, I'll admit, but the meaning is more obvious.

For "inverted", I'm still a bit confused even reading the test. A little help text or brief use case description could help to explain and motivate these I think.

bin/hledger-prices.hs
```diff
@@ -30,27 +32,59 @@ Print all prices from the journal.
 showPrice :: MarketPrice -> String
 showPrice mp = unwords ["P", show $ mpdate mp, T.unpack . quoteCommoditySymbolIfNeeded $ mpcommodity mp, showAmountWithZeroCommodity $ mpamount mp]
 
+divideAmount' :: Amount -> Quantity -> Amount
+divideAmount' a d = a' where
+    a' = (a `divideAmount` d) { astyle = style' }
+    style' = (astyle a) { asprecision = precision' }
+    extPrecision = (1+) . floor . logBase 10 $ realToFrac d
+    precision' = extPrecision + asprecision (astyle a)
+
```
simonmichael: Is this the improve precision part ? What problem were you seeing ? Is this something that should be in the standard amount routines ?

bin/hledger-prices.hs
ony: I used to think about transaction as a temporary account to which we deposit and from which we withdraw money.
I guess I owe some use cases here to make it clear.

```
2017/1/1
    assets:usd  $-107
    assets:eur  100 EUR

2017/1/2
    assets:usd  $-107  @@ 100 EUR
    assets:eur  100 EUR

2017/1/3
    assets:usd  $-107
    assets:eur  100 EUR @ $1.07
```

All transactions represent a valid record of currency exchange. But first two and the last one bear a different price information. 

```
P 2017-01-01 $ 0.93457943925233644859813084112149532710280373831775700934579439252 EUR
P 2017-01-02 $ 0.935 EUR
P 2017-01-03 EUR $1.07
```

You can argue that last version is more preferred because you buy EUR for USD.
But consider that I'm abroad and pay with my credit card:

```
2017/1/1 Auchan supermarket
    assets:usd  $-107
    expenses  70 EUR
    expenses  30 EUR

2017/1/2 Auchan supermarket
    expenses  70 EUR
    expenses  30 EUR
    assets:usd  $-107
```
If you'll look at bank receipt you probably will see only one amount/rate. I.e. second transaction that will have two postings with prices will match operation more distantly than the first one.

bin/hledger-prices.hs
ony: You are correct. In standard `divideAmount` precision inherited from first amount which effectively cuts off important information when price is inferred from ratio of amounts in different commodities.

I'd like to see standard amount routines to include `multiplyAmount` and/or maybe implement `Num` class. And probably should include log10 based arithmetic for estimating precision required for reverse transformation.

But I'm not sure about correctness of this implementation. It gives me result that banks in my country usually use to represent price. I.e. with inverted costs I get EUR price in UAH with mostly 4 digits after dot.

tests/bin/prices.test
ony: Sorry. I thought that I've put this lines initially by accident to speedup tests locally.

bin/hledger-prices.hs
simonmichael: Useful discussion, highlighting that, although the essential maths are the same, how you write the journal entry - eg the order of postings, or which posting you write the price on - can affect report output. 

But, sorry, I don't follow what you mean with the second part ("consider that I'm abroad...") ?

What's your opinion on consistently using "price" and "transaction price" (or something better) rather than "cost" ?

As the world's most knowledgeable user of this command, may I ask you: in your real world usage, when/why do you use this command, the --cost flag, and the --inverted-cost flag ?

ony: > But, sorry, I don't follow what you mean with the second part ("consider that I'm abroad...") ?

I referred to a different kind of transactions where you pay in supermarket from an account of different currency. I.e. when withdraw from card is in one currency and there is several expense postings in other one.
In that case I would prefer to have currency exchange rate only near to withdrawal posting which had to be specified in an "inverted" form and thus `--costs` is not enough to catch that.

> What's your opinion on consistently using "price" and "transaction price" (or something better) rather than "cost" ?

I still don't feel to fully understand wording of `--cost` and `--value` to effectively contribute to decision about dropping mention of "cost".

>    -B --cost               convert amounts to their cost at transaction time
>                           (using the transaction price, if any)
>   -V --value              convert amounts to their market value on the report
>                           end date (using the most recent applicable market
>                           price, if any)
> 

And the fact that I'm not an English native speaking person this makes it even hard for me to be sure in how to distinguish "cost" and "price". For me difference is that "cost" usually refers to actually committed expenses in past and "price" usually refers to some indicator rather than real expense. I.e. "cost" may differ from "price" and that's why I thought `--cost` used to refer to application of past "price" while `--value` were used to estimate current "value" according to recent "price" that can be used to sell something.

> As the world's most knowledgeable user of this command, may I ask you: in your real world usage, when/why do you use this command, the --cost flag, and the --inverted-cost flag ?

Maybe I'm the only one who uses this "plugin" :) . My use cases:
1. See how rate of my main currency changes to have a rough estimation of future conversion rates. I.e. look into other currencies in term of my main currency. `P 2017-01-01 EUR $1.07`
1. Copy-paste some rates from prev. transaction to set market prices for reports with options like `-B`.
```
2017/1/1 My Bank | buy some currency for upcoming trip
  assets:bank  $-108
  assets:cash  100 EUR @ $1.07  ; reported by `--inverted-cost`
  expenses:bank  $1

2017/2/1 Supermarket | spend some money from my
  assets:bank  $-107 @@ 100 EUR  ; according to bank report and receipt. re
  expenses:grocery  50 EUR
  expenses:snacks  50 EUR
```
Effectively
```
zsh% ./bin/hledger-prices --costs
P 2017-01-01 EUR $1.07
P 2017-02-01 $ 0.935 EUR
zsh% ./bin/hledger-prices --inverted-costs
P 2017-01-01 $ 0.935 EUR
P 2017-02-01 EUR $1.070
```
Mostly I use `--costs` to get rates with which I've got exchange. And with `--costs --inverted-costs` I get source for picking prices to put into my journal.

simonmichael:
@ony: Thanks for your comments. Re costs vs prices, believe me it's confusing for native english speakers too. :) The language is fuzzy. 

Here's my basic feedback. As a potential user, the flags are not clear. I can't tell from the names and docs what they do and when I would use them.

This makes me not entirely comfortable approving this as it is; I want to help raise the standard of all our docs and UX, especially in the main repo.

On the other hand this is described as an [experimental addon](http://hledger.org/manual.html#experimental-add-ons), and addons are supposed to be the place where development is unconstrained and I don't get in the way. Also I'm not using this myself yet, and you are. Maybe I should be sending pull requests to you instead. Do you have the necessary access to self-approve/merge it yourself ?

I'm cc'ing this to the list for wider comment, in case anybody has thoughts on this PR or on the larger topic of merging policy.

simonmichael:
PS it's also possible I've gone blind from staring at this stuff and am just overthinking it. What do the rest of y'all think.

simonmichael:
Thanks!

Issue #565, opened 2017-06-05
Balancesheet displays only last account name


markokocic:
Balancesheet commands now displays only last part of the account name, making it confusing.
```
marko@comp:~/src/ledgers$ cat test.journal

01/01 Test1
        assets:bank1:giro  50
        equity:start

01/01 Test2
        assets:bank2:giro  150
        equity:start

marko@comp:~/src/ledgers$ hledger bs -f test.journal
Balance Sheet

Assets:
                  50  giro
                 150  giro
--------------------
                 200

Liabilities:
--------------------
                   0

Total:
--------------------
                 200

```
Balance command works as expected.
```
marko@comp:~/src/ledgers$ hledger bal -f test.journal
                 200  assets
                  50    bank1:giro
                 150    bank2:giro
                -200  equity:start
--------------------
                   0
marko@comp:~/src/ledgers$
```

simonmichael:
Thanks for the report. We need a functional test for the above, and a better fix for #552.

Issue #552, opened 2017-05-23
hledger-1.2 balancesheet --flat regression


simonmichael:
As of 1.2, balancesheet ignores the --flat flag.

simonmichael:
Hi @mstksg - would you mind reviewing this fix ? (I meant to send a PR). I can't remember why you added this, and it breaks bs --flat.

simonmichael:
This fix was no good, alas. Reopening.

mstksg:
will look at this :)

PR #566, opened 2017-06-05
Fixing bs/is/cf --flat regressions

mstksg:
Should properly address #552, and also #565.  Looks like the issue was that I was trying to duplicate the code in `Balance.hs`, but did not duplicate the logic completely (left out an extra conditional).  Would be nice if all the code duplication doesn't have to be here...i'll try to see if there's a way to consolidate the two in the future.


simonmichael:
You were fast, and I was slow, sorry. Thanks @mstksg !

PR #567, opened 2017-06-06
Show colorized amounts for totals of bs/is/cf

mstksg:
Small patch, just changes to show the "total" field in bs/is/cf outputs in red if they are negative, using `cshowAmount` instead of `showAMount`.

### DIFF
--- a/hledger/Hledger/Cli/BalanceView.hs
+++ b/hledger/Hledger/Cli/BalanceView.hs
@@ -100,7 +100,7 @@ multiBalanceviewQueryReport
     -> ([Table String String MixedAmount], [[MixedAmount]], Sum MixedAmount)
 multiBalanceviewQueryReport ropts q0 j t q = ([tabl], [coltotals], Sum tot)
     where
-      singlesection = "Cash" `isPrefixOf` t -- TODO temp 
+      singlesection = "Cash" `isPrefixOf` t -- TODO temp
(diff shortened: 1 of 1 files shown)


hledger/Hledger/Cli/BalanceView.hs
```diff
@@ -129,7 +129,7 @@ balanceviewReport BalanceView{..} CliOpts{command_=cmd, reportopts_=ropts, rawop
         unless (no_total_ ropts' || cmd=="cashflow") . mapM_ putStrLn $ -- TODO temp
           [ "Total:"
           , "--------------------"
-          , padleft 20 $ showMixedAmountWithoutPrice (getSum amt)
+          , padLeftWide 20 $ showamt (getSum amt)
```
simonmichael: Why switch to padLeftWide... for better layout of a multi-line amount ? Do you have an example by chance ? If it's noticeable, we should mention it in commit history, release notes etc.

simonmichael:
Thanks!

hledger/Hledger/Cli/BalanceView.hs
mstksg: I switched to `leftPadWide` because for some reason, it would not pad the amounts if they were negative/red.  From the examples I tried, this did not affect the behavior in the normal non-colored cases.

simonmichael: Oh, I see. Cool.

PR #576, opened 2017-06-20
Fix date of 1.2 release in CHANGES files

jdreaver:
I feel a little silly making a small PR for this, but the date did throw me off when I was wondering if version 1.2 was recent or not.

Thanks for making `hledger` by the way! I use it all the time.


simonmichael:
Well spotted, thanks @jdreaver !

PR #578, opened 2017-06-27
Update brick to 0.19

felixonmars:


simonmichael:
Thanks! I think we probably should support older brick for a while also, if necessary using CPP like [here](https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Read/CsvReader.hs#L45).

simonmichael:
cv https://github.com/fpco/stackage/issues/2617

felixonmars:
Just updated. As a n00b in Haskell I am not very sure of my skills, let me know if I am not doing it correctly :P

simonmichael:
Thanks! This was merged as ffe8b47.

PR #581, opened 2017-07-06
Staging commit: This is only to demonstrate balance sheet with equity.

nniro:
This is only a small demonstration to see the incidence of adding the default equity account to the balance sheet report.

### DIFF
--- a/hledger/Hledger/Cli.hs
+++ b/hledger/Hledger/Cli.hs
@@ -14,6 +14,7 @@ module Hledger.Cli (
                      module Hledger.Cli.Add,
                      module Hledger.Cli.Balance,
                      module Hledger.Cli.Balancesheet,
+                     module Hledger.Cli.Balancesheetequity,
                      module Hledger.Cli.Cashflow,
                      module Hledger.Cli.Help,
                      module Hledger.Cli.Histogram,
@@ -44,6 +45,7 @@ import Hledger.Cli.Accounts
 import Hledger.Cli.Add
 import Hledger.Cli.Balance
 import Hledger.Cli.Balancesheet
+import Hledger.Cli.Balancesheetequity
 import Hledger.Cli.Cashflow
 import Hledger.Cli.Histogram
 import Hledger.Cli.Help
--- a/hledger/Hledger/Cli/Balancesheetequity.hs
+++ b/hledger/Hledger/Cli/Balancesheetequity.hs
@@ -1,0 +1,40 @@
+{-# LANGUAGE QuasiQuotes, RecordWildCards, NoCPP #-}
+{-|
+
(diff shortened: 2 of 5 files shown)


simonmichael:
I like it. Let's fix the functional tests (use make [func]test to run those locally) and also update the manual (in commands.m4.md). Maybe use singular "Equity" as the report section title ? I think it's more idiomatic.

nniro:
there is one issue with adding Equity in the balance sheet report though : whenever there is going to be an income and/or an expense the report will no longer balance until the results are consolidated into the equity account.

I think we really need to implement a mean for the users to close accounts (transfer all the funds into another account).

in my accounting classes, we used a special account in the expenses category called summary of results.
We consolidated all income and expense accounts to it then consolidated that special account to the equity account(s).

This is but one method though. With a command to close accounts, the users could directly close the income and expense accounts to their equity account(s) if they want to. We may even implement helper functions to automate the consolidation of all income and expense accounts (or of course a query) to an account of their choice.

simonmichael:
This is something good to explore. Maybe check how beancount does it. We want to keep things easy for simple use cases too.

cwarden:
@nniro, have you looked at the [hledger-equity add-on](https://github.com/simonmichael/hledger/blob/817292a1dd00cd487ce65aa534f63873f1e9e18c/bin/hledger-equity.hs) to generate closing transactions?

nniro:
@cwarden, yes, this is the kind of operation that we need for this. We'll just need to make it generic (by requiring a query/AccountName for the debit and another for the credit accounts) and be part of hledger-lib.

something like this : 
`closeAccounts :: Day -> Query -> Journal -> T.Text -> T.Text -> [Transaction]
closeAccounts day qryToClose j description specialConsolAccountName = 
        let ledger = ledgerFromJournal qryToClose j in map (closeAccount day) $ ledgerTopAccounts ledger
        where   closeAccount day acct =
                        let
                                amount = sum $ amounts $ aibalance acct
                                credit = post specialConsolAccountName amount
                                debit = post (getAcctCode (aname acct)) (divideAmount amount (-1))
                        in
                        newTransaction j day description [debit, credit]`

At that point, we'll just have to decide how we want the user interface to implement this for the end user. I'll check other solutions for ideas.

simonmichael:
@nniro feel free to update this PR, or I'll close it.

simonmichael:
Update: extensive discussion of this in #hledger today. As [wikipedia](https://en.wikipedia.org/wiki/Balance_sheet#Types) says, for a balance sheet for personal use typically doesn't show equity, whereas for a large business it does. I think both use cases are common and important and actually I lean towards adding a new command, balancesheetequity(bse), for the second case.

PR #582, opened 2017-07-06
The tests code before implementing a fix for the -H bug.

nniro:
Check the individual patches description for more details.


nniro:
As a side note, when you add empty_=True to the defreportopts of the failing test it makes it pass. This way I can be pretty much sure that the test will pass once the fix is made.

nniro:
I will send another pull request with all the patches including the fix for the bug

PR #583, opened 2017-07-06
A fix for a bug with the -H option. It did not show anything when a period with no transactions was chosen rather than show the ending balances.

nniro:
These set of patches add a testing context for the fix and the fix itself.
The testing context is to make sure the fix does in fact behave like it should. They were added for BalanceReport and MultiBalanceReport only.

this is an excerp of one of the patches description :

This patch fixes a bug that happened when using the -H option on
a period without any transaction. Previously, the behavior was no output
at all even though it should have shown the previous ending balances of
past transactions. (This is similar to previously using -H with -E, but
with the extra advantage of not showing empty accounts)

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -929,6 +929,10 @@ abspat pat = if isnegativepat pat then drop (length negateprefix) pat else pat
 --     expenses:supplies  $1
 --     assets:cash
 --
+-- 2008/10/01 take a loan
+--     assets:bank:checking $1
+--     liabilities:debts    $-1
+--
 -- 2008/12/31 * pay off
 --     liabilities:debts  $1
 --     assets:bank:checking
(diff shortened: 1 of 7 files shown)


nniro:
Ah I didn't notice the warnings that samplejournal2 is already initialized elsewhere in the project. I'll need someone to tell me where to put the new samplejournal (or just rename it to samplejournal3). At first, I wanted to update samplejournal with my new entry and then change all the tests that failed to reflect the new journal, but I wasn't sure if that would be accepted or what.

simonmichael:
Status check ? I think you're still working on this and I don't need to do anything at the moment.

nniro:
Finally got travis happy.

simonmichael:
I see examples/sample.journal hasn't been changed. We have both samplejournal and sample.journal because unit tests needed a pure value, functional tests needed a file. Until now they have been the same, for simplicity and ease of moving tests between unit and functional. I would probably keep it that way, unless you think it's fine to let them diverge ? If so maybe we should rename one of them, samplejournal -> testjournal or something.

Re unit tests vs functional tests: I did find it quite hard to understand/replicate/test this after being away from it; functional tests would have helped me process it more quickly. But unit tests run faster and more focussed, and we should probably try them some more, so I'm fine with them.

simonmichael:
I have rebased and merged. Thanks for the fix, and good catch! I wonder how long it has been undetected.

PR #584, opened 2017-07-13
Exclude Fixed Assets from Cashflow Report

cwarden:
Add "fixed" to the regex used to exclude asset accounts from cashflow
reports.  Assumes accounts are set up something like:
```
assets:fixed assets:equipment
assets:fixed assets:vehicles
assets:fixed assets:buildings
```


hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -293,9 +293,9 @@ journalEquityAccountQuery _ = Acct "^equity(:|$)"
 -- | A query for Cash (-equivalent) accounts in this journal (ie,
 -- accounts which appear on the cashflow statement.)  This is currently
 -- hard-coded to be all the Asset accounts except for those containing the
--- case-insensitive regex @(receivable|A/R)@.
+-- case-insensitive regex @(receivable|A/R|fixed)@.
```
simonmichael: Perhaps :fixed would be better ? So as not to match assets:unfixed, assets:non-fixed etc.

PR #585, opened 2017-07-16
Update Arch Linux command

funilrys:
`hledger` is is now part of Arch Linux [community] so it's now available directly from `pacman`


simonmichael:
Thanks!

PR #590, opened 2017-07-23
Implement find Add-on

cwarden:
Simple add-on to print transactions with their location using
`hledger find [query]`.

### DIFF
--- a/bin/hledger-find.hs
+++ b/bin/hledger-find.hs
@@ -1,0 +1,41 @@
+#!/usr/bin/env stack
+{- stack runghc --verbosity info
+   --package hledger-lib
+   --package hledger
+   --package here
+   --package text
+-}
+
+{-# OPTIONS_GHC -Wno-missing-signatures -Wno-name-shadowing #-}
+{-# LANGUAGE QuasiQuotes #-}
+
+import Data.String.Here
+import Hledger
+import Hledger.Cli
+
+------------------------------------------------------------------------------
+cmdmode = hledgerCommandMode
+  [here| find
+Print transactions with their locations.
+With a query, only matched transactions are shown.
+Reads the default journal file, or another specified with -f.
+FLAGS
+  |]
+  [] -- [flagNone ["strict"] (\opts -> setboolopt "strict" opts) "makes date comparing strict"] --
+  [generalflagsgroup1]
+  []
+  ([], Just $ argsFlag "[QUERY]")
+------------------------------------------------------------------------------
+
+main :: IO ()
+main = do
+  opts <- getHledgerCliOpts cmdmode
+  withJournalDo opts $
+   \CliOpts{rawopts_=_rawopts,reportopts_=ropts} j -> do
+    d <- getCurrentDay
+    let
+      q = queryFromOpts d ropts
+      ts = filter (q `matchesTransaction`) $ jtxns $ journalSelectingAmountFromOpts ropts j
+      location t = "; " ++ (showGenericSourcePos $ tsourcepos t) ++ "\n"
+      showTransactionWithLocation t = location t ++ showTransactionUnelided t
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks. This is quite close to hledger print, should it be a separate command or an optional feature of print ? When do you use this ?

cwarden:
I had started working adding it to `hledger print`, but wasn't sure how to best implement it.  Adding another boolean to `showTransactionHelper` didn't seem like the right approach.  If you have a suggestion, I can take another stab at it.

I use it after `hledger print` when I've found a transaction I want to edit so an option to `print` (maybe `-L`/`--location`) would work too, e.g.
```
$ hledger -f ./examples/sample.journal print gift
2008/06/01 gift
    assets:bank:checking              $1
    income:gifts
$ !! -L
; "/home/cwarden/git/hledger/./examples/sample.journal" (lines 22-24)
2008/06/01 gift
    assets:bank:checking              $1
    income:gifts                     $-1
```

As an add-on, I'd probably use it with an alias like `alias hf="fc -s print=find"`

simonmichael:
@cwarden, I think I do like the idea of adding this as an option on the print command. I like your approach of inserting a comment above the transaction. Eventually, I hope such comments will be parsed as attached to the transaction. Something else we might want to show in this way would be transaction id. 

We could format them as tags, but I don't think that would be of much benefit. Better a standardish format that editors can parse and jump to.

It may or may not be worth using the valuable -L short flag, assuming that this won't be used too often.

For your use case above, maybe we can do better, eg open $EDITOR focussed at the first matched txn, as hledger-ui's E does ? If so, is there still a use case for the above ? Probably.

simonmichael:
PS showTransaction & co. probably can be done better, but for now I'm fine with adding another flag or three to showTransactionHelper, it's internal.

cwarden:
Here's a more detailed use case.  I use the pending flag to mark two types of transactions that I need to revisit: 1) receivables and payables that need to be settled by a second transaction, and 2) transactions to which I want to add more detail.

I'll use `hledger find -P` to list these pending transactions.

If it's a transaction that needs to be settled by another transaction such as when I receive payment for an invoice, then after I add the new transaction, I need to find the pending transaction to remove the pending flag. 

If it's a transaction that needs more detail such as a mortgage payment that I want to break down into principal, interest, and escrow, I'll go directly to the pending transaction to remove the pending flag.

Although I generally do want to open an editor and go to one of these transactions, it's not necessarily the first one.

simonmichael:
Nice, thanks. 

I'm imagining an edit command that opens a specific txn for editing. You'd have to filter out just the one you want, by date or txn id or something, to pick the right edit position. Though I suppose it could open multiple txns in a temp buffer, with their ids/locations attached, and write them all back to the right place, if necessary.

cwarden:
Outputting the transaction id and adding an edit command is a great idea, @simonmichael.

simonmichael:
I needed this the other day, and found it useful. Let's make it an option of the print command.

simonmichael:
Would @cwarden or anyone else like to pick up this PR ?

simonmichael:
Inactive, closing for now.

ony:
> I'm imagining an edit command that opens a specific txn for editing. ...

> I needed this the other day, and found it useful. Let's make it an option of the print command.

Should we have this as a separate wishes for these? I don't mind having `vim` to open file at specific position when I find inconsistency (I hate calling `/^03\/16 ` and then dashing through transactions on that day).

simonmichael:
Up to you.. I would open them if someone was likely to work on them soon.

You know that E in hledger-ui will open your editor at the error line, right ? Best used in hledger-ui --watch, after an error appears.

cwarden:
This functionality was implemented in the hledger-print-location addon in efa0d4391089c4a81b51bb5231723fb592cd6ec4.

Issue #523, opened 2017-03-23
allow flexible naming of main account types


simonmichael:
High-level commands like balancesheet/incomestatement/cashflow, and more parts of hledger in future, need to know which accounts are assets, liabilities, equity, revenues, and expenses. Until now this has been done with hard-coded anglo-centric account name patterns. We need to support names in other languages.

The current plan is to write [account directives](http://hledger.org/manual.html#account-directive) for the top-level accounts and define the account type there. Eg:
```
account aktivum
  type: A

account forpliktelse 
  type: L

account aksje
  type: Q

account omsetning 
  type: R

account utgift
  type: E
```

Ie an account tag named `type`, which would also apply to subaccounts.

Can we foresee any usability issues ? Any other ideas ?

Related: 
[support the account directive & account metadata #217](https://github.com/simonmichael/hledger/issues/217)

simonmichael:
Related: #217.

simonmichael:
Alternatives & older mockups:
```
# using a keyword/subdirective rather than a specially-named tag:
account aktivum
  asset          # assets and its subaccounts are asset accounts (by default ? need to override for "contra" accounts ?)

account forpliktelse
  liability      # this and subs are liability accounts

account aksje
  equity

account omsetning
  revenue

account utgift
  expense

account annen   # a special account, not one of the standard types

# a single directive listing the five top-level account names in ALQRE order. Account names may not contain comma.
account-types aktivum, forpliktelse, aksje, omsetning, utgift

# six top-level account names, separated by double space. The first five are the A/L/Q/R/E accounts.
account-roots  assets  liabilities  equity  revenues  expenses  special assets
```

simonmichael:
```
# accounts directive, one-line form. Defines top-level accounts & types.
accounts assets, liabilities, equity, revenues, expenses, special assets
accounts  assets  liabilities  equity  revenues  expenses  special assets

# accounts directive, optional multi-line form.
# Defines account tree, or at least the roots.
# First five root accounts are ALQRE.
# Otherwise, order of names might influence default account ordering in reports.
# Comments and tags allowed.
accounts
 assets      ; this and subaccounts are assets
  bank       ; phone:111-111-1111, url:http://blahblah
   checking
     ; acctno:12345
     ; routingno:123127745
   saving
  cash
 liabilities
  debts
 equity
 revenues
  salary
 expenses
  food
  home
 special assets
  book tokens
  prepaid classes

```

simonmichael:
[...]

PR #597, opened 2017-08-01
Allow users to customize balancesheet, incomestatement, cashflow, and also to provide their own custom compound reports

mstksg:
Attempts to offer a solution to several open musings (#523); was mostly a quick afternoon's work so it's not too big a deal if it should be scrapped, i'm not too personally attached :)  Just an idea!

Adds a new "compound report" directive:

```
report balancesheet
  title     Balance Sheet
  aliases   bs balsheet
  type      historical
  subreport Assets        ^assets?(:|$)
  subreport Liablities    ^(debts?|liabilit(y|ies))(:|$)
```

Which can be called by the command line:

```
hledger report balancesheet
hledger report bs -MTA --depth=2
hledger report balshet -WV
```

The original `balancesheet`, `cashflow`, and `incomesheet` are now basically aliases:

````
hledger balacesheet <opts>
    => hledger report balancesheet <opts>
````

If multiple `report` directives are given with the same name, the last one seen is used.

`balancesheet`, `cashflow`, and `incomesheet` are treated specially; you may fill it out incomplete (and leave out `type` or `title` or `aliases`) and the "default" ones are used.  So you can do:

```
report balancesheet
  aliases akfo
  subreport Assets        ^aktivum?(:|$)
  subreport Liablities    ^forpliktelse(:|$)
```

And it will assume `type historical`, `title Balance Sheet`, and `aliases bs akfo`.  Just for convenience.

You can also call `hledger report`, with no arguments, to get a list of all recognized compound reports:

```
> hledger report
Currently defined compound balance reports:
    balancesheet (Balance Sheet): Defined at /home/robert/.hledger.journal:7:1
    cashflow (Cashflow Statement): Built-in report
    incomestatement (Income Statement): Built-in report
    customreport (My Custom Report): Defined at /home/robert/.hledger.journal:15:1
```

### CODE REVIEW

**Comment by simonmichael** [comment]
Pretty nice! I have a bunch of thoughts, maybe I'll chat you on IRC a bit for efficiency.

**Comment by simonmichael** [comment]
The proposal above is at the reporting/presentation level, which allows for some nice control of the output. #523 is more at the model level, giving the program a deeper understanding of what the accounts mean, which can be useful for other things. We might need something of both.

Aside from the three standard reports, can we find some examples of other useful reports this would enable ?

I don't love the idea of having the custom report commands at second level. It might be okto making them visible as top-level commands, shown in the main commands list, etc. 

I think pretty soon you would want more features for these reports, eg a depth limit. Would it be better to go further and allow full hledger command lines ?

Ledger allows command options, proposed in https://github.com/simonmichael/hledger/issues/349 . Possibly it allows full commands too.

I can think of a few problems/use cases this proposal touches:

A. user uses non-english account names and wants the standard balancesheet/cashflow/incomestatements reports (which require specific account names) to work. Also perhaps wants to see non-english titles in the output. This is the focus of #523.

B. user wants to define custom reports (or general commands) for use at any time, preferably in a way that's low hassle, robust, portable, shareable. This is a bit like git's command aliases. (I feel I wrote about this somewhere but haven't found it.)

C. user wants to define custom reports (or commands ?) for use with a specific journal, probably tied to that journal's accounts and purpose. The journal entries and the custom reports are maintained together, shared together, etc.

Ideally, we wouldn't require every non-anglo user to set up a custom config so the standard reports work. Another possibility would be to select account names/report titles based on locale, at least for the default out-of-box behaviour.

**Comment by mstksg** [comment]
It's difficult to have these commands viewable at the top level, because they aren't known until the journal files are completely parsed (and includes are processed).  The only way it'd be possible would be to parse all of the journals every time `help` is called or desired...it might be kind of unsavory, but I think it's technically possible.  We'd have to give `help` an additional command line argument to give the location of the journal file that would contain these commands.

Alternatively instead of

```
report balancesheet
report incomestatement
report cashflow
```

as ways to specify custom reports, we could instead *fix* things to have only three reports available, and only have the three directives:

```
balancesheet
  title     Balance Sheet
  subreport Assets        ^assets?(:|$)
  subreport Liablities    ^(debts?|liabilit(y|ies))(:|$)

incomestatement
  ...

cashflow
  ...
```

So `balancesheet`, `incomestatement`, and `cashflow` themselves become three disjoint directives, each just sharing the same syntax/fields for their body.

Then no more `hledger report <name>` command, just back to bs/is/cf like usual.  This would still allow users to localize and customize their reports, except disable adding additional reports.

simonmichael:
I now like the original form of this proposal - custom second level report commands definable in journal. I have had a few cases where I wanted such custom reports, like is/bs but with different account groupings.

I wouldn't want to replace the top level bs/cf/is commands yet. That could be explored again later. I'd rather add this as simply and non-complectingly as possible.

Would @mstksg or anyone else like to pick this up again ?

simonmichael:
Inactive, closing for now.

PR #600, opened 2017-08-02
Fix build of hledger-ui with megaparsec-6

hpdeifel:


simonmichael:
Wow, I thought I'd tested that. Thanks!

PR #601, opened 2017-08-08
Fix build of hledger-web with megaparsec-6.0

hpdeifel:


simonmichael:
Rebased and merged. Thank you very much, sorry for the slow response.

Issue #474, opened 2017-01-13
hledger-ui, hledger-web etc. should support --pivot


simonmichael:
The --pivot feature works for hledger CLI, but not for the other hledger tools, and the docs don't mention this (hence the bug label). We'd like it to work there too.

simonmichael:
WOrks now.

PR #606, opened 2017-08-29
ui, web: Enable --pivot and --anon in ui and web (#474)

zarybnicky:
This is a very simple change, fixing #474.

I've exposed pivot and anon from Cli.Utils and used them in the custom withJournalDo that -ui and -web have. Other tools can already use --pivot and --anon (at least the tools in /bin), as they use the withJournalDo from Cli.Utils (either directly or using a wrapper, as hledger-budger does).

Though I'd suggest changing the signature of withJournalDo to `CliOpts -> (Journal -> IO ()) -> IO ()` and partially applying the whole UiOpts/WebOpts, so that there can be one unified withJournalDo.
(I haven't done that now, as that changes the API and requires more changes.)


simonmichael:
Thank you!

Issue #430, opened 2016-11-15
"Add transaction" button not keyboard-accessible


ndarilek:
As a keyboard user, I attempt to tab to this button but it never seems to receive focus. Also, when I do place focus on it, space/enter doesn't appear to trigger it.

Can anyone else duplicate? Granted, I've only run this within Sandstorm, but I don't imagine that would make a difference.

Thanks.

simonmichael:
Hmm, you're right. The "Add a transaction" link is generated by some bootstrap or jquery modal thing, starting at https://github.com/simonmichael/hledger/blob/master/hledger-web/templates/default-layout-wrapper.hamlet#L97 . I went as far as adding tabindex=4 to all of those divs and the h3, but still couldn't tab to the link (tested in safari). Any help welcome.

There is a specific keyboard shortcut that activates the link: `a` (assuming the web page has the focus).

ndarilek:
That looks like the dialog itself. Where is the button that opens it?

simonmichael:
The "Add a transaction" link is the h3 at line 102. I think due to the .modal-title class. I don't know why it's so complicated, it uses http://getbootstrap.com/javascript/#modals .

the-solipsist:
Since the upgrade from v0.26 to v1.1, the 'a' key works for me in hledger-web.  Is this still a problem?

simonmichael:
But the problem reported ("I attempt to tab to this button but it never seems to receive focus. Also, when I do place focus on it, space/enter doesn't appear to trigger it.") is still unsolved, right ?

squiddle:
i do not know (much|any) haskell but i think the add transaction link is rendered here: 

https://github.com/simonmichael/hledger/blob/master/hledger-web/Handler/JournalR.hs#L40

ocdtrekkie:
@zarybnicky should probably claim the bounty or whatever since they fixed it like a year ago.

simonmichael:
Agreed, @zarybnicky please go ahead.

zarybnicky:
Oh, I've completely forgotten about this, thanks for tagging me. I'll have a look at how to claim the bounty.

(I was about to start trying to replicate the issue on Sandstorm but apparently, according to #425, it's resolved there as well.)

ocdtrekkie:
@zarybnicky Yeah, turns out I didn't actually update the Sandstorm grain I opened to test it. :/

simonmichael:
[...]

PR #607, opened 2017-08-29
web: Make "Add transaction" button tabbable (#430)

zarybnicky:
This fixes the rest of #430.

Adding a href to an anchor makes it into a button, adding it into the tab order and making it clickable. While having data-toggle="modal" adds JS onclick to the link, making it clickable, it is adding href="#" that makes it selectable via tab and clickable via space.


simonmichael:
Excellent!

PR #608, opened 2017-08-30
lib: Fix filtering by payee and note (#598)

### LINKED ISSUE #598: Filter by payee
Hello,
I am trying to filter my ledger by payee. 

The "tags" section of the documentation suggests to use the `payee:` implicit tag. I tried:

    hledger register payee:"A Payee"

with no luck.

However,

    hledger register desc:"A Payee"

seems to correctly return the entries having "A Payee" as the payee.

I am wondering whether I am doing something wrong, the documentation has a glitch, or I bumped into a bug.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47848940-filter-by-payee?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zarybnicky:
Fixes #598.

Neither parseQuery nor matchPosting/matchTransaction were set up to accept payee:/note: filters. Also, transactionNote was non-functional (it returned payee) and payeeAndNoteFromDescription didn't actually fulfill its documentation, namely "When there's no |, payee == note == description."

Edit: It's also occurred to me that in the Query datatype, Desc and Code are obsolete. I'd suggest either replacing them with either `Tag "desc"`, or maybe with an `ImplicitTag` type.

Edit2: After looking at the code in more detail (and fixing broken tests, namely not:tag:.), it seems to me that the whole Query module could use some more datatypes instead of string matching - eg. aggregate implicit tags into a `data ImplicitTag` (with Desc and Code there as well) or using a parser instead of parseQueryTerm with a cascade of T.stripPrefix.

### DIFF
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -187,16 +187,18 @@ transactionPayee :: Transaction -> Text
 transactionPayee = fst . payeeAndNoteFromDescription . tdescription
 
 transactionNote :: Transaction -> Text
-transactionNote = fst . payeeAndNoteFromDescription . tdescription
+transactionNote = snd . payeeAndNoteFromDescription . tdescription
 
 -- | Parse a transaction's description into payee and note (aka narration) fields,
 -- assuming a convention of separating these with | (like Beancount).
 -- Ie, everything up to the first | is the payee, everything after it is the note.
 -- When there's no |, payee == note == description.
 payeeAndNoteFromDescription :: Text -> (Text,Text)
-payeeAndNoteFromDescription t = (textstrip p, textstrip $ T.tail n)
+payeeAndNoteFromDescription t
+  | T.null n = (t, t)
+  | otherwise = (textstrip p, textstrip $ T.drop 1 n)
   where
-    (p,n) = T.breakOn "|" t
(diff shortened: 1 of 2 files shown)


zarybnicky:
Travis says: Invalid symmetric difference expression master...2292800f4224c95525e540e0cdb6c20af0a7a9b6
Let me try again...

zarybnicky:
And I see I forgot to check the tests... I'll look into it soon, during the weekend.

simonmichael:
Thanks for the fixes! Actually I hadn't thought of making payee and note searchable like first-class fields (tag:payee=... is what I expected to work). But after thinking about it, it makes sense. I think we should just say "you can pivot on these field names (description, payee, note..) or any tag name", and drop the "implicit tags" concept.

Aside: the payee/note feature is a complication, but I've wanted it often enough that I *think* it's worth keeping. I've been brainstorming different syntax/semantics for it, but I think the current behaviour is reasonable. We should probably add some tests. I will update docs soon.

simonmichael:
Rebased and merged as 72cf6a8.

PR #617, opened 2017-10-01
bin: budget: re-factor to use hledgerCommandMode

ony:
This is preparation to make it part of of hledger.

### DIFF
--- a/bin/hledger-budget.hs
+++ b/bin/hledger-budget.hs
@@ -5,7 +5,7 @@
   --package cmdargs
   --package text
 -}
-{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE OverloadedStrings, QuasiQuotes #-}
 {-
 
 hledger-budget REPORT-COMMAND [--no-offset] [--no-buckets] [OPTIONS...]
@@ -142,10 +142,26 @@ $ hledger budget -- bal --period 'monthly to last month' --no-offset --average
 import Control.Arrow (first)
 import Data.Maybe
 import Data.List
+import Data.String.Here
 import System.Console.CmdArgs
 import Hledger.Cli
-import Hledger.Cli.Main (mainmode)
 import Hledger.Data.AutoTransaction
+
+budgetmode :: Mode RawOpts
+budgetmode = (hledgerCommandMode
+    [here| budget
+Perform various reporting commands taking into account budgeting entries in journal.
+    |]
+    [] -- ungroupped flags
+    [("\nBudgeting", budgetFlags), generalflagsgroup2] -- groupped flags
+    [] -- hidden flags
+    ([], Nothing)
+    ) { modeGroupModes = Group
+        { groupUnnamed = map fst actions
+        , groupNamed = []
+        , groupHidden = []
+        }
+    }
 
 budgetFlags :: [Flag RawOpts]
 budgetFlags =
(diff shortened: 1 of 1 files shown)


ony:
To integrate as builtin command I suspect we need to adapt those hacks for handling CLI to work with modes deeper than 1 level.
I've tried to use positional arguments but to allow any flags I need to put `--` (double dash) after command.

ony:
Sample of the hack:
```diff
From c48a8b15197ea2e9285ab5118a41f0d1ff72f6c7 Mon Sep 17 00:00:00 2001
From: Mykola Orliuk <virkony@gmail.com>
Date: Mon, 2 Oct 2017 01:24:38 +0200
Subject: [PATCH] bin: budget: ugly hack for CliOpts

---
 bin/hledger-budget.hs | 7 +++++--
 1 file changed, 5 insertions(+), 2 deletions(-)

diff --git a/bin/hledger-budget.hs b/bin/hledger-budget.hs
index fe57cf2b..b15125ac 100755
--- a/bin/hledger-budget.hs
+++ b/bin/hledger-budget.hs
@@ -170,7 +170,7 @@ budgetFlags =
     ]
 
 actions :: [(Mode RawOpts, CliOpts -> Journal -> IO ())]
-actions = first injectBudgetFlags <$>
+actions = first (injectBudgetSubCmd . injectBudgetFlags) <$>
     [ (balancemode, balance)
     , (balancesheetmode, balancesheet)
     , (cashflowmode, cashflow)
@@ -182,6 +182,9 @@ actions = first injectBudgetFlags <$>
 injectBudgetFlags :: Mode RawOpts -> Mode RawOpts
 injectBudgetFlags = injectFlags "\nBudgeting" budgetFlags
 
+injectBudgetSubCmd :: Mode RawOpts -> Mode RawOpts
+injectBudgetSubCmd mode = mode { modeValue=[("command", "budget"), ("budget-cmd", head (modeNames mode))] }
+
 -- maybe lenses will help...
 injectFlags :: String -> [Flag RawOpts] -> Mode RawOpts -> Mode RawOpts
 injectFlags section flags mode0 = mode' where
@@ -245,6 +248,6 @@ main = do
 
 budget :: CliOpts -> Journal -> IO ()
 budget opts journal =
-    case find (\e -> command_ opts `elem` modeNames (fst e)) actions of
+    case find (\e -> stringopt "budget-cmd" (rawopts_ opts) `elem` modeNames (fst e)) actions of
         Just (_, action) -> budgetWrapper action opts journal
         Nothing -> print budgetmode
-- 
2.14.1
```

simonmichael:
@ony thanks. I see you kept all the balance, print, register etc. subcommands. I was thinking there was a single "budget" command, and the others could be dropped, but I guess not. 

I haven't tried the latest yet. I think your idea is that you run `hledger budget print`, `hledger budget balance` etc., and get variants of those commands that see automated postings and periodic transactions ? Plus, they do something I don't understand with an alternate account tree inferred just from the periodic transactions ?

Does having automated postings or periodic transactions rules in your journal break the regular commands ?

The additional level of commands might be ok temporarily but I don't think we want that long term. 

Do you have a tutorial or worked example showing the new kind of reports this enables ?

simonmichael:
I installed the latest and am slowly figuring some things out. I misunderstood this PR as making the command built in, when actually it's just a step in that direction. Merging.

simonmichael:
And this may not be the right place to continue this discussion, but.. adding to my notes:

The budget command's subcommands rewrite all account names to budget categories. That is, to the nearest (same or closest parent) account name that is used in any periodic transaction. If the account name or a parent is not mentioned in a periodic transaction, that account is rewritten to `<unbucketed>`.

I am reminded of the rewrite command, which adds automated postings, and does only that (IIRC). So to run reports on that, you pipe it into another hledger invocation: `hledger rewrite ... | hledger -f- print`. What do you think of making the budget command more like that: `hledger budget | hledger -f- bal` - can it be made to work ? Then we wouldn't have the subcommands.

simonmichael:
PS or, of course, we may turn on more of the required functionality by default (such as periodic txns and automated postings), so the budget command has less to do.

I'd still like to understand/see it prove its usefulness, before we do that - not just blindly add those features because we're all familar with them from Ledger. Right now I don't know if even you are successfully using this in battle.

ony:
In the morning I considered other approach. Instead of having dedicated command we can have generic flags like `--budget-offset` (add up transactions) and `--budget-buckets` (re-write account names) and `--budget` that is equivalent to specifying both of them.
This way we'll match with Ledger CLI more closely.

Pros:
- No need to handle multi-level sub-commands like `hledger budget balance`.
- Instead of writing `hledger budget balance --no-offset` you can write `hledger --budget-buckets balance`.

Cons:
- No control on which sub-commands will be affected. I.e. apart from reporting commands we'll hit for example `print`.
- Misleading absence of effect on external commands `hledger --budget ui`.

simonmichael:
Sounds interesting. But you can implement (or expose) the flags for the just commands you want, no ?

ony:
Yes. I can inject these flags in specific modes and corresponding handlers. I'll experiment in that direction [branch `feature/budget-cli` in my repo](https://github.com/ony/hledger/tree/feature/budget-cli)

ony:
Created #620 pull request. I decided that it is easier to move it as is without integrating much into current generic flags handlings.

PR #620, opened 2017-10-05
Very simple plugging in budget as builtin

ony:
Move `hledger-budget` to be part of buil-in commands.

### DIFF
--- a/bin/hledger-budget.hs
+++ b/bin/hledger-budget.hs
@@ -165,7 +165,8 @@ Perform various reporting commands taking into account budgeting entries in jour
 
 budgetFlags :: [Flag RawOpts]
 budgetFlags =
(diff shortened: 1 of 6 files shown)


simonmichael:
Nice, @ony. I am travelling and haven't been able to spend more time with this yet. Are there docs yet ? I can merge this when it comes with docs clear enough for a hledger newcomer and when two people say "I'm using this and it's useful".

simonmichael:
docs: https://github.com/simonmichael/hledger/blob/master/bin/hledger-budget.hs#L10

simonmichael:
What about this one @ony.. do we close it and drop the hledger-budget addon, since it is now pretty much integrated ?

simonmichael:
> What about this one @ony.. do we close it and drop the hledger-budget addon, since it is now pretty much integrated ?

I'll do that tomorrow, unless you object.

Issue #481, opened 2017-01-15
balance assertion errors should show a more accurate line number


simonmichael:
Currently they show the line number where the transaction starts. Better to show the line number of the posting with the failing assertion.

jeevcat:
I was thinking about having a look at this issue as something easy to get me started contributing. However, as far as I can tell, there are currently no line numbers being printed for balance assertion errors?
I'm looking [here](https://github.com/simonmichael/hledger/blob/08acad3f42a78cc84b870d66d23f15473be40e07/hledger-lib/Hledger/Data/Transaction.hs#L393) and at the following strings at 417 and 419.

simonmichael:
Great. That's actually a different error (unbalanced transaction). [Here's](https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Data/Journal.hs#L538) the balance assertion error.

simonmichael:
I notice it says column 0 which looks a little odd, I wonder how hard it would be to show just the line number, or (gold plated version) the actual column of the = sign.

jeevcat:
I have this working now, but it only finds the column using the output of `postingAsLines` or `show transaction`. However, this isn't always the same as what is written in a journal file. 

I can see that I can get the transaction's 1-based position in the input stream from a Transaction. How can I access the current input stream from the `checkBalanceAssertion` function?

simonmichael:
To do it robustly I guess we'd have to save the position of the balance assertion during parsing, similar to Transaction.tsourcepos. Perhaps:

type BalanceAssertion = Maybe (Amount, GenericSourcePos)
data Posting = Posting {... pbalanceassertion :: BalanceAssertion ...}

simonmichael:
PS if you think it's worth the trouble. Or we could just not show a column number.

jeevcat:
Can we go with `type BalanceAssertion = (Maybe Amount, Maybe GenericSourcePos)` instead?

In some places I'm unable to populate the source position easily (like in [Equity.hs](https://github.com/simonmichael/hledger/blob/b7f8ca2cdaa0fb23e419a689788e8ce8a3f11471/hledger/Hledger/Cli/Commands/Equity.hs#L69) and also [CSV parsing](https://github.com/simonmichael/hledger/blob/b7f8ca2cdaa0fb23e419a689788e8ce8a3f11471/hledger-lib/Hledger/Read/CsvReader.hs#L700).

simonmichael:
[...]

PR #625, opened 2017-10-14
Balance assertion errors now show line of failed assertion posting.

jeevcat:
Addresses #481.

This works, but I'm very new to Haskell and I'm not sure I'm breaking any rules or best practices. Please let me know if there's a better way to do anything, especially where I'm increasing the `GenericSourcePos`'s line number.

Also, this currently prints column 0. Should we try for a custom message instead?


simonmichael:
Nice @jeevcat. As you can see on the issue page our travis check has caught a functional test that needs updating. You can run those locally, see http://hledger.org/contributing.html#tests .

jeevcat:
Yeah I just saw this too. I'm working on updating the tests now.

jeevcat:
Looks like it's passing the test now. Is there anything I should change to make my changes better conform with the current codebase?

simonmichael:
Looks good to me. Thanks and welcome!

PR #626, opened 2017-10-15
Rewrite of BalanceAssertion type to track its source position.

jeevcat:
Addresses further concerns in #481 regarding column number.

I changed the BalanceAssertion type such that it now tracks the line number and column of the balance assertion's `=`. I changed all references to BalanceAssertions to reflect this.

I removed the previous code for finding the line number, as it was no longer required.

Changed test to also check column number.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -513,7 +513,7 @@ journalCheckBalanceAssertions j =
 -- | Check a posting's balance assertion and return an error if it
 -- fails.
 checkBalanceAssertion :: Posting -> MixedAmount -> Either String ()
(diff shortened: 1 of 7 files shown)


simonmichael:
Nice! Now that I see it, maybe my idea to put the position in BalanceAssertion rather than Posting was wrong, since the assertion is not really complete without the account and date as well. Thoughts ? If you're happy with it, I'll merge.

jeevcat:
That's true. In that case a Nothing BalanceAssertion will always be accompanied by a nullsourcepos in the Posting.

If you think it's best, I'll change it back to something like: `{... pbalanceassertion :: Amount, pbalanceassertionsourcepos :: Maybe GenericSourcePos ...}`?

simonmichael:
I guess it's fine. Thank you!

PR #628, opened 2017-10-16
(WIP) More csv fields

MatthiasKauer:
See https://github.com/simonmichael/hledger/issues/627 for discussion.


simonmichael:
This looks definitely on the right track. Could you squash these into a single, cleaner commit and I'll take another look.

MatthiasKauer:
Ok, I'll squash and open a new PR.

Issue #624, opened 2017-10-13
Tracking investments in hledger


MatthiasKauer:
Hi,
I just converted my trading-related transactions from last year to hledger.
In that process, I learned a few things that I felt are worth sharing.

First of all, I want to point out that everything worked for me.

Some things could still be improved.
For instance, I would like to rewrite the guide from http://hledger.org/investments.html.
Buying and selling stocks does not only involve asset accounts.
Capital gain is an income and must be included in tax reports.
https://github.com/simonmichael/hledger/issues/377 also discusses that and my examples look similar.

```
2015/01/01  Initial purchase
    asset:invest:AAPL   AAPL 5 @ USD 130
    asset:invest:cash

2017/01/01  Realize profits
    income:capgain
    asset:invest:AAPL   AAPL -5 @ USD 130   ; Is there any way to deduce or validate "@130" here?
    asset:invest:AAPL   AAPL 5 @ USD 150

2017/01/01  Sell (Note: should be combined with 'realize profits' in most cases)
    asset:invest:AAPL   AAPL -5 = AAPL 0 ; ensure that no more shares are left
    asset:invest:cash   USD 750
```

The guide should also have a dividend example even though it is straightforward.
I also suggest a reference to Martin Blais' guides for Beancount.
They explain modeling investments in double-entry accounting quite well.

Some questions I came across and features I'd find helpful in hledger:
- CSV import should be able to generate transactions with more than 2 accounts. I imagine this would only require an extension of the field names that are currently allowed. It should be simpler than https://github.com/simonmichael/hledger/issues/507.
- Tracking cost basis would be really helpful. Currently, I have to look up the cost basis of a lot and add it. This can probably not be fully automated because there are so many rules when positions are sold partially. However, it can be checked that the cost the user enters exists at least. I believe (but have not verified) that this is what ledger does. This is discussed, for instance, in https://github.com/simonmichael/hledger/issues/82 and https://github.com/simonmichael/hledger/issues/488. How hard is that feature to implement?
- `register` could show where transactions came from. That would make some reports easier.
- Can hledger calculate inline? Say `Expense:AAPL:Commission   USD $(750-741.9)`? Note: this should be in addition to deducing the last entry. Does this feature make sense to you?
- Mark-to-market performance as discussed in https://github.com/simonmichael/hledger/issues/329 would be nice but, for me, it is just "nice to have". I'd suggest tracking only realized performance, i.e., actual transactions for now.
[...]

PR #629, opened 2017-10-16
Docs investment guide

MatthiasKauer:
As discussed in https://github.com/simonmichael/hledger/issues/624.
I have only checked the markdown output on github.


simonmichael:
This looks really good. Thank you!

PR #630, opened 2017-10-19
(WIP) allow more postings in csv-imported transactions

MatthiasKauer:
Refers to #627

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -512,8 +512,10 @@ journalfieldnamep = do
 journalfieldnames = [
    "account1"
   ,"account2"
-  ,"amount-in"
-  ,"amount-out"
+  ,"account3"
+  ,"account4"
+  ,"amount3"
+  ,"amount4"
   ,"amount"
   ,"balance"
   ,"code"
@@ -686,6 +688,29 @@ transactionFromCsvRecord sourcepos rules record = t
       ,"the parse error is:      "++show err
       ]
 
+    -- TODO: Are different currencies also needed?
+    -- TODO: Remove amount-in / amount-in
+    -- TODO: Error messages are not helpful
+    -- TODO: Handle cases where only amount or only amount are defined
+
+    amount3    = getAmount rules record currency "amount3"
+    amount4    = getAmount rules record currency "amount4"
+    account3    = maybe "" render (mfieldtemplate "account3")
+    account4    = maybe "" render (mfieldtemplate "account4")
+
+    posting1 = Just posting {paccount=account1, pamount=amount1, ptransaction=Just t, pbalanceassertion=balance}
+    posting2 = Just posting {paccount=account2, pamount=amount2, ptransaction=Just t}
+
+    posting3 = case (T.pack account3) of
+                  ("")  -> Nothing
+                  (acct) -> Just posting {paccount=acct, ptransaction=Just t,
+                    pamount=amount3}
+    posting4 = case (T.pack account4) of
+                  ("")  -> Nothing
+                  (acct) -> Just posting {paccount=acct, pamount=amount4, ptransaction=Just t}
+
+    postings = [posting1, posting2, posting3, posting4]
+
     -- build the transaction
     t = nulltransaction{
       tsourcepos               = genericSourcePos sourcepos,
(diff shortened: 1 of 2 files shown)


simonmichael:
@MatthiasKauer could you amend this commit to fix the failing test and re-push it (force-push to this same PR is fine).

simonmichael:
PS also rebase against latest master if possible.

MatthiasKauer:
I have taken a look and I'm not sure how this case should be supported in the future.

My idea would be to remove support for "amount-in" and "amount-out" entirely.
The csv file from the test 

    10/2009/09,Flubber Co,50,
    11/2009/09,Flubber Co,,50

could then be parsed by rules
```
account1 Assets:MyAccount
date %1
date-format %d/%Y/%m
description %2
amount1 %3
amount2 %4
currency $
```

First of all, *are you okay with such a breaking change?*

If so, could you give me a pointer to the implementation?
As I understand it now, getAmount returns a `MixedAmount` or throws an error.
I need to allow some such errors to occur for some of the amounts so `amount2` isn't mandatory.
Having neither amount1 nor amount2 should constitute an error though, so the group of amounts needs to be checked.
Once this is done, I need to create postings with empty amount such that hledger later infers the amount. Is this possible or are you enforcing the amount? If it is enforced can I somehow use the same deduction function hledger uses to calculate the missing amount here. With values in amt1, amt2, amt3, and the existence of amt4, I would then infer the value of amt4.

If postings without amount are allowed, I would simply output the following, however:

```
2009/09/10 Flubber Co
    Assets:MyAccount             $50
    income:unknown

2009/09/11 Flubber Co
    Assets:MyAccount
    expenses:unknown        $50
```

MatthiasKauer:
I have played with the REPL some more now and also read the [docs for Amount](http://hackage.haskell.org/package/hledger-lib-1.4/docs/Hledger-Data-Types.html#t:Amount) 
and [MixedAmount](http://hackage.haskell.org/package/hledger-lib-1.4/docs/Hledger-Data-Types.html#t:MixedAmount).
It appears that an empty Mixed amount can be created via `Mixed []` but it prints `0` and I'm not sure how it would print in a posting yet.
An alternative I have found is encoding the error in a maybe.
```
either (\x -> Nothing) (Just . Mixed . (:[])) $ runParser (evalStateT (amountp <* eof) mempty) "" $ pack "USD -10"
<interactive>:49:10: warning: [-Wunused-matches]
    Defined but not used: ‘x’
Just USD -10
```
This loses the error message, however, and `either` should be just as flexible as `maybe` anyway, right?

New plan: Parse all amount values into a list `[Either (ParseError Char MPErr) Amount]`. Ensure that the number of "right" values is at least "number of specified accounts - 1". If so, calculate the implied amount somehow. If not, output an error containing all individual parse errors.

What do you think?

simonmichael:
Making amount-in/amount-out unnecessary sounds nice! I'd be fine with that if there's another way to do it and things are still robust. Amount inferring is done by balanceTransaction and journalBalanceTransactions. Note how fearsomely complicated these are, partly to support [balance assignment](http://hledger.org/manual.html#balance-assignments). Perhaps it's simpler when reading CSV, or perhaps not. Experimentation needed for sure.

MatthiasKauer:
Although far from finished, I have made further progress with this.
I can now handle the situation where amount1 and amount2 (previously amount-in/out) are specified.

Current problems:

- postings are not in the correct order. I intend to fix that by creating a type for (Account, Amount, Account Idx) instead of the pairs (Account, Amount) I now use and sorting by Idx later on.
- Error messages are not there yet. This will hopefully not be too hard.
- Rule validation should be adjusted
- Docs should be adjusted
- my usage of `costOfMixedAmount (-amt)))` should be tested

MatthiasKauer:
@simonmichael Should I continue with the points I discussed above?

simonmichael:
I need to spend some time on this to comment usefully. Thanks for the reminder, this is top of queue.

simonmichael:
Do these commits both increase the number of postings that can be generated (allowing 2, 3 or 4), and also remove the need for amount-in/amount-out ? Are these things necessarily connected ?

simonmichael:
I'm not sure what the "transformed to (Maybe Account, Either Amount)" commit is for, either from the code changes or the description ?
Why must the "amount" keyword be dropped ? We don't really want to break all existing CSV rules.

MatthiasKauer:
Thanks for getting back to me. Let me address your comments.

> Do these commits both increase the number of postings that can be generated (allowing 2, 3 or 4), and also remove the need for amount-in/amount-out ? Are these things necessarily connected ?

Increasing the number of postings is the intention. amount-in/amount-out have been replaced with amount1/amount2 because I found it tricky to handle them in the presence of amount2.

> Why must the "amount" keyword be dropped?

Fair point. We should just treat amount as an alias of amount1, shouldn't we?

MatthiasKauer:
> I'm not sure what the "transformed to (Maybe Account, Either Amount)" commit is for, either from the code changes or the description ?

Here I am trying to deal with the fact that account1, 2, 3 and amount1, 2, 3 may or may not be specified in the rules file and/or not be present in the csv file. If an account is not specified, we don't need an amount for it but maybe we should warn if someone specifies amount3 without account3.
A missing amount, however, has to be supported to have the functionality that amount-in/amount-out fulfilled previously. In this commit, I am trying to calculate the one missing amount which is implied by the others. 

I don't know a priori which of the amounts would be missing so I worked on this list of Maybe and Either pairs.
This is probably not idiomatic at all given my limited Haskell experience.

simonmichael:
Thanks @MatthiasKauer. With your comments and a second read-through I understand better. I know this is WIP; it breaks some tests as you know, it doesn't tackle docs yet, and I find it still a bit too tricky to know what's going on. Perhaps these commits would be better combined into one, with more of the old stuff cleaned out. Or we could find some self-complete, semantically meaningful, but smaller steps. I will see if I can come up with anything, unless you beat me to it.

MatthiasKauer:
Thank you for taking another look at this @simonmichael 
I realize that I have been somewhat lax with my commit messages here. I apologize and will try to include more information in the next commits or the accompanying messages here.

Regarding squashing commits: I would do that just before merging. Now, the separation still helps me to understand my thought process along the way and lets me back out to intermediate states.
You can see the overall diff on github too, no?

As I mentioned above, I think the main functional challenge is currently that postings are not in the correct order (meaning account3 may appear before account1).
After that, there are many fixes on the boundary: error reporting, docs, testing.

We could accept that postings are not in order but we don't have to. I believe it can be fixed by creating a triple type for (Account, Amount, Account Idx) instead of the pairs (Account, Amount) I now use and sorting by Idx later on. Is that a good approach?

Although I won't (be able to) stop you from implementing the changes yourself, I am quite interested in bringing this to a conclusion. Should I continue with what I suggested? Or should we break this up into smaller steps like: 1) replace amount-in/out with amount(1)/amount2 functionality, 2) allow additional accounts with amounts that must be present (cannot be implied) 3) allow implied amounts in all accounts

What do you think?

hledgerlist:
We are adding new tricks to something that is already quite tricksy. I don't like to step on your toes or abandon work, but  I am quite doubtful that we will succeed in making this robust and intuitive, by adding more complexity in the code, especially when I don't understand the proposed new functionality in detail. 

My approach would be to document first; it's cheaper than coding and code review. So I have been thinking about how it can work for the user, and how to document it. I am currently cleaning up the existing CSV docs as a way to prime the pump.

simonmichael:
Cleanups now deployed at http://hledger.org/csv.html - just expanded the Description so far. I am done for today. Happy to comment on end-user doc ideas/drafts. 

I won't have much time for code review in the coming week. As a simplifying constraint, let's say I'll prioritise reviewing "merge-ready" commits - rebased against master, free of regressions, clear etc.

simonmichael:
PS and please excuse me for getting a bit more "efficient" with this, but I can't afford more. If you can attract more eyeballs/reviewers to this topic, that might be another way to get more input.

simonmichael:
Design thoughts posted, back on #627.

simonmichael:
Closing, see #627.

Issue #487, opened 2017-01-17
Use of digit groups separator requires presence of decimals


ony:
Consider example:
```sh
% COLUMNS=80 hledger reg -f -
commodity EUR
	format 1,000.00 EUR

commodity UAH
	format 1.000,00 UAH

2017/1/1
	expenses  1,000 EUR
	expenses  1.000 EUR
	expenses  1,000.00 EUR
	assets

2017/1/1
	expenses  1,000 UAH
	expenses  1.000 UAH
	expenses  1.000,00 UAH
	assets
```
This results in
```
2017/01/01                      expenses                  1.00 EUR      1.00 EUR
                                expenses                  1.00 EUR      2.00 EUR
                                expenses              1,000.00 EUR  1,002.00 EUR
                                assets               -1,002.00 EUR             0
2017/01/01                      expenses                  1,00 UAH      1,00 UAH
                                expenses                  1,00 UAH      2,00 UAH
                                expenses              1.000,00 UAH  1.002,00 UAH
                                assets               -1.002,00 UAH             0
```
Which is pretty unexpected.
It should be:
```
2017/01/01                      expenses              1,000.00 EUR  1,000.00 EUR
                                expenses                  1.00 EUR  1,001.00 EUR
                                expenses              1,000.00 EUR  2,001.00 EUR
                                assets               -2,001.00 EUR             0
2017/01/01                      expenses                  1,00 UAH      1,00 UAH
                                expenses              1.000,00 UAH  1.001,00 UAH
                                expenses              1.000,00 UAH  2.001,00 UAH
                                assets               -2.001,00 UAH             0
```

I have to be very careful when I write numbers with thousands and always put decimals separator to avoid treating my numbers smaller than they are in fact.

`hledger` should honor decimals and digit groups specification of commodity format.

I don't mind to use space as a separator, but this isn't allowed yet. See #330 .

simonmichael:
It's not hard at all to confuse the style inferrer. Does specifying the style with a [commodity directive](http://hledger.org/journal.html#commodity-directive) help ?

Given a clear specification of what *should* happen above, we could improve style inference, maybe. I'm not sure it's the right approach though.

simonmichael:
Oh you gave the expected result. I admit I haven't spent the time to understand what's happening here. Given the difficulty of inferring style correctly for all possible confusing inputs, I'm not even sure yet if this is a bug, or something we can't reasonably handle. Improvements to code or docs welcome.

[...]

PR #635, opened 2017-10-28
Bug/487 digits groups separator ambiguity

### LINKED ISSUE #399: Can't Use Thousand Separators Without Using A Decimal Separator
Thousand separators in an amount number can't be used unless the number also has a decimal separator. If there is only one punctuation mark, it's always considered to be a decimal separator even if one or more of the following are present:
- A "D directive" with:
  - A sample amount using only a thousand separator.
  - A sample amount using both thousand and decimal separators.
- A commodity directive that defines a format similar to the sample amounts listed above.
- A preceding transaction with an amount that uses a format similar to the sample amounts listed above.

For example:

```
D JPY 111,111,111
# also doesn't work:
# D JPY 111,111,111.0
# D JPY 1,111
# D JPY 1,111.0

2016/01/01 default_commodity
  (asset:no_comma)  100
  (asset:comma)  100,500
```

The balance sheet for this journal adds up to ""200,500" or "200.500" for either commodity, rather than "100,600".

Using a commodity directive also has incorrect totals. An unexpected value rounding also occurs with the format "JPY 111,111,111".

```
commodity JPY
  format JPY 111,111,111 # also causes unexpected rounding

#  format JPY 111,111,111.0
#  format JPY 1,111
#  format JPY 1,111.0

2016/01/01 yen
  (asset:commodity:no_comma)  JPY 100
  (asset:commodity:comma)  JPY 100,500
```

Using a preceding dummy transaction with a sample amount has the same output as the cases above:

```
2016/01/01 init
  (asset:commodity:init)  JPY 111,111,111
  (asset:commodity:init)  JPY -111,111,111
  # same output with the formats tried in other examples

2016/01/01 yen
  (asset:commodity:no_comma)  JPY 100
  (asset:commodity:comma)  JPY 100,500
```
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37935078-can-t-use-thousand-separators-without-using-a-decimal-separator?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ony:
Trust latest `commodity` or `D` (default commodity) directives in case of ambiguity in parsing number.

This kinda resolves #399 and its duplicate #487. We still may get issues caused by misinterpreting numbers `1,000` and `1.0` both as `1` if no `commodity` directive or default commodity `D` directive were provided in same journal file earlier.


hledger-lib/Hledger/Read/Common.hs
```diff
@@ -145,6 +146,12 @@ setDefaultCommodityAndStyle cs = modify' (\j -> j{jparsedefaultcommodity=Just cs
 getDefaultCommodityAndStyle :: JournalParser m (Maybe (CommoditySymbol,AmountStyle))
 getDefaultCommodityAndStyle = jparsedefaultcommodity `fmap` get
 
+getDefaultDecimalHint :: JournalParser m (Maybe Char)
```
simonmichael: Let's add short haddocks explaining what these do.

simonmichael:
I trust your judgement that this works better. We should have haddocks for the two get*Style functions, and an update to the manual explaining how things work now.

ony:
@simonmichael , as you probably noticed expecting that `format` directive will affect interpretation is natural. But actual behavior is pretty hard to document. Because mostly it something like:
"If you have issue with misinterpreting decimals try putting `commodity` directive in the begining of the file" and that was a first thought of me, and as I remember yours also.

If you think we should have a separate directive for that (probably still currency-specific) it should be easy to add that.

simonmichael:
The doc looks reasonable to me. Merging, thank you!

PR #637, opened 2017-10-31
doc: fix readme markdown formatting on github

mickdekkers:
An update to GitHub's markdown parser made it so a space is required between hash characters and the text of titles for them to be displayed correctly.


simonmichael:
Thanks!

PR #639, opened 2017-11-02
Allow to use space as a digit groups separator

### LINKED ISSUE #330: empty space as digit group?
I find it easier to read: `100 000`, than `100.000`, also because space is easier to type. The currency here in Chile is very 'low' compared to USD, so we normally have to use a lot of `.` (miminal wage per month is around `CLP 150 000`).

I tried to write (since we don't use decimals):

```
D $ 1 000,
```

But of course didn't work, so I was wondering if it was possible to be implemented :)
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31711049-empty-space-as-digit-group?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ony:
Note that this pull request shares some common history with #635, but still can go independently.

Resolves #330

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -24,6 +24,7 @@ import Prelude.Compat hiding (readFile)
 import Control.Monad.Compat
 import Control.Monad.Except (ExceptT(..), runExceptT, throwError) --, catchError)
 import Control.Monad.State.Strict
+import Data.Char
 import Data.Data
 import Data.Default
 import Data.Functor.Identity
@@ -599,7 +600,7 @@ rawnumberp = do
     (firstSep, groups) <- option (Nothing, []) $ do
         leadingDigits <- some digitChar
         option (Nothing, [leadingDigits]) . try $ do
-            firstSep <- oneOf sepChars
+            firstSep <- oneOf sepChars <|> whitespaceChar
             groups <- some digitChar `sepBy1` char firstSep
             return (Just firstSep, leadingDigits : groups)
 
@@ -614,11 +615,14 @@ rawnumberp = do
         return (lastSep, fromMaybe [] digits)
 
     -- make sure we didn't leading part of mistyped number
-    notFollowedBy $ oneOf sepChars
+    notFollowedBy $ oneOf sepChars <|> (whitespaceChar >> digitChar)
 
     return $ dbg8 "rawnumberp" (firstSep, groups, extraGroup)
     <?> "rawnumberp"
 
+-- | Parse a unicode char that represents any non-control space char (Zs general category).
+whitespaceChar :: TextParser m Char
+whitespaceChar = charCategory Space
 
 -- test_numberp = do
 --       let s `is` n = assertParseEqual (parseWithState mempty numberp s) n
--- a/hledger-lib/doc/hledger_journal.5.m4.md
+++ b/hledger-lib/doc/hledger_journal.5.m4.md
@@ -267,7 +267,8 @@ Amounts consist of a number and (usually) a currency symbol or commodity name. S
   `3 "green apples"`\
   `-$1,000,000.00`\
   `INR 9,99,99,999.00`\
-  `EUR -2.000.000,00`
+  `EUR -2.000.000,00`\
+  `1 999 999.9455`
 
 As you can see, the amount format is somewhat flexible:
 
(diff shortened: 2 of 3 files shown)


simonmichael:
I trust your judgement that this works better. We should have haddocks for the two get*Style functions, and an update to the manual explaining how things work now.

simonmichael:
Oops, sorry that was for #635.

ony:
@simonmichael ok. I was about to write that the most important is 05bd421 and last commit is mostly a sample of how to add new alternative digits group separator without allowing it for decimals.

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -580,11 +581,14 @@ rawnumberp = do
         return (lastSep, fromMaybe [] digits)
 
     -- make sure we didn't leading part of mistyped number
-    notFollowedBy $ oneOf sepChars
+    notFollowedBy $ oneOf sepChars <|> (whitespaceChar >> digitChar)
```
simonmichael: Space separators are blowing my mind just a little..

What's the reason you don't add space to sepChars ?
Does this accept other whitespace chars, like tab or line feed ?

ony: > Space separators are blowing my mind just a little..

I used space groups separator because that's one of the way most calculators do and that it doesn't mix like `,` with `.` which in some locales flipped. And it is hoarder to distinguish comma and dot in some fonts. Note that in some programming languages underscores are used to visually emphasize magniture like `1_000` which I believ is an attempt to get visually similar character to space.

> What's the reason you don't add space to sepChars ?

Space special and that's one ifts benefit. I think allowing amounts like `$ 1,000 45 @ 2 33 EUR` would be awful.
`sepChars` are generic and can be used either for groups or for decimals separator.

> Does this accept other whitespace chars, like tab or line feed ?

line-feed is a control character and doesn't fit Zs unicode category. Same for tabs.

See [wiki](https://en.wikipedia.org/wiki/Unicode_character_property)

> Includes the space, but not TAB, CR, or LF, which are Cc

simonmichael: I think support for space digit group separator is nice. https://en.wikipedia.org/wiki/Decimal_mark#Digit_grouping indicates increasing support for it.

So it sounds like it accepts a (single) genuine space only. Wikipedia also suggests unicode thin space, but I'm not sure if we really want to go there - non-obvious unicode characters always end up causing problems somewhere..

Ah I see, we should probably split up sepChars some time.

simonmichael:
Looking good, needs to be added to manual.

hledger-lib/Hledger/Read/Common.hs
ony: @simonmichael with this change we support [any kind of space within Zs category](https://www.compart.com/en/unicode/category/Zs) including "thin space" `U+2009`. I prefer to use non-breakinging space `U+00A0` in word processors (like LyX) to ensure that numbers are not teared apart.

Here we actually have 3 sets of characters:
- that can be used either as digit group separator or decimal part separator `sepChar`
- that can only be used as digit groups separator
- that can only be used as decimal part separator (currently empty)

Since last two groups always used in one place there is no much point in factoring out them.
The only problem is validation if number is finished correctly and not followed by something. With space you cannot just check that number is not followed by space because it might be used to separate number from commodity/comment/price. Thus we have that `whitespaceChar >> digitChar` instead of just `whitespaceChar`.

ony:
> Looking good, needs to be added to manual.

Done. I slightly changed original wording to avoid confusion causing "I decided to use commas for thousands" to be in a head without telling it somehow to hledger.
Also I removed plurals from "commas", "periods" to avoid assumption like use of double-space or double-comma is allowed.

hledger-lib/Hledger/Read/Common.hs
simonmichael: Huh. So this change allows ascii space and *16* other kinds of unicode space.

Although we allow unicode in the journal, somehow this feels likely cause complications, hair-pulling, teeth-gnashing in future. I can also see the value of a non-breaking space in your use case. I guess I won't object if the list is for this.

ony: > Although we allow unicode in the journal, somehow this feels likely cause complications, hair-pulling, teeth-gnashing in future. I can also see the value of a non-breaking space in your use case. I guess I won't object if the list is for this.

I'm not sure I understand your concerns. Among all of the space characters ascii space is the most dangerous one because it used as delimiter of number and currency which at this moment can also contain non-ascii space characters because of `spacenonewline`.

I'd suggest to delegate most of the decision about interpreting characters to parser (`megaparsec`).

I think giving an unequal **treating** of whitespace characters actually is worse. Imagine that copying from posts from blogs and word processors that tend to flip those characters as they want resulting in journal where two spaces delimiter doesn't separate account from amount anymore because they were replaced with some long spaces.

Use of [categories](https://hackage.haskell.org/package/base-4.9.1.0/docs/Data-Char.html#t:GeneralCategory) will help to get rid of that burden at least for space separators (not that name of category pretty matching usage).

P.S. I've noticed that `spacenonewline` allows [vertical tab character (`\v`)](https://www.compart.com/en/unicode/U+000B) which appear like:
```shell
zsh% echo -e 'A\vB'
A
 B
```

simonmichael:
Thanks!

Issue #402, opened 2016-09-14
Problem : Parser errors are not informative.


crocket:
```
commodity BTC
  note Bitcoin
```

When I parse the above journal with hledger, it should report that `note` is not a supported subdirective of commodity directive.

Instead, the parser reports the following error.

```
hledger: ledger.journal:1:1:
unexpected 'c'
expecting end of input or transaction or directive
```

The same happens when `account` directive contains unsupported subdirectives and when the end of a journal file is not a new line. When the end of a journal file is not a newline, the parser reports an error at the beginning of the last transaction or any last directive.
##

simonmichael:
Agreed, thanks.

karan-ta:
I started using timedot format and created a new file with contents :  
2016/2/2
client1   .... ....
research  .

but i got error - 
using conversion rules file /karanahuja.timedot.rules
hledger: "/karanahuja.timedot" (line 2, column 1):
unexpected 'i'
expecting journal transaction or directive or end of input

simonmichael:
That looks strange. I can't reproduce here with hledger 1.2..

ony:
@crocket I agree that errors sometimes quiet hard to understand because of excessive backtracking in journal parser.

Note that it is hard to identify "not supported" because it means that we should already partially support that sub-directive in parser. On the other hand most of the parser able to generate meaningful error message with their expectations. Example:

```
hledger: -:2:3:
unexpected 'n'
expecting "format"
```

ony:
With newer `MegaParsec` you can get even better error report:

```
hledger: -:2:3:
unexpected "note B"
expecting "format"
```

simonmichael:
The specific error reported by this issue has been improved: https://github.com/simonmichael/hledger/pull/640#issuecomment-347354365

PR #640, opened 2017-11-03
Improve error reporting for directives

ony:
- Factor out `wordsp`.
- Limit scope of directives backtracking.

Resolves simonmichael/hledger#402

### DIFF
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -81,6 +81,8 @@ import Control.Monad.State.Strict
 import qualified Data.Map.Strict as M
 import Data.Monoid
 import Data.Text (Text)
+import Data.String
+import Data.List
 import qualified Data.Text as T
 import Data.Time.Calendar
 import Data.Time.LocalTime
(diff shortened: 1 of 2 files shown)


ony:
There is some issue with Travis:

```
$ ./.travis-maybe-skip.sh
fatal: Invalid symmetric difference expression master...8b50b568b351bd59264eae3be40af7a12f793d65
```

While this pull request references 7133dea

simonmichael:
A before/after example would make this a no brainer, but this sounds good. Thanks!

simonmichael:
```
$ cat a.j
commodity BTC
  note Bitcoin
$ hledger-1.4 -f a.j print
hledger-1.4: a.j:1:1:
unexpected 'c'
expecting end of input or transaction or directive

$ hledger -f a.j print
hledger: /Users/simon/src/hledger/a.j:2:3:
unexpected "note B"
expecting "format"

```
This is much better. Would be better still if it didn't imply the format subdirective is required.

ony:
> This is much better. Would be better still if it didn't imply the format subdirective is required.

But it actually required there unless it is an empty line. If you would drop indentation for `note Bitcoin` you'll get other error:

```
hledger: -:2:1:
unexpected 'n'
expecting end of input or transaction or directive
```

I guess there is a way to add "or comment or end of line", but I think it will make that error even more "technical".

PR #643, opened 2017-11-17
Allow hledger-budget to generate transactions past the end date of th…

adept:
…e last real transaction if -b/-e are supplied


simonmichael:
Sounds useful, thanks!

ony:
Thanks. I used to put empty transaction in future to achieve this.

PR #645, opened 2017-11-18
added --forecast to balancemode, print, register

adept:
See PR #644 for inspiring discussion

### DIFF
--- a/hledger-lib/Hledger/Data/AutoTransaction.hs
+++ b/hledger-lib/Hledger/Data/AutoTransaction.hs
@@ -155,4 +155,6 @@ runPeriodicTransaction pt = generate where
         case parsePeriodExpr errCurrent periodExpr of
             Left e -> error' $ "Failed to parse " ++ show (T.unpack periodExpr) ++ ": " ++ showDateParseError e
             Right x -> x
-    generate jspan = [base {tdate=date} | span <- interval `splitSpan` spanIntersect effectspan jspan, let Just date = spanStart span]
+    generate jspan = 
+      let effectspan' = spanIntersect effectspan jspan 
+          in [base {tdate=date} | span <- interval `splitSpan` effectspan', let Just date = spanStart span, spanContainsDate effectspan' date]
--- a/hledger/Hledger/Cli/Utils.hs
+++ b/hledger/Hledger/Cli/Utils.hs
@@ -31,7 +31,7 @@ import Data.List
 import Data.Maybe
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
-import Data.Time (Day)
+import Data.Time (Day, addDays)
 import Data.Word
 import Numeric
 import Safe (readMay)
@@ -70,6 +70,7 @@ withJournalDo opts cmd = do
           . anonymiseByOpts opts
           . journalApplyAliases (aliasesFromOpts opts)
         <=< journalApplyValue (reportopts_ opts)
+        <=< journalAddForecast opts
   either error' f ej
 
 -- | Apply the pivot transformation on a journal, if option is present.
@@ -116,6 +117,31 @@ journalApplyValue ropts j = do
                 | otherwise
                 = id
     return $ convert j
+
+-- | Run PeriodicTransactions from journal from today or journal end to requested end day.
+-- Add generated transactions to the journal
+journalAddForecast :: CliOpts -> Journal -> IO Journal
+journalAddForecast opts j = do
+  today <- getCurrentDay
+  let startDate =
+        -- Create forecast starting from end of journal + 1 day, and until the end of requested reporting period
+        -- If end is not provided, do 180 days of forecast
+        case spanEnd (jdatespan j) of
+          Just jEnd -> addDays 1 jEnd
+          Nothing   -> today
+      endDate = fromMaybe (addDays 180 today) $ periodEnd (period_ ropts)
+      dates = DateSpan (Just startDate) (Just endDate)
+      withForecast = [makeForecast t | pt <- jperiodictxns j, t <- runPeriodicTransaction pt dates] ++ (jtxns j)
+      makeForecast t = txnTieKnot $ t { tdescription = T.pack "Forecast transaction" }
+      ropts = reportopts_ opts
+  if forecast_ ropts 
+    then return $ journalBalanceTransactions' opts j { jtxns = withForecast }
+    else return j
+  where      
+    journalBalanceTransactions' opts j =
+      let assrt = not . ignore_assertions_ $ inputopts_ opts
+      in
+       either error' id $ journalBalanceTransactions assrt j
 
 -- | Write some output to stdout or to a file selected by --output-file.
 -- If the file exists it will be overwritten.
--- a/hledger/Hledger/Cli/Utils.hs
+++ b/hledger/Hledger/Cli/Utils.hs
@@ -123,12 +123,10 @@ journalApplyValue ropts j = do
 journalAddForecast :: CliOpts -> Journal -> IO Journal
 journalAddForecast opts j = do
   today <- getCurrentDay
-  let startDate =
-        -- Create forecast starting from end of journal + 1 day, and until the end of requested reporting period
-        -- If end is not provided, do 180 days of forecast
-        case spanEnd (jdatespan j) of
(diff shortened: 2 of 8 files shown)


hledger/Hledger/Cli/Utils.hs
```diff
@@ -117,6 +118,31 @@ journalApplyValue ropts j = do
...
+  today <- getCurrentDay
+  let startDate =
+        -- Create forecast starting from today or end of journal, whatever is earliest, and until the end of requested reporting period
+        -- If end is not provided, do 180 days of forecast
+        case spanEnd (jdatespan j) of
+          Just jEnd -> min jEnd today
+          Nothing -> today
+      endDate = fromMaybe (addDays 180 today) $ periodEnd (period_ ropts)
+      dates = DateSpan (Just startDate) (Just endDate)
+      withForecast = [makeForecast t | pt <- jperiodictxns j, t <- runPeriodicTransaction pt dates] ++ (jtxns j)
```
ony: > Regarding `--forecast`: i think that this switch should "expand" periodic transaction, and automatic transactions might be triggered in addition with `--auto`. Do you envision problems/limitations here?

I see. I thought that you were going to expand on a whole period of journal and thus overlap with real expenses. That's why I suggested to drop real transactions from journal and show only generated from periodic transactions, but for `balance` report it is really useful to have them.

Potential problem here is that you may get forecast overlapping with real expenses if your last transaction/today is not at the end of budget period.

```
2016/12/20
    expenses:housing:rent  $600
    assets:cash

~ monthly from 2016/1/25
    income  $-1000
    expenses:housing  $700
    assets
```

As a forecast based on budget I would expect to see $600 under `expenses:housing:rent` and $100 under `expenses:housing`.
But if you'll consider "periodic transactions" as independent of budgeting - then it is totally fine.

hledger/Hledger/Cli/Utils.hs
```diff
@@ -117,6 +118,31 @@ journalApplyValue ropts j = do
...
+journalAddForecast :: CliOpts -> Journal -> IO Journal
+journalAddForecast opts j = do
+  today <- getCurrentDay
+  let startDate =
+        -- Create forecast starting from today or end of journal, whatever is earliest, and until the end of requested reporting period
+        -- If end is not provided, do 180 days of forecast
+        case spanEnd (jdatespan j) of
+          Just jEnd -> min jEnd today
+          Nothing -> today
+      endDate = fromMaybe (addDays 180 today) $ periodEnd (period_ ropts)
```
ony: Why we need this ~6 extra months outside of report period?

hledger/Hledger/Cli/Commands/Balance.hs
```diff
@@ -283,6 +283,7 @@ balancemode = (defCommandMode $ ["balance"] ++ aliases) { -- also accept but don
      ,flagReq  ["format"] (\s opts -> Right $ setopt "format" s opts) "FORMATSTR" "use this custom line format (in simple reports)"
      ,flagNone ["pretty-tables"] (\opts -> setboolopt "pretty-tables" opts) "use unicode when displaying tables"
      ,flagNone ["sort-amount","S"] (\opts -> setboolopt "sort-amount" opts) "sort by amount instead of account name"
+     ,flagNone ["forecast"] (\opts -> setboolopt "forecast" opts) "generate forecast transactions"
```
ony: You can add it to `reportflags` in `hledger/Hledger/Cli/CliOptions.hs` to get all report commands affected similar to how `--yearly` and `--costs` are defined and can be seen even in `stat` and `rewrite`.

We may want to introduce `budgetingFlags` or something like that if we want target subset of reports.

adept: Thanks, i missed reportflags. Fixing

hledger/Hledger/Cli/Utils.hs
adept: I'm of the opinion that for the best result end date have to be explicitly specified, but if not, there should be some sensible default. Not sure that 180 days is sensible, though.

Note that it only applies if -e is not given

hledger/Hledger/Cli/Utils.hs
adept: You have a point. It is best to start forecast immediately after the end of the journal (on the next day). Updating pr

hledger/Hledger/Cli/Utils.hs
ony: Oh. Sorry I've missed that it actually adds a default report end date.

> I'm of the opinion that for the best result end date have to be explicitly specified, but if not, there should be some sensible default. Not sure that 180 days is sensible, though.

I'm not sure if it makes sense but I thought that it might be something like `startDate` (forecast) + "longest period among all periodic transactions".
But it will require more changes to get `Interval` parsing extracted out of `runPeriodicTransaction`.

So I guess it should be fine to leave 180 days as a default value for `--forecast` without explicit `-e`.

hledger/Hledger/Cli/Utils.hs
ony: I actually didn't thought about that. I referred to overlap happending during reporting. If you use  `--forecast -M` and example I provided you'll get under category `expenses:housing` amount of`$1300` for `2016/12` because from periodic transaction we'll generate same `$700` amount though in a same budget period we already paid `$600`.

```
           || 2016/12 
===========++=========
 assets    ||   $-300 
   cash    ||   $-600 
 expenses  ||   $1300 
   housing ||   $1300 
     rent  ||    $600 
 income    ||  $-1000 
-----------++---------
           ||       0 
```

It feels like for budget-based forecasting it is a bit wrong and you should just expect that by the end of `2016/12` you'll spend `$700` (extra `$100` on top of real posting `$600`).

But "budget period" that's a notion that we are missing right now. I don't think therer is any actions at this moment regarding this issue.

ony:
@simonmichael , if you don't have any comments to address you can merge, I guess.

adept:
squashed

adept:
I've pushed more changes here: 
- bugfix to ensure runPeriodicTransactions honors given DateSpan
- change for runPeriodicTransactions to honor "effective from" date, so now "yearly from 2017-11-12" would first occur on 2018-11-12, and not on 2018-01-01 (works for all intervals)

hledger-lib/Hledger/Data/AutoTransaction.hs
```diff
@@ -155,6 +155,20 @@ runPeriodicTransaction pt = generate where
         case parsePeriodExpr errCurrent periodExpr of
             Left e -> error' $ "Failed to parse " ++ show (T.unpack periodExpr) ++ ": " ++ showDateParseError e
             Right x -> x
+    -- If transaction is specified with effective span, e.g. "yearly from 2017-10-15", make sure it happens
+    -- on the specified start date and then every <interval> after that (2018-10-15, 2019-10-15, ...).
+    -- This is done by computing offset of the start date from the beginning of the nearest <interval>
+    -- and shifting each computed transaction date by this offset
+    effectiveFrom = spanStart effectspan
+    startOfEarliestEnclosingInterval = spanStart $ head $ interval `splitSpan` (DateSpan effectiveFrom Nothing)
```
ony: This feature is already here. You can trigger it with a doctest:

```haskell
-- >>> mapM_ (putStr . show) $ runPeriodicTransaction (PeriodicTransaction "every 2nd day of month" ["hi" `post` usd 1]) (mkdatespan "2017/02/01" "2017/04/02")
-- 2017/02/02
--     hi           $1.00
-- <BLANKLINE>
-- 2017/03/02
--     hi           $1.00
-- <BLANKLINE>
```

adept:
Ah, i did not know this. Is syntax the same for yearly expressions? it is a bit awkward to compute day-of-year for that

ony:
@adept , there is no `every 1st Jan of year` as I know. But you can add it :)

adept:
Yeah, I like your suggestion for warning/disallowing "yearly from YYYY/MM/DD"...

I also agree that it is bad to introduce different semantics for yearly inside runPeriodicTransactionOnly

adept:
I've created #652 and #653 on the heels of discussion here and will remove all the tweaking I introduced in runPeriodicExpression from this PR

simonmichael:
Superseded by #654.

PR #646, opened 2017-11-18
Auto switch

adept:
See PR #644 for inspiring discussion. Created on top of PR #645

### DIFF
--- a/bin/hledger-budget.hs
+++ b/bin/hledger-budget.hs
@@ -13,7 +13,6 @@ import Data.List
 import Data.String.Here
 import System.Console.CmdArgs
 import Hledger.Cli
-import Hledger.Data.AutoTransaction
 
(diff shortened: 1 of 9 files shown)


ony:
I can see benefit of having automated postings enabled on demand only if some people use them for budgeting and by some reason don't want to see "budget" accounts in reports.
If someone don't use them at all assuming option `--auto` to be always enabled shouldn't affect their reports.

@simonmichael , I don't clearly remember similar discussion regarding `rewrite`. Should we still keep away from executing automated postings that already in journal?

adept:
@ony i tried changing default for auto_ to True in ReportOptions and can confirm that whole of functest passes, so there is no adverse effects from enabling it at all times. Not sure if it has any associated performance penalties though

simonmichael:
Superseded by #654.

PR #651, opened 2017-11-22
Support every 2nd Thursday of month

ony:


hledger-lib/Hledger/Data/Dates.hs
```diff
@@ -756,7 +775,20 @@ reportinginterval = choice' [
...
                             string "month"
-                          return $ DayOfMonth n
+                          return $ DayOfMonth n,
+                       do string "every"
+                          many spacenonewline
+                          n <- fmap read $ some digitChar
+                          thsuffix
+                          many spacenonewline
+                          wd <- weekday
+                          many spacenonewline
```
adept: I think this "many spacenonewline" could be omitted

hledger-lib/Hledger/Data/AutoTransaction.hs
```diff
@@ -146,6 +146,53 @@ renderPostingCommentDates p = p { pcomment = comment' }
...
+-- 2017/01/01
+--     hi           $1.00
+-- <BLANKLINE>
+-- 2017/02/01
+--     hi           $1.00
+-- <BLANKLINE>
+-- 2017/03/01
+--     hi           $1.00
+-- <BLANKLINE>
+-- >>> mapM_ (putStr . show) $ runPeriodicTransaction (PeriodicTransaction "every 30rd day of month from 2017/1 to 2017/5" ["hi" `post` usd 1]) nulldatespan
```
simonmichael: I guess it accepts st|th|rd suffix interchangeably ?

hledger-lib/Hledger/Data/AutoTransaction.hs
```diff
@@ -146,6 +146,53 @@ renderPostingCommentDates p = p { pcomment = comment' }
...
+-- 2017/01/30
+--     hi           $1.00
+-- <BLANKLINE>
+-- 2017/03/30
```
simonmichael: I suspect it's better to adjust to the last day of shorter months (eg 2/28 or 2/29), rather than skipping the month ?

simonmichael:
Needs to be added to docs, http://hledger.org/manual.html#period-expressions I guess.

adept:
Re iso standard - are you talking about iso 8601? Repeating/interval syntax there is very human unfriendly :(

hledger-lib/Hledger/Data/AutoTransaction.hs
adept: All "every Nth" parsers do this currently/already, could be considered feature :)

adept:
It will affect -p argument as well, allowing new period expression there

cfellinger:
Hai,

there are two parts to this reaction:

I remind us of english biased interfaces and of non-existing dates.


be aware of english biased interfaces
=====================================

On Thu, Nov 23, 2017 at 06:24:47PM +0000, Dmitry Astapov wrote:

> Re iso standard - are you talking about iso 8601? Repeating/interval
> syntax there is very human unfriendly :(

Well at least it's not english based.  Really bimonthly and from--to
need translation for non-english speaking kids.  This iso standard
does use letters (DMY) to indicate time span, P to indicate the start
of the period declaration and R to specify the repeat count, that's a
whooping five letters to remember.  Even I can manage that:), whereas
bimonthly and such, do they need a hyphen?  Does wekelijks work?  And
kwartaal? Or am I stuck with weekly and quarterly (or is it quarter)?

I haven't spelled the current proposals, but I'm always put
off bij the use of slash (/) instead of the iso standard dash (-).
It makes me wonder of the order of the date fields, and the
impossebility to cover ALL the current date formats in use
in the world.  Let's use dashes with the sane iso order :]

I do sympathise with trying to mimick natural language period
and date specifiers, but I see many problems if there were
other languages besides *american* english to deal with.

All those problems are gone when using the iso standard. And

  R3/2015-11-01/P3M

doesn't look that complicated compared to

  every three months from 2015-11-01, 3 times in a row

and all the variants that amount to the same.


be aware of non-existing dates
==============================

Then there is the problem of getting to non existing dates
like "jan the 31 next month".  Do you bark, do you silently
move over to the first of the next month, or the the 28e,
and does that one (28) stick. Does the rounding depnnd on
the direction of the repeater (upto versus down-to).

Iso doesn't help there.  In our own periodic scheduler on top of
mledger I added a rounding indicator to stear the handling of such
cases.  The periodic scheduler barks loudly (crashes:) when non
existing dates are requested.  Unless such a rounding indicator was
given.  The increment isn't cummulative, it's always calculated from
the start date, so there is no cummulative error to worry about.

The rounding indicator starts with ± (plus minus sign) followed with
down/inwards/outwards/up.  Yes its english, my kids are now old enough
so they can cope:). But perhaps ± followed by v<>^ would do better, or
simply use utf arrow symbols without the ±, like ← and → for increment
direction agnostic rounding only.

Though iso doesn't specify the handling of such invalid date
traversals its compact notation is more suited to adding the above
rounding indicators.  A language interface would be better served
with something like "rounding down".  Further complicating any
input translation from other languages.

-- 
groetjes, carel

adept:
Nothing in hledger prevents adding ISO8601 parser in addition to existing notations, keeping both camps happy. I dont know if there are existing implementation in haskell, though.

As for the slashes/dashes, this parser as all the rest in hledger, supports variety of date separators, including dash, so full or partial iso8601 dates are supported here and elsewhere.

PS
Unless I am misremembering things, ISO8601 would not easity allow you to specify an equivalent of "2017-" aka "from 2017", since periods without end date require repeating intervals, so for something basic as this you would already need extensions to the ISO syntax.

cfellinger:
On Fri, Nov 24, 2017 at 01:20:19AM +0000, Dmitry Astapov wrote:

> Nothing in hledger prevents adding ISO8601 parser in addition to
> existing notations, keeping both camps happy. I dont know if there
> are existing implementation in haskell, though.

Yes, but no, but yes, the point was actually that keeping the other
camp happy is impossible, the various formats overlap and you can't
parse them correctly unless you know which format to expect. That's
why iso-8601 came along.  Again, I'm all for local dates, but maybe
have an indicator it's in a local format and maybe have a format
specifier in the journal to disambiguate it further.

> As for the slashes/dashes, this parser as all the rest in hledger,
> supports variety of date separators, including dash, so full or
> partial iso8601 dates are supported here and elsewhere.

And here I yokingly suggested to use slash for local formats and
reserve dash together with 4 digits year to indicate iso format.

> PS
> Unless I am misremembering things, ISO8601 would not easity allow
> you to specify an equivalent of "2017-" aka "from 2017", since
> periods without end date require repeating intervals, so for
> something basic as this you would already need extensions to the ISO
> syntax.

Yes you're right, and it's even worse, repeated dates are missing too.
One can use repeating durations (periods) to mimick it.  Fortunately
both problems are solved when we would embrace the latest and greatest
yet to be issued version, see

   http://www.loc.gov/standards/datetime/ISO_DIS%208601-1.pdf

and especially

  http://www.loc.gov/standards/datetime/ISO_DIS%208601-2.pdf

If you thought prior iso 8601 versions were barock, you will love this one :]

It might serve as an interchange/internal format, though it will take
some years to spot all ambiguities and such, but it seems hard to use
without the specs at hand.  Plain english repeat/period expressions
are easier to cough up.

The only thing going for iso-8601 repeat/period notation was its parsability
and its compactness: no pesky spaces, so it's clear where the date spec ends
and the next part (transaction description or such) starts.  We could use a
special marker (say two spaces) to counter that.

ISO 8601 repeat/periods are no longer my favourite :((((
And plain ISO 8601 dates are not that great either:
plenty of people don't use the gregorian calendar.

-- 
groetjes, carel

simonmichael:
Good points, thanks @cfellinger (and now I remember where I learned about [ISO 8601 time intervals](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals)).

The current [period expressions syntax] probably won't ever go away. Adding support for more of ISO 8601 is just a discussion right now. But it's worth thinking carefully about additions to period expressions.

simonmichael:
More on 4: "It should work similarly" - I don't actually know how this will affect existing commands, eg balance and register. This needs testing to make sure it doesn't do something obviously wrong or confusing. I'm guessing it won't affect the boundary dates, and will affect the length of the periods, but since it can specify varying-length periods, what happens there ?

And I forgot:

5. what real problem is this solving ?

I think the answer is: this allows more expressive periodic transactions, which are a proposed feature intended to be useful for budgeting and forecasting.

(So maybe we should settle the question of periodic transactions, before spending more time on this ?)

adept:
@simonmichael you are correct about (5), it was added first and foremost for periodic transactions, and "bled through" to other parts that use periodic expressions. One potential alternative might be to allow "extra forms" just in periodic expressions, and nowhere else.

simonmichael:
Special purpose syntax within period expressions may be harmless.. but I'm curious how irregular periods behave with balance/register. In any case I'll postpone merging this and 652 until periodic transactions settle.

ony:
Closing since already included in #654 and have not much of use without `--forecast` mode .

hledger-lib/Hledger/Data/AutoTransaction.hs
ony: Makes sense. I followed an idea that if we've asked to have it `every 30rd day of the month from 2017/2` then we want excactly that day of a month and we aware that not all months have that day. Same with `every 5th tue of the month in 2017`.
Anyway there is still som undecided stuff regarding periods. Not always we want flexible intervals that spans outside of requested `DateSpan`.

PR #652, opened 2017-11-22
Added support for DayOfYear

adept:


hledger-lib/Hledger/Data/Dates.hs
```diff
@@ -751,12 +771,49 @@ reportinginterval = choice' [
...
+                            many spacenonewline
+                            string "of"
+                            many spacenonewline
+                            string "year"
+                          return $ DayOfYear (read m) n,
+                       do string "every"
+                          many spacenonewline
+                          ("",m,"") <- choice' [month, mon]
+                          many spacenonewline
+                          n <- fmap read $ some digitChar
```
ony: I'm curious if it is possible to simplify this 3 cases with something like

```
do string "every"
  ...
  (m, d) <- choice' [
    ...
    ]
```

Maybe there is permutation/unordered combinator for list of parsers...

adept: I'v been looking for one as well, and failed. But we can factor out optional "of smthng" tail out of 4 parsers, and it will be less verbose...

adept:
real-world use case is tied to --budget/--forecast: you need this if you want to generate events that happen every year, but not on first of Jan

hledgerlist:
Was there something you needed this for in your budgeting ? Real examples are great for docs too.

simonmichael:
Same questions as [651](https://github.com/simonmichael/hledger/pull/651#issuecomment-346825940).

What is the proposed change ? Add "every nov 29th" to period expressions. 

How will this affect things ? Similar to existing "every Nth day of month" form. Needs some testing for intuitive behaviour with the balance command.

What real problem is this solving ? I think, more expressive periodic transactions. Suggesting same conclusion as #651.

adept:
Yes, this is correct - purpose is more expressiveness for periodic transactions. 

I'll add a functest for balance with this period expression

simonmichael:
Superseded by #654/#656.

PR #653, opened 2017-11-22
Make runPeriodicTransaction stricter: it will reject period start days...

adept:
… that do not line up with requested interval


ony:
Eh.... This looks inconvenient as for me... What I meant when I were saying about making stricter against `monthly from 2017/5/4` and `yearly from 2017/5` is to make it as a parase error because granularity of period is more wider than granularity of date range. I.e. `monthly from 2017/5/1` is also wrong way to specify period because `monthly` doesn't honor days anyway. And this makes sense not only for `runPeriodicTransaction`.

adept:
Well, this check could certainly be lifted into parsePeriodExpr. I was trying for the smallest change of existing behavior (however strange it might be)

simonmichael:
Superseded by #654/#656.

PR #656, opened 2017-11-26
Fix backtracking in periodexpr parser

ony:
- Simplify ~and make more strict~ doctests for periodexpr.
- ~Test that expose backtracking issue. Borrowed from simonmichael/hledger#654 .~
- Besides consuming leading space consume ending space for periodexpr also.
- Drop implicit option (def, def) behaviour of periodexpr. I.e. disallow hledger reg -p '' and auto-transaction with heading just '~'.
- Slightly re-factor periodexpr.
- _Ensure that reportinginterval doesn't consume trailing space. Useful if  we'll start disallowing periods like "every1stjan2009-"._


simonmichael:
Ready for rebase against latest master.

ony:
@simonmichael , perfect.  Re-based.

PR #658, opened 2017-11-28
lib: speed up bal/bs by removing needless numeric comparisons

adept:
accountsFromPostings is currently doing excessive work when
adding up postings in each account. There is no need to sort postings
by amount since summing up does not depend on order. It is enough to
sort (for grouping purposes) by accountname only.

### DIFF
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -10,6 +10,7 @@ account, and subaccounting-excluding and -including balances.
 module Hledger.Data.Account
 where
 import Data.List
+import Data.List.Extra (groupSort, groupOn)
 import Data.Maybe
 import Data.Ord
 import qualified Data.Map as M
@@ -63,10 +64,9 @@ nullacct = Account
 accountsFromPostings :: [Posting] -> [Account]
 accountsFromPostings ps =
   let
-    acctamts = [(paccount p,pamount p) | p <- ps]
-    grouped = groupBy (\a b -> fst a == fst b) $ sort $ acctamts
-    counted = [(a, length acctamts) | acctamts@((a,_):_) <- grouped]
-    summed = map (\as@((aname,_):_) -> (aname, sumStrict $ map snd as)) grouped -- always non-empty
+    grouped = groupSort [(paccount p,pamount p) | p <- ps]
+    counted = [(aname, length amts) | (aname, amts) <- grouped]
+    summed =  [(aname, sumStrict amts) | (aname, amts) <- grouped]  -- always non-empty
     nametree = treeFromPaths $ map (expandAccountName . fst) summed
     acctswithnames = nameTreeToAccount "root" nametree
     acctswithnumps = mapAccounts setnumps    acctswithnames where setnumps    a = a{anumpostings=fromMaybe 0 $ lookup (aname a) counted}
@@ -132,7 +132,7 @@ clipAccountsAndAggregate d as = combined
     where
       clipped  = [a{aname=clipOrEllipsifyAccountName d $ aname a} | a <- as]
       combined = [a{aebalance=sum (map aebalance same)}
-                  | same@(a:_) <- groupBy (\a1 a2 -> aname a1 == aname a2) clipped]
+                 | same@(a:_) <- groupOn aname clipped]
 {-
 test cases, assuming d=1:
 
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -75,6 +75,7 @@ import Data.Array.ST
 import Data.Functor.Identity (Identity(..))
 import qualified Data.HashTable.ST.Cuckoo as HT
 import Data.List
+import Data.List.Extra (groupSort)
 -- import Data.Map (findWithDefault)
 import Data.Maybe
 import Data.Monoid
@@ -752,8 +753,7 @@ journalInferCommodityStyles j =
 commodityStylesFromAmounts :: [Amount] -> M.Map CommoditySymbol AmountStyle
 commodityStylesFromAmounts amts = M.fromList commstyles
   where
-    samecomm = \a1 a2 -> acommodity a1 == acommodity a2
-    commamts = [(acommodity $ head as, as) | as <- groupBy samecomm $ sortBy (comparing acommodity) amts]
+    commamts = groupSort [(acommodity as, as) | as <- amts]
     commstyles = [(c, canonicalStyleFrom $ map astyle as) | (c,as) <- commamts]
 
 -- | Given an ordered list of amount styles, choose a canonical style.
--- a/hledger-lib/hledger-lib.cabal
+++ b/hledger-lib/hledger-lib.cabal
@@ -83,6 +83,7 @@ library
     , uglymemo
     , utf8-string >=0.3.5 && <1.1
     , HUnit
+    , extra
   exposed-modules:
       Hledger
       Hledger.Data
@@ -171,6 +172,7 @@ test-suite doctests
     , uglymemo
     , utf8-string >=0.3.5 && <1.1
     , HUnit
+    , extra
     , doctest >=0.8
     , Glob >=0.7
   other-modules:
@@ -259,6 +261,7 @@ test-suite hunittests
     , uglymemo
     , utf8-string >=0.3.5 && <1.1
     , HUnit
+    , extra
     , hledger-lib
     , test-framework
     , test-framework-hunit
--- a/hledger-lib/package.yaml
+++ b/hledger-lib/package.yaml
@@ -70,6 +70,7 @@ dependencies:
 - uglymemo
 - utf8-string >=0.3.5 && <1.1
 - HUnit
+- extra
 # for ledger-parse:
 #- parsers >= 0.5
 #- system-filepath


simonmichael:
Nice. Here's one way to show quick benchmarks:
```
$ quickbench -v -w hledger-1.4,hledger 'hledger -f examples/1000x1000x10.journal bal'
```
(or omit last argument to run standard tests in bench.sh).

simonmichael:
Are we sure this doesn't change the order of the accounts returned ? Eg, might siblings within the tree no longer be alphabetised ?

adept:
Grouping method is not changed (we are grouping by names, as before, and input to groupBy is sorted by name, as before). 

Then, grouped is fully contained in this function : it is used as input to counted or summed, and both of these collapse each group to single entry. After that processing is unchanged. So I don't see how anything related to display could be affected by this change - it should exactly preserve existing semantics.

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -64,7 +64,7 @@ accountsFromPostings :: [Posting] -> [Account]
 accountsFromPostings ps =
   let
     acctamts = [(paccount p,pamount p) | p <- ps]
-    grouped = groupBy (\a b -> fst a == fst b) $ sort $ acctamts
+    grouped = groupBy (\a b -> fst a == fst b) $ sortBy (\a b -> compare (fst a) (fst b)) $ acctamts
```
ony: Whole expression sounds like a `groupSortOn` from [`Data.List.Extra` of "`extra` package](https://hackage.haskell.org/package/extra-1.6/docs/Data-List-Extra.html).

adept: This would internally generate (accountname, (accountname, amount)) list, which sounds a bit silly and might outweight the cost of adding one extra package to the build ... 

Shortening the code to `groupBy ((==) `on` fst) . sortBy (compare `on` fst)` sounds nice, though

ony: The reason why it was written like that is to re-use computations made at sort. Though extra thumbs shouldn't introduce much of performance degradation.

If it have significant impact and we are ok to trade readability we can use:

```hledger
grouped = groupSortBy (compare `on` fst)
```

This implementation assumes that comparison is not very complex and can be re-computed in a similar way we do here.

adept: I might be a bit dense here ... how exactly does it re-use computations made at sort?

ony:
```haskell
groupSortOn f = map (map snd) . groupBy ((==) `on` fst) . sortBy (compare `on` fst) . map (f &&& id)
```
With this approach we calculate `f` only once when we require value during `compare` within `sortBy`. Alternative approach would be:

```haskell
groupSortOn f = groupBy ((==) `on` f) . sortBy (compare `on` f)
```

But then it will assume that `f` is lighter than introducing new tuple. And there is no way to optimize it outside without use of memoization.

That's how I justify why `groupSortOn` introduce temporary tuple which actually can be dropped by compiler.

adept: Maybe we are talking across each other. Are you proposing to replace this:
```
acctamts = [(paccount p,pamount p) | p <- ps]
grouped = groupBy ((==) `on` fst) $ sortBy (compare `on` fst) $ acctamts
counted = [(a, length acctamts) | acctamts@((a,_):_) <- grouped]
summed = map (\as@((aname,_):_) -> (aname, sumStrict $ map snd as)) grouped -- always non-empty
```

With this: 
```
grouped =  map (\grp -> (paccount (head gpr), map pamount grp)) $ groupSortOn paccount ps
counted = [(a, length amts) | (a, amts) <- grouped]
summed = map (\(aname,as) -> (aname, sumStrict $ map snd as)) grouped -- always non-empty
```
?

This would make sense, but i will still maintain that just writing `grouped = groupSortOn fst acctamts` (where `acctamts :: (AccountName, Amount)`) would just introduce overhead of extra package and extra tuple-of-tuple built for no real benefit.

ony: @adept , no I don't suggest to extend this change to many lines of code. I like small and precise hit :)

I just tried to suggest a ways to render your idea with hiding internals outside of that line of code as much as possible.

If you'll do `git grep 'group.*sort' ':/**/*.hs'` you'll find other place where this pattern of `group . sort` repeats. I just felt that someone already had to implement that `groupSort`-kind of functions.

I didn't considered that adding of package `extra` may introduce overhead.

adept: Omg, I got `groupSortOn` and `groupSortBy` horribly mixed up, thank you for being patient with me and not calling me an idiot I certainly was. I'll push more changes later

ony: @adept I think it is easy to mix `On` with `By`.
Now when I understand your last snipped of code it looks like `groupSort` is actually what we want here.

```haskell
groupSort :: Ord k => [(k, v)] -> [(k, [v])]
groupSort = map (\x -> (fst $ head x, map snd x)) . groupOn fst . sortOn fst
```
I.e.
```haskell
grouped =  map (\grp -> (paccount (head gpr), map pamount grp)) $ groupSortOn paccount ps
```
Can be written as
```haskell
grouped = groupSort $ map (paccount &&& pamount) ps
```

adept:
Indeed, this is the best fit

adept:
@ony: I like it better now

adept:
@sm: there is no bench.sh in my clone, and no rules to build one. What am I missing?

adept:
@sm: benchmark vs current master (called 1.4 here):
```
$ quickbench -n3 -v -w hledger-1.4,hledger-658 'hledger -f examples/100000x1000x10.journal bal'
Running 1 tests 3 times with 2 executables at 2017-11-28 21:38:31 GMT:
1: hledger-1.4 -f examples/100000x1000x10.journal bal
	[13.60s]
2: hledger-1.4 -f examples/100000x1000x10.journal bal
	[13.51s]
3: hledger-1.4 -f examples/100000x1000x10.journal bal
	[13.07s]
1: hledger-658 -f examples/100000x1000x10.journal bal
	[12.90s]
2: hledger-658 -f examples/100000x1000x10.journal bal
	[12.85s]
3: hledger-658 -f examples/100000x1000x10.journal bal
	[12.79s]

Best times:
+----------------------------------------++-------------+-------------+
|                                        || hledger-1.4 | hledger-658 |
+========================================++=============+=============+
| -f examples/100000x1000x10.journal bal ||       13.07 |       12.79 |
+----------------------------------------++-------------+-------------+

```

simonmichael:
Ah, my mistake. Here's what I have in it:
```
$ cat bench.sh
hledger -f examples/100x100x10.journal print
hledger -f examples/1000x1000x10.journal print
hledger -f examples/10000x1000x10.journal print
#hledger -f examples/10000x1000x10.journal print ff
hledger -f examples/100000x1000x10.journal print
hledger -f examples/100000x1000x10.journal print ff

hledger -f examples/100x100x10.journal register
hledger -f examples/1000x1000x10.journal register
hledger -f examples/10000x1000x10.journal register
#hledger -f examples/10000x1000x10.journal register ff
hledger -f examples/100000x1000x10.journal register
hledger -f examples/100000x1000x10.journal register ff

hledger -f examples/100x100x10.journal balance
hledger -f examples/1000x1000x10.journal balance
hledger -f examples/10000x1000x10.journal balance
#hledger -f examples/10000x1000x10.journal balance ff
hledger -f examples/100000x1000x10.journal balance
hledger -f examples/100000x1000x10.journal balance ff

```

simonmichael:
Here's what I got just now (below). We shouldn't entirely trust quickbench but I think we can say the end-user performance impact is minimal. The new code reads nicely, but is also less obvious; most contributors will not be familiar with the extra library. However it is a nice and useful lib, likely to stay well maintained, and I'm ok with adding it and making it part of our standard toolkit. This should be mentioned in the history. Would you mind splitting this into two commits, depend-on-extra-lib and code-cleanup-slight-speedup (if you mention the speedup but let's quantify it for users reading history/changelog). 

```
~/src/hledger$ quickbench -n5 -N3 -w hledger-1.2,hledger-20171127,hledger-20171128-658 'hledger -f examples/1000x1000x10.journal bal'
Running 1 tests 5 times with 3 executables at 2017-11-28 16:34:43 PST:

Best times 1:
+--------------------------------------++-------------+------------------+----------------------+
|                                      || hledger-1.2 | hledger-20171127 | hledger-20171128-658 |
+======================================++=============+==================+======================+
| -f examples/1000x1000x10.journal bal ||        0.26 |             0.31 |                 0.31 |
+--------------------------------------++-------------+------------------+----------------------+

Best times 2:
+--------------------------------------++-------------+------------------+----------------------+
|                                      || hledger-1.2 | hledger-20171127 | hledger-20171128-658 |
+======================================++=============+==================+======================+
| -f examples/1000x1000x10.journal bal ||        0.26 |             0.31 |                 0.31 |
+--------------------------------------++-------------+------------------+----------------------+

Best times 3:
+--------------------------------------++-------------+------------------+----------------------+
|                                      || hledger-1.2 | hledger-20171127 | hledger-20171128-658 |
+======================================++=============+==================+======================+
| -f examples/1000x1000x10.journal bal ||        0.27 |             0.31 |                 0.30 |
+--------------------------------------++-------------+------------------+----------------------+
~/src/hledger$ quickbench -n5 -N3 -w hledger-1.2,hledger-20171127,hledger-20171128-658 'hledger -f examples/10000x1000x10.journal bal'
Running 1 tests 5 times with 3 executables at 2017-11-28 16:35:06 PST:

Best times 1:
+---------------------------------------++-------------+------------------+----------------------+
|                                       || hledger-1.2 | hledger-20171127 | hledger-20171128-658 |
+=======================================++=============+==================+======================+
| -f examples/10000x1000x10.journal bal ||        1.92 |             1.86 |                 1.81 |
+---------------------------------------++-------------+------------------+----------------------+

Best times 2:
+---------------------------------------++-------------+------------------+----------------------+
|                                       || hledger-1.2 | hledger-20171127 | hledger-20171128-658 |
+=======================================++=============+==================+======================+
| -f examples/10000x1000x10.journal bal ||        1.90 |             1.87 |                 1.83 |
+---------------------------------------++-------------+------------------+----------------------+

Best times 3:
+---------------------------------------++-------------+------------------+----------------------+
|                                       || hledger-1.2 | hledger-20171127 | hledger-20171128-658 |
+=======================================++=============+==================+======================+
| -f examples/10000x1000x10.journal bal ||        1.90 |             1.90 |                 1.87 |
+---------------------------------------++-------------+------------------+----------------------+
```

simonmichael:
PS can you point me to where work is being saved in the latest commit ? As I read the diff it seems like the new code also does a sort.

hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -752,8 +753,7 @@ journalInferCommodityStyles j =
 commodityStylesFromAmounts :: [Amount] -> M.Map CommoditySymbol AmountStyle
 commodityStylesFromAmounts amts = M.fromList commstyles
   where
-    samecomm = \a1 a2 -> acommodity a1 == acommodity a2
-    commamts = [(acommodity $ head as, as) | as <- groupBy samecomm $ sortBy (comparing acommodity) amts]
+    commamts = [(acommodity $ head as, as) | as <- groupSortBy (comparing acommodity) amts]
```
ony: Isn't this is the same case you shown earlier?

```haskell
commamts = groupSort [(acommodity as, as) <- amts]
```

hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -75,6 +75,7 @@ import Data.Array.ST
 import Data.Functor.Identity (Identity(..))
 import qualified Data.HashTable.ST.Cuckoo as HT
 import Data.List
+import Data.List.Extra(groupSortBy)
```
ony: Small suggestion: you may want to put space after module name `import Data.List.Extrta␣(groupSortBy)`

ony:
> PS can you point me to where work is being saved in the latest commit ? As I read the diff it seems like the new code also does a sort.

Difference is that `groupSort :: Ord k => [(k, v)] -> [(k, [v])]` do sorting only on a first part of the tuple which is also used for grouping. You cannot get rid of sorting if you plan to use lists here.

As I understand intention of this change is dropping one of the comparison there were shown in top of profiler report.
To get some memory saving and as result performance speedup you'll need to use `Map` to aggregate results, I guess.

I hope someday to see something like:

```haskell
import qualified Data.Map.Strict as M

balanceReport :: [Posting] -> [(Account, Amount)]
balanceReport = map (second getSum) . runReport paccount (Sum . pamount)

newtype Report k a = Ord k => Report { reportEntries :: (M.Map k a) }
instance Monoid a => Monoid (Report k a) where
   mempty = M.empty
   mappend a b = mconcat [a, b]
   mconcat = Report . M.unionsWith mappend . map reportEntries

runReport :: (Ord k, Monoid a) => (Posting -> k) -> (Posting -> a) -> [Posting] -> [(k, a)]
runReport f g =  M.toAscList . reportEntries . mconcat . map (Report . uncurry singleton . (f &&& g))
```

hledger-lib/Hledger/Data/Journal.hs
adept: Indeed, thanks

adept:
Updated.

@simonmichael : any reason not to add bench.sh to the repo (given that there are makefile rules that depend on it)?

simonmichael:
Rebased and merged as cf0dfa9f. Thanks!

You're right, I have cleaned up and added bench.sh: 30e4d265. These notes should be added to the dev guide.

adept:
Thanks!

Issue #660, opened 2017-11-30
Included journals get a different error reporting


ony:
I never got a nice error reports for my journal and I realized that only after making some improvements in reporting of parser expectations.

```shell
zsh% hledger reg -f invalid.journal 
hledger: /tmp/ws/invalid.journal:1:1:
unexpected 'i'
expecting end of input or transaction or directive

zsh% hledger reg -f - <<< 'include invalid.journal' 
hledger: SourcePos {sourceName = "-", sourceLine = Pos 2, sourceColumn = Pos 1} in included file "invalid.journal":
ParseError {errorPos = SourcePos {sourceName = "./invalid.journal", sourceLine = Pos 1, sourceColumn = Pos 2} :| [], errorUnexpected = fromList [Tokens ('n' :| "")], errorExpected = fromList [], errorCustom = fromList []}
```

Issue #509, opened 2017-02-11
-B/--cost cancels the formatting applied by a commodity directive


simonmichael:
Eg see very precise amounts in report linked at https://twitter.com/LedgerTips/status/830238616261586945 . Adding a `commodity $0.00` directive to the generated journal to limit precision has no effect, because the primary amount is bytes. It should affect amounts converted with -B also.

PR #666, opened 2017-12-04
Small fixes to journal handling (styles, error, commodity scope)

ony:
- Small additional fix for issue spot by @dashed in #487 to inherit effect of `commodity` directive on parsing included files.
- Fix style canonicalization when `--cost` option applied ( #509 ).
- Get pretty error report for included journals ( #660 ).

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -732,8 +732,11 @@ journalApplyCommodityStyles j@Journal{jtxns=ts, jmarketprices=mps} = j''
 -- from the posting amounts (or in some cases, price amounts) in this
 -- commodity if any, otherwise the default style.
 journalCommodityStyle :: Journal -> CommoditySymbol -> AmountStyle
-journalCommodityStyle j c =
-  headDef amountstyle{asprecision=2} $
+journalCommodityStyle j = fromMaybe amountstyle{asprecision=2} . journalCommodityStyleLookup j
+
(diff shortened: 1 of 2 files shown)


hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -802,7 +805,10 @@ journalConvertAmountsToCost j@Journal{jtxns=ts} = j{jtxns=map fixtransaction ts}
       fixtransaction t@Transaction{tpostings=ps} = t{tpostings=map fixposting ps}
       fixposting p@Posting{pamount=a} = p{pamount=fixmixedamount a}
       fixmixedamount (Mixed as) = Mixed $ map fixamount as
-      fixamount = canonicaliseAmount (jinferredcommodities j) . costOfAmount
+      fixamount = applyJournalStyle . costOfAmount
+      applyJournalStyle a
+          | Just s <- journalCommodityStyleLookup j (acommodity a) = a{astyle=s}
+          | otherwise = a
```
simonmichael: What would it take for this to recurse forever ?
Does it ever need to look up more than twice ?

simonmichael: My mistake, I thought journalCommodityStyleLookup was recursive.

simonmichael:
Thanks @ony!

dashed:
@ony thanks for promptly looking into this! 👍 I tried out `master` branch and can confirm that it seems to work.

PR #672, opened 2017-12-11
remove autocomplete transaction date, validate on web form submission

### LINKED ISSUE #433: hledger-web: missing error when adding a transaction with an invalid date  [$20]
The add-transaction-popup simply disappears with no changes to the journal being made requiring the user to type in everything again with a correct date.

<bountysource-plugin>

---
There is a **[$20 open bounty](https://www.bountysource.com/issues/39402412-hledger-web-missing-error-when-adding-a-transaction-with-an-invalid-date?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

efx:
Resolve #433. 
I am not interested in the bounty so you can transfer that amount to a higher priority issue.
This also validates transaction date upon form submission. As noted in
the issue, this removes hledger web's "smart date" functionality. This
functionality can be restored given the demand but should be tracked in
another issue.
date-filled


simonmichael:
Thanks. I've done some quick testing:

In firefox, the placeholder text is out of position:
firefox

In safari, there's no placeholder text, calendar popup, and no js errors reported:
safari

In brave, the placeholder text is not styled like the other fields (light gray etc.):
brave

In all browsers, the date field is no longer auto-focussed when the add form opens. This is important, so you can press a and immediately start typing a date.

Help welcome from others to help efx get this polished and robust across browsers.

simonmichael:
Correction: In safari, the new date picker doesn't activate.

The new placeholder text is mm/dd/yyyy. If this must be hard coded we'd prefer it to be yyyy-mm-dd, consistent with how dates are displayed on the page.

efx:
@simonmichael Thanks for keeping me accountable! I did not test with other browsers as I should've. :). I should get some time later in the week to address these comments. I'll also include screenshots for the various browsers. Are there particular browsers you target? (beyond latest stable Chrome, Firefox, Safari )

simonmichael:
If you have access to an IE, that would be good. If not, just do as I do and check the browser support chart, trying not to use anything too bleeding edge. I didn't see exactly how that date picker works, is it something built in ?

efx:
@simonmichael roger that. `type=date` for inputs is an addition to the HTML specification: https://html.spec.whatwg.org/dev/input.html#date-state-(type=date) so I think it is built in (or dependent on how the browser vendor implements it.

simonmichael:
https://caniuse.com/#search=date <https://caniuse.com/#search=date>

Seems not supported in IE 11 or Safari. I guess that's ok if the field works about as well as before without it and losing auto suggestion of (today, tomorrow) is not a big deal. It should show some placeholder text though.

I wonder if there's a date widget in whatever js libs we are currently using.

efx:
ah, good point. I can look for a minimal polyfill and/or check if the existing libraries have a fallback.

efx:
So the [input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) format will display according to the user's locale. When retrieving the value, it should appear as `2017-12-01`.  @simonmichael Do you think it is a requirement to display the date in the above format?

simonmichael:
I'm not sure, let's get it building and I'll try the latest on demo.hledger.org.

simonmichael:
PS if it built locally but failed on travis, let's investigate that further.

efx:
I will reopen tomorrow with a cleaner rebase.

PR #674, opened 2017-12-13
lib: Allow balance-only entries in csv reader

Nadrieril:
This PR enables the csv reader to import rows with no amount but with a balance entry.
I hope it is not too dirty :)

PS: This codebase is quite a pleasure to edit, thanks to all the maintainers

### DIFF
--- a/hledger-lib/Hledger/Data/Transaction.hs
+++ b/hledger-lib/Hledger/Data/Transaction.hs
@@ -228,6 +228,7 @@ postingAsLines elideamount onelineamounts ps p = concat [
     shownAmounts
       | elideamount    = [""]
       | onelineamounts = [fitString (Just amtwidth) Nothing False False $ showMixedAmountOneLine $ pamount p]
+      | null (amounts $ pamount p) = [""]
       | otherwise      = map (fitStringMulti (Just amtwidth) Nothing False False . showAmount ) . amounts $ pamount p
       where
         amtwidth = maximum $ 12 : map (strWidth . showMixedAmount . pamount) ps  -- min. 12 for backwards compatibility
@@ -254,7 +255,7 @@ showPostingLines p = postingAsLines False False ps p where
 tests_postingAsLines = [
    "postingAsLines" ~: do
     let p `gives` ls = assertEqual (show p) ls (postingAsLines False False [p] p)
-    posting `gives` []
+    posting `gives` [""]
     posting{
       pstatus=Cleared,
       paccount="a",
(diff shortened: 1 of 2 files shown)


hledger-lib/Hledger/Data/Transaction.hs
```diff
@@ -228,6 +228,7 @@ postingAsLines elideamount onelineamounts ps p = concat [
     shownAmounts
       | elideamount    = [""]
       | onelineamounts = [fitString (Just amtwidth) Nothing False False $ showMixedAmountOneLine $ pamount p]
+      | null (amounts $ pamount p) = [""]
```
Nadrieril: I had to change this because otherwise postings with empty amount would not be printed, even though they should be in the case of a balance-only transaction. I don't really see why a posting with no amounts (but potentially an account/balance/...) should not be printed in general actually, so I figured this change made sense.

simonmichael:
Welcome @Nadrieril. Thanks for the PR, and I'm glad you found the code pleasant!

hledger codebase is currently in "condition yellow", which means I am focussed on fixes rather than new features perhaps until the year end release. Some of our other reviewers may have more feedback for you. It will be helpful to give a small reproducible example of this in action, brief explanation of the real world problem requiring this, discussion of any interactions with existing features, and at least a draft of changes to the manual for this.

Nadrieril:
I would argue this is more a bug than a feature. I at least was surprised that a csv entry with only a balance value would be an error, since balance-only ledger transactions are perfectly valid.
The real world problem I have is that I am trying to hledgerize old accounts data but I don't need to import more than the monthly balance for now. Since this data is in spreadsheets it makes sense to try to use the csv import feature.
There should be no interaction with existing features since this is a localized change affecting only the csv importer. I also don't believe the manual needs to be changed since it did not mention this restriction. I only have a slight concern about the change I mentioned in the code comment above; I am not familiar enouigh with the codebase to see if it could be a problem (do postings with empty amounts happen often, and when they do, should they be silently erased ?).

Here is a small example:
```
date,from_account,to_account,description,amount,currency,balance
2017/01/01,assets:bank,expenses:unknown,"Balance import",,"$",1000.00
```
with the obvious rules file.

Expected output of `hledger -f example.csv print`:
```
2017/01/01 Balance import
  assets:bank  = $1000.00
  expenses:unknown
```
Instead current hledger outputs an error.

adept:
JFYI: I built this branch and checked, and it did not break conversion of any of my 320 input csv files

simonmichael:
This all sounds fair enough. I have merged this as 9e6e0603 (rewritten somehow during the usual increasing desperate git rebase attempts). Thanks!

Nadrieril:
Thanks !

PR #677, opened 2017-12-22
stack: bump default resolver to lts-10.0

rainbyte:
Current lts includes last ghc version and hledger builds ok with it


simonmichael:
Merging this, though I expect we'll continue with current policy of updating stack.yaml to latest nightly every month or so. Thanks!

PR #680, opened 2017-12-23
[WIP] use "native" datepicker with fallback

efx:
See #672 for previous discussion, etc. I messed up the rebase so introduce a clearer version of those changes here.
This also validates transaction date upon form submission. As noted in
the issue, this removes hledger-web's "smart date" functionality. This
functionality can be restored giving the demand but should be tracked in
another issue.
This embeds @brianblakely's nice https://github.com/brianblakely/nodep-date-input-polyfill.


hledger-web/Foundation.hs
```diff
@@ -297,19 +299,11 @@ getLastMessage = cached getMessage
 addform :: Text -> ViewData -> HtmlUrl AppRoute
 addform _ vd@VD{..} = [hamlet|
 
-<script language="javascript">
```
efx: I removed this attribute because it is deprecated: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

simonmichael:
I deployed 7444bb6 at http://demo.hledger.org/ and tested in Safari (only). The date picker pops up. Clicking a date doesn't enter it in the field. 

We don't have to use the native date picker if it's too much trouble. I'm open to using a more mature datepicker addon that is attractive and just works everywhere.

simonmichael:
Oops I'll be more precise. To reproduce: press a. Add form appears with date picker showing. Click a date. Date is entered into the field. Clear the field. Now clicking dates has no effect, field remains empty.

efx:
Odd, this appears to be https://github.com/brianblakely/nodep-date-input-polyfill/issues/55. A way around that would require some JS acrobatics that would negate the value of a simple polyfill.
I think your idea to use another premade one is probably better overall. This one seems sensible: https://github.com/AuspeXeu/bootstrap-datetimepicker. I'll wire that up sometime early next week.

efx:
screenshot_2017-12-26_14-06-40. Updated to just use https://github.com/uxsolutions/bootstrap-datepicker as it fits the scope and needs of hledger-web better.

simonmichael:
Deployed at http://demo.hledger.org?add=1 . Nice. Simple and clean, no obvious blockers. Seems to work for me in firefox and safari. What browser versions are supported ?

today/tomorrow/yesterday are still accepted (and even update the selection if the date picker is open).

Wishes:
- show the date picker whenever the date input field is focussed (rather than the browser's input history dropdown).
- populate the date field with today's date in YYYY-MM-DD format by default ? 
- put initial focus in the description field ?

simonmichael:
- pressing a to show add form, then escape to cancel, leaves the date popup visible on the main screen.

efx:
@simonmichael I think this supports most browsers, but am not entirely sure.
I'll see about the autofocus and population items and   try fixing the bug you mentioned in the 2nd comment.

PR #684, opened 2017-12-30
fix datepicker not closing, linting

efx:
Added defacto JavaScript linting with [eslint](https://eslint.org/docs/user-guide/configuring). I also included [an editor
config file](http://editorconfig.org/#example-file) to aid common style.


simonmichael:
Looks better. Thanks!

PR #696, opened 2018-01-25
doc: remove raw markup macro

### LINKED ISSUE #692: seeing markup on website
On http://hledger.org/hledger-web.html, some of the styling markup shows through:
hledger-web-markup

I have seen this in Firefox and Brave.

efx:
Alrighty, this should actually fix #692.  I actually verified it locally this time by running `make site-preview` :smile: 
Saw it here as well: http://hledger.org/hledger-ui.html
saw the `:::` macro in hledger-web, hledger-ui so removed. Not sure if
that had special behavior in somewhere in the build chain. Is this some kind of hakyll syntax?

### DIFF
--- a/hledger-web/hledger-web.m4.md
+++ b/hledger-web/hledger-web.m4.md
@@ -61,20 +61,20 @@ With `--serve`, it just runs the web app without exiting, and logs
 requests to the console.
 
 By default the server listens on IP address 127.0.0.1, accessible only to local requests.
-You can use `--host` to change this, eg `--host 0.0.0.0` to listen on all configured addresses. 
+You can use `--host` to change this, eg `--host 0.0.0.0` to listen on all configured addresses.
 
 Similarly, use `--port` to set a TCP port other than 5000, eg if you are
 running multiple hledger-web instances.
 
 You can use `--base-url` to change the protocol, hostname, port and path that appear in hyperlinks,
-useful eg for integrating hledger-web within a larger website. 
+useful eg for integrating hledger-web within a larger website.
 The default is `http://HOST:PORT/` using the server's configured host address and TCP port
 (or `http://HOST` if PORT is 80).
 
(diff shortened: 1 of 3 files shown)


simonmichael:
I actually have no idea. Does everything still look fine under `make site-preview` ?

simonmichael:
Also, in hledger help web [--pager|--man|--info], after `make Shake && ./Shake all && stack build hledger` ?

efx:
Interesting... Yes, things looked good with site-preview. Running the `make Shake` now.

simonmichael:
I see in the source file, hledger-web/hledger-web.m4.md, it's just a div in a chunk of HTML that's supposed to be conditionally included (for web docs) by the _web_() m4 macro, which is defined in doc/lib.m4. [Here](https://github.com/simonmichael/hledger/blob/master/Shake.hs#L295) in Shake.hs is where we call m4 and pandoc to process this file into site/hledger-web.md. Perhaps something is going wrong in this command ?

efx:
Ah, thanks for the pointers. The `:::` doesn't look like anything markdown. Either way, the snippet was just embedded in those 2 files so it seems fine just patching it there.

simonmichael:
Oh I now see your fix is in site/doc/1.5/hledger-ui.md, the archived docs snapshot (I was confused by the additional whitespace fixes). Made by [./Shake site/doc/VER/.snapshot](https://github.com/simonmichael/hledger/blob/master/Shake.hs#L333), which should be just copying the site/*.md files. So I agree, that looks good even if we might not have found the root cause. Thanks!

simonmichael:
Still present in the dev version of those docs. Ah, latest pandoc generates [fenced divs](http://pandoc.org/MANUAL.html#extension-fenced_divs) by default when generating markdown. This commit should fix it.

efx:
Excellent find!

Issue #704, opened 2018-02-05
need to read scientific notation numbers in CSV


simonmichael:
Coinbase uses scientific notation for very small amounts, which breaks our CSV reader. I think it makes sense for use to add support for this when reading CSV. Would anyone like to take a crack at it ?
```
2018-01-01 20:28:12 -0800,3.0e-08,...
```

flip111:
Is there a recommendation of library to use ?

simonmichael:
Haskell can read it natively I believe. We'll have to handle read errors gracefully, still allow the use of e or E as commodity symbol (I do that), etc. numberp in Hledger.Read.Common seems the right neighbourhood.

simonmichael:
Oops no, that’s for journal format, which is a separate question.

simonmichael:
All good points. I need it only for CSV right now. Limiting scope that way could possibly keep this a small task. I have not seen anybody needing this in journal format yet, though I do like the idea of allowing there just for generality, assuming we find solutions to all the horrible issues you raise. :)

simonmichael:
Now I see the commit comments were showing up on PR #706. Never mind, I continue the main discussion here on the original issue #704. 

I guess next steps are to merge #706, add our scientific notation to the journal > Amounts doc, and make sure we do warn if it looks a digit group separator is used in scientific notation.

simonmichael:
Merged. For the record (and noted in scientific.test): E-notation with only a digit group separator is indeed wrongly parsed as a decimal point. Eg, assuming you use comma as thousands separator and period as decimal point:
```
2018/1/1
  (a)   $1,000e1   ; looks like $10000, but parsed as $10.
```
If you declare the thousands separator with a commodity directive, you get a useful (though unclear) parse error instead:
```
commodity $1,000.00
2018/1/1
  (a)   $1,000e1   ; gives a parse error
```
But setting the thousands separator implicitly by the first amount does not give the useful error:
```
2018/1/1
  (a)   $1,000.00  ; implicitly declare comma as thousands separator
2018/1/1
  (a)   $1,000e3  ; should be parsed as $1000000, is parsed as $1000
```
I've mentioned it in [docs](http://hledger.org/manual.html#amounts).

PR #706, opened 2018-02-09
Simple prototype of scientific number notation support

ony:
This is follow-up change for #704


simonmichael:
Looking good, with this patch hledger successfully parses coinbase CSV when it contains scientific notation. Nice tests. I expect some corner cases remain to be clarified like you mentioned on #704, but let's wait and see.

The "some strange effect of default" issue seems to be unrelated to this PR so could be removed, I have opened #716 for it.

This should be mentioned in [Amounts](http://hledger.org/manual.html#amounts); let me know if you want help with that.

simonmichael:
Merged as b377bf etc., thanks.

PR #713, opened 2018-03-05
ensure stack's nix config works on nixos

### LINKED ISSUE #712: hledger sets `nix enabled` to false by default
The `stack.yaml` file in the root of the repo sets `nix.enabled` to [`false`](https://github.com/simonmichael/hledger/blob/master/stack.yaml#L6).

This makes builds on NixOS fail by default.

Normally, to use stack on NixOS, you add [`nix.enabled`](https://docs.haskellstack.org/en/stable/nix_integration/#additions-to-your-stackyaml) to `~/.stack/config.yaml`.  However, with `nix.enabled` set to `false` in hledger's `stack.yaml`, the settings in `~/.stack/config.yaml` are ignored.

Ideally, `nix.enabled` would not be set in hledger's `stack.yaml` file so that `stack` would fall back to the `nix.enabled` setting specified in `~/.stack/config.yaml`.

efx:
Fix #712 
I tested this on Arch Linux + a nix install and it does use nix for the stack installation when have mentioned config in my `~/.stack/config.yaml`.
@cdepillabout would you mind testing this on NixOS?
Thanks for the improvement too!

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -3,7 +3,6 @@
 resolver: lts-10.3
 
 nix:
-  enable: false
   pure: false
   packages: [perl gmp ncurses zlib]
 


cdepillabout:
@efx Thanks!

I just tested this.  It works on NixOS as well. :+1:

efx:
Excellent, thank you for testing @cdepillabout.  @simonmichael I can add to this PR with the appropriate documentation / and or reformat the commit message. Should I use `site: ` or something?

PR #718, opened 2018-03-12
hledger-lib: add missing Semigroup instance for Journal

peti:
GHC 8.4.1 makes Semigroup a superclass of Monoid. See
https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid
for details.

I could not run the test suite with GHC 7.10.3 because of https://github.com/joelburget/easytest/issues/3, but the test suite did succeed with the other compilers.


simonmichael:
Merged as 4c5a58ec and fixup 7635f56a. Thanks!

PR #724, opened 2018-03-25
Use skipMany/skipSome for parsing spacenonewline

cocreature:
Using `many` and `some` only to then ignore the result will still allocate the list of characters. By using `skipMany/skipSome` we can avoid this and thereby slightly reduce memory allocations.

The gains are sadly not particularly large but running a balance report with `+RTS -s` shows a decrease in total bytes allocated as well as max residency (and also a minor improvement in runtime performance but that’s probably too small to be significant):

Old:
```
     650,905,632 bytes allocated in the heap
      74,621,888 bytes copied during GC
       9,646,752 bytes maximum residency (9 sample(s))
          64,864 bytes maximum slop
              29 MB total memory in use (0 MB lost due to fragmentation)
```
New:
```
     646,328,488 bytes allocated in the heap
      73,869,168 bytes copied during GC
       9,375,360 bytes maximum residency (9 sample(s))
          70,016 bytes maximum slop
              28 MB total memory in use (0 MB lost due to fragmentation)
```

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -660,7 +660,7 @@ smartdate = do
 smartdateonly :: SimpleTextParser SmartDate
 smartdateonly = do
   d <- smartdate
-  many spacenonewline
+  skipMany spacenonewline
   eof
   return d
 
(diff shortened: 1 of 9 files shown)


simonmichael:
Thanks!

PR #731, opened 2018-04-15
CsvReader: print more information about amount-in/out errors

ehildenb:
I had issues debugging why my CSV reading file wasn't working.

Actually printing out the amount-in and amount-out helped me debug my issues.

I've also included some error message formatting changes.


simonmichael:
Thanks!

PR #732, opened 2018-04-16
Fix #729

### LINKED ISSUE #729: auto-postings don't work well with balance assertions
Consider the following file:

```
=trigger
 (target)  1

2018/04/13
  trigger  1
  source

2018/04/13
  [target]  =1
  [should be zero]
```
Result is the following:
```
$ hledger -f test.journal bal -N --auto
                  -1  should be zero
                  -1  source
                   2  target
                   1  trigger
```
However if the journal is first processed, the result is correct:
```
$ hledger -f test.journal print --auto | hledger -f- bal -N
                  -1  source
                   1  target
                   1  trigger
```
It appears that balance assertion is calculated before any auto-postings are made.

adept:
This PR moves auto posting generation to a point before journalFinalise (which will infer missing amounts and check balance assertions). This makes auto posting identical in behavior to similar postings written manually in the appropriate transactions (and should fix #729).

Rather than add yet another bool argument to all the parsers that will drive auto posting generation, I opted to provide them with full InputOpts instead. I moved Reader type definition to Hledger.Read.Common to make it work and avoid circular module dependencies.

I also changed auto posing generator slightly to inject generated postings immediately after the one that triggered them. This yields more natural ordering and, I think, will help with mixing auto postings and balance assertions

### DIFF
--- a/dev.hs
+++ b/dev.hs
@@ -51,7 +51,7 @@ timeReadJournal msg s = timeit msg $ either error id <$> readJournal Nothing Not
 main = do
   -- putStrLn $ regexReplaceCI "^aa" "xx" "aa:bb:cc:dd:ee"
 
-  (_t0,_j) <- timeit ("read "++journal) $ either error id <$> readJournalFile Nothing Nothing True journal
+  (_t0,_j) <- timeit ("read "++journal) $ either error id <$> readJournalFileWithOpts def journal
   return ()
   -- printf "Total: %0.2fs\n" (sum [t0,t1,t2,t3,t4])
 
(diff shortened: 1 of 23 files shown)


adept:
Note that travis build fails for hledger-ui, but this was not introduced in this PR, and i have no idea how to fix it

simonmichael:
Woohoo! Looks good, thank you!

I was just looking at the hledger-ui issue. It's confusing because of hledger-ui not building in stackage nightly and platform differences. I'll sort it out.

simonmichael:
@adept: it just needs the parameter removed from [UI/Main.hs#84](https://github.com/simonmichael/hledger/blob/master/hledger-ui/Hledger/UI/Main.hs#L84).

adept:
Fixed UI/Main.hs  Glad you like it :)

simonmichael:
I overlooked the hledger-api breakage. Here's a diff fixing it, probably quicker for you to commit than I to figure out how to add to this PR..

simonmichael:
I made [readJournalFile[s]WithOpts](https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Read.hs#L263) a while back, and was too lazy to replace all uses of the old functions. We may have some redundancy now.

adept:
I've finished the transition to readJournalFile(s)WithOpts and it looks like after rebase checks are passing. Merge? :)

adept:
@simonmichael i've looked at it and i think that equivalent changes are already in the RP (part of the first commit)

simonmichael:
Oops. Thank you!

PR #738, opened 2018-04-18
lib, app: better descriptions for forecast transactions

adept:
Include period in the description of generated forecast transaction. I think that this greatly improves the output of "print" and "register" with --forecast.


simonmichael:
Ok, cool.

Issue #670, opened 2017-12-08
rewrite tools/pandoc-*.hs as lua scripts


simonmichael:
Shake.hs will soon depend on pandoc 2, which has built in support for [filters written in lua](http://pandoc.org/lua-filters.html). If we rewrite our [haskell pandoc filters](http://pandoc.org/filters.html) in lua ([tools/pandoc-*hs](https://github.com/simonmichael/hledger/tree/master/tools), just [the ones being used by Shake.hs](https://github.com/simonmichael/hledger/blob/master/Shake.hs#L82)), we'll remove another source of complexity and fragility from the doc toolchain. Need someone to help with this.

simonmichael:
Getting-started ideas:

Pick one filter, eg https://github.com/simonmichael/hledger/blob/master/tools/pandoc-drop-html-blocks.hs
Get that filter working on command line, cf linked docs
Get a similar lua filter working on command line
Modify Shake.hs to use the lua filter instead of the haskell one
Watch compilation errors: `stack exec --package base-prelude -- ghcid Shake.hs`
Repeat

ehildenb:
Anyone working on this? I can give a try.

simonmichael:
Please do!

ehildenb:
@simonmichael I think you can close due to #741.

simonmichael:
True, and PS it's working great! No more waiting for pandoc filters to compile, and one less reason to need to wait for pandoc to compile.

@ehildenb, if you have any time to look at replacing hakyll-std with just pandoc, that would be quite useful too. If not, I'll try it one of these days.

PR #740, opened 2018-04-21
Pandoc 2 fixes

ehildenb:
This fixes a few issues with the files in `site/hakyll-std` regarding the updates that happenned to Pandoc.

Basically, pandoc decided to monadify-everything, which means we need to build the document using `runPure`.

Additionally, `unpack` needs to be called inside the `fromMaybe` so that both arguments are of type `String`.

I think ideally this file would be phased out.

These fixes need to get in before https://github.com/simonmichael/hledger/issues/670 can make progress.

### DIFF
--- a/site/hakyll-std/TableOfContents.hs
+++ b/site/hakyll-std/TableOfContents.hs
@@ -54,7 +54,7 @@ markupHeader :: Tree Block -> H.Html
 markupHeader (Node (Header _ (ident, _, keyvals) inline) headers)
   | headers == [] = H.li $ link
   | otherwise     = H.li $ link <> (H.ol $ markupHeaders headers)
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks! I'm just going to merge this. FWIW I feel like hakyll-std should be phased out.

ehildenb:
I just call `./Shake website`, and it triggers the hakyll-std build. Building `pandoc` takes forever (somewhere between 5 and 10 minutes), and then the rest goes ok.

I made these fixes because I couldn't test the lua scripts with the `./Shake website` build failing. I plan on also removing hakyll-std, not sure which I will get done first.

ehildenb:
If you're still having problems, could try adding the `--package pandoc` flag back to `site/hakyll-std/hakyll-std.hs` (revert commit 41ae44b).

PR #741, opened 2018-04-23
[WIP] Docs/lua filters

ehildenb:
@simonmichael this is *not* ready for merge. This addresses #670 .

The first commit marked `!!!` (1e4cb7a4) adds the files which are not tracked but generated for the website to the repo. I will remove this commit before merging, it's just there to make it easier to provide evidence that the switch to Lua filters was correct.

The commit marked `sss` (1001d8e8) should be squashed into the previous commit. It is a proposed change to the drop in HTML blocks, instead of adding an empty string, use pandoc's designated `Null`.

The second commit marked `!!!` (8f061dc) is evidence that changing the Lua filter to do something clearly incorrect did not change the generated documentation. I'm not sure if the particular filter is not needed anymore, or if the first commit marked `!!!` does not add enough files to be tracked.

Before merging, I also want to delete the relevant lines in the `Shake.hs` file (instead of uncommenting them).

If you check that I'm tracking enough extra files with the first `!!!`, approve the change of the `sss`, and approve either me removing the filter associated with dropping notes (the second `!!!`) or figure out why making it clearly incorrect doesn't break things, I can clean up the branch and make ready for merge.


simonmichael:
@ehildenb thanks for this. Apologies, I may be I'm not too smart today. I don't really get what's going on with the second !!!. Please submit a cleaned-up patch set as you think best and I'll merged.

ehildenb:
@simonmichael I've updated the PR. I decided to make the change to `Null` I proposed above, removed the filter which seemed to be doing nothing, and removed all the `tools/pandoc-*.hs` filters from the repo. Also removed references to them in the `./Shake.hs` file.

I think you should do a full rebuild of all documentation on your system before merging, and if anything that is tracked changes, inspect then commit it, to make sure I didn't miss/break anything.

simonmichael:
Merged. Thanks!

Issue #744, opened 2018-04-26
retire hakyll-std/reduce hakyll & pandoc dependency


ehildenb:
In PR https://github.com/simonmichael/hledger/pull/740 there are references to phasing out Hakyll std, or at least parts of it. In the file `site/hakyll-std/hakyll-std.hs` it also mentions that it should go away.

@simonmichael How are we interested in doing that? Do we want to completely remove Hakyll as a dependency, or just remove a site-specific build file `hakyll-std` (I'm not super familiar with Hakyll, so am not sure if this is possible)? If we remove Hakyll as a dependency, will the site be built with manual calls to `pandoc` instead?

simonmichael:
Looks like our comments crossed! Quite interested! Yes it seems like we could do everything needed with Shake and pandoc and so I would love to get rid of hakyll and especially hakyll-std entirely.

ehildenb:
Alright, my plan is to migrate things over to the Shake file in a series of commits. First the file-copying stuff that Hakyll is doing. The hard part will be the `relativizeUrls` function that Hakyll provides. We can either:

1.   Keep importing the `relativizeUrls :: String -> String -> String` function that Hakyll provides. Still have dependence on Hakyll in the Shake file then.
2.   Write our own relativizeUrls either as a standalone Haskell function, or as a Pandoc Lua filter.
3.   Just make sure all the URLS on the website are already relativized in the original markdown.

Which would you prefer?

simonmichael:
I'd prefer to not depend on hakyll or pandoc (in source form) at all, as they take forever to build and tend to have upgrade issues. Re urls, I'm in favour of the simplest thing that could possibly work, not sure how much of a problem this will be, let's test with one page maybe.

simonmichael:
Though it does depend on pandoc source, http://hackage.haskell.org/package/pansite looks like it might be worth a look..

simonmichael:
Actually that's a web app, and probably too new. Never mind.

ehildenb:
Not sure depending on Pandoc source is the problem, as much as invoking `stack exec ghc -- --package pandoc`, which causes rebuilds every time. I think it it's installed with the stack file directly it shouldn't cause rebuilts of Pandoc when we change things that involve Pandoc. Not sure though, not super familiar with stack.

[...]

PR #755, opened 2018-05-06
Hakyll std

ehildenb:
Addressing #744.

Unfortunately the long pandoc build times are still present, I couldn't get it to work without adding `--package pandoc` to the invocation of `ghc` which builds `tools/pandoc-site.hs`. They only happen on `./Shake website-render` though.

Dependencies on Hakyll and Blaze completely removed. Instead a pandoc filter and the Shake file are used to build the website.

I have tested locally, and everything under `site/_site` builds exactly the same except for:

1.   Links now have the `title="LINK_NAME"` attribute added, because of how pandoc renders links (due to b5c320c).
2.   The table of contents ordered lists now have the attribute `type="1"` added (and HTML layout is slightly different), because pandoc renders ordered lists with the type added by default (due to fb42d6a).
3.   The spacing on table of contents is different (due to 770b58b).
4.   Useless file `coins.svg` is also copied into each website `images/` directory as well, can't avoid this without listing out many more files in `./Shake.hs` (due to 1254557).
5.   Large parts of the HTML are indented due, and many apostrophes are replaced with `&#39;` because of how pandoc renders documents (due to 3514b78).

I think this is ready for merge, future action needed is to replace `tools/pandoc-site.hs` with a Lua script version. This will remove the long pandoc build times. Alternatively, figure out how to make either `hpack` or `stack` build `tools/pandoc-site.hs` correctly so it caches dependencies for us (including the pandoc dependencies). I fought with it a bit, but just don't understand enough about how these things work.


simonmichael:
Very nice! I will look asap.

simonmichael:
Looks great ehildenb. I'll just paste some notes here.

Can you add a module doc to pandoc-site.hs summarising what it does.

Is the "from https://github.com/blaenk/blaenk.github.io" comment still accurate ?

Perhaps we can avoid adding site-building deps to stack-ghc8.2.yaml, which forces a slow pandoc build on folks trying to build hledger with it.

Does the double slash in "site/_site//*.html" in Shake.hs mean something ? Or is it just a workaround for dropDirectory2 ?

In the Makefile, the site-* rules need cleanup. I'll do this.

Docs we'll need to update: probably the developer workflows and contributor guide on the wiki.

simonmichael:
It seems to build everything cleanly in about a minute on my system and I expect is already more robust than what we had. Merging, let's refine in master. Thank you!

ehildenb:
The double slack `//` in Shake means "match any number of path components, including zero". So `site/_site//*.html` will match `site/_site/index.html`, and will also match `site/_site/doc/1.2/csv.html`. It's so that the same rule can be used to build all the html files using pandoc.

PR #764, opened 2018-05-11
Fix doctest error. 

Xitian9:
Fix doctest error introduced by 49c8c093acf8fd212f1e78561969792238f00fae.


simonmichael:
Thank you very much!

PR #769, opened 2018-05-14
Parse comments at beginning of periodic transactions.

### LINKED ISSUE #759: can't comment a posting in a periodic transaction
This posting is commented and should be ignored, but isn't:
```
~ weekly
    ; (a)  1
```
```
$ hledger reg --forecast
hledger: could not balance this transaction (real postings are off by 1)
2018/05/14 Forecast transaction (weekly)
    ; (a)               1
```
This affects periodic transactions, not regular ones.

Xitian9:
Fixes #759

### DIFF
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -457,11 +457,13 @@ modifiertransactionp = do
   postings <- postingsp Nothing
   return $ ModifierTransaction valueexpr postings
 
+-- | Parse a periodic transaction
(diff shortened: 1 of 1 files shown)


Xitian9:
It might be a better approach to allow comments in the first line of a periodic transaction, which would allow unification of treatment of comments between periodic and normal transactions. However, this would require disallowing semicolons in periodic expressions. I don't think there are any circumstances when they should be there anyway, but I'd be hesitant to make this change without consultation.

simonmichael:
Thanks! I think it makes sense for periodic transactions to parse as similarly to regular transactions as possible. I don't know of any reason semicolon would need to work differently there.

simonmichael:
Thanks!

PR #773, opened 2018-05-15
lib: fix performance regression introduced in 30ae9914

awjchen:
In commit 30ae9914, I introduced a performance regression: the memory usage of `hledger -f examples/10000x10000x10.journal register +RTS -s > /dev/null` (compiled with `-O2`) increased from `92 MB` to `104 MB`. The change contained in this pull request reverses the regression when applied to both 30ae9914 and the current master branch.


simonmichael:
I think this commit breaks three functional tests. (I'm not sure why travis is not completing here.)

awjchen:
Ah, I forgot the leading newline. Hopefully this time it works.

On another note, how do I run the functional tests locally?

simonmichael:
Guessing it works. Thanks!

PR #777, opened 2018-05-17
Refactor comment parsers

awjchen:
In this PR I refactor the comment parsers (in `Hledger/Read`).

I have done this to prepare for a discussion and potential implementation of changes to the allowed syntax for comments, tags, date-tags, and bracketed dates. The intention of this PR, however, is to preserve the current syntax and behaviour.

The primary refactoring in this PR, taking place in the first four commits, is the weakening of several parsers from `ErroringJournalParser` (`StateT Journal`, `ParsecT`, and `ExceptT`) or `JournalParser` (`StateT Journal`, `ParsecT`) to `SimpleTextParser` (`Parsec`). From brief checks (`+RTS s`), this change is associated with a slight decrease in performance (slightly longer runtime, slightly higher memory usage when parsing comments). This is surprising to me, as I was expecting these changes to at least preserve performance. (Perhaps it is something to with strictness? Or maybe I have done something silly.)

The fifth commit removes the redundant parsing alternatives following `followingcommentsp` (and `followingcommentandtagsp`), in `Read/JournalReader.hs`. These alternatives are redundant because they parse a case that is already attempted in `followingcommentsp`.

The sixth commit is just cleanup and polish. I hope my code style is acceptable for the project; if not, I am willing to go back and reformat it.

In this PR, I have built and tested with `stack --stack-yaml=ghc7.10.yaml`, so hopefully I will not break backwards compatibility again.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -95,10 +95,10 @@ where
 --- * imports
 import Prelude ()
 import Prelude.Compat hiding (readFile)
-import Control.Arrow ((***))
 import Control.Monad.Compat
 import Control.Monad.Except (ExceptT(..), runExceptT, throwError) --, catchError)
 import Control.Monad.State.Strict
+import Data.Bifunctor
 import Data.Char
 import Data.Data
 import Data.Default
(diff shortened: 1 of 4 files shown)


simonmichael:
Nice cleanups!

Issue #749, opened 2018-05-03
space as digit group separator should not be parsed as decimal point


simonmichael:
With a journal containing only the below, the amount is parsed as 1. I think (hope) no country uses space for decimal point, so we should prevent this:
```
2018-01-01
  (a)   USD1 000

;$ hledger -f d.j reg amt:1
;2018/01/01                      (a)                       USD1 000      USD1 000
```

And possibly the same issue, the following commodity directive should complain about a missing decimal point (now required), instead it gives a parse error:
```
commodity 1 000  USD

2018-01-01
  (a)   USD1 000

; hledger: d.j:1:12:
; unexpected space
; expecting digit
```
A correct commodity directive does work as intended:
```
commodity 1 000.00  USD

2018-01-01
  (a)   USD1 000

;$ hledger -f d.j reg amt:1000
;2018/01/01                      (a)                   1 000.00 USD  1 000.00 USD
```

ony:
Space have special meaning and cannot be decimal separator:
```haskell
    (firstSep, groups) <- option (Nothing, []) $ do
        leadingDigits <- some digitChar
        option (Nothing, [leadingDigits]) . try $ do
            firstSep <- oneOf sepChars <|> whitespaceChar
            groups <- some digitChar `sepBy1` char firstSep
            return (Just firstSep, leadingDigits : groups)
```

Isn't `USD1` actually treated as a commodity?

simonmichael:
Contrary to the haddock, `firstSep <- oneOf sepChars <|> whitespaceChar` allows a space (and possibly other things in the [unicode Space category](http://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Char.html#t:GeneralCategory)).

ony:
I see problem is in `fromRawNumber` actually we get `(Just ' ', ["1", "000"], Nothing)` same as before we assume that if there is two groups - last one is decimals regardless of what separator were used.

simonmichael:
I've added three tests related to this. I think the numberp/rawnumberp parsing code is a bit hard to follow and we could maybe make it clearer. 

- https://github.com/simonmichael/hledger/blob/master/tests/journal/amounts-and-commodities.test#L96
- http://hledger.org/journal.html#amounts
- http://hledger.org/journal.html#declaring-commodities
- https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Read/Common.hs#L758

awjchen:
[...]

PR #780, opened 2018-05-20
lib: fix parsing of amounts with a whitespace digits separator 

awjchen:
This PR addresses a problem which was found in issue #749. This PR does not address that issue.

The problem was that we fail to parse the following journal
```
commodity 1 000  USD

2018-01-01
  (a)   USD1 000
```
failing with the following error
```
hledger: test.journal:1:12:
unexpected space
expecting digit
```
We fail to parse this next journal as well
```
2018-01-01
  (a)   USD1 000 ; comment
```
```
hledger: test2.journal:2:9:
unexpected 'U'
expecting '=', '{', end of input, or newline
```
It seems that the issue is caused by the use of a space as the digits separator and trailing whitespace after the amount. The current parser consumes the trailing whitespace as a separator and then expects at least another digit, which it does not find. This patch groups together the separator and the following digit group. With this patch, the above two journals parse successfully.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -763,8 +763,9 @@ rawnumberp = do
         leadingDigits <- some digitChar
         option (Nothing, [leadingDigits]) . try $ do
             firstSep <- oneOf sepChars <|> whitespaceChar
-            groups <- some digitChar `sepBy1` char firstSep
-            return (Just firstSep, leadingDigits : groups)
+            secondGroup <- some digitChar
+            otherGroups <- many $ try $ char firstSep *> some digitChar
+            return (Just firstSep, leadingDigits : secondGroup : otherGroups)
 
     let remSepChars = maybe sepChars (`delete` sepChars) firstSep
         modifier
--- a/tests/journal/amounts-and-commodities.test
+++ b/tests/journal/amounts-and-commodities.test
@@ -136,17 +136,17 @@ $ hledger -f- bal -V -N
 #>=1
 
 # 11. After a space-grouped amount, a posting comment should parse.
-#<
-#2018-01-01
-#  (a)   USD 1 000  ;comment
-#
-#$ hledger -f- print
-#> //    # any stdout, no stderr, 0 exit code
+<
+2018-01-01
+  (a)   USD 1 000  ;comment
+
+$ hledger -f- print
+> //    # any stdout, no stderr, 0 exit code
 
 # 12. After a space-grouped amount, trailing whitespace should parse.
-#<
-#2018-01-01
-#  (a)   USD 1 000 
-#
-#$ hledger -f- print
-#> //    # any stdout, no stderr, 0 exit code
+<
+2018-01-01
+  (a)   USD 1 000 
+
+$ hledger -f- print
+> //    # any stdout, no stderr, 0 exit code


ehildenb:
Can you add a test or two for this case? Perhaps the above examples.

simonmichael:
I added some functional tests for #749 in tests/journal/amounts-and-commodities.test. You can just uncomment them and `make functest` or `COLUMNS=80 shelltest --execdir --hide-successes -j8 tests/journal/amounts-and-commodities.test [-i...]`.

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -763,8 +763,9 @@ rawnumberp = do
         leadingDigits <- some digitChar
         option (Nothing, [leadingDigits]) . try $ do
             firstSep <- oneOf sepChars <|> whitespaceChar
-            groups <- some digitChar `sepBy1` char firstSep
-            return (Just firstSep, leadingDigits : groups)
+            secondGroup <- some digitChar
+            otherGroups <- many $ try $ char firstSep *> some digitChar
```
ony: Hmm... Interesting how this is different from previous code that suppose to produce list of one or more elements with non-zero list of digit characters separated by `firstSep`? Am I missing something?

awjchen: Yes, the new code also parses a list of one or more numbers separated by `firstSep`, where each number is a non-zero list of digit characters. The difference is only that I have added `try` so that a separator and the following digits are parsed together as an atomic unit. The old code failed to parse the above examples because it interpreted trailing whitespace as a separator.  With the new code, whitespace will only be interpreted as a separator if digits follow it.

awjchen:
I've uncommented tests 11 and 12 in `tests/journal/amounts-and-commodities.test` and updated this PR.

Test 10 is also relevant, but as it is, it does not yet pass since we have not fully resolved issue 749.

hledger-lib/Hledger/Read/Common.hs
ony: Hmm... Yep. Confirmed with test-case

```
### Failure in /home/nikolay/contrib/hledger/hledger-lib/Hledger/Read/Common.hs:757: expression `parseTest rawnumberp "1 000 USD"'
expected: RawNumber (Just ' ') ["1","000"] Nothing
 but got: 1:2:
          unexpected space
          expecting digit
```

Apparently `sepBy` in `megaparsec` expects that if separator were parsed successfully then next entry is guaranteed to parse correctly.

Though I still cannot understand error message. `notFollowedBy $  (whitespaceChar >> digitChar)` should say something about unexpected space followed by digit.

awjchen: It seems that the problem is that the remaining text that the `notFollowedBy ...` line is given to parse is ` 000 USD`, which triggers its second alternative. The first section of `rawnumberp` doesn't consume the ` 000` due to `try` in `option (...) . try $ do` that encloses the failure on the `groups <- some digitChar ...` line.

awjchen:
Having the approval of @ony (and myself), I will merge this PR.

simonmichael:
Thanks @awjchen for the fix and @ony for the review. The results look good. 

In general, use whichever kind of test works best for you, we don't have a firm policy. But it's a good point about doctests being part of the API documentation. hunittests run fastest but are a chore to write, shell tests run pretty fast and are easy to write and understand, doctests are easy to write and provide useful documentation but alas run extremely slowly. Please feel free to add notes to https://github.com/simonmichael/hledger/wiki/Tests.

simonmichael:
easytests are new and might be easier than hunittests.

ony:
@awjchen , doctests are simple unit-tests for developers that helps to catch errors that propagates further. Especially useful for parsers. They also help to understand some corner cases of API.

In this particular case issue were isolated because of extra check `notFollowedBy` in the end. But if that line would be missing you'll get error in a next parser or will get wrong interpretation.

awjchen:
Ah, I haven't before considered whether issues were propgated or isolated. Thanks for the tips!

PR #781, opened 2018-05-20
Lua pandoc site

ehildenb:
Uses a lua filter to build the website.

Generated files under `site/` are exactly as before.

Instead of using a single bullet-list item `toc` to mark where to place TOC, a single paragraph `$toc$` is used.

Uncommented extra-deps in stack-ghc8.2.yaml, and added the `haddock-library` one. Not sure there are needed anymore, because there are no more source-level dependencies on Pandoc AFAIK.


simonmichael:
Great! I haven't looked closely yet, but is there a way we can get the extra-deps out of stack-ghc8.2.yaml ? As mentioned they caused me problems when trying to test compilation across GHC versions. Are we still building something that depends on pandoc as a library ? If so perhaps it can have its own stack yaml file, or become a stack script.

ehildenb:
I will take a look and see if it's needed anymore. Could be that it's entirely superfluous.

simonmichael:
Looking good @ehildenb. 

Was the removal of screenshots in {1.0,1.1}/hledger-ui.md intentional in https://github.com/simonmichael/hledger/pull/781/commits/7ffed7025311d7ba0be7dbf7ff766c3ff415d6b6 ?

Looking at the commits in I think sequential order (starting at the top on this page), it's odd that pandoc-site is switched to lua, and then comes several more commits to the old hs file. I'm not quite sure how that works. Also we have the addition and removal of extra deps. Could you do another pass with rebase just to simplify the history a bit before merging. If you prefer, I can take a crack at it.

simonmichael:
PS what's the .ci file for, again ?

ehildenb:
@simonmichael well, the old way of generating the website would look for a single bullet list which started with the word `toc`, and then the entire bullet list would be replaced with the table of contents. Because the screenshot code was indented in those markdown files it was included as a part of the first bullet list element, and removed in the generated content, so I figured I didn't need to retain it in the source. If you want, I'll put those chunks of html back unindented so that they are passed through to the final website.

Also, the order of that Github shows commits in does not correspond to the git DAG order of the commits. Github uses the timestamps of commits to decide the order to print them in (so that people's non-linear histories with lots of messy merges and stuff display nicely no matter what). If you look at `git log`, or even better, `git hist` (see https://github.com/ehildenb/dotfiles/blob/76b160b4037153736d6cc01c4af7aaf7ee9c1ad1/git/gitconfig#L48 for the alias, also called `git gr` or `git gra` by many) you'll see that the git DAG order makes sense. If you want me to modify the timestamps of the commits, I can do that too, though I don't think it's needed.

The `.ci` file is used to track test-sets for https://github.com/cynic-net/git-ci/tree/record-build-artifacts as a testing setup for any build artifacts which might not already have tests. Basically, it allows me to say "go to each commit in this branch and build the website, then build a separate git history which contains the generated website". That's how I've been able to give detailed commit-wise explanations of how the website generation changed in https://github.com/simonmichael/hledger/pull/755 or make claims like "none of the generated files under `site/` were changed" above. I can remove the `.ci` file from the repo if you like, probably won't need that particular one again after this PR.

simonmichael:
I (mostly) see, thanks. I followed your git-ci link but I'm non the wiser. It's fine to leave .ci in if useful to you, let's add a comment linking to the tool and saying what it does.

Being able to build hledger and docs without having to build pandoc and hakyll is a great improvement for speed, robustness, and ease of onboarding new devs. Thanks a lot! I will add pandoc to the list of recommended tools when I find it.

simonmichael:
I'm getting a lua error with this one file for some reason, on the ubuntu production server but not on my macbook. pandoc is version 2.1.1.
```
$ ./Shake site/_site/doc/1.0/hledger.html
pandoc --from markdown --to html site/doc/1.0/hledger.md --template site/site.tmpl --metadata=siteRoot:../.. --metadata=title:hledger --lua-filter tools/pandoc-site.lua --output site/_site/doc/1.0/hledger.html
Error running filter tools/pandoc-site.lua:
Error while running filter function: Error during function call: could not read argument 1: Could not read list: Could not get Block value: expected table but got nil.
Error when running Shake build system:
* site/_site/doc/1.0/hledger.html
user error (Development.Shake.cmd, system command failed
Command: /bin/sh -c 'pandoc --from markdown --to html site/doc/1.0/hledger.md --template site/site.tmpl --metadata=siteRoot:../.. --metadata=title:hledger --lua-filter tools/pandoc-site.lua --output site/_site/doc/1.0/hledger.html'
Exit code: 83
Stderr:
Error running filter tools/pandoc-site.lua:
Error while running filter function: Error during function call: could not read argument 1: Could not read list: Could not get Block value: expected table but got nil.
)
```

ehildenb:
Hmmm, I've run on Arch and on Nix and it seems to be fine. I'm using pandoc 2.1.3 on my dev server, what is the version on your Macbook?

simonmichael:
Oops, I meant to say pandoc version (the failing one) is 2.2.1. pandoc version on my mac (working) is 2.1.2.

simonmichael:
Now with pandoc 2.1.3 on linux, I can still reproduce it, like so:
```
$ pandoc --version
pandoc 2.1.3
Compiled with pandoc-types 1.17.4.2, texmath 0.10.1.1, skylighting 0.7.1
Default user data directory: /home/simon/.pandoc
Copyright (C) 2006-2018 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```
```
$ cat a.lua
-- minimised version of pandoc-site.lua

local headers = {}

function Header(h)
  table.insert(headers, h)
  return h
end

function Pandoc(doc)
  print(headers)
  print(pandoc.utils.hierarchicalize(headers))
end
```
```
$ pandoc site/doc/1.0/hledger.md --lua-filter a.lua
table: 0xe881878
Error running filter a.lua:
Error while running filter function: Error during function call: could not read argument 1: Could not read list: Could not get Block value: expected table but got nil.
```
It dies in hierarchicalize. Though, not if I comment out the Header function. 

Perhaps hslua version matters ? pandoc --version doesn't report it, but it looks as if this pandoc was built with hslua-0.9.5.1.

simonmichael:
Should be reported in pandoc issue tracker, perhaps.

ehildenb:
Hmmm, maybe instead of just returning `h`, should be rebuilding a `pandoc.Header(...)` or the like.... I just traveled a bunch, but can investigate more later.

simonmichael:
That would be great, thanks.

simonmichael:
Currently blocking updates to hledger.org.

simonmichael:
Perhaps it's this: https://pandoc.org/lua-filters.html#module-pandoc.utils

> hierarchicalize (blocks)
>    Convert list of blocks into an hierarchical list. An hierarchical elements is either a normal block (but no Header), or a Sec element.

simonmichael:
Although, the immediately following example suggests hierarchicalizing Headers works just fine.

I am a bit rusty debugging weakly-typed languages!

simonmichael:
Testing with that example data, it still fails:
```
  headers =
     {
        pandoc.Header(2, pandoc.Str 'first'),
        pandoc.Header(2, pandoc.Str 'second'),
     }
```
So I am thinking that pandoc example is out of date. If you agree, would you mind opening an issue with them ?

simonmichael:
#792 gives the same error here. I'm running:
```
$ pandoc --version
pandoc 2.1.3
Compiled with pandoc-types 1.17.4.2, texmath 0.10.1.1, skylighting 0.7.1
```

simonmichael:
https://github.com/jgm/pandoc/issues/4679

simonmichael:
Workaround until https://github.com/hslua/hslua/issues/63 is fixed:
Install pandoc with 32-bit lua, eg: `stack install pandoc --flag=hslua:lua_32bits`

PR #782, opened 2018-05-21
lib: fix issue where spaces were allowed as decimal points

### LINKED ISSUE #749: space as digit group separator should not be parsed as decimal point

awjchen:
This PR fixes #749 by implementing the changes discussed therein. That is, we ensure that a whitespace character is never used as a decimal point.

This PR also enables the tests that were prepared for issue #749.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -714,10 +714,12 @@ fromRawNumber suggestedStyle negated raw = (quantity, precision, mdecimalpoint, 
     -- unpack with a hint if useful
     (mseparator, intparts, mdecimalpoint, frac) =
             case raw of
-                -- just a single punctuation between two digits groups, assume it's a decimal point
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks!

PR #790, opened 2018-05-28
feat(hledger-ui): added neovim as a supported editor

Aerex:
When neovim is set as EDITOR hleger will jump to the correct line number of the transaction; before hledger will open journal at top of the file


simonmichael:
Thanks! Sorry I lost track of this.

PR #792, opened 2018-05-31
tools/pandoc-site: switch to manually gathering top-level headers

ehildenb:
An attempt to fix the issues of #781 . @simonmichael can you try it and see if it works on your machines? It generates the same website as before for me.


simonmichael:
Responded on #781, did not work.

Issue #794, opened 2018-06-02
base-compat 0.10.x breaks hledger-lib build


peti:
~~~
    Hledger/Read/Common.hs:725:62: error:
        Ambiguous occurrence ‘<>’
        It could refer to either ‘Sem.<>’,
                                 imported from ‘Prelude.Compat’ at Hledger/Read/Common.hs:97:1-39
                                 (and originally defined in ‘Data.Semigroup’)
                              or ‘Data.Monoid.<>’,
                                 imported from ‘Data.Monoid’ at Hledger/Read/Common.hs:110:1-18
~~~

simonmichael:
Thanks! It looks like we missed this in https://github.com/simonmichael/hledger/pull/788. I thought I ran `make buildtest-all`, but it seems not. I guess we need to run this or similar as part of CI somehow, though that is already slow and disk-space-limited.

peti:
It's possible that the base-compat 0.10.x version simply hadn't been released yet at the time you ran those tests.

simonmichael:
Oh I see.. right. Or I just didn't have it installed, I don't think these tests force the latest version of deps. (Maybe they should, if they could..)

Separately, I guess this is a breakage caused by removal of upper bounds, and base-compat is a package where I should keep specifying upper bounds.

simonmichael:
I'll merge this when I've verified make buildtest-all passes with base-compat-0.10.1 installed (or if you observe that before I do let me know).

simonmichael:
Merged your and my changes, which should make us somewhat robustly build and doctest warning-free across the last four GHC versions again. Thanks!

peti:
What shall we do about Hackage? I see two options:

1. Edit the version constraint of `base-compat` to `> 0.8.1 && < 10`.
2. Release the updated code as a new version.

simonmichael:
Yes, perhaps I'll do a revision on 1.9.1 at least, adding the upper bound on base-compat. I have been resisting spending brain cycles on a minor release.

simonmichael:
And also on [fsnotify](https://github.com/simonmichael/hledger/issues/797#issuecomment-394587766).

simonmichael:
Done.

simonmichael:
But this causes problems in stackage nightly (https://github.com/commercialhaskell/stackage/issues/3706). So I'll probably do a minor release of all packages soon.

PR #795, opened 2018-06-02
Fix hledger-lib build with ghc 8.0.x and base-compat 0.10.x.

peti:
We don't need to import Data.Monoid because Prelude.Compat exports "<>"
already. In fact, importing that module causes build failures:

    Hledger/Read/Common.hs:725:62: error:
        Ambiguous occurrence ‘<>’
        It could refer to either ‘Sem.<>’,
                                 imported from ‘Prelude.Compat’ at Hledger/Read/Common.hs:97:1-39
                                 (and originally defined in ‘Data.Semigroup’)
                              or ‘Data.Monoid.<>’,
                                 imported from ‘Data.Monoid’ at Hledger/Read/Common.hs:110:1-18

Fixes https://github.com/simonmichael/hledger/issues/794.

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE PackageImports #-}
 {-|
 
 Date parsing and utilities for hledger.
@@ -73,9 +74,9 @@ module Hledger.Data.Dates (
 where
 
 import Prelude ()
-import Prelude.Compat
+import "base-compat" Prelude.Compat
 import Control.Monad
-import Data.List.Compat
+import "base-compat" Data.List.Compat
 import Data.Default
 import Data.Maybe
 import Data.Text (Text)
--- a/hledger-lib/Hledger/Data/StringFormat.hs
+++ b/hledger-lib/Hledger/Data/StringFormat.hs
@@ -2,7 +2,7 @@
 -- hledger's report item fields. The formats are used by
 -- report-specific renderers like renderBalanceReportItem.
 
-{-# LANGUAGE FlexibleContexts, TypeFamilies #-}
+{-# LANGUAGE FlexibleContexts, TypeFamilies, PackageImports #-}
 
 module Hledger.Data.StringFormat (
           parseStringFormat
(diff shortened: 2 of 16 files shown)


simonmichael:
We should probably update all four base-compat bounds to >=0.10.1 && <0.11 too I guess.

peti:
OK, done.

simonmichael:
It looks like we can't use base-compat 0.10.* yet with hledger-api, since servant deps don't allow it. I guess you can add the older one in stack.yaml extra-deps ? Also add the newer one there, for folks building with cabal or older stack snapshots.

Also ==0.10.* isn't quite the same thing, wouldn't it be better to depend on >=0.10.1 ?

simonmichael:
Ie in stack.yaml:
```
extra-deps:
  ...
  - base-compat-0.10.1
  # for hledger-api:
  - base-compat-0.9.3
```

simonmichael:
..and in the package.yamls:
```
- base-compat >=0.10.1 && <0.11
```

simonmichael:
Phew, details. Let me try again, shouldn't it be maybe like this:

stack.yaml: add the old base-compat to extra-deps
```
  # for hledger-api:
  - base-compat-0.9.3
```
stack-ghc*.yaml: add the new base-compat to extra-deps
```
  - base-compat-0.10.1
```
hledger{,-lib,-ui,-web}: require the latest base-compat
```
  - base-compat >=0.10.1 && <0.11
```

simonmichael:
With stack.yaml, it's also failing on:
```
Test suite failure for package hledger-lib-1.9.99
    doctests:  exited with: ExitFailure 1
Full log available at /home/simon/src/hledger.org/.stack-work/logs/hledger-lib-1.9.99-test.log



    Hledger/Data/StringFormat.hs:17:1: error:
        Ambiguous module name ‘Prelude.Compat’:
          it was found in multiple packages:
          base-compat-0.10.1 base-compat-batteries-0.10.1
       |
    17 | import Prelude.Compat
       | ^^^^^^^^^^^^^^^^^^^^^
```

peti:
> Also ==0.10.* isn't quite the same thing, wouldn't it be better to depend on >=0.10.1 ?

Why? AFAIK, base-compat 0.10.0 works fine, so why would you like to exclude it?

peti:
https://hackage.haskell.org/package/servant-server-0.13.0.1 adds support for `base-compat-0.10.x`. Maybe that simplifies matters?

peti:
I pushed an updated version. Running test builds now.

Issue #796, opened 2018-06-03
Symbols inside numbers in currencies


Ekleog:
Hello,

So I am trying to add the Yen as a currency in hledger. I thus put this:
```ledger
commodity 円
    format 1億0,000万0,000円
```
which is the common way of writing Yen in japanese.

However, `hledger` then replies:
```
unexpected '0'
expecting end of input, newline, or the rest of right-symbol amount
```

As hledger appears to support UTF-8 pretty well (as eg. a `format 1,0000,0000円` works), I guess this is due to having a non-digit non-dot-or-comma symbol inside the number.

Is there a way to have symbols other than dot-and-comma be considered for inclusion as digit separators?

As usual, thank you for your work!

simonmichael:
Wow. :) Yes your guess is correct. Where could we learn more about this usage, the allowed symbols and their meanings etc, eg on wikipedia ?

simonmichael:
/me reads https://en.wikipedia.org/wiki/Japanese_numerals .. mind blown.

I would definitely consider adding more support for this, if a developer steps up to: design and build it, verify all tests pass, and stick around to resolve the issues that will show up later.

Ekleog:
Hmm, according to https://ja.wikipedia.org/wiki/%E5%86%86_(%E9%80%9A%E8%B2%A8)#%E5%86%86%E3%81%AE%E6%B5%81%E9%80%9A%E9%AB%98 the total amount of currency of Yen is 96兆0,377億円 ; so I guess for Yen it should be enough to add 兆, 億 and 万 as dot-and-comma-like symbols.

Do you think this would be an acceptable solution, or that it'd be better to have something more generic, in order to accommodate other currencies that may have “weird” (non-pejorative menaing) numbering schemes?

simonmichael:
That sounds ok, though it seems unlikely to be the last word on the matter.

alerque:
Having symbols other than comma, period, and space inside the currency is not unique to the Yen. Swiss Francs would be another example. 

I just tried to import my books from a year when I had some extra travels (from Money Manager Ex) and ran into lots of issues with currencies. I even thought I could make it work by adopting only ISO-4217 symbols, but that's a not quite a go either. Basically it doesn't seem possible to get properly localized currency displays yet.

* KZT (Kazakhstan Tenge) adopts spaces for thousands separators only if there are ten or more thousands (`₸1000,00` but `₸10 000,00`). Space separators work, but it's all or nothing.
[...]

Issue #793, opened 2018-06-01
clarify amount parsing/rendering/commodity directives


simonmichael:
Here are some notes and actions aimed at improving amount parsing, rendering, and commodity directives, from this [mail thread](https://groups.google.com/forum/#!msg/hledger/GqDwXF1LAJY/atAj60JeBgAJ) which spun off from [#698](https://github.com/simonmichael/hledger/issues/698#issuecomment-390783216).

## problems with amount parsing/rendering (in current master)

- numbers are parsed loosely, accepting any of two decimal separators
- we don't warn about inconsistent choice of decimal separator across amounts
- ambiguous-separator amounts can be silently misparsed (a single digit group separator is interpreted as a decimal separator)
- commodity directives, the recommended solution, have unclear semantics
  - they are used for: declaring commodity symbols, resolving input decimal separator ambiguity, controlling output style
  - the directive's scope for each of of these is non-obvious. Should subfiles be affected ? other files ? transactions in the same file preceding the directive ?
- commodity directives used to resolve the input decimal separator also limit output style
  - they force digit group separators to appear in output
  - they force the output decimal separator to match the input
- D directives do all that commodity directives do and more, adding complexity
- allowing commodities to have different input decimal separators within a file is excessive flexibility
- allowing commodities to have different output decimal separator within a report is excessive flexibility
- lack of clarity makes this annoying to learn, costly to support; ongoing issue reports

## goals

- be convenient and intuitive
- be i18n-aware
- detect and report errors, avoid guessing
- don't require learning detailed rules
- in every situation, do something that's sensible at least in hindsight
- keep as much backward and sideways compatibility as possible
- end confusion and bug reports about basic number parsing and rendering

## a short term plan

- clarify commodity directives' scope, use them for describing both input and output
  - a directive for a commodity sets its input decimal point for the rest of the current file, exclusive of subfiles
  - a directive for a commodity declares its symbol's validity for the rest of the current file, exclusive of subfiles
  - the first directive for a commodity across all files sets its output style in the report
- refine decimal separator parsing
  - commodities and input decimal separators are always declared together, and per file
  - at the start of each file (and each included file), no commodities or input decimal separators are known
  - parsing a commodity directive
    - declares the commodity and its input decimal separator for the current file
    - the directive's amount must have a separator
    - if the separator is ambiguous is it assumed to be the decimal separator. Could display a warning.
    - if this is the first directive for this commodity among all files: also declares the commodity's output style for the report
[...]

PR #798, opened 2018-06-03
JPY separators

Ekleog:
OK so I've taken a first stab at doing #796.

However, there's something I don't get about unit tests: in my second commit I added a clearly-wrong test (with a willingly changed digit), and yet `stack --nix test` doesn't fail (though I had to locally reset the `stack.yml` snapshot to `[…]-15` instead of `-25` in order to remove the need for ghc8.4.2, but I guess this shouldn't matter much).

From my reading of https://github.com/simonmichael/hledger/wiki/Developer-workflows it should however have been tested and failed (at least the place where I've added the test looks very much like a unit test), so I'm not sure what I'm doing wrong in the adding-a-test or the testing procedure?

Disclaimer: I'm not used to haskell development, and even though I did a few toy projects with it, I still don't really grok how the toolchain actually works.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -700,7 +700,7 @@ fromRawNumber raw mExp = case raw of
     in  Right (quantity, precision, mDecPt, Nothing)
 
   WithSeparators digitSep digitGrps mDecimals -> case mExp of
-    Nothing -> 
+    Nothing ->
       let mDecPt = fmap fst mDecimals
           decimalGrp = maybe mempty snd mDecimals
           digitGroupStyle = DigitGroups digitSep (groupSizes digitGrps)
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -822,7 +822,10 @@ isDecimalPointChar :: Char -> Bool
 isDecimalPointChar c = c == '.' || c == ','
 
 isDigitSeparatorChar :: Char -> Bool
-isDigitSeparatorChar c = isDecimalPointChar c || c == ' '
+isDigitSeparatorChar c =
+  isDecimalPointChar c || c == ' ' ||
+  -- Japanese support
+  c == '兆' || c == '億' || c == '万'
 
 
 data DigitGrp = DigitGrp {
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -317,7 +317,7 @@ formatdirectivep expectedsym = do
   Amount{acommodity,astyle} <- amountp
   _ <- lift followingcommentp
   if acommodity==expectedsym
-    then 
+    then
       if asdecimalpoint astyle == Nothing
       then parserErrorAt pos pleaseincludedecimalpoint
       else return $ dbg2 "style from format subdirective" astyle


simonmichael:
It looks like some bit-rot in the tests, I don't believe TESTS is defined any more.

The travis test failed because of the warning about the unused jpy definition. `make travistest` is a little better at finding things that will fail on travis.

simonmichael:
PS, thanks for diving in!

I haven't understood the above docs in detail; how would you explain 1兆2,345億6,789万0,123円 to westerners ? Is there some way to simulate how it would sound when read aloud ?

FYI the code until now assumes one digit group separator in all numbers. And it's probably about to assume the same digit group separator among all numbers in a file (see #793). Your changes so far should allow numbers to be parsed, but the style won't be preserved in the output.

Ekleog:
> I haven't understood the above docs in detail; how would you explain 1兆2,345億6,789万0,123円 to westerners ? Is there some way to simulate how it would sound when read aloud ?

Hmm, so I'm not japanese, so I can't positively assert anything, only learning the language and about to go for a quite long period over there :)

First, 1兆2,345億6,789万0,123円 is 1234567890123円 : the kanji are only used for grouping. Grouping is done by 4-digits, as is done in pronunciation, and commas are put at thousands, because otherwise it'd be written １兆２千３百４十５億６千７百８十９万百２十３円 for correct pronunciation, which is waaaay too long.

> Can any of those three characters appear at the start or end of a number ? The current code might parse them as commodity symbols in that case.

Then, I'm almost sure numbers can't *start* with a digit grouping: even though for 十, 百 or 千 (10, 100 and 1000) they can be used without a preliminary number, 万, 億 and 兆 (1,0000 , 1,0000,0000 and 1,0000,0000,0000) are always used with a preceding 1 when they are used.

However, I guess they *can* appear at the end of digits: １万円 is 10,000¥. So I guess I've got some more work to do here.

> FYI the code until now assumes one digit group separator in all numbers. And it's probably about to assume the same digit group separator among all numbers in a file (see #793). Your changes so far should allow numbers to be parsed, but the style won't be preserved in the output.

Hmm, I'm not sure about that? I already have
```ledger
commodity €
    format 1.000,00€
commodity $
    format $1,000.00
```
in my ledger file, and output is alright, with both € and $ in the right display format.

So somehow it already manages to do the right thing. I'll have to look into how to make it handle multiple digit separators, I was sure indian rupees quoted in the docs had mixed dots and commas, but it turns out I was wrong.

Thanks for the pointer!

> It looks like some bit-rot in the tests, I don't believe TESTS is defined any more.
> 
> The travis test failed because of the warning about the unused jpy definition. make travistest is a little better at finding things that will fail on travis.

Hmm… I'm going to try and fix the tests, then, thanks!

Ekleog:
Well, I don't understand at all the errors `stack --nix test` now gives me, as when I try to perform type inference in my head I really believe this *should* type… I guess I'm reaching my limits of Haskell knowledge here, so depending on your answer I'll likely stop trying to unbreak these tests, and just put my added test somewhere else.

Anyway, thank you already for your first feedback!

simonmichael:
Thanks for the info. Investigating the disabled tests would be welcome, but it's probably best handled as a separate task.

Ekleog:
OK, so upon more reading it looks like it'll require a complete refactoring of the parser, around `rawnumberp` [(here)](https://github.com/Ekleog/hledger/blob/4606759edd6c40394b90a6b5a6f1100192005d5f/hledger-lib/Hledger/Read/Common.hs#L749) especially, as the current return type doesn't allow to model multiple digit group separators.

So I unfortunately won't have time to do this in the near future, leaving the PR open in case someone else comes along and wants to know what's to be done. I'll try to get back to it as soon as possible, but don't expect better than a few months at best :/

simonmichael:
Understandable. Thanks for the prototyping.

PR #804, opened 2018-06-07
Display the line on which a parse error occurs

awjchen:
This PR works towards improving the error messages in hledger by displaying the line of source on which a parse error occured. This was done by (1) embedding the source positions and the source itself into megaparsec's custom error type, and (2) by using megaparsec's `parseErrorPretty'` function, which was modified to suit our needs. 

In order to ensure that all parse errors go through megaparsec's machinery, this PR removes the `ExceptT String` layer of the parser and the `ErroringJournalParser` type. This required a rewrite/refactoring of the comment parsers and of the `includedirectivep` parser. (Unfortunately, the new comment parsers are slightly slower and use slightly more memory.)

In order to show an example parse error, one of the last commits refines the date parser, producing parse errors like this:

```
> cat test.journal 
include a.timeclock
> cat a.timeclock 
i 2016-01/31 12:00:00 a:aa
o 2016-01-01 16:00:00
> hledger -f test.journal reg 
hledger: in file included from test.journal:1:20,
a.timeclock:1:3:
  |
1 | i 2016-01/31 12:00:00 a:aa
  |   ^^^^^^^^^^
invalid date (mixing date separators is not allowed): 2016-1/31

```

More work will be needed to refine the other error messages, and I expect that the implementation of the custom error messages will probably need to change based on the needs we discover during this work.

Question: I've basically copied some of the source of megaparsec into hledger. What do we need to do to satisfy licensing stuff?

Smaller question: Have I overdone the comments for the new comment parsers? I'm not sure if I've commented on the right things.

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -91,7 +91,6 @@ import Data.Time.Calendar
 import Data.Time.Calendar.OrdinalDate
 import Data.Time.Clock
 import Data.Time.LocalTime
-import Data.Void (Void)
 import Safe (headMay, lastMay, readMay)
 import Text.Megaparsec
 import Text.Megaparsec.Char
@@ -313,7 +312,7 @@ earliest (Just d1) (Just d2) = Just $ min d1 d2
 
 -- | Parse a period expression to an Interval and overall DateSpan using
 -- the provided reference date, or return a parse error.
-parsePeriodExpr :: Day -> Text -> Either (ParseError Char Void) (Interval, DateSpan)
+parsePeriodExpr :: Day -> Text -> Either (ParseError Char CustomErr) (Interval, DateSpan)
 parsePeriodExpr refdate s = parsewith (periodexpr refdate <* eof) (T.toLower s)
 
 maybePeriod :: Day -> Text -> Maybe (Interval,DateSpan)
@@ -373,13 +372,13 @@ fixSmartDateStr :: Day -> Text -> String
 fixSmartDateStr d s = either
                        (\e->error' $ printf "could not parse date %s %s" (show s) (show e))
                        id
-                       $ (fixSmartDateStrEither d s :: Either (ParseError Char Void) String)
+                       $ (fixSmartDateStrEither d s :: Either (ParseError Char CustomErr) String)
 
 -- | A safe version of fixSmartDateStr.
-fixSmartDateStrEither :: Day -> Text -> Either (ParseError Char Void) String
+fixSmartDateStrEither :: Day -> Text -> Either (ParseError Char CustomErr) String
 fixSmartDateStrEither d = either Left (Right . showDate) . fixSmartDateStrEither' d
 
-fixSmartDateStrEither' :: Day -> Text -> Either (ParseError Char Void) Day
+fixSmartDateStrEither' :: Day -> Text -> Either (ParseError Char CustomErr) Day
 fixSmartDateStrEither' d s = case parsewith smartdateonly (T.toLower s) of
                                Right sd -> Right $ fixSmartDate d sd
                                Left e -> Left e
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -183,6 +183,7 @@ instance NFData PostingType
 type TagName = Text
 type TagValue = Text
 type Tag = (TagName, TagValue)  -- ^ A tag name and (possibly empty) value.
+type DateTag = (TagName, Day)
 
 -- | The status of a transaction or posting, recorded with a status mark
 -- (nothing, !, or *). What these mean is ultimately user defined.
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -14,6 +14,7 @@ Some of these might belong in Hledger.Read.JournalReader or Hledger.Read.
 
 --- * module
 {-# LANGUAGE CPP, BangPatterns, DeriveDataTypeable, RecordWildCards, NamedFieldPuns, NoMonoLocalBinds, ScopedTypeVariables, FlexibleContexts, TupleSections, OverloadedStrings #-}
+{-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE PackageImports #-}
 
@@ -81,12 +82,10 @@ module Hledger.Read.Common (
   -- ** comments
   multilinecommentp,
   emptyorcommentlinep,
-  followingcommentp,
-  followingcommentandtagsp,
 
-  -- ** tags
-  commentTags,
-  tagsp,
+  followingcommentp,
+  transactioncommentp,
+  postingcommentp,
 
   -- ** bracketed dates
   bracketeddatetagsp
@@ -98,6 +97,7 @@ import "base-compat-batteries" Prelude.Compat hiding (readFile)
 import "base-compat-batteries" Control.Monad.Compat
 import Control.Monad.Except (ExceptT(..), runExceptT, throwError) --, catchError)
 import Control.Monad.State.Strict
+import Data.Bifunctor (bimap, second)
 import Data.Char
 import Data.Data
 import Data.Decimal (DecimalRaw (Decimal), Decimal)
@@ -383,6 +383,8 @@ datep' mYear = do
     Nothing -> fail $
       "partial date "++dateStr++" found, but the current year is unknown"
       where dateStr = show month ++ [sep] ++ show day
+
+{-# INLINABLE datep' #-}
 
 -- | Parse a date and time in YYYY/MM/DD HH:MM[:SS][+-ZZZZ] format.
 -- Hyphen (-) and period (.) are also allowed as date separators.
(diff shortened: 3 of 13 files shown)


simonmichael:
Sounds really good. I'm curious how much megaparsec code you had to copy, and why ?

simonmichael:
Also, just curious, do you know why megaparsec requires us to define a custom type to have these fancy errors ? What we're doing seems like something generic megaparsec could support out of the box.

awjchen:
Thanks for the review!

>Also, just curious, do you know why megaparsec requires us to define a custom type to have these fancy errors ? What we're doing seems like a generic use case that megaparsec could support out of the box.

> I see the megaparsec code now. megaparsec is BSD so we may use it freely, but even though it's only a small amount we should probably include their copyright notice and LICENSE text somewhere (click on the license on hackage for details). Do you think it would make sense to contribute to megaparsec instead ? If not [yet], it might be convenient to keep it separate, eg in Text.MegaParsec.Custom (there is some other conventional naming scheme which I don't recall).

About the need for a custom type, no, I'm not sure why. Since I'm not sure I hesitate to suggest a change to megaparsec. Maybe I'll take a look at the megaparsec issues and pull requests to see if this has been discussed before.

I will include the megaparsec copyright and license stuff. I'll try to look around at other projects to see how they do these things.

I agree that it would be convenient to keep the custom parse errors module separate.

> Naming. Something more precise than CustomErr would be nice, but I see it's not easy. ParseError is used and HLParseErr isn't great.

Hmm. I'm still having a hard time coming up with a name other than `CustomErr`, because to me the type is just a way to extend megaparsec's error handling in whatever way we need. Perhaps it would be better to make reference to parsing in the name of the type? But maybe this is not necessary because the only contexts in which `CustomErr` currently appears (outside of the ParseErrors module) are `MonadParsec CustomErr s m` and `ParseError Char CustomErr`, which already reference parsing.

> I try to keep each term in hledger jargon to a single meaning, and Interval is already pretty heavily used for http://hledger.org/manual.html#report-intervals. Could we find something else for this ? Span is also taken (date span). Range ? Extent ? Region ?

Perhaps we could rename `parseErrorAtInterval` to `parseErrorBetween`?

> I see some INLINABLE's added to existing parsers. Not needed now but personally I would give such things separate commits in case we want to check the performance impact now or in future.

Makes sense. Thanks for the pointer! (FYI, if I remove the INLINABLEs I see a decrease in speed of about 2%.)

> Can you explain how we allow terminating the entire parse at any point, now that ErroringJournalParser is gone ? I missed it on my read through.

Ah, I forgot about the ability of `ErroringJournalParser` to terminate parsing at any point. If we remove `ErroringJournalParser`, we would no longer have that ability. However, I feel that we should consider removing the ability to terminate parsing at any point because we currently don't make use of it. To me, it seems like (a small amount of) unnecessary complexity; though, it does not seem to be a burden to keep around. The only problem I see is in keeping our error messages consistent between two error reporing mechanisms, if we wish to do so, because that would require us to duplicate some of megaparsec's machinery for include files; for example, (1) keeping track of a stack of source positions (currently done with `pushPosition`), and (2) making the source text available to the parsers throwing errors (currently done with `region`).

simonmichael:
Hmm, I thought we did make use of it, eg everywhere we throwError ?  The point (as I think you know) is to produce a specific parse error and stop when we encounter something definitely wrong, rather than let the parser continue trying other things and then giving a less useful error.

simonmichael:
(If all our parse errors are still as specific as before this change, I guess I'm misunderstanding and there's no problem.)

awjchen:
Ah, I overlooked the specificity of parse errors when removing `ErroringJournalParser`. You are right that that this PR makes the parse errors less specific than before, but this loss of specificity is restricted to one location: `commoditydirectiveonelinep` in `Hledger/Read/JournalReader.hs`. This is because we only lose parse error specificity if we backtrack, and because `commoditydirectiveonelinep` contains the only instance of `throwError` (via `parserErrorAt`) used within a `try` block. In this case, the loss of specificity can be amended by reducing the scope of the try block enclosing `commoditydirectiveonelinep` to exclude the final conditional that calls `throwError`. With this fix, I think that our parser errors will still be as specific as before this PR.

Regarding the uses of `throwError`: all current instances of `throwError` are used for one of two purposes: (1) to re-throw a parse error obtained during "re-parsing" (in `followingcommentandtagsp` and `includedirectivep`), or (2) throwing a parse error at a specific source position via `parserErrorAt` (`commoditydirectiveonelinep`, `formatdirectivep`, and `defaultcommoditydirectivep`). The first category of `throwError` was eliminated by rewriting/refactoring, and the second by changing `parserErrorAt` to use the `ErrorFailAt` custom parse error.

So, I believe there is no _current_ need for `ErroringJournalParser` and `ExceptT/throwError`. I think I can appreciate how they could be used to simplify backtracking logic, but I would prefer to use them only after finding "ordinary" parser combinators insufficient for the task at hand.

simonmichael:
I want to believe you but.. are there not many places where we can backtrack, not just at throwError calls ? I'm thinking of calls to fail, try, etc. And perhaps more than that. I'm a bit unclear on this.

I haven't spent time trying to come up with some examples of less-specific parse errors in the new code. I'm not sure how much testing of parse error messages we have currently. Perhaps we should have a test that just checks for regressions in a collection of common parse errors.

awjchen:
Yes, I agree that we should have a set of tests for parse errors.

It is true that there are other places in the parser where backtracking might be possible, and indeed I have found another case where this PR introduces additional backtracking. However, unless I have missed something, I believe there are no other instances.

Here is my reasoning:

1. I would like to show that there is no loss in parse error specificity due to the changes in this PR. However, I will find two cases where this is not true, but I verify the other cases.
2. `ExceptT` errors can be used to increase parse error specificity by unconditionally terminating parsing, whereas parse errors thrown in the `ParsecT` layer are subject to backtracking. We are only concerned about the loss in specificity from (a) an increase in backtracking caused by (b) switching from errors in the `ExceptT` layer to errors in the `ParsecT` layer.
3. Regarding (b): the only way we throw errors in the `ExceptT` layer is through the functions imported from `Control.Monad.Except`, from which we import `ExceptT`, `runExcept`, and `throwError`.
4. Only `ExceptT` and `throwError` (and the only parser helper function that uses them: `parserErrorAt`) can be used to throw errors.
5. Regarding (a): Backtracking may only occur when a parser fails without consuming any input. The only user-available mechanism by which parsers can be made not to consume input on failure is through the `try` combinator.
6. In our parsers, we only make use of `try` via `choice'`, `choiceInState` and `try` itself.
7. Therefore, we only need to care about uses of `throwError` and `ExceptT` that occur either (i) before a parser has had the opportunity to consume any input, or (ii) somewhere "underneath" `choice'`, `choiceInState`, or `try`.
8. A negative example is `includedirectivep`, which makes two direct uses of `throwError`. These uses of `throwError` do not satisfy (i) because the `string "format"` parser must succeed and consume input before reaching the points at which `throwError` is used. They also do not satisfy (ii) because they are not contained within any backtracking within `includedirectivep`, nor is `includedirectivep` contained within any backtracking within its sole caller, nor `directivep`, and so on with `addJournalItemP`, and then `journalp` itself. Therefore, using e.g. `fail` (from the `ParsecT` layer) here, rather than `throwError`, will cause a parse error that will be "propagated" to a failure in the top level parser, `journalp`, with no opportunity for backtracking into other alternatives.
9. Other than in the following point (point 10) and `commoditydirectiveonelinep`, I have checked all other uses of `throwError` and `ExceptT` using the same process in point 8 and have found no introductions of additional backtracking. These locations are `followingcommentandtagsp`, `orRethrowIOError`, `formatdirectivep`, `defaultcommoditydirectivep`.
10. However, I have found an instance where removing the `ExceptT` layer potentiailly introduces additional backtracking: `journalp` is itself called in `includedirectivep` under a `choiceInState`. So then, switching to a "regular" parse error would introduce backtracking into the other alternatives (`timeclockfilep` and `timedotfilep`). This applies to all removals of the `ExceptT` errors under `journalp`(!).
11. But then again, the problem raised in point 10 may not be a big issue in practice. This is because, if I understand correctly, when merging parse errors from backtracking, megaparsec prefers the longest match; that is, megaparsec chooses the parse error that occurs at the position furthest from the start of the input. So, if a journal file contains an error and fails to parse at a certain point, it is likely that the other parsers will either fail with the same error at the same point, or fail at an earlier point due to a difference in format. In these two cases, the original, intended error is preserved.
12. There may be other ways of working around the problem in point 10; for example, by giving users the option to declare the filetype to be either "journal", "timeclock", or "timedot".

simonmichael:
That looks like smart reasoning. We should keep in mind the goal is not just "don't be less specific than before" but also "allow us to make it as specific as we want in future". If you think we can generate clear and specific errors with just JournalParser, then let's try it. I would like to start a collection of parse error tests as well. tests/journal/parse-errors.test ? I'd also love to get this merged soon to unblock other PRs.

simonmichael:
PS I feel like there must be some whole-parse-terminating function built in to megaparsec. Perhaps `failure` is it ?

awjchen:
> We should keep in mind the goal is not just "don't be less specific than before" but also "allow us to make it as specific as we want in future".

That's a good point.

> If you think we can generate clear and specific errors with just JournalParser, then let's try it. 

I won't be completely sure that we will be able to generate specific parse errors to our satisfaction with just `JournalParser` until I've done the work to apply these new custom errors to all the parsers, but given our current parsing needs, I think that being careful about using `try` and backtracking should take us a long way towards our goal. So, yes, let's try it!

> I would like to start a collection of parse error tests as well. tests/journal/parse-errors.test ?

Cool, I see you've already started. I think I'll need some time to think about how to test for the specificity of parse errors, since I don't really have a clear idea of what to do.

> I'd also love to get this merged soon to unblock other PRs.

I've done some work to address most the concerns we discussed above. I have
- included the megaparsec copyright and license with the copied code
- moved the custom parse errors module to `Text.Megaparsec.Custom`
- renamed `parseErrorAtInterval` to `parseErrorAtRegion` (no new commit, since I did this with `git rebase`), and
- reduced the backtracking of `commoditydirectiveonelinep`

I still haven't thought of a better name for `CustomErr`, but since other PRs are waiting on this one, I suppose we can do this renaming later.

>PS I feel like there must be some whole-parse-terminating function built in to megaparsec. Perhaps `failure` is it ?

I also tried to find such a function in megaparsec, but could not find any. Looking at the source, `failure` seems to use the same mechanism as `fail`, so it looks like just a "regular" parse error.

simonmichael:
Re testing parse errors, what I have in mind is just some tests of full parse error messages, so that when they change we'll get some warning and a chance to make sure they have not regressed.

simonmichael:
Sounds good! Thank you!

simonmichael:
Here's the date parser's output before and after, for the record:
```
$ cat a.j
2018/1/
$ hledger-1.9 -f a.j print
hledger-1.9: /Users/simon/src/PLAINTEXTACCOUNTING/hledger/a.j:1:8:
bad date: 2018/1/

$ hledger -f a.j print
hledger: /Users/simon/src/PLAINTEXTACCOUNTING/hledger/a.j:1:8:
  |
1 | 2018/1/
  |        ^
unexpected newline
expecting day

```

PR #807, opened 2018-06-09
allow descriptions/comments in periodic transactions

simonmichael:


simonmichael:
I will hold this until after #804 lands, but any comments welcome. It improves the docs, and makes periodic transaction rules better at generating realistic transactions, which is useful when using them as a data entry aid.

simonmichael:
PS functional tests are not yet passing.

simonmichael:
Also:

It parses periodic transactions eagerly, along with the rest of the journal (not just when --forecast or --budget is used).

It enables the use of relative dates in periodic transactions, which previously was not allowed.

barrucadu:
I think it would be good to add a test for a periodic transaction with a description to forecast.test, maybe to budget.test too.

simonmichael:
Done. I didn't bother for budget.test since --budget doesn't use the descriptions.

hledger-lib/Hledger/Read/JournalReader.hs
```diff
@@ -457,14 +457,45 @@ modifiertransactionp = do
...
+  -- Instead, read until two or more spaces and reparse that. More use of two spaces is not ideal.
+  pos <- getPosition
+  periodtxt <- lift singlespacedtextp
+  d <- liftIO getCurrentDay
+  (interval, span) <-
+    case parsePeriodExpr d periodtxt of
+        Right (i,s) -> return (i,s)
+        Left e -> 
+          -- Show an informative error. XXX a bit unidiomatic, check for megaparsec helpers  
+          fail $ -- XXX
```
simonmichael: @awjchen if you have any time or advice, this function could use your help. It does reparsing, and I replaced two throwError calls with fail, which might allow too-vague parse errors (not sure).

awjchen: Hey,

I've attempted some changes in https://github.com/awjchen/hledger/commit/fcf95c1db5957f58e7bf2766d79f8b7e505d61a4 to eliminate the "re-parsing". I used the `match` combinator directly on `periodexprp`, which required generalizing the types of the parsers in `Data/Dates.hs` -- I hope this is okay.

I haven't looked at the this date handling stuff before, so I've only tried to make small changes while trying to preserve behaviour. I've tested the changes with `stack install --fast --test && make functest`, but I wouldn't know if I've messed something up.

simonmichael:
Thank you very much! I have tweaked it to require at least one space between the period expression and following fields, if any.

The boundary between period expression and description is now thinner and it might be possible for the period expression parser to consume part of the description. And two spaces is a bit more human readable. Overall though, I think it's better not to require two spaces. And all parse errors are still good enough I expect.

This has landed in master as c4d45cff.

simonmichael:
Actually, here are some problems. Consider this monthly rule, which at the time of writing generates some transactions in 2018:
```
~ monthly  We will change this in 2019.
  (a)  1

$ hledger -f a.j bal -QN --forecast
Balance changes in 2018/07/01-2018/12/31:

   || 2018q3  2018q4 
===++================
 a ||      3       3 
```

An unlucky choice of wording in the description can change the period expression (and the description), despite the capitalisation and the double space, which could be surprising.
```
~ monthly  In 2019 we will change this.
  (a)  1

$ hledger -f a.j bal -QN --forecast
Balance changes in -:

  || - 
==++===

$ hledger -f a.j bal -QN --forecast -e 2020
Balance changes in 2019:

   || 2019q1  2019q2  2019q3  2019q4 
===++================================
 a ||      3       3       3       3 
```
Another example:
```
~ monthly  next year we will change this.
  (a)  1

$ hledger -f a.j bal -QN --forecast -e 2020
Balance changes in 2019:

   || 2019q1  2019q2  2019q3  2019q4 
===++================================
 a ||      3       3       3       3 
```
And capitalising this one even gives an error:
```
~ monthly  Next year we will change this.
  (a)  1

$ hledger -f a.j bal -QN --forecast -e 2020
hledger: Prelude.read: no parse
```

simonmichael:
We could disallow double spaces in period expressions. Just within periodic transaction rules, or always. 

Then, users could use double space after the period expression to prevent the problem above. 

We could document and recommend this but not require it, to provide a smoother experience to users not aware of the rule. Or we could go back to always requiring the double space. Or, do that only in strict mode.

awjchen:
I'm not sure what the right design is, but I've just went and attempted to restore the behaviour before my above change.

The change I made above allowed the period expression parsers to parse double-spaces. This was because I changed the code to parse with `periodexprp` directly, instead applying it to the result of `singlespacedtextp`, and because the period expression parsers use `skipMany spacenonewline` to parse spaces. Changing the code to use `optional spacenonewline` instead of `skipMany spacenonewline` (https://github.com/awjchen/hledger/commit/f0a1888cd678b145679ab238c7728e9deffd70e5) allows at most one space at a time, and seems to restore the old behaviour.

simonmichael:
@awjchen can you remind me of the status here ? Do we have one subsequent change by you awaiting review, https://github.com/awjchen/hledger/commit/f0a1888cd678b145679ab238c7728e9deffd70e5 ? To get that showing up for review maybe it should be a PR or I should recreate the forecast-improvements branch or something ?

awjchen:
I'm thinking of abandoning the approach I used in https://github.com/awjchen/hledger/commit/f0a1888cd678b145679ab238c7728e9deffd70e5. It requires many manual changes to the parsers in `Data/Dates.hs` and introduces an invariant (i.e. that we should only ever parse single spaces in those parsers).

It seems it would be more natural to do things the way you originally did them in https://github.com/simonmichael/hledger/pull/807/commits/e542ca93f24d2b3fbcebe56fb992e7a8687d32b0, which is to first parse single-spaced text, then "re-parse" that text with the parsers in `Data/Dates.hs` as they are.

So, I've been trying to extend our custom parse errors to support "re-parsing" in https://github.com/awjchen/hledger/commit/ec2ec921e4f32aa023cce09c880d5aee4de02de8. I've applied those changes to the number parsers in https://github.com/awjchen/hledger/commit/cd77d0bd204212a476fd6beac58814c1304b5b61 [WIP] and, so far, it seems to work.

(I'm applying "re-parsing" to the number parsers in anticipation of the changes for issue 793, since it could be the case that we would have to parse the commodity before even beginning to interpret a number.) 

What do you think of this way of implementing "re-parsing"?

If it's fine, I would use it to reimplement the behaviour in your original commit https://github.com/simonmichael/hledger/pull/807/commits/e542ca93f24d2b3fbcebe56fb992e7a8687d32b0, undoing my workaround https://github.com/awjchen/hledger/commit/fcf95c1db5957f58e7bf2766d79f8b7e505d61a4.

simonmichael:
I pushed some (disabled) tests to master for this (see #823). 

I think the next goal is to consume only up to 2+ spaces when parsing the period expression, as it did before. Should it also stop at a semicolon (comment character), so eg `~ monthly ;in 2020` or `~ monthly;in 2020` would work ?

simonmichael:
I think yes, that would be expected to work. 

And, should these delimiters be enforced for all period expressions, including eg -p (periodexprp) or just the ones in periodic transaction rules (periodictransactionp) ?

simonmichael:
PS no pressure but if you get motivated to work on this, please do; I won't get to it for a bit. We could use the old code or your new re-parsing helpers. I'd like to focus just on the periodic transaction rule parsing for now to reduce scope.

awjchen:
Sure, I can try to work on this.

Though, I might need some time to think about how to proceed, since megaparsec 7.0.0 is making a lot of changes (https://github.com/mrkkrp/megaparsec/blob/master/CHANGELOG.md) to the parse errors. It looks like these changes will clash with or make redundant what I've been (ab)using megaparsec's custom parse errors for (e.g. it looks like pretty-printing parse errors with the offending line will now be handled by megaparsec with `ParseErrorBundle`!).

If I didn't know that these changes were incoming, I would use the "re-parsing" combinators introduced in this PR. But now that I know, I'm thinking of taking a conservative approach, where I would try to avoid unexpected clashes by implementing the functionality within an `ExceptT` layer (i.e. what we had before). Alternatively, I suppose I could try to build something on top of the master branch of megaparsec, or I could delay this work until the next version of megaparsec is released.

What kind of an approach do you think is best?

simonmichael:
Thanks for the heads-up. I expect we'll want to use the new goodies, if we can still build with enough old GHC versions. If MP is still changing perhaps we should wait, or if it's near-stable we could experiment with the latest. There's really no urgency on this issue from users, it's just a loose end that's bugging me (and generating some useful insights).

simonmichael:
megaparsec 7 is out. \o/

PR #811, opened 2018-06-17
Add screenshots for Sandstorm

ocdtrekkie:
I haven't tested this, but the formatting should be right. These PNGs have been shrunk using tinypng.com to ensure they don't significantly increase the size of the Sandstorm metadata. This should add these screenshots to the app market next time HLedger Web for Sandstorm is built.

I used demo.hledger.org to get a good image of what a used HLedger should look like, including an account view to show a graph, and a populated add transaction form to hint at what that process should look like.


ocdtrekkie:
As a bonus, I just tossed in a README.md for the .sandstorm directory which documents the process for building the Sandstorm version.

simonmichael:
Thanks!

PR #800, opened 2018-06-04
lib: WIP attempt to unbreak tests

Ekleog:
Just pushing this as a PR in case someone wants to have a look at it; I don't get the typing error output by haskell.

PR #812, opened 2018-06-17
lib: Fix compile errors in tests

zarybnicky:
A continuation of #800.

Unfortunately, the tests don't pass - there are three test failures here, where I don't know what the intended result is:
```
: [Failed]
values are equal
expected: Amount {acommodity="$", aquantity=10.00, ..}
 but got: Amount {acommodity="$", aquantity=10, ..}
: [Failed]
Equal comment
expected: " tcomment1\n tcomment2\n ttag1: val1\n"
 but got: "tcomment1\ntcomment2\nttag1: val1\n"
: [Failed]
Posting is parsed well
expected: " a: a a \n b: b b \n"
 but got: "a: a a\nb: b b\n"
```

I also saw a ton of disabled and commented out tests - I suppose those should eventually be enabled and uncommented again?

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -881,7 +881,6 @@ journalDateSpan secondary j
       pdates   = concatMap (catMaybes . map (if secondary then (Just . postingDate2) else pdate) . tpostings) ts
       ts       = jtxns j
 
--- #ifdef TESTS
 test_journalDateSpan = do
  "journalDateSpan" ~: do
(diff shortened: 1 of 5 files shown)


simonmichael:
@zarybnicky, thanks for pushing this forward. A little while back I got those tests passing, but I wasn't happy with my first solution or with the nail biting/teeth clenching process required to get there, which led to a [massive yak shave](https://github.com/simonmichael/hledger/compare/510b4c1...e9c5d13a) to improve test running and output, and a simpler fix.

PR #814, opened 2018-06-18
doc: Remove non-ASCII from man- and infopages

### LINKED ISSUE #813: compilation fails on embedded manuals with locale set to C
Since the last few commits, I've started having this error while trying to compile hledger:
```
    /home/inuits/hledger/hledger/Hledger/Cli/DocFiles.hs:58:6: error:
        * Exception when trying to run compile-time code:
            /home/inuits/hledger/hledger/embeddedfiles/hledger_journal.5: hGetContents: invalid argument (invalid byte sequence)
          Code: makeRelativeToProject "embeddedfiles/hledger_journal.5"
                  >>= embedStringFile
        * In the untyped splice:
            $(makeRelativeToProject "embeddedfiles/hledger_journal.5"
                >>= embedStringFile)
       |
    58 |     ($(makeRelativeToProject "embeddedfiles/hledger_journal.5" >>= embedStringFile)
       |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Digging into it further, I've found several non-ASCII characters in *.m4.md files. AFAIK, having non-ASCII in man or info pages means that they are non-portable (even though most modern Linux boxes come with *.UTF-8 locales). I'm guessing I've run into it due to my Czech locale and running `make` with `LC_ALL=C`.

zarybnicky:
Fixes #813. For some reason, my whitespace distribution is wildly different in *.txt files, so I've committed only the relevant parts of them.

### DIFF
--- a/hledger-lib/hledger_journal.5
+++ b/hledger-lib/hledger_journal.5
@@ -649,8 +649,8 @@ Write the price per unit, as \f[C]\@\ UNITPRICE\f[] after the amount:
 .nf
 \f[C]
 2009/1/1
-\ \ assets:euros\ \ \ \ \ €100\ \@\ $1.35\ \ ;\ one\ hundred\ euros\ purchased\ at\ $1.35\ each
-\ \ assets:dollars\ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ \ ;\ balancing\ amount\ is\ \-$135.00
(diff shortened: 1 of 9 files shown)


zarybnicky:
This was just a stopgap and not what's actually needed, more info in #813.

Issue #657, opened 2017-11-26
hledger-web only processes first file -f argument


dbp:
If I run: `hledger-web --server -f file1.ledger -f file2.ledger`, the web UI only shows data from the first file, and the title at the top of the page is `file1.ledger`. This isn't great, especially because I can't run hledger-web from standard in (which granted would probably complicate things, as it would necessitate a read-only mode), and seems to contradict the documentation, which writes:

> Like hledger, it reads data from one or more files in hledger journal, timeclock, timedot, or CSV format specified with -f, or $LEDGER_FILE, or $HOME/.hledger.journal (on windows, perhaps C:/User/USER/.hledger.journal).

Perhaps this limitation has something to do with needing to know where to store added entries, but it would be great if it were fixed. 

There are at least two use cases:

- I having archival .ledger files and then current ones, and ideally want to be able to show them all in the web ui sometimes. 
- I usually process bank transaction when statements are issued (to make it easier to not duplicate things), but I wanted to look at the most current data available, so I downloaded all transactions since last statement. I don't want to add this to my ledger file, but I would like to be able to, temporarily, look through them in the web ui.

zarybnicky:
@simonmichael While solving the Sandstorm permission problem and resurrecting the edit/import forms, I might as well solve this one. What's the standard way of handling multiple files now? Append to the last one specified?

simonmichael:
I suspect we add to the primary (first) file. I would check what the add command does. We should have pretty solid support in hledger-lib for multi-file journals now.

simonmichael:
This got fixed by 6a966836 and released in hledger-web 1.10.

It still only shows the first file's name as the title, pull requests welcome for that.

Issue #641, opened 2017-11-08
hledger-web 1.4 showing zero amount accounts by default


ohmyohmyohmy:
hledger-web 1.4 is showing zero amount accounts by default. So, there's no option to turn it off.

Running on Windows 7.

Thanks!

simonmichael:
Hi, you're right. It was [hard-coded](https://github.com/simonmichael/hledger/blob/master/hledger-web/Handler/Common.hs#L76), as part of hledger-web's goal of being a simple UI that just works without surprises for non-power-users. Is it causing you a lot of trouble ?

ohmyohmyohmy:
It's just not pretty having a whole bunch of Accounts Receivables (about 50) showing zero, on my balance sheet is all.

Plus, I was expecting zero amount accounts to be hidden by default as stated in the doc:

>  -E --empty
>     show items with zero amount, normally hidden

simonmichael:
That does sound annoying. And you're right that it's at least a docs bug. 

On the other hand: I imagine non-expert users starting up hledger-web, and not seeing a full picture of their assets (eg) just because some of them happen to be currently empty, and this seems like a usability/discoverability problem to me.

I'm not using hledger-web a lot at the moment, but maybe what's needed is a discoverable toggle within the UI ? Maybe somebody can work on this ?

ohmyohmyohmy:
Yeah, I see what you mean. Though, I imagine anyone using hledger-web would have started with hledger or ledger-cli first, and then try hledger-web out of curiousity, so the concern with non-experts (which is who I am by the way) may be a non-issue.

Like me, I started with hledger and then tried hledger-web, so I sorta expected both to act the same way, and have zero amount accounts hidden by default and the option to turn it on with -E when I want it to. I actually thought I did something wrong, and was checking my ledger file to see if I entered something, a special comment maybe, that would turn zero-amount accounts on (told you I was a non-expert).

A simple checkbox at the top-left corner with the words "Show accounts with zero amounts" then "refresh", would help, I guess?

simonmichael:
Some possible resolutions:

A. change hledger-web --help/docs to match the current behaviour (always show zero-balance accounts)

B. change hledger-web to behave like hledger (show zero-balance accounts only if you start it with a -E command line flag. Won't be available to GUI users double-clicking the hledger-web icon.)

C. enhance hledger-web so zero-balance accounts can be toggled in the UI (as in hledger-ui), making them discoverable

[...]

Issue #425, opened 2016-10-31
hledger-web on sandstorm is not satisfying [$30]


simonmichael:
Negative feedback is accumulating at https://apps.sandstorm.io/app/8x12h6p0x0nrzk73hfq6zh2jxtgyzzcty7qsatkg7jfg2mzw5n90 , need better Sandstorm integration to fix this.

ndarilek:
A thought I had: Is there any current Git integration? Can the web interface make a commit whenever a transaction is added or modified, or does it already?

One feature I've played with for some more recent Sandstorm apps I've developed is backing them with a Git repo. If the web interface could detect or be told that the ledger is in a repository, and if I can expose that repo via Git, then the ledger could be updated both via the web interface and by the repository. Not sure how to handle incoming pushes while the web interface is open, so the idea obviously needs some additional thought.

simonmichael:
No, we don't have that feature yet. It would be nice. filestore is a haskell library that can talk to git and other VCS's.

AaronM04:
FWIW, my hledger-web on sandstorm.io + git workflow is as follows:

* create all transactions in the web interface (there's no way to import a ledger file, and no way to update it except by adding transactions with the web interface).

* periodically use the Download Backup button in the middle of the top bar, and then run a shell script I created that moves the .zip file from my Downloads directory and commits the ledger file, commits, and finally pushes it to GitLab.

simonmichael:
> there's no way to import a ledger file

Actually there is; just reverse the backup I think. It was described recently on the hledger mail list.

AaronM04:
@simonmichael Cool, I didn't know about that. 💯  What's the link to this thread?

simonmichael:
https://groups.google.com/d/msg/hledger/5BTfowLOkpE/TqVuRjaoDQAJ

simonmichael:
The [sandstorm reviews situation](https://apps.sandstorm.io/app/8x12h6p0x0nrzk73hfq6zh2jxtgyzzcty7qsatkg7jfg2mzw5n90) is getting pretty serious. I would almost prefer we were removed from Sandstorm until ready rather than keep accumulating bad first impressions and bad reviews. @ndarilek is there a way to mark an app as experimental/work in progress/not ready for review ?

simonmichael:
Also, is there any way to contact reviewers, to understand their issues better ? Eg "Could not find a way to enter a single transaction".

simonmichael:
[...]

PR #821, opened 2018-06-24
web: Import & export, permissions from CLI or headers

zarybnicky:
More or less a complete overhaul of hledger-web - the diff is quite scary...

I've attempted to split it into well-defined commits but at the end, I was moving around large chunks of code, so the resulting diff is more-or-less "delete everything & write it anew"...

I've added two new CLI flags: `--capabilities`, which statically controls the capabilities of the application, and `--capabilities-header`, which overrides the previous flag and enables reading the (comma-separated) capabilities from the header given as a parameter.

Also, I've resurrected, finished, and refactored the import and edit forms (of which only one of them had been working at all, not sure which one), added a separate 'Manage journal files' page with a list of all the files the current journal is composed of (and edit/upload/download buttons for each of them).

In addition, having all of the forms as separate endpoints for both GET and POST plays well with yesod-form's error messages - after submitting the add form, I redirect the user to /add (with just the add form), where the error message is displayed next to the relevant field.

This PR also addresses a few other issues:
* originally #657, which has since been fixed in master
* #641, by defining a new keybinding "e", which toggles the hiding/showing of empty accounts
* #433, by redirecting to /add and displaying all errors there (also account/amount parse errors, ...)
* #337 - same as above
* part of #425 - this doesn't solve onboarding or the rest, only the import/edit/export part of it.
* partially solves #328 by adding the edit form, but for that a 'Delete' button per-transaction is still necessary
* part of #327 - CSV export/import is still not implemented, but that should be easy
* #33 should've been closed a while ago IMO - the add form is still visible in text-only browsers at the bottom of the page, but the other forms are on separate pages.

I haven't actually tested it on Sandstorm, only locally - I haven't really looked into what's needed for that. I've defined the permissions in the `*.capnproto` file, edited the `launcher.sh` file, so the only thing remaining is to test it. I've however ran out of time in the short term, so I've finished all the changes to code and opened this PR...


zarybnicky:
I forgot to mention - there are no immediate eye catching differences. Only when you start it with `hledger-web --capabilities=view,add,manage`, then there is an extra wrench icon next to the search bar, which opens the 'Manage' UI

zarybnicky:
Ok, building with `make buildtests-all` says that I need to either update yesod or make a CPP'd type alias for WidgetT vs WidgetFor...

simonmichael:
Looks good! Thank you!

Providing an invalid capability name to --capability causes it to serve "Internal Server Error", it should give a command-line error.

When started without the view capability, perhaps it could also hide the journal's file name.

--capability sounds like the right technical term, but I don't love it for the user experience. Is there something more friendly and familiar - --access, --permission[s], ... ?

I will go ahead and merge, as a release is due tomorrow.

I will just say the addform (and perhaps the others in the past) were in the default template for a reason: to make them always available so that you could hit a to start data entry no matter where you navigate in the app. Lack of this may become more noticeable if we add more screens. But perhaps the old way was ugly and unsustainable.

simonmichael:
PS I see this does even more than I noticed, I'll have to read the linked issues. And, I forgot about docs. We should document the changes in the hledger-web manual (and the web help dialog if you think it appropriate).

simonmichael:
I think I also misread the CI status. `stack test hledger-web` no longer builds:
```
[1 of 4] Compiling TestImport       ( tests/TestImport.hs, .stack-work/dist/x86_64-osx/Cabal-2.2.0.1/build/test/test-tmp/TestImport.o ) [Foundation changed]

/Users/simon/src/PLAINTEXTACCOUNTING/hledger/hledger-web/tests/TestImport.hs:10:1: error:
    Could not find module ‘Foundation’
    Use -v to see a list of the files searched for.
   |
10 | import Foundation
   | ^^^^^^^^^^^^^^^^^
```

After that we should also check `make buildtest-all` passes. I can run that one if necessary.

zarybnicky:
I'll respond more in the evening but for now - `make buildtest-all` works with the last two commits on this branch, where I've fixed these build issues - that's what the green checkmark should mean, right?

zarybnicky:
Nevermind, I don't remember changing the testsuite - I only recall considering removing it, as right now, it is AFAIK useless.

simonmichael:
@zarybnicky: I disabled hledger-web's test suite for now.

How would you feel about changing the zero toggle key to Z, for consistency with hledger-ui and to free up the valuable e/E keys for edit operations ? See https://github.com/simonmichael/hledger/issues/641#issuecomment-344285093 for more context.

zarybnicky:
@simonmichael I was about to get into this again, at least for today :)

re Z/E: Yes, I saw that comment earlier today. I have no strong feelings either way, I'll include `z` in the next hledger-web PR - together with #834 and perhaps more.

re testsuite: Currently, the tests do nothing even if they compile (well, they test that the app starts at all...). I'm not sure how valuable `yesod-test` is here - though I have next to no experience with Selenium or other integration test frameworks for webapps (I only ever used unittests for both frontend JS and backend). Though considering that we've added permissions to the mix, it might be worth filling the testsuite out.

> Providing an invalid capability name to --capability causes it to serve "Internal Server Error", it should give a command-line error.
Will look into it.

> When started without the view capability, perhaps it could also hide the journal's file name.
An oversight on my part, I knew I forgot something...

> --capability sounds like the right technical term, but I don't love it for the user experience. Is there something more friendly and familiar - --access, --permission[s], ... ?
See #834, which I've opened for further discussion.

> I will just say the addform (and perhaps the others in the past) were in the default template for a reason: to make them always available so that you could hit a to start data entry no matter where you navigate in the app. Lack of this may become more noticeable if we add more screens. But perhaps the old way was ugly and unsustainable.
I'm aware of that. I've tried leaving it in, but Bloodhound (the JS auto-completion library) refused to work when there were two addForms on the page (on /add, after POSTing a badly filled out add form). I'm sure there's a way to fix that, but supporting two add forms on a single page requires a lot of JS changes - this was simply the easier solution.

What I could add is a `Manage journal` modal window - keeping the `/manage` page for JS-disabled browsers, but adding an onClick handler and a keybinding (`m`?) for opening it.

zarybnicky:
@ocdtrekkie If you do, can you look if the protobuf settings for permissions are correct? I meant to look into testing Sandstorm before, but I never got around to it...

ocdtrekkie:
Yeah, I've got a big to-do backlog at the moment, but I'd absolutely test those out.

PR #823, opened 2018-06-26
Allow "re-parsing" with custom parse errors, some commodity cleanups

awjchen:
This PR allows the use of "re-parsing" with the custom parse errors, and applies this "re-parsing" to two parsers:
- the number parser, to delay the interpretation of numbers (in anticipation of changes potentially required issue #793), and
- the period transaction parser, to separate out the handling of the whitespace rules for period expressions (restoring the original behaviour in PR #807).

Additionally, this PR disallows a comma `,` from being interpreted as a commodity, adds tests for this case, and removes unused commodity code.

### DIFF
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -265,14 +265,14 @@ indentedlinep = lift (skipSome spacenonewline) >> (rstrip <$> lift restofline)
 -- >>> Right _ <- rjp commoditydirectivep "commodity $\n  format $1.00"
 -- >>> Right _ <- rjp commoditydirectivep "commodity $\n\n" -- a commodity with no format
 -- >>> Right _ <- rjp commoditydirectivep "commodity $1.00\n  format $1.00" -- both, what happens ?
-commoditydirectivep :: JournalParser m ()
+commoditydirectivep :: Monad m => JournalParser m ()
 commoditydirectivep = commoditydirectiveonelinep <|> commoditydirectivemultilinep
 
 -- | Parse a one-line commodity directive.
 --
 -- >>> Right _ <- rjp commoditydirectiveonelinep "commodity $1.00"
 -- >>> Right _ <- rjp commoditydirectiveonelinep "commodity $1.00 ; blah\n"
-commoditydirectiveonelinep :: JournalParser m ()
+commoditydirectiveonelinep :: Monad m => JournalParser m ()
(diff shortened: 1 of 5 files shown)


hledger-lib/Hledger/Read/JournalReader.hs
```diff
@@ -467,7 +466,12 @@ periodictransactionp = do
   -- a period expression
   pos <- getPosition
   d <- liftIO getCurrentDay
-  (periodtxt, (interval, span)) <- lift $ first T.strip <$> match (periodexprp d)
+  periodExcerpt <- lift $ excerpt_ singlespacedtextp
```
simonmichael: With this code, two spaces enforces the separation of period expression and description. But if you forget and write one space, and are unlucky with wording, it's still possible to get unintended overparsing. Eg `~ monthly In 2019 we will change this` is parsed as `monthly In 2019`. 

Which will be simplest and least surprising overall - allow one space, but document the above and recommend two spaces for robustness; or require two spaces ? I lean towards the latter. Any other options ?

simonmichael: Another option: instead of english-like period expressions, use some more distinctive and complete syntax like https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals . Dropping the old syntax is probably out of the question though.

awjchen:
I'm very sorry for the long delay. For the last bit I've had a lot more to do than usual.

I've made the suggested changes, that is, I've removed the `INLINABLE` pragma for `excerpt_` and I've corrected the wording in tests 21 and 23.

I think that the simplest and least surprising of the two alternatives is to require a double-space for separation of the period expression and description, since if we allowed a single space to separate the two I think we would probably introduce a lot of corner cases.

If a double-space is not a robust enough separator (e.g. in the case where the second space is forgotten), I think we technically have the option to specify a new separator between the period expression and description, without breaking compatibility, since descriptions were added only recently to periodic transactions. However, I'm not yet sure how to choose such a separator.

simonmichael:
Thanks! I think requiring  double space is the way to go. It is our idiom, for better or worse.

simonmichael:
Hi @awjchen. I've lost track of this. Should I just close it ? I think you said you plan to revisit it but I can't find where.

simonmichael:
(Ie, start fresh.)

awjchen:
Yes, I do plan on revisiting this. With all the changes to the codebase since this PR was submitted, I think it's best to close this PR and start fresh.

PR #839, opened 2018-07-15
Wrap $(cmd_version xxxx) with double quotes to solve installation failure

yang-ling:
This install script failed on my laptop. My OS is ubuntu 18.04.

I found the problem is `cmd_version xxxx` like `cmd_version hledger` will return empty string if
hledger is not installed, and then `cmpver $(cmd_version hledger) $HLEDGER_VERSION`
will become `cmpver $HLEDGER_VERSION`, and then `$(cmpver "$(cmd_version hledger)" $HLEDGER_VERSION) = 2` will be false.

Wrapping `$(cmd_version xxxx)` with double quotes can solve this, and it succeeds on my machine.

Trailing white spaces are trimmed automatically by my editor

### DIFF
--- a/hledger-install/hledger-install.sh
+++ b/hledger-install/hledger-install.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Easy hledger installation script for POSIX systems.
-# Uses cabal if installed and stack is not, 
+# Uses cabal if installed and stack is not,
 # or stack, which it will install if needed (or if --force-install-stack is used).
 # Requires bash and some other POSIX tools.
 # This is based on get-stack.sh which is copyright (c) 2015-2017, Stack contributors.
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks!

simonmichael:
@yang-ling do you think the other calls to cmd_version could have a similar bug ?

yang-ling:
@simonmichael I think the other calls are fine.

There are 2 other calls.

1. `[[ $(cmd_version "$1") == "$2" ]]`  in function `has_cmd_version`.  `has_cmd_version` has never been called.
2.  `echo "$1" $(cmd_version "$1") is installed at $(cmd_location "$1")` in function `print_cmd_version`. In this case, `$(cmd_version "$1")` will never return empty string, because it is wrapped by a if-else block. The `if-statement` is `[[ $(cmd_location "$1") ]]`. If the program is not installed, this `if-statement` will be false, and then `echo "$1 not found"` will be called.

simonmichael:
Thanks!

Issue #828, opened 2018-06-28
Why can't hledger query for comments?


peti:
I have a journal with 10 years worth of financial data that contains lots of useful information in posting comments, i.e.: 

~~~
2018/06/25 EDEKA     ; Kartenzahlung
    Vermögen:Girokonto         EUR -53.68 = EUR 8574.21    ; Debitkarte 123456789
    Ausgaben:Lebensmittel
~~~

Now, I wanted to make a report that show all transactions paid with my debit card, but not my wife's. That seemed easy enough, because I assumed that I could run something along the lines of

    hledger reg comment:123456789

but it turned out that hledger cannot query by posting comments. That was quite a surprise to me. Why is there no `comment:` and/or `pcomment:` selector that can be used for this purpose?

adept:
As a person with double-digits years of ledger/hledger journal, I do feel an occasional need to "grep" for something. On the other hand, I agree with @simonmichael that "from the position of hledger's current design, that was a wrong use of comments" and "if comment becomes just another data field, results will start to depend on its contents and there's no longer a way to add some guaranteed-harmless annotations".

Interestingly, the example provided even have the comment that looks like a tag, if only it was written "Debitkarte:123456789"

I do see a middle ground, though (warning: i did not think long and hard about this). An ability to regexp-match whole transaction body could be added to print (and potentially other commands as well) and will both give a grep-like prefilter ability (in "user beware" style - it would be anyone's guess what happens to balance assertions for example) and avoid giving comments a first-class place among other data fields.

simonmichael:
Yes, if I was writing that comment I would usually think "this looks like something I might like to query, make it a tag" and write `; Debitkarte:123456789`.

The print --grep (I don't know what else you'd call it) idea is interesting. Somewhat equivalently, I have long wanted a query syntax for searching all fields, which could be made to also match comments. It would be a kind of ambivalent half support for comment searching. It would be hard to match on the comment field only. The "non-affecting comment" principle would be weakened.

simonmichael:
Or (does not require blank lines between entries, cleans up the entries, loses the things print loses, eg inter-transaction comments):

    hledger print | awk -vRS='' -vORS='\n\n' /PATTERN/

simonmichael:
I'm still off on this tangent, making notes. How could print --grep look as a query ?

    `entry:REGEX` - matches transactions whose journal entry contains REGEX

requires saving journal text with each transaction

[...]

PR #842, opened 2018-07-18
hledger-lib: implement "comment:regexp" query

peti:
The `comment:` selector will match any transaction or posting comment. It therefore
represents a super set of the matches `desc:`, `note:`, `tag:`, and `payee:` will
perform.

Closes https://github.com/simonmichael/hledger/issues/828.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -29,6 +29,7 @@ module Hledger.Query (
   queryIsReal,
   queryIsStatus,
   queryIsEmpty,
+  queryIsComment,
   queryStartDate,
   queryEndDate,
   queryDateSpan,
@@ -96,6 +97,7 @@ data Query = Any              -- ^ always match
                               --   and sometimes like a query option (for controlling display)
            | Tag Regexp (Maybe Regexp)  -- ^ match if a tag's name, and optionally its value, is matched by these respective regexps
                                         -- matching the regexp if provided, exists
+           | Comment Regexp   -- ^ match any transaction or posting comment
     deriving (Eq,Data,Typeable)
 
 -- custom Show implementation to show strings more accurately, eg for debugging regexps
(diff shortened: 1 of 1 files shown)


simonmichael:
Notes: 
- the main discussion of this is on #828. Maybe we should try to keep it there, I don't know.
- thanks for this PR, a useful proof of concept
- this makes `comment:` search: transaction comments, transaction-preceding comments, posting comments, and descriptions
- I think these are the benefits it brings: 
  - being able to attach and search free-form notes on postings, slightly more ergonomically than using tags
  - being able to search comments, in case you have put data there that might be handy to search
  - being able to search descriptions and comments at once with a single query term

- I'm still against this in its present form, for reasons previously mentioned. I feel it negates the original purpose of comments, further increases the number of places to store data, and makes hledger and the process of recording data more complicated, for insufficient justification.
- I think there might be other ways we could achieve the above goals.

peti:
> I'm still against this in its present form.

What do I have to change to make you agree to this PR?

simonmichael:
Update: we had another short IRC chat yesterday. @peti got annoyed and abandoned ship. So far, two people are somewhat in favour of my position (@adept, f-a), noone has supported peti's, and the list doesn't care about this. I have been thinking about it. 

Some notes. First, here are the types of comment in hledger, for reference:

-   transaction-preceding comment
    Lines beginning with ; immediately before a transaction.
-   transaction comment
    Text following ; after the description field, or indented on following lines. Displayed by print. Can define searchable tags.
-   posting comment
    Text following ; after the amount/price/balance assertion fields, or indented on following lines. Displayed by print. Can define searchable tags.
-   directive comment (not implemented)
    Lines beginning with ; immediately before a directive, and text following ; after it, or indented on following lines.
-   file comment
    All other lines beginning with ; or # or \*, not attached to transactions or directives.
    Also lines between a \`comment\` directive and \`end comment\` or end of file.
    
Several ideas aimed at improving the situation for someone with peti's wish:

-   make tags more convenient
    -   a TAG:[PAT] query could be a shortcut for tag:TAG[=PAT]
    -   a desc: tag could set the transaction/posting description
    -   a desc:VALUE tag could be written more briefly as ; :VALUE

-   posting descriptions
    -   postings could have a new optional description field, separated from the amount/price by 2+ spaces, replacing or augmenting the transaction description

-   searchable comments
    -   all transaction/posting comments could be searchable with cmt: (peti's proposal, in essence)
    -   there could be two kinds of transaction/posting comments, light/soft vs deep/hard
        -   transaction/posting comments beginning with ; can define tags and are searchable with cmt:
        -   transaction/posting comments beginning with ;; (and all file comments) can't define tags and aren't searchable

Tangential thoughts on comment visibility:

-   print currently shows transaction and posting comments
-   it should also show transaction-preceding comments
-   should it also show file (inter-transaction) comments ?
-   should it have options to show/hide some/all kinds of comments ?
    
Summary: 

-   I can understand the urge to search comments, I sometimes feel it too.
-   It's important to avoid feature creep and keep complexity under control.
-   It's important to present clear and intuitive concepts, data models and best practices to users.
-   I find the ; vs ;; idea attractive.
-   But no matter how we slice it, searchable comments means adding another data field, another concept, and another decision for users to make when recording journal entries.
-   If someone wants to push this forward, please do the work of writing the docs and let's see how it reads. My best guess is we will judge it more complicated and not quite worth it, but I keep an open mind.

Issue #303, opened 2015-11-28
Support wildcards in includes


rodlogic:
It would be nice to be able to specify an open ended list of journal files instead of an explicit file. For instance: `include capone/*.journal`
##

simonmichael:
Resolved by #843/#846.

PR #846, opened 2018-07-23
allow glob patterns in 'include' directive

### LINKED ISSUE #843: Allow globs in 'include' statements
I see from the [documentation](http://hledger.org/journal.html#including-other-files) that globs cannot be used in 'include' statements. I have also confirmed that this does not work with v1.10.

Would this be difficult to add to hledger? I would be willing to have a go at implementing this, if it's an uncontroversial feature.

jbweston:
Closes #843 and #303.

I am opening this as a "work in progress" to make sure I am on the right path, and get some feedback (especially on style)

#### Questions

+ ~~All the functional tests pass, however it seems that the doctests are failing. The root cause is that the `Glob` and `filemanip` packages both export modules called `System.FilePath.Glob`, and doctest does not know how to deal with this. I am a Haskell noob, so I am unsure how to proceed.~~

+ ~~Do we care that this change may potentially break existing journal files? Notably journal files that have filenames that are also valid glob patterns. IMO this does not seem relevant, but I wanted to bring it up.~~

+ Currently we are setting/resetting the parser state inside `parseChild`. Should this be factored out to the call site instead?

+ We are updating the journal after reading each child. Should we instead create all the child journals in a list, and then make a single call like `put $ (fold childrenj) <> parentj`? I was unsure how much I should assume about what `newJournalWithParseStateFrom` does...

+ ~~When parsing should fail I have used the general `fail` from `Control.Monad.Except`, however I see that elsewhere in `JournalReader.hs` there are explicit calls to `parseErrorAt`. I decided to use `fail` because I just want the current position reported anyway, so I can avoid threading the parser position around, or calling `getPosition`.~~

### DIFF
--- a/tests/journal/include.test
+++ b/tests/journal/include.test
@@ -37,7 +37,7 @@ include b.timedot
 >>>=0
 
 # 3. include glob patterns
-printf '2018/01/01\n (A)  1\n' >xy.journal; printf '2018/02/01\n  (B)  1' >yy.journal; hledger -f - print; rm -rf x.journal y.journal
+printf '2018/01/01\n (A)  1\n' >xy.journal; printf '2018/02/01\n  (B)  1' >yy.journal; hledger -f - print; rm -f xy.journal yy.journal
 <<<
 include *y.journal
 >>>
--- a/tests/journal/include.test
+++ b/tests/journal/include.test
@@ -35,3 +35,16 @@ include b.timedot
     (b.bb)            1.00
 
 >>>=0
+
+# 3. include glob patterns
+printf '2018/01/01\n (A)  1\n' >xy.journal; printf '2018/02/01\n  (B)  1' >yy.journal; hledger -f - print; rm -rf x.journal y.journal
+<<<
+include *y.journal
+>>>
+2018/01/01
+    (A)               1
+
+2018/02/01
+    (B)               1
+
+>>>=0
--- a/tests/journal/include.test
+++ b/tests/journal/include.test
@@ -48,3 +48,15 @@ include *y.journal
     (B)               1
 
 >>>=0
+
+# 4. include invalid glob patterns
+hledger -f - print
+<<<
+include [.journal
+>>>=1
+
+# 5. include nonexsitant file
+hledger -f - print
+<<<
+include doesnotexist.journal
+>>>=1
--- a/tests/journal/include.test
+++ b/tests/journal/include.test
@@ -37,14 +37,14 @@ include b.timedot
 >>>=0
 
 # 3. include glob patterns
-printf '2018/01/01\n (A)  1\n' >xy.journal; printf '2018/02/01\n  (B)  1' >yy.journal; hledger -f - print; rm -f xy.journal yy.journal
+printf '2018/01/01\n (A)  1\n' >ab.journal; printf '2018/01/01\n  (B)  1' >bb.journal; hledger -f - print; rm -f ab.journal bb.journal
 <<<
-include *y.journal
+include *b.journal
 >>>
 2018/01/01
     (A)               1
 
-2018/02/01
+2018/01/01
     (B)               1
 
 >>>=0
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -94,6 +94,7 @@ import Text.Megaparsec.Char
 import Text.Megaparsec.Custom
 import Text.Printf
 import System.FilePath
+import System.FilePath.Glob hiding (match)
 
 import Hledger.Data
 import Hledger.Read.Common
@@ -189,15 +190,18 @@ includedirectivep :: MonadIO m => JournalParser m ()
 includedirectivep = do
   string "include"
   lift (skipSome spacenonewline)
-  filename <- T.unpack <$> takeWhileP Nothing (/= '\n') -- don't consume newline yet
+  fileglob <- T.unpack <$> takeWhileP Nothing (/= '\n') -- don't consume newline yet
 
   parentpos <- getPosition
 
   curdir <- lift $ expandPath (takeDirectory $ sourceName parentpos) ""
-                   `orRethrowIOError` (show parentpos ++ " locating " ++ filename)
-   -- </> correctly handles case when 'filename' is absolute
-  let filepaths = [curdir </> filename]
-  -- read child inputs
+                   `orRethrowIOError` (show parentpos ++ " locating " ++ fileglob)
+
+  filepaths <- if isLiteral (compile fileglob)
+                  -- </> and globDir1 correctly handle case when 'fileglob' is absolute
+                  then pure [curdir </> fileglob]
+                  else liftIO $ globDir1 (compile fileglob) curdir
+
   forM_ filepaths $ parseChild parentpos
 
   void newline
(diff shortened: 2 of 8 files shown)


jbweston:
I will also need to modify the documentation to indicate that globs are now supported

tests/journal/include.test
```diff
@@ -35,3 +35,16 @@ include b.timedot
     (b.bb)            1.00
 
 >>>=0
+
+# 3. include glob patterns
+printf '2018/01/01\n (A)  1\n' >xy.journal; printf '2018/02/01\n  (B)  1' >yy.journal; hledger -f - print; rm -rf x.journal y.journal
```
simonmichael: Shouldn't it be rm -rf xy.journal yy.journal ?

simonmichael: Also, the -r is overkill perhaps.

simonmichael:
I'm having trouble reproducing the doctest issue. My doctest --version says 0.13.0. But I think the fix may be to specify the Glob package like in tests/doctests.hs.

jbweston:
> I think you happen to have filemanip installed by chance, and doctest does not follow the normal package visibility rules. I can look for a solution.

I admit my ignorance as to how `stack` works. I thought that 

> I'm having trouble reproducing the doctest issue. My doctest --version says 0.13.0. But I think the fix may be to specify the Glob package like in tests/doctests.hs.

I see this when I run `make test` and also `make pkgtest`, however *not* when I run `make doctest`! However the error mesage includes these lines:
```
Test suite failure for package hledger-lib-1.10.99
    doctests:  exited with: ExitFailure 1
```
which made me conclude that the problem was indeed with doctests.

In any case your suggestion (implemented in 437d99bbfe3d7594a5b4e530fccfc81adb4d9b16) fixes this.

tests/journal/include.test
jbweston: fixed in e133e79

jbweston:
> >  Do we care that this change may potentially break existing journal files? Notably journal files that have filenames that are also valid glob patterns. IMO this does not seem relevant, but I wanted to bring it up.

> Generally yes we would care, but what are some examples you have in mind ?

Nothing more than what I mentioned: that filenames containing glob patterns will now be interpreted as glob patterns. Here is the [list of patterns](https://hackage.haskell.org/package/Glob-0.9.2/docs/System-FilePath-Glob.html) that `Glob` supports. I can totally imagine a filename containing brackets `[..]`, which would be interpreted as a character range. `Glob` supports restricting the class of patterns that it parses, so we could restrict to just  `*` and `?`, for example. This would already solve my personal use-case, as well as #303.

jbweston:
> + {} syntax (alternatives) didn't work. Your docs will say what's supported I expect.

 Indeed alternatives are not supported by `Glob`.

jbweston:
> + File names in the include directive seem to be case sensitive even on a case-insensitive file system (tested on a mac). Ideally includes should mimic the case behaviour of the underlying file system, I guess ? An old bug.

When you say that this is an old bug,
do you mean that at the moment this is the behavior of includes on Mac?
I searched, but can't see an open issue for this.
I guess this is a limitation of the `Glob` library. I looked, but the only
other library that I could see for globbing is (`filemanip`](https://hackage.haskell.org/package/filemanip),
but it is not clear to me what happens if it detects an invalid glob pattern
(AFAICT it just ignores them?), and the maintainer appears unresponsive (no comments
on open issues, and the last commit was nearly 4 years ago).

The only sane solutions appear to be:

+ fix the Glob library
+ roll our own globbing (we can make our lives easier by restricting to only `*` and `?` patterns)

jbweston:
> + Files from globs seem to be included in reverse order (compared to how ls lists them). You can see this if each contains a transaction on the same date.

This is weird; I will investigate, however this is probably just the way `Glob` gives the filenames back

jbweston:
> + Cycles, eg a file including itself, are not detected and hledger hangs. This is an old bug, perhaps more likely to be triggered now. If you feel like looking at that next, please do..

Sure. I guess this should be in a separate pull request?

jbweston:
Does Github provide a way to thread conversations? There are now like 4 orthogonal discussions to be had, and it's kind of cumbersome to quote the relevant context in each new comment!

simonmichael:
> I admit my ignorance as to how stack works. I thought that it would only bring in packages that are explicitly mentioned in a package.yaml, and that each project is isolated from other ones.

That's right, and I don't think doctest uses stack, or it bypasses that mechanism somehow.

> I can totally imagine a filename containing brackets [..], which would be interpreted as a character range. Glob supports restricting the class of patterns that it parses, so we could restrict to just * and ?, for example.

Got it. I think such filenames are very rare, among hledger users anyhow, and we can be progressive this time. I'd just support the full syntax.

> do you mean that at the moment this is the behavior of includes on Mac?
I searched, but can't see an open issue for this.

Yeah, that's what I discovered on mac just today. There's no github issue for it. If it's a Glob limitation, I would open a bug and hope for a fix. It's not a big deal and I agree that Glob seems the best option overall.

> This is weird; I will investigate, however this is probably just the way Glob gives the filenames back

If that were so I'd report that as a bug too.

> Sure. I guess this should be in a separate pull request?

Yes, sounds good.

> Does Github provide a way to thread conversations? 

Nope! At least the conversation's on just one page..

simonmichael:
PS https://hackage.haskell.org/package/Glob/docs/System-FilePath-Glob.html#v:compile <- good link to include in docs

jbweston:
> > This is weird; I will investigate, however this is probably just the way Glob gives the filenames back

> If that were so I'd report that as a bug too.

3df3e5b works around this by calling `sort` on the result that glob gives us back

jbweston:
>>    do you mean that at the moment this is the behavior of includes on Mac?
    I searched, but can't see an open issue for this.

> Yeah, that's what I discovered on mac just today. There's no github issue for it. If it's a Glob limitation, I would open a bug and hope for a fix. It's not a big deal and I agree that Glob seems the best option overall.

I am hesitant to open a bug upstream because I am not 100% sure what the observed behaviour is, and what the expected behaviour is.

I borrowed a mac briefly and ran the following in a Bash terminal:
```shell
$ touch a.x b.x
$ ls ?.x
a.x      b.x
$ ls ?.X
ls: ?.X: No such file or directory
```
which is case sensitive. (However `rm a.X` removes `a.x`, so it's case *in*sensitive)

jbweston:
The last 2 commits add a short line to the documentation of the `include` directive.

33f3198 adds the output of `./Shake manuals`, which I presumed I should include, as these files seem to be under version control?

I also couldn't see exactly where to put this changelog entry; `hledger-lib/CHANGES` didn't seem quite right, as that is apparently for API changes to the library, and `hledger/CHANGES` only seems to contain changes from previous releases, not a running-update of the changes being implemented in the `dev` version...

simonmichael:
> I am hesitant to open a bug upstream because I am not 100% sure what the observed behaviour is, and what the expected behaviour is.

You're right, I see the same thing with ls. Let's not worry about it.

> 33f3198 adds the output of ./Shake manuals, which I presumed I should include, as these files seem to be under version control?

Thanks, I don't always update those generated manuals every time but bonus points for doing that.

> I also couldn't see exactly where to put this changelog entry

Extra bonus points for thinking about this! Yes hledger/CHANGES is where I would mention this. I usually generate these from the commit log all at once just before release (like the example command in the comments for `make changes-show`). That way I think less and keep it somewhat consistent. I'm open to better ways of doing it, just add a pending section to the changelog if you like.

Let me know when you're ready by removing the WIP from the title, and I'll merge this

simonmichael:
Thank you very much!

I suppose we could now add support for `-f 'GLOB'` (though not `-f GLOB`) if we wanted.

PR #854, opened 2018-07-31
more auto posting/transaction modifier cleanups

simonmichael:
This continues some cleanups already in master. I had to change a test, but I think it's doing the right thing ?


simonmichael:
Assuming ok.

PR #857, opened 2018-08-02
raise error on recursive includes

### LINKED ISSUE #853: Hledger does not detect cycles when resolving 'include' directives
Brought up by @simonmichael in #846,  [here](https://github.com/simonmichael/hledger/pull/846#pullrequestreview-139763653).

### Steps to Reproduce
```
$ echo 'include b.journal' > a.journal
$ echo 'include a.journal' > b.journal
$ hledger -f a.journal print
```

### Observed Behaviour
Hledger runs until it runs out of memory. This is also a bug in ledger, except that it fails due to too many open file handles long before it exhausts the available memory.

### Expected Behaviour
There are only 2 possibilities AFAICT: an error should be reported, or the `include` directive that would result in a cycle being formed should be ignored (because we're already parsing the file in question). I think it is reasonable to expect people to "break" cycles manually, and so reporting an error is the correct thing to do.

jbweston:
Closes #853.

### DIFF
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -216,6 +216,10 @@ includedirectivep = do
             else parseErrorAt parserpos $ "No existing files match pattern: " ++ filename
 
     parseChild parentpos filepath = do
+        parentfilestack <- fmap sourceName . statePos <$> getParserState
+        when (filepath `elem` parentfilestack)
+            $ parseErrorAt parentpos ("Cyclic include: " ++ filepath)
+
         childInput <- lift $ readFilePortably filepath
                              `orRethrowIOError` (show parentpos ++ " reading " ++ filepath)
 
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -213,7 +213,7 @@ includedirectivep = do
         filepaths <- liftIO $ sort <$> globDir1 fileglob curdir
         if (not . null) filepaths
             then pure filepaths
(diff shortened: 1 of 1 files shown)


jbweston:
It turns out MegaParsec keeps its own stack of source positions (which we were pushing onto from `includedirectivep` anyway), and we can easily get a stack of include filenames from this.

simonmichael:
Well spotted!

I see one of the tests failed, perhaps because of the use of "timeout" ?

jbweston:
> I see one of the tests failed, perhaps because of the use of "timeout" ?

I also saw the tests failing sporadically on my machine, but when I manually ran the testcases outside of `shelltestrunner` everything seemed fine. I think that this is related to the fact that `shelltestrunner` runs tests in parallel, but AFAICT the tests are conducted in the same directory. I have to be more careful that filenames that I write/delete are unique between tests.

---

On a related note, I saw that `timeout` was part of the GNU coreutils, so I naively assumed that this (or something similar) would exist on MacOS.  It seems I was wrong. [stack overflow](https://stackoverflow.com/questions/3504945/timeout-command-on-mac-os-x) provides a couple of solutions. In general what tools can we assume developers will have on their machines?

jbweston:
The checks are passing now, as I modified the tests in the way that I mentioned above

simonmichael:
I think maybe you haven't pushed the latest changes here yet.

Yes timeout is not available by default on OSX or windows. I try to use as few external tools as possible. 

And yes if we must generate files, it's important to generate unique filenames in case of parallel tests, you'll see some other tests using $$ for this. 

+1 for the test, but if it's hard to make it robust and cross platform I'm ok with leaving this one out, regressions are probably a low risk here.

jbweston:
@simonmichael I removed the tests

Issue #861, opened 2018-08-20
Allow arithmetic expressions in balance assertions


sequba:
It'd be nice if one could write:

``
2013/1/2
  a    $1  = $2 + $2
  b
``

amitaibu:
Here's a real world use case for why and how we would use this feature: We have credit line in our bank of for example $100 - meaning we are able to go $100 below 0. 

So, if our balance shown in the bank is 0, than we look at it as if we in fact had $100. The reason we do this, is that if for example our balance was `$-60`, then we would treat it as `$40` - and it won't be in the red.

The way I would have wanted to use it, would be in a CSV rules file, have something like:

```
fields date, description, bal
balance (%bal +100)
```

alerque:
@amitaibu I'm not sure your use is a good case study for this feature. I for one wouldn't lift a finger to "fix" that case because I strongly believe it isn't broken.

> So, if our balance shown in the bank is 0, than we look at it as if we in fact had $100.

But you wouldn't actually have $100, so why should the app be expected to report it as such? What you actually have below $0 is a liability debt to the bank in the amount that they have extended you credit. If you really insisted on misrepresenting your financial status you would just add an entry showing a $100 deposit into the account. Of course with double entry accounting you would balance that against a liability account, say for the credit your bank extends. Summing that with your actual account would show the correct *real* balance.

I'm not saying math expressions shouldn't be allowed –I would find them useful– but perhaps you should address that case study another way with existing tools such that the values are what they say on the tin. :wink:

amitaibu:
Yeah, I guess my above (simplified) example isn't passing correctly what I'm doing which is a mix of keep tracking of past events and some forecasting. I'll hide that comment, as it probably just adds noise - sorry about that, and thanks @alerque for pointing it out.

Issue #195, opened 2014-06-22
Hledger does not handle balance assertions when multiple commodities


thdox:
Please see below:
- the test file
- works correctly with ledger (c++)
- fails with hledger

$ cat hledger_test_16

    2000/01/01 * Initial
        Actif:Courant:BnpCc                 -102411,12 F = -102411,12 F
        Actif:Courant:BnpCc                   21036,81 € = 21036,81 €
        Equity
     
    2002/01/11 * Passage à l'euro
        Actif:Courant:BnpCc                  102411,12 F = 0,00 F
        Actif:Courant:BnpCc                  -15612,47 € = 5424,34 €
        Revenu:Devise                       -102411,12 F
        Revenu:Devise                         15612,47 €

$ ledger --no-pager -f hledger_test_16 bal

               5424,34 €  Actif:Courant:BnpCc
             102411,12 F
             -21036,81 €  Equity
            -102411,12 F
              15612,47 €  Revenu:Devise
    --------------------
                       0

$ hledger -f hledger_test_16 bal

    hledger: Balance assertion failed for account Actif:Courant:BnpCc on 2000-01-01
    In transaction:
    2000/01/01 * Initial
        Actif:Courant:BnpCc  -102411,12 F
        Actif:Courant:BnpCc    21036,81 €
                              102411,12 F
        Equity                -21036,81 €
      
    After posting:
       Actif:Courant:BnpCc      21036,81 €  ;
     
    expected balance is 21036,81 €, actual balance was -102411,12 F
      21036,81 €.

thdox:
Hi Simon, I've tested this against hledger 0.24.

It works correctly with test cases hledger_test_16a and hledger_test_16c (with an 'a' and a 'c' at the end of filename).
It does not work for hledger_test_16 (without letter at the end of filename).

I suspect this does not work because of your last comment "The ability to assert the full multi-commodity balance can be added again using the == syntax, if needed."
I do not fully catch this last comment. I went to http://hledger.org/manual.html, and a search can not find any occurrence of '=='.

I've tried syntax below, but this fails.

    2002/01/11 * Passage à l'euro
        Actif:Courant:BnpCc                  102411,12 F = 0,00 F
        Actif:Courant:BnpCc                  -15612,47 € == 5424,34 €
        Revenu:Devise                       -102411,12 F
        Revenu:Devise                         15612,47 €

Can you help me here by providing a syntax example that would fit with hledger_test_16 test case ?

Thanks

[...]

Issue #183, opened 2014-05-04
Support simple expressions in postings


alexanderkjeldaas:
When working with differenciated VAT (in EU), an invoice might say:

Food A      (12% vat)   EUR   50
Food B      (12% vat)   EUR   10
...
...
Plastic bag  (25% vat) EUR   0.1
Transportation (25% vat) EUR 4

Total:   EUR 1423.00
including VAT 25%       EUR 0.82
including VAT 12%       EUR 240

In this scenario, it is useful to be able to write any of the following:

```
expenses:food:12%   1423.10 - 0.1 - 4
expenses:other:25%     0.1 + 4
cash
```

That is, if a few entries have a different vat setting, being able to use some simple calculations is good.  The idea is that the posting should only include numbers that are easy to find on the receipt and it should be succinct.

It could have been written:

```
expenses:food:12%   1423.10
expenses:food:12%        -0.1
expenses:food:12%        -4
expenses:other:25%     0.1
expenses:other:25%     4
cash
```

but that is pretty hackish and unclear I think.

In this example I have accounts that have fixed VAT amounts, so I can transfer the VAT at the end of the reporting period:

```
expenses:food:12%     -41243  = 0
expenses:food             41243/1.12
vat:incoming:12%
```

Another way to do this is like this:

```
expenses:food:12%   (1423.10 - 0.1 - 4)/1.12
vat:incoming:12%        240
expenses:other:25%     (0.1 + 4)/1.25
vat:incoming:25%       0.82
cash                           -1423
```

In that example, only numbers from the receipt are used, and the posting must balance.  Without the ability to calculate non-vat food and other expenses, a calculator must be used, which makes writing the posting 3x as much work.  A posting where the numbers are calculated is also less clear.  It is hard to know whether the numbers correspond to the receipt.

The following simple 2-minute hack in perl implements this partially as a preprocessor, but more principled support would be great.

```
#!/usr/bin/perl -p

if (m/(\s+[a-zäöåA-ZÖÄÅ0-9:%]+\s\s+)(-?[0-9.()]+(\s*[-+*\/]\s*[0-9.()]+)+)/) {
    $_=$1.eval($2)."\n";
}
```

simonmichael:
Needs discussion/developers, closing.

simonmichael:
On #913 @sorsasampo wrote:
[...]

PR #871, opened 2018-09-21
Exact assertions & multi commodity balances

ag-eitilt:
I'm not entirely sure how "lay the groundwork for #556" became "implement #183 and #861", especially since doing so was only a side effect of putting multiple commodities on a single line.  That, at least, can be traced to what @simonmichael mentioned about halfway down #195, but that still begs the question of why I implemented exclusive balance assertions before what I actually set out to fix.  Either way, it seems to work, unless I missed testing some strange interaction somewhere.  And with minimal journalfile breakage, as well: commodities including parentheses need to be quoted now, but that might actually bring things more in line with ledger, depending on what its arithmetic expressions disallow.


ag-eitilt:
Ah, didn't notice that test suite.  Not quite sure what happened to the rewrites, but I'll track that down next.

simonmichael:
Thanks for the PR! Tip: use `make functest` to run those functional tests locally. You'll need to install shelltestrunner.

This looks like a WIP branch, which might get cleaned up before merge, so I'll review with that in mind. For reviewers, can you summarise the effect of this PR, eg: "it allows posting amounts and balance assertion amounts to be written as arithmetic expressions" ?

simonmichael:
I think I'm right in saying the impact of these commits is:

1. amounts (in postings, balance assertions, balance assignments) can be written as addition/subtraction expressions of multiple amounts: `$10 - $0.25 + EUR 5`.
This means that a posting can describe a multi-commodity change (a MixedAmount, not just an Amount). And balance assertions/balance assignments can check/set more than one commodity at once.

2. parentheses can be used to enclose these amount expressions (making them Ledger-parseable) and to enclose subexpressions: `($10 - ($0.25 + EUR 5))`.

3. balance assertions can be written with `==` to assert that no commodities other than the ones written are present: `== $10.25 + EUR 5` means the balance contains no other commodities. Assertions still exclude subaccounts.

ag-eitilt:
@simonmichael Exactly, and having those in bullet points makes it so much clearer.  Thanks!  The fix for auto transactions also added:

* Values in amount expressions may be multiplied by raw quantities: `$10 * 0.33`.
Multipliers at the left side of an amount expression or marked with a commodity symbol still need to be handled.

I'm beginning to see what you meant about all the complex interactions with assertions...  Luckily I *think* the largest difficulties are worked out, but we'll see how it goes from here.  And unfortunately, `make functest` relies on Stack, so I've been running the tests manually.  Cabal works and I've learned some nice things about `new-build`, but the program's definitely not optimized for it.

hpdeifel:
A quick test suggests that I could use this to parse amount expressions in `hledger-iadd`, I'd love to have that in `hledger-lib` instead of my custom hack. Especially with multiplication, which is quite tricky (see  https://github.com/hpdeifel/hledger-iadd/issues/34#issuecomment-425258425).

Unit handling is a bit broken though, I'd expect (2€*3€) to be 6€² :slightly_smiling_face:

ag-eitilt:
All right, I think this is everything for this PR.  The future work on pending postings and subaccounts with assertions will be based on this branch, but unless you prefer fewer, more general PRs, this is both able to stand on its own and is not intrinsically tied to either, so this seems a good point to freeze the patch for review.

To consolidate the changes:

* Balance assertions may check that accounts contain *only* the amounts specified rather than ignoring unlisted commodities: `== $10`
  - This introduces breaking changes to the API around the `BalanceAssertion` type; it is no longer a type alias with an included `Maybe` but a plain dictionary, and users are expected to add the `Maybe` themselves if required.  Additionally, the function `partialbalanceassertionp` has been renamed to `balanceassertionp` as it may now parse exact balances as well.  No incompatibilites are introduced with existing journal files.

* Posting amounts and balance assignments/assertions may now be specified by expressions in addition to individual values: `$10 + $5`
  - This allows multiple commodities to be listed in a single posting, by writing it as the sum of each: `$10 + 5 CAD`
  - Addition, subtraction, and multiplication are all supported, as is surrounding either parts of the expression for control over operation distribution or the entire expression for compatibility with original ledger: `$10 - ($3 + $2) * 2`
    + Any commodities using parentheses now need to be quoted, which unfortunately has the chance of breaking some journals.
    + Addition and subtraction don't affect the commodities of either value, but multiplication retains only that of the right-hand side: `$2 * 3 CAD` becomes `6 CAD`; see below.
    + Multiplying by a value comprised of multiple commodities applies each separately: `2 * ($3 + 4 CAD)` results in `$6 + 8 CAD`

On multiplication, there are two details of my implementation that will probably be A., contentious and B., needing clear documentation.  First, as @hpdeifel said, multiplying by values tagged with their own commodities doesn't attempt any unit multiplication; instead, the commodity of the multiplier simply replaces that of the multiplicand.  Part of that is logical (what does "Euros squared" actually mean?), part of that is logistical (especially without ledger's commodity equivalences, there's no way of saying that `m` * `a` = `F`), part of that is historical (that's the way auto-postings handled it before the PR), and part of that is ideological (we're working with financial balances, not scientific calculations).  If more of y'all would also prefer multiplying commodities as well as quantities, then we can work on it, but otherwise I do think this is the best way of handling it.

Second, multipliers aren't treated as first-class amounts.  Ones *with* an actual amount to the left are multiplied as expected, and transaction modifiers headed by a multiplier work as expected (`*.5 + $10` can be applied to `$10` and `10 CAD` to create `$15` and `5 CAD + $10` respectively), but doing the same thing wouldn't make sense in true postings.  They actually threw the tests off a couple times before I fixed the order of application (`*.5 + $10` almost never makes sense as half a unit of the default commodity plus ten dollars).  Instead, those are silently dropped from the balance; that example would wind up as simply `$10`.  It might be better to raise an error, but given the parse tree, I'm not sure it's realistically possible to distinguish a posting in a transaction from one in a modifier at that stage, and I'm not sure which stage would be the best to run the check.  If y'all do want the error, I don't have anything against diving back in to add it, though.  For clarity, that does mean that the only multiplier(s) in an auto-posting that are applied to the base value are those to the left of the first addition or subtraction.

simonmichael:
@ag-eitilt congrats on all the progress. You've done a lot!

I would have preferred to discuss this more cheaply, with mockups and docs, as I tried to suggest previously. I know sometimes it's easier to think in code.  But for something substantial like this, it's pretty risky. More work is required to get this merge-ready, hopefully you understand.

Here are some next steps I think should happen:
- history cleanup. Re-record as clear commits against latest master, with less of the dev noise, for easier review.
- simplification. Can these two enhancements (arithmetic amount expressions and all-commodity balance assertions) be separated so we can deal with two smaller PRs one at a time ? I think so.
- docs. A draft of the new user documentation covering these changes (anywhere; commit, gist, comment..) will save a lot of time. It will speed up review. Or it may reveal aspects that aren't right for users yet.

Excellent progress.

simonmichael:
PS I would attempt the docs first. It is really the quickest way to get something into hledger.

simonmichael:
Hi @ag-eitilt.. I expect you are busy and will get back to this when you next have some time. But just in case you are burning the midnight oil writing an epic documentation tome, which would be overkill, I'm checking in to see the above made sense, and to offer help if needed, eg with drafting docs.

ag-eitilt:
Yeah, just been busy!  Thursday and friday are really the only time I have to focus on coding projects, and I'm not sure how much I'll get done this week either.  We'll see.  The only question I'd really have just now would be if there's any project style it would be best for me to follow when I begin cleaning/splitting the history; I noticed the `lib`/`doc`/... tags halfway through and will apply them, but is there anything else to be aware of or is it all common sense after that?

simonmichael:
Prefixing commits with [component:](https://github.com/simonmichael/hledger/wiki/Issues#components) is nice (it helps me make changelogs), but it's not essential. It can easily be tweaked last minute, also.

Fewer, perhaps larger commits with less backtracking will be good. Sometimes squashing all into one big commit can be acceptable; not sure how well that would work here (after splitting the features).

Nothing else comes to mind right now.

site/doc/1.11/journal.md
```diff
@@ -356,6 +356,38 @@ when an amountless posting is balanced using a price's commodity, or
...
+
+-   addition, subtraction, and multiplication using the standard ASCII
+    operators (`+`, `-`, `*`), either adjacent to the amount(s) or
+    separated by spaces
+-   parenthesised (sub-)expressions, to be evaluated before any
+    operations outside them. Note, however, that all commodity symbols
+    must be attached to an amount inside the parentheses: `($10 * 0.2)`
+    works, `$(10 * 0.2)` does not.
+-   multiple commodities in the same expression
+
```
simonmichael: Good so far

site/doc/1.11/journal.md
```diff
@@ -356,6 +356,38 @@ when an amountless posting is balanced using a price's commodity, or
...
+Multiplication of commodities is handled a bit differently than may be
+expected; if the amount to the right of the operator includes an
+explicit commodity, that commodity replaces any to the left for
+compatibility with [automated postings](#automated-postings). If that
+right value includes multiple commodities itself (parenthesised to obey
```
simonmichael: I don't think we should discard commodities. Can we instead disallow multiplying by amounts, ie only multiplying by a number will be legal.

site/doc/1.11/journal.md
```diff
@@ -356,6 +356,38 @@ when an amountless posting is balanced using a price's commodity, or
...
+
+As examples, with the value each expression simplifies to (if
+applicable):
+
+`$47.18-$7.13` -> `$40.05`
+`$47.18 - ($20 + $7.13)` -> `$20.05`
+`$6.10 + £4.35 * 0.8 €` -> `$6.10 + 3.48 €`
+`($6.10 + £4.35) * 0.8 €` -> `8.36 €`
+`$6.10 * (£0.2 + 0.8 €)` -> `£1.22 + 4.88 €`
+
```
simonmichael: I think folks will immediately want division (`/`) as well, can you add that ? We'll need to handle division by zero in some reasonable manner.

site/doc/1.11/journal.md
```diff
@@ -460,6 +492,28 @@ that no matter how many assertions you add, you can't be sure the
...
 
+
+The asserted balance checks only the amount(s) of each listed commodity's
+balance within the (possibly larger) account balance. We could call this a
+partial balance assertion. This is compatible with Ledger, and makes it
+possible to make assertions about accounts containing multiple commodities
+without needing to manually track every commodity the account contains.
+
+To instead assert a balance to the exclusion of all other commodities, use the
+exact assertion form `==EXPECTEDBALANCE`. This ensures that the account does
```
simonmichael: There is some repetition here. You could delete lines 490 to 502 or something ?

site/doc/1.11/journal.md
```diff
@@ -1242,9 +1296,14 @@ something.):
...
-be ordinary fixed amounts. Fixed amounts with no commodity symbol will
-be given the same commodity as the matched transaction's first posting.
+the matching posting(s), multiplied by N". They can also be ordinary
+fixed amounts (which are inserted unchanged) or [amount
+expressions](#amount-expressions) headed by either (only the
+multiplication to the left of the first addition or subtraction will be
+applied to the posting value).
+
+Any postings in real transactions which attempt to use this
+multiplication-headed form will ignore the value instead.
```
simonmichael: I didn't understand "headed by" onward exactly. Shall we can simplify again, and delete those lines ? Ie we will accept an amount, an amount expression, or a numeric multiplier (`*N`). (Adding support for `\N`, `+N`, `-N` might be worthwhile later ?)

site/doc/1.11/journal.md
ag-eitilt: That's because the original version snuck back in through one of the merges...  It's fixed in the split branch.

ag-eitilt:
So, this week is going to be mostly focused on smaller refactoring as well, starting with splitting out the assertation portion.  I'll leave this open for now to not lose the conversation, but once the amount expressions get cleaned up, I'll close this PR; that seems the best way to cut through the tangle.

site/doc/1.11/journal.md
ag-eitilt: Blame a couple years of linguistics.  That's the passage that was primary in the "first draft" naming of the commit.  I'll revisit this.

site/doc/1.11/journal.md
ag-eitilt: Adding it won't be hard at all, except for divide-by-zero; the operation parser wound up being amazingly extensible.  I'm tempted to just make it silently fail, but it's probably better to work through the error reporting system to ensure sanity.  I've not looked at that yet, though, so I might put division off until the end of the refactor, if you think that's all right.

site/doc/1.11/journal.md
ag-eitilt: On the "those can wait", I might actually put the first step of the PR together without multiplication entirely.  Addition and subtraction are a much more contained change, and seem to generate much less discussion (while being more frequently requested).  We should definitely keep working out the questions around multiplication, but splitting things further like that means there's less pressure to resolve it so we can find something we're all happier with.

site/doc/1.11/journal.md
mildred: division can be a problem too... What would be the amount for `$0.05 / 3` ?

cfellinger:
On Fri, Oct 12, 2018 at 12:12:04PM +0000, mildred wrote:
> mildred commented on this pull request.
...
> +As examples, with the value each expression simplifies to (if
> +applicable):
> +
> +`$47.18-$7.13` -> `$40.05`
> +`$47.18 - ($20 + $7.13)` -> `$20.05`
> +`$6.10 + £4.35 * 0.8 €` -> `$6.10 + 3.48 €`
> +`($6.10 + £4.35) * 0.8 €` -> `8.36 €`
> +`$6.10 * (£0.2 + 0.8 €)` -> `£1.22 + 4.88 €`
> +
> 
> division can be a problem too... What would be the amount for `$0.05 / 3` ?

this should fail, unless you explicitly specified the rounding to be
applied!  e.g. at home in our meta-ledger input we use ++ and -- at
the end of such expressions to indicate ROUND_UP or ROUND_DOWN.

  '$0.05 / 3'     >> runtime error
  '$0.05 / 3 --'  >> $0.01
  '$0.05 / 3 ++'  >> $0.02
  '€1 / 3'        >> runtime error
  '€1 / 3 --'     >> €0.33
  '€1 / 3 ++'     >> €0.34

ofcourse ++ and -- are only allowed once at the end of an expression
There is no rounding towards zero or fancy things like fair rounding,
only plain UP and DOWN.  At home we always have two decimal digits but
maybe the number of decimal digits should be commodity specific.

fair rounding are schemes to compensate for rounding ups preference to
round up, i.e halfway (.5) always rounds up.  Bankers might like them,
but they complicate things, and explicit seems better then implicit.

-- 
groetjes, carel

site/doc/1.11/journal.md
simonmichael: > I looked into the transaction modifiers and found that this is exactly how they currently operate. It's undocumented behaviour

You're right! I'd rather enforce that the multiplier is a commodityless number for now. Quietly discarding information from one side of an expression is something we should avoid. I'll clarify this in docs. I agree that it should work the same in transaction modifiers and in amount expressions.

site/doc/1.11/journal.md
simonmichael: 1.6666666666666666...

simonmichael: I meant: 0.016666666666666666...

simonmichael:
We don't currently support rounding strategies, that would be a separate discussion. We store arbitrary decimal numbers (up to 255 decimal places IIRC), and we display them with the number of decimal places inferred or specified for each commodity.

site/doc/1.11/journal.md
ag-eitilt: First take of it at #913, pending further discussion.

simonmichael:
Status: keeping this PR open as the parent, but the code is superseded by #902 and #913.

simonmichael:
Actually I'm going to close this, for clarity. We can always follow the links back. The "exact multi-commodity assertions" part has been merged as #902, and the "amount expressions and multi-commodity postings" part is still WIP, currently at #913.

PR #872, opened 2018-09-21
cli: command to list included files

adept:
This adds "hledger includes" that will list all included files.

This is useful for auditing and automation (I plan to build some Shake rules around rebuilding included files from CSV's)

### DIFF
--- a/hledger/Hledger/Cli/CliOptions.hs
+++ b/hledger/Hledger/Cli/CliOptions.hs
@@ -278,6 +278,7 @@ standardAddonsHelp = [
   ,("vty", "start the curses-style interface")
   ,("web", "start the web interface")
   ,("accounts", "list account names")
+  ,("files", "list included files")
   ,("balance-csv", "output a balance report as CSV")
   ,("close", "show a transaction entry zeroing all accounts")
   ,("print-unique", "print only transactions with unique descriptions")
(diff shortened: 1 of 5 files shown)


hledger/Hledger/Cli/Commands/Includes.hs
```diff
@@ -0,0 +1,57 @@
+{-|
+
+The @includes@ command lists account names:
+
+- by default, it shows names of all the included journals and journal itself.
+
+- with '--omit-self', it shows names of all included files only
```
simonmichael: Is the option worthwhile ? I guess the main journal will always be the first item ?

adept: not sure if it is worthwhile honestly. Let me sleep on this ;)

adept:
"files" is actually better that includes, i think. Let me change it.

simonmichael:
PS don't forget an entry in hledger_commands.m4.md.

adept:
Here we go, take 2

simonmichael:
Looks good!

What do you use the regex argument for ? Is it necessary to be case sensitive ? (I think this would be the only case sensitive one in hledger)

adept:
Basically to get files from specific subdirectories only (eg autogenerated
files vs manually created ones).

Given that filesystems are more often case sensitive than not I think it
makes sense to keep the matching case sensitive.

simonmichael:
Fair enough. I’ll merge this soon if there are no further comments - thanks.

PR #873, opened 2018-09-22
cli: "close" can print only opening or closing transaction

adept:
Gives "hledger close" ability to print just the opening transaction or just the closing transaction.

Useful to automate/script "period closing" activities


simonmichael:
Combining `hledger close --only-opening --only-closing` is the same as `hledger close` I guess.

Would `--opening`/`--closing` also work, for less typing ?

Can you say any more about how you use this in practice ?

adept:
`--opening` and `--closing` are definitely just as good

I use per-year journals and `hledger close` to generate end-of-year carry-over transactions. Closing transaction goes into separate file that is !included in (year-1) journal, opening transactions goes into file that gets included into (year) journal. Ability to supply --opening / --closing means simpler scripting.

adept:
Added test for `close --opening --closing`

adept:
I am out of good ideas :( 

* "opening-close" with alias to "close-open"?
* "end-of-period" ?
* "carry-over" ?
* "reopen" ?

hledger/Hledger/Cli/Commands/Close.hs
```diff
@@ -40,6 +42,9 @@ calculated as of end of yesterday, and the opening transaction is dated today.
 To close on some other date, use: `hledger close -e OPENINGDATE ...`.
 (-p or date: can also be used, the begin date is ignored.)
 
+You can chose to print just one of the transactions with `--opening-only` 
+or `--closing-only`.
```
simonmichael: --opening/--closing now

simonmichael:
Cf [hledger list thread](https://groups.google.com/d/msg/hledger/-B5PPlrEzkU/bnax9XuJAAAJ). I said I was leaning towards renaming to `endperiod`, but there's a few ways you could imagine using this command:

1. carrying asset/liability/equity balances between files while allowing both single or multi-file reports (what I use it for)
2. "closing the books" in the accounting sense: transfering revenues/expenses to assets/liabilities
3. recording the closure of a real-world account, eg a checking account, with some extra convenience (ensuring all the balance has been moved elsewhere, eg)
4. other things we haven't thought of ?

3 isn't quite practical yet, you'd need to specify the transfer account (I forgot it's hard coded to equity:opening/closing balances). But maybe it's useful to keep the more generic "close" name, and we should try that a bit longer.

PR #874, opened 2018-09-22
lib: fix balance --budget in the presence of --value

### LINKED ISSUE #862: Cumulative budget forecast gives incorrect percentages when assets are included
I just ran into the following weird bug. Simplified example:

```
P 2018/01/26 SHARE €10

2018/05/17 Dividend
    revenue:dividend       €-10
    assets:pension          €10

2018/05/17 Buy
    assets:pension:shares   1 SHARE @ €10
    assets:pension          €-10

~ monthly from 2018/06 to 2018/07
    assets:pension                    €1
    assets:bank
```

Gives the following forecast:

```
hledger -f bug.journal bal -M --budget --cumulative --forecast --value
Budget performance in 2018/05/01-2018/06/30:

                ||                      May                        Jun
================++=====================================================
 <unbudgeted>   ||    €-10                       €-10
 assets:bank    ||       0                        €-1 [ 100% of   €-1]
 assets:pension ||     €10                       €11 [10100% of    €1]
----------------++-----------------------------------------------------
                ||       0                          0 [             0]
```

10100% of €1 is €101, not €11, so this can't be right. It looks like it might accidentally multiply the asset by the asset price twice maybe?

adept:
This fixes #862 

I thnk that `amountValue` should not transplant original price of the amount into the amount value, as this will incorrectly inflate amount further down the line whenever `amountValue` of `costOfAmount` is applied on the result.

Fix is verified using sample from #862

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -400,9 +400,7 @@ canonicaliseAmount styles a@Amount{acommodity=c, astyle=s} = a{astyle=s'}
 amountValue :: Journal -> Day -> Amount -> Amount
 amountValue j d a =
   case commodityValue j d (acommodity a) of
-    Just v  -> v{aquantity=aquantity v * aquantity a
-                ,aprice=aprice a
-                }
+    Just v  -> v{aquantity=aquantity v * aquantity a}
     Nothing -> a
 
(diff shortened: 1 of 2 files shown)


simonmichael:
That's interesting. Tests pass, so sounds good to me. Thanks!

PR #877, opened 2018-09-27
allow account types (asset, liability..) to be declared

simonmichael:


hpdeifel:
Nice! I use german account names and the hardcoded english names were always annoying to me. I have a custom shell alias that passes `--alias` to hledger for all ALERX accounts but had  to remember to use it only for `bs`, `is` and friends as I wanted to see the german names in `bal` or `reg`...

Replacing those aliases with account declarations worked really well in my limited testing, but I haven't tried to do anything fancy with non-toplevel accounts.

About the syntax:

What's the reason for using single letters (ALERX) instead of just spelling out the words (Assets, etc...)? After all you're probably going to need this only once per journal, so the extra typing overhead wouln't be significant. On the other hand i think complete words be much more readable (especially if you don't know the english names off the top of your head and wonder what those letters might stand for).

I also kinda hate the syntactic significance of one vs. two spaces, especially because the difference between

```
account Einnahmen  R
account Ausgaben   X
account Vermögen   A
account Schulden   L
```

and

```
account Einnahmen R
account Ausgaben  X
account Vermögen  A
account Schulden  L
```

is just that "Einnahmen" isn't considered a revenue account. But I think that ship may have already sailed.

simonmichael:
Thanks for the feedback.

> What's the reason for using single letters (ALERX) instead of just
> spelling out the words (Assets, etc...)? After all you're probably
> going to need this only once per journal, so the extra typing
> overhead wouln't be significant. On the other hand i think complete
> words be much more readable (especially if you don't know the
> english names off the top of your head and wonder what those letters
> might stand for).

- Since the goal was to fix anglocentrism, I figured single letters
  would be more pleasant for non-english-speakers than a bunch of
  english words.

- I figure these will be used more often than we'd think - you might
  have multiple journals for different purposes, you might set up a new
  chart of accounts each year - so reading and typing out the full words
  repeatedly would feel tedious.

- if you write the full word there can be confusion about spelling.
  ASSET or ASSETS ?

I take your point that some folks might find the full word, even if
not in their language, clearer than just the first letter.

Should we allow a choice then ?
(A|L|E|R|X|ASSET|LIABILITY|EQUITY|REVENUE|EXPENSE), case insensitive ?

> I also kinda hate the syntactic significance of one vs. two spaces,
> especially because the difference between

> ```
> account Einnahmen  R
> account Ausgaben   X
> account Vermögen   A
> account Schulden   L
> ```

> and

> ```
> account Einnahmen R
> account Ausgaben  X
> account Vermögen  A
> account Schulden  L
> ```

> is just that "Einnahmen" isn't considered a revenue account. But I
> think that ship may have already sailed.

I hear that. Double space is pretty baked in to *ledger at present,
and we use it in a few places by now. I'm open to other ideas.

Another option is to support comments and tags like txns/postings:
```
account Assets ; type:A
```

I want to add this, but I felt we should also have a shorter syntax
for such a common declaration. The above feels a bit boilerplatey.

Another option is one-letter tags:
```
account Assets      ; A:
account Liabilities ; L:
account Revenues    ; R:
```

tonyxiao:
could this be used to associate generic tags with accounts?

simonmichael:
Yes, we’ll do that too

simonmichael:
Experimentation will continue, but this is now merged.

hpdeifel:
Sorry for the slow replies.

>     * Since the goal was to fix anglocentrism, I figured single letters
>       would be more pleasant for non-english-speakers than a bunch of
>       english words.

I still think that as a non-native english speaker, I'll have no clue what "ALERX" stands when I need it and would have to look it up in the documentation. Also, I'm not going to need those often, so spelling them out would be absolutely fine.

simonmichael:
Noted. A new wrinkle: we might drop the A letter/ASSET word syntax in favour of some tag like
```
account asset ; type:A
```
and the same would apply there, we could allow `type:ASSET`. 

Pro: 
- less special syntax. You don't need to declare many types so the extra typing/noise is not a big deal.
- more Ledger compatible (presumably)
- easier to make ledger-mode compatible (just need to teach ledger-mode to ignore comments when autocompleting)

Unknown:
- upper case or case insensitive ?
- ok to reserve very common "type" for this tag name for ? What word is left for account type as in "checking/cash/credit card/brokerage/..." ?

Issue #879, opened 2018-09-29
Plan for megaparsec 7 support? blocking Arch Linux packagers


felixonmars:
Looks like megaparsec changed the related part again... Do we have a plan to support the new version soon?

simonmichael:
Yes: https://github.com/simonmichael/hledger/issues/853#issuecomment-420054999. Have you had a chance to start on this @awjchen ?

Is it causing problems @felixonmars ?

felixonmars:
Ah thanks, didn't find that comment.

It is currently blocking the megaparsec update on Arch. I have recalled the batch already so it is not in a hurry :)

awjchen:
I've been working on re-implementing the `ExceptT` layer, but I had decided to delay moving to megaparsec 7 since some of our dependencies (cassava-megaparsec, config-ini) still required megaparsec <7. It now looks like they have all been updated to use megaparsec 7, so I'll prioritize this work.

simonmichael:
This was merged, next hledger release will require megaparsec 7.

PR #880, opened 2018-09-30
Re-implement the 'ExceptT' layer of the parser and switch to megaparsec 7 [WIP]

awjchen:
Details can be found in the commit messages.

This PR also fixes an issue found in #853.
Error messages for cyclic include files now appear as follows:
```
$ cat a.j 
include b.j

$ cat b.j 
include c.j

$ cat c.j 
include a.j

$ hledger stats -f a.j
hledger: in file included from a.j,
in file included from b.j,
c.j:1:12:
  |
1 | include a.j
  |            ^
Cyclic include: a.j

$ hledger stats -f b.j
hledger: in file included from b.j,
in file included from c.j,
a.j:1:12:
  |
1 | include b.j
  |            ^
Cyclic include: b.j

$ hledger stats -f c.j
hledger: in file included from c.j,
in file included from a.j,
b.j:1:12:
  |
1 | include c.j
  |            ^
Cyclic include: c.j

```


awjchen:
I've added a commit switching hledger to megaparsec 7, addressing #879.

I ran into a problem while trying to build on ghc 7.10 (we must use config-ini-0.2.3.0 for megaparsec 7, but it uses `-fno-warn-redundant-constraints`, which was only introduced in ghc 8.0) that I don't know how to resolve, so I've marked this PR as a WIP. Any suggestions would be greatly appreciated!

Also, benchmarks with `stack bench hledger` do not indicate a regression:

Before
```
read bench/10000x1000x10.journal        [1.27s]
print                                   [1.61s]
register                                [2.26s]
balance                                 [0.18s]
Total: 5.31s                 
```
After adding the `ExceptT` layer
```
read bench/10000x1000x10.journal        [1.14s]
print                                   [1.59s]
register                                [2.22s]
balance                                 [0.18s]
Total: 5.13s                 
```
After switching to megaparsec 7
```
read bench/10000x1000x10.journal        [1.03s]
print                                   [1.60s]
register                                [2.28s]
balance                                 [0.20s]
Total: 5.11s                 
```

simonmichael:
Thanks for this, and for checking performance.

It's fine to ignore GHC 7.10 at this point. We can remove it officially from the supported GHCs later.

Here's my take on some related issues:

|          |              |                                                                                                          |                |
| -------- | ------------ | -------------------------------------------------------------------------------------------------------- | -------------- |
| issue/pr | author       | title/description                                                                                        | waiting for    |
| \#807    | simonmichael | allow descriptions/comments in periodic transactions (regression in periodic transaction parsing)        | \#880          |
| \#823    | awjchen      | Allow "re-parsing" with custom parse errors, some commodity cleanups (& fix 807)                         | ?              |
| \#853    |              | Hledger does not detect cycles when resolving 'include' directives                                       | \#880          |
| \#879    |              | Plan for megaparsec 7 support? blocking Arch Linux packagers                                             | \#880          |
| \#880    | awjchen      | Re-implement the 'ExceptT' layer of the parser and switch to megaparsec 7 \[WIP\] (fixing 807? 823? 853) | review, rebase |

simonmichael:
More feedback coming soon I hope.

hledger-lib/Text/Megaparsec/Custom.hs
```diff
@@ -129,6 +151,136 @@ customParseErrorPretty source err = case findCustomError err of
...
+-- then apply the pretty-printer for custom parse errors.
+
+finalParseErrorPretty :: FinalParseErrorBundle' CustomErr -> String
+finalParseErrorPretty bundle =
+     concatMap printIncludeFile (NE.init (sourceFileStack bundle))
+  <> customParseErrorPretty (errorSource bundle) (finalParseError bundle)
+  where
+    printIncludeFile path = "in file included from " <> path <> ",\n"
+
+
```
simonmichael: Phew. This stuff looks pretty tricky and opaque to a new reader. The comments are definitely appreciated.

I wonder why megaparsec stopped tracking the stack of files.

hledger-lib/package.yaml
```diff
@@ -57,7 +57,7 @@ dependencies:
 - easytest
 - filepath
 - hashtables >=1.2.3.1
-- megaparsec >=6.4.1 && < 7
+- megaparsec >=6.4.1
```
simonmichael: Huh.. I thought it would be megaparsec >=7. Do we support both 6 and 7 ?

stack-ghc7.10.yaml
```diff
@@ -26,8 +26,8 @@ extra-deps:
 - base-orphans-0.7
 - bifunctors-5.5.2
 - brick-0.37.1
-- cassava-megaparsec-1.0.0
-- config-ini-0.2.2.0
+- cassava-megaparsec-2.0.0
```
simonmichael: https://www.stackage.org/package/cassava-megaparsec hasn't been in stackage for a long time, which complicates our install instructions. We'll try and improve that.

simonmichael:
(A little more testing tomorrow.)

hledger-lib/package.yaml
awjchen: Ah, this is a mistake on my part. The code in this PR will only work with megaparsec 7, so the bounds should be megaparsec >= 7.0.0

hledger-lib/Text/Megaparsec/Custom.hs
awjchen: Hmm, I think I can do better with the comments so that the code is less opaque. I will do some work on them, but it might be one or two days before I have the chance to. Is there anything  particularly opaque that I should make sure to address?

From reading the blog post and changelog for megaparsec 7, it seems that support for stacks of include files was dropped to make megaparsec faster and simpler: stacks of include files made the manipulation of source positions very hairy. Indeed, the bug we found in #853 could be an example of this, in my opinion.

simonmichael:
> Hmm, I think I can do better with the comments so that the code is less opaque. I will do some work on them, but it might be one or two days before I have the chance to. Is there anything particularly opaque that I should make sure to address?

Nothing particular, unless we can think of a better name than "Bundle". After some more time/experience we might find more ways to clarify.

simonmichael:
I confirm this fixes the wrong parsing error we were triggering on #853. 

I thought this might be related to #823 as well, but I'm not sure. Anyway I guess I'll wait till this lands before looking further at #807/#823.

awjchen:
I've added some commits expanding the comments and updating the megaparsec bounds.

In this PR I was focused only on #853 and #879. I haven't thought of #807 or #823 in a while, so I'm not yet sure how to proceed with them, but I think that work should be done next.

simonmichael:
Great. Thanks @awjchen!

simonmichael:
> I ran into a problem while trying to build on ghc 7.10 (we must use config-ini-0.2.3.0 for megaparsec 7, but it uses -fno-warn-redundant-constraints, which was only introduced in ghc 8.0)

Reported: https://github.com/aisamanra/config-ini/issues/1#issuecomment-428377098

simonmichael:
Fixed by #916.

PR #882, opened 2018-10-01
Remove Twitter Section of Site

### LINKED ISSUE #870: twitter widget needs updating
No tweets are appearing on the home page. Apparently we're using an old style and they just turned it off.

https://twittercommunity.com/t/deprecating-widget-settings/102295/4

nikhiljha:
Fix #870

### DIFF
--- a/site/index.md
+++ b/site/index.md
@@ -224,21 +224,3 @@ a file format for human-friendly approximate time logging
 </div> <!-- row -->
 </div> <!-- container -->
 
-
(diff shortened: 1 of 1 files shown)


simonmichael:
I would have said maybe keep the links to facilitate searching Twitter manually, but they don't bring up recent results, so are basically useless and I agree we might as well remove them. 

Not sure what to replace this with yet.

Thanks!

PR #886, opened 2018-10-05
WIP: Web: Prevent Adding Transactions with Invalid Date

### LINKED ISSUE #433: hledger-web: missing error when adding a transaction with an invalid date  [$20]

nikhiljha:
Fixes #433 

Works in Chrome, Firefox, Edge by preventing the user from entering invalid dates (will work until the year 9999 😄).

I also removed the serverside date validation code and let Yesod handle it instead.

If Safari should be supported, there's a somewhat lightweight polyfill available: https://www.npmjs.com/package/date-input-polyfill

screen shot 2018-10-05 at 4 25 44 pm

TODO:
- Add the Safari Polyfill.
- Remove the old date picker button.

### DIFF
--- a/hledger-web/Hledger/Web/Widget/AddForm.hs
+++ b/hledger-web/Hledger/Web/Widget/AddForm.hs
@@ -87,15 +87,12 @@ addForm j today = identifyForm "add" $ \extra -> do
       nulltransaction {tdate = date, tdescription = desc, tpostings = postings}
 
     dateFS = FieldSettings "date" Nothing Nothing (Just "date")
-      [("class", "form-control input-lg"), ("placeholder", "Date")]
+      [("class", "form-control input-lg")]
     descFS = FieldSettings "desc" Nothing Nothing (Just "description")
       [("class", "form-control input-lg typeahead"), ("placeholder", "Description"), ("size", "40")]
     acctFS = FieldSettings "amount" Nothing Nothing (Just "account") []
     amtFS = FieldSettings "amount" Nothing Nothing (Just "amount") []
-    dateField = checkMMap (pure . validateDate) (T.pack . show) textField
-    validateDate s =
-      first (const ("Invalid date format" :: Text)) $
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks for looking at this. I have installed it at http://demo1.hledger.org, to compare with http://demo.hledger.org, and tested in firefox 63 on mac. Some quick feedback:

- ?add=1 doesn't bring up the form. Compare http://demo.hledger.org/journal?add=1 and http://demo1.hledger.org/journal?add=1
- the placeholder text is too low, not centered vertically
- the placeholder text I see is `mm / dd / yyyy`, which is not the format I use or that we see in the journal file.
- this needs to at least degrade gracefully, allowing data entry somehow, in all common browsers, and certainly safari. hledger-web is supposed to be something that just works.

simonmichael:
- when I submit the form with just a date and description, I see the form again with those fields populated and the other required fields (account, amount) highlighted, rendered within the page, not as a dialog. 

This is standard yesod forms behaviour, but I don't think hledger-web has ever done this before. Is it because of your change ?

Preserving the data and highlighting the missing fields are both very good. 

Having the form appear first as a dialog and then in the page is not so good, but maybe not the end of the world.

- it seems impossible to enter an invalid date now, which prevents #433. But the above enhanced behaviour is the real fix we wanted. Maybe we can get that without changing the date picker ?

Issue #878, opened 2018-09-28
--anon seems to break balancesheet


simonmichael:
Adding to --anon to balance does not change the numbers, as expected. Adding it to balancesheet changes them dramatically, at least with my personal ledger. Could someone confirm ?

cocreature:
I can confirm this. The reason why this happens is that when `assets` is anonymized the query done by the balancesheet hits the wrong accounts. I’m not entirely sure what it is hitting in this case (I would have expected it to match nothing at all). An easy fix would be to not anonymize account prefixes that are used in fixed queries such as `assets`, `liabilities` and probably a few others. Ofc that does end up leaking a bit of information but I’m not sure that’s actually a problem.

PR #899, opened 2018-10-11
Anonymize original postings

cocreature:
I took another look at #878 and figured out that the reason why the balancesheet doesn’t just show zeros is that the original postings are present for inferred postings and are not anonymized. That causes the query to match for all postings with inferred amounts. By also anonymizing the original posting we now get zeros as expected and we avoid accidentally leaking information (the problem probably isn’t limited to balancesheet queries).


simonmichael:
Bah those "original postings" are a pain. Thanks for the fix!

Issue #893, opened 2018-10-09
transaction modifiers miss inferred amounts in transaction


jkr:
Running with a stack build of up-to-date master.

Transaction modifiers seem to interact oddly with inferred amounts. The following works as expected:

```
= ^expenses:groceries
    [budget:groceries]                           *-1
    [assets:bank:checking]                        *1

10/7 * MARKET
    expenses:groceries:food                      $20
    assets:bank:checking
```

```
$ hledger -ffoo.ledger reg --auto
2018/10/07 MARKET               ex:groceries:food              $20           $20
                                [budget:groceries]            $-20             0
                                [as:bank:checking]             $20           $20
                                assets:bank:checking          $-20             0
```

But the following does not:

```
= ^expenses:groceries
    [budget:groceries]                           *-1
    [assets:bank:checking]                        *1

10/7 * MARKET
    expenses:groceries:food
    assets:bank:checking                         $-20
```

```
$ hledger -ffoo.ledger reg --auto
hledger: could not balance this transaction - can't have more than one balanced virtual posting with no amount (remember to put 2 or more spaces before amounts)
2018/10/07 * MARKET
    expenses:groceries:food
    [budget:groceries]
    [assets:bank:checking]
    assets:bank:checking                $-20
```

simonmichael:
Thanks for the report.

jkr:
I'll poke around as a way of familiarizing myself with the codebase, and see if I can help out. Would `HLedger.Data.TransactionModifer` be where I should start? (It looks like the inference logic is in `H.D.Transaction`, right?)

simonmichael:
@jkr, that would be great. Yes that sounds right.

jkr:
I figured out the issue: it was apply the modifiers before it finalized the journal with `journalFinalise`. A simple solution (apply modifiers after `journalFinalise`) is here:

https://github.com/simonmichael/hledger/compare/master...jkr:modifiers_fix

BUT, this produces its own problem. Because the application happens after the finalization, it no longer checks to make sure they balance. So we can now have:

```
= ^expenses:groceries
    [budget:groceries]                           *-1
    [assets:bank:checking]                        *2

10/7 * MARKET
    expenses:groceries:food                      $20
    assets:bank:checking                        -$20
```

```
$ stack exec hledger -- print -f /tmp/foo.ledger --auto
2018/10/07 * MARKET
    expenses:groceries:food             $20
    [budget:groceries]                 $-20
[...]

PR #900, opened 2018-10-11
Make transaction modifiers work correctly with inferred amounts

jkr:
This solves #893 by having `journalFinalise` run twice with automated transactions: the first run takes care of reordering and inferring amounts (but does not check assertions), the second does not reorder. It was necessary to make reordering a boolean flag in `journalFinalise`. Note that if `--auto` is not specified, the finalization is only run one time, as before.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -523,19 +523,21 @@ filterJournalTransactionsByAccount apats j@Journal{jtxns=ts} = j{jtxns=filter tm
 -- use.  Reverse parsed data to normal order, standardise amount
 -- formats, check/ensure that transactions are balanced, and maybe
 -- check balance assertions.
-journalFinalise :: ClockTime -> FilePath -> Text -> Bool -> ParsedJournal -> Either String Journal
-journalFinalise t path txt assrt j@Journal{jfiles=fs} =
-  journalTieTransactions <$> 
-  (journalBalanceTransactions assrt $
-   journalApplyCommodityStyles $
-   j {jfiles        = (path,txt) : reverse fs
-     ,jlastreadtime = t
-     ,jdeclaredaccounts = reverse $ jdeclaredaccounts j
-     ,jtxns         = reverse $ jtxns j         -- NOTE: see addTransaction
-     ,jtxnmodifiers = reverse $ jtxnmodifiers j -- NOTE: see addTransactionModifier
-     ,jperiodictxns = reverse $ jperiodictxns j -- NOTE: see addPeriodicTransaction
-     ,jmarketprices = reverse $ jmarketprices j -- NOTE: see addMarketPrice
-     })
+journalFinalise :: ClockTime -> FilePath -> Text -> Bool -> Bool -> ParsedJournal -> Either String Journal
+journalFinalise t path txt reorder assrt j@Journal{jfiles=fs} =
+  let j' = if reorder
+           then j {jfiles        = (path,txt) : reverse fs
+                  ,jlastreadtime = t
+                  ,jdeclaredaccounts = reverse $ jdeclaredaccounts j
+                  ,jtxns         = reverse $ jtxns j         -- NOTE: see addTransaction
+                  ,jtxnmodifiers = reverse $ jtxnmodifiers j -- NOTE: see addTransactionModifier
+                  ,jperiodictxns = reverse $ jperiodictxns j -- NOTE: see addPeriodicTransaction
+                  ,jmarketprices = reverse $ jmarketprices j -- NOTE: see addMarketPrice
+                  }
+           else j
+  in journalTieTransactions <$>
+     (journalBalanceTransactions assrt $ journalApplyCommodityStyles j')
+
 
 journalNumberAndTieTransactions = journalTieTransactions . journalNumberTransactions
 
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -22,6 +22,9 @@ module Hledger.Data.Journal (
   journalCommodityStyles,
   journalConvertAmountsToCost,
   journalFinalise,
+  journalReverse,
+  journalSetTime,
+  journalSetFilePath,
   journalPivot,
   -- * Filtering
   filterJournalTransactions,
@@ -518,6 +521,23 @@ filterJournalTransactionsByAccount apats j@Journal{jtxns=ts} = j{jtxns=filter tm
       (negatives,positives) = partition isnegativepat apats
 
 -}
+
+journalReverse :: Journal -> Journal
+journalReverse j =
+  j {jfiles            = reverse $ jfiles j
+    ,jdeclaredaccounts = reverse $ jdeclaredaccounts j
+    ,jtxns             = reverse $ jtxns j
+    ,jtxnmodifiers     = reverse $ jtxnmodifiers j
+    ,jperiodictxns     = reverse $ jperiodictxns j
+    ,jmarketprices     = reverse $ jmarketprices j
+    }
+
+journalSetTime :: ClockTime -> Journal -> Journal
+journalSetTime t j = j{ jlastreadtime = t }
+
+journalSetFilePath :: FilePath -> Text -> Journal -> Journal
+journalSetFilePath path txt j = j {jfiles = (path,txt) : jfiles j}
+
 
 -- | Do post-parse processing on a parsed journal to make it ready for
 -- use.  Reverse parsed data to normal order, standardise amount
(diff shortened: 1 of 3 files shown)


hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -523,19 +523,21 @@ filterJournalTransactionsByAccount apats j@Journal{jtxns=ts} = j{jtxns=filter tm
...
-   journalApplyCommodityStyles $
-   j {jfiles        = (path,txt) : reverse fs
-     ,jlastreadtime = t
-     ,jdeclaredaccounts = reverse $ jdeclaredaccounts j
-     ,jtxns         = reverse $ jtxns j         -- NOTE: see addTransaction
-     ,jtxnmodifiers = reverse $ jtxnmodifiers j -- NOTE: see addTransactionModifier
-     ,jperiodictxns = reverse $ jperiodictxns j -- NOTE: see addPeriodicTransaction
-     ,jmarketprices = reverse $ jmarketprices j -- NOTE: see addMarketPrice
-     })
+journalFinalise :: ClockTime -> FilePath -> Text -> Bool -> Bool -> ParsedJournal -> Either String Journal
```
simonmichael: Could you move/copy the nice commit doc explanation into the haddock.

hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -523,19 +523,21 @@ filterJournalTransactionsByAccount apats j@Journal{jtxns=ts} = j{jtxns=filter tm
...
+                  ,jlastreadtime = t
+                  ,jdeclaredaccounts = reverse $ jdeclaredaccounts j
+                  ,jtxns         = reverse $ jtxns j         -- NOTE: see addTransaction
+                  ,jtxnmodifiers = reverse $ jtxnmodifiers j -- NOTE: see addTransactionModifier
+                  ,jperiodictxns = reverse $ jperiodictxns j -- NOTE: see addPeriodicTransaction
+                  ,jmarketprices = reverse $ jmarketprices j -- NOTE: see addMarketPrice
+                  }
+           else j
+  in journalTieTransactions <$>
+     (journalBalanceTransactions assrt $ journalApplyCommodityStyles j')
```
simonmichael: Bikeshedding: maybe confining the effect of reorder to a smaller block of code makes it clearer ?
```haskell
journalFinalise t path txt reorder assrt j@Journal { jfiles = fs } =
  journalTieTransactions
  <$> (journalBalanceTransactions assrt $
       journalApplyCommodityStyles j{
          jfiles            = (path, txt) : mayberev fs
          , jlastreadtime     = t
          , jdeclaredaccounts = mayberev $ jdeclaredaccounts j
          , jtxns             = mayberev $ jtxns j         -- NOTE: see addTransaction
          , jtxnmodifiers     = mayberev $ jtxnmodifiers j -- NOTE: see addTransactionModifier
          , jperiodictxns     = mayberev $ jperiodictxns j -- NOTE: see addPeriodicTransaction
          , jmarketprices     = mayberev $ jmarketprices j -- NOTE: see addMarketPrice
          })
  where mayberev = if reorder then reverse else id
```

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -273,11 +286,24 @@ parseAndFinaliseJournal' parser iopts f txt = do
...
+          runFin reorder ignore = journalFinalise t f txt reorder ignore
+          fj = if auto_ iopts
+               then applyTransactionModifiers <$>
+                    runFin True False pj >>=
+                    runFin False (not $ ignore_assertions_ iopts)
+               else runFin True (not $ ignore_assertions_ iopts) pj
+      in
+        case fj of
+          Right j -> return j
+          Left e  -> throwError e
```
simonmichael: IIRC this code is already a bit hard to grok, so I don't love that we are adding more complexity. 

The commit doc is clearer than the code comment, perhaps you can move more of it into the code (as haddock or ordinary comment).

I really don't love the idea of duplicating this twice. Or that we have these two similar looking functions, one with no docs. I wonder if we can clean this up.

simonmichael:
(It seems like parseAndFinaliseJournal' can go away, it's only used by the timeclock and timedot readers. Right @awjchen ?)

jkr:
> (It seems like parseAndFinaliseJournal' should go away, it's only used by the timeclock and timedot readers. Right @awjchen ?)

Per my ack search when I came across it, that's all.

hledger-lib/Hledger/Data/Journal.hs
jkr: Yeah -- I like this. But in this case, we'd be adding `(path,txt)` to `jfiles` twice in the `--auto` case, right?

simonmichael: True. It's also setting jlastreadtime. "reorder" doesn't quite cover what it's about. "finallyfinalise" ? Split this function into two ?

simonmichael: I don't see exactly how to clarify it right now.. if you don't see a way, it's ok to leave it how it was.

jkr:
I added a commit (8cbead6) which only does the double run if `--auto` is set AND the `jtxnmodifiers` is non-null. So even if auto is default, there will be no performance hit unless there are modifiers to apply.

(Informal tests showed a slight performance hit when using `--auto` before, though I haven't run it on large enough journals to know how it increases with size.)

hledger-lib/Hledger/Data/Journal.hs
jkr: `journalFinalise` is only used in Read.Common's `parseAndFinaliseJournal` (and the primed version). Since it's a single-use function, I wonder if combining all its steps is creating more confusion. I.e., just combine them in `parseAndFinaliseJounal` as we need them, instead of adding a bunch of boolean flags.

simonmichael: Perhaps!

jkr: See 6a4ece6 below for a try at this splitting up (docs aren't added, and it isn't added to the primed function).

Pros: more explicit what's going on; no opaque boolean flags; no running unnecessary steps
Cons: at first blush, a more confusing `parseAndFinaliseJournal` function

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -297,9 +297,19 @@ parseAndFinaliseJournal' parser iopts f txt = do
           runFin reorder ignore = journalFinalise t f txt reorder ignore
```
simonmichael: Did you mean to stop calling journalFinalise here, and drop that function ?

jkr: I was just adding it to the unprimed version as an example. If it seems like a good idea, I'll go through, add docs to the smaller functions, and integrate it into the primed version. (BTW, I still don't understand the primed one -- it seems to relate to ExceptT error catching, but I note that the program compiles fine with all tests passing if it's replaced in TimeClock and TimeDot with the unprimed one.)

jkr: Oh -- oops. I see what you mean. One moment.

simonmichael:
> I was just adding it to the unprimed version as an example. If it seems like a good idea, I'll go through, add docs to the smaller functions, and integrate it into the primed version.

Yep good idea, it looks much clearer.

> (BTW, I still don't understand the primed one -- it seems to relate to ExceptT error catching, but I note that the program compiles fine with all tests passing if it's replaced in TimeClock and TimeDot with the unprimed one.)

ErroringJournalParser is a variant of JournalParser that can abort parsing by throwing an exception. It's our most general and featureful kind of journal parser. Some months ago we removed it thinking it was unneeded, and it has recently been re-added.

awjchen:
> (BTW, I still don't understand the primed one -- it seems to relate to ExceptT error catching, but I note that the program compiles fine with all tests passing if it's replaced in TimeClock and TimeDot with the unprimed one.)

> (It seems like parseAndFinaliseJournal' should go away, it's only used by the timeclock and timedot readers. Right @awjchen ?)

It's true that one can replace `parseAndFinaliseJournal'` with `parseAndFinaliseJournal` in TimeClock (and Timedot). This is because the TimeClock parsers are defined for all `JournalParser m`, while `ErroringJournalParser n` is just a synonym for  `JournalParser m` where `m = ExceptT FinalParseError n`.

But using `parseAndFinaliseJournal` instead of `parseAndFinaliseJournal'` introduces an `ExceptT` layer that is subsequently ignored. By using `parseAndFinaliseJournal'` we are being more specific, using only what we need. For this reason, I would prefer to keep `parseAndFinaliseJournal'`, all else equal; however, more practically, I don't see anything wrong with getting rid of it, especially if there is some reason to do so.

simonmichael:
A practical reason would be getting rid of duplicated code and one more type folks don't have to think about. (We could also factor out the common code). I think there's no harm in standardising and using the more powerful type for the time parsers even though they don't make use of it yet.

awjchen:
Agreed, those sound like good reasons to me.

PR #902, opened 2018-10-12
multi-commodity assertions

ag-eitilt:
More focused PR derived from #871, containing only the changes required to allow assertions over the entire contents of an account.  Written primarily as groundwork and proof-of-concept for #556.

### DIFF
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -236,8 +236,11 @@ instance Show Status where -- custom show.. bad idea.. don't do it..
   show Pending   = "!"
   show Cleared   = "*"
 
+-- | The amount to compare an account's balance to, to verify that the history
+-- leading to a given point is correct or to set the account to a known value.
 data BalanceAssertion = BalanceAssertion {
-      baamount   :: Amount,
+      baamount   :: Amount,             -- ^ the expected value of a particular commodity
+      baexact    :: Bool,               -- ^ whether the assertion is exclusive, and doesn't allow other commodities alongside 'baamount'
       baposition :: GenericSourcePos
     } deriving (Eq,Typeable,Data,Generic,Show)
 
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -236,21 +236,26 @@ instance Show Status where -- custom show.. bad idea.. don't do it..
   show Pending   = "!"
   show Cleared   = "*"
 
-type BalanceAssertion = Maybe (Amount, GenericSourcePos)
+data BalanceAssertion = BalanceAssertion {
+      baamount   :: Amount,
+      baposition :: GenericSourcePos
+    } deriving (Eq,Typeable,Data,Generic,Show)
+
+instance NFData BalanceAssertion
 
 data Posting = Posting {
-      pdate             :: Maybe Day,         -- ^ this posting's date, if different from the transaction's
-      pdate2            :: Maybe Day,         -- ^ this posting's secondary date, if different from the transaction's
+      pdate             :: Maybe Day,               -- ^ this posting's date, if different from the transaction's
+      pdate2            :: Maybe Day,               -- ^ this posting's secondary date, if different from the transaction's
       pstatus           :: Status,
       paccount          :: AccountName,
       pamount           :: MixedAmount,
-      pcomment          :: Text,              -- ^ this posting's comment lines, as a single non-indented multi-line string
+      pcomment          :: Text,                    -- ^ this posting's comment lines, as a single non-indented multi-line string
       ptype             :: PostingType,
-      ptags             :: [Tag],             -- ^ tag names and values, extracted from the comment
-      pbalanceassertion :: BalanceAssertion,  -- ^ optional: the expected balance in this commodity in the account after this posting
-      ptransaction      :: Maybe Transaction, -- ^ this posting's parent transaction (co-recursive types).
-                                              -- Tying this knot gets tedious, Maybe makes it easier/optional.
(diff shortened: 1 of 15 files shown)


hledger/hledger.cabal
```diff
@@ -193,7 +193,7 @@ executable hledger
     , safe >=0.2
     , shakespeare >=2.0.2.2
     , split >=0.1
-    , statistics
+    , statistics <0.15
```
simonmichael: I'd merge this commit with the previous "Minor infrastructure fixes", and rename to "set statistics upper bound; update gitignore" (or split those)

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -677,11 +677,8 @@ priceamountp = option NoPrice $ do
 
 partialbalanceassertionp :: JournalParser m BalanceAssertion
 partialbalanceassertionp = optional $ do
-  sourcepos <- try $ do
-    lift (skipMany spacenonewline)
-    sourcepos <- genericSourcePos <$> lift getSourcePos
-    char '='
-    pure sourcepos
+  sourcepos <- genericSourcePos <$> lift getSourcePos
+  char '='
```
simonmichael: Having a hard time understanding this commit.. is it incomplete ? Message needs clarification.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -238,11 +238,18 @@ instance Show Status where -- custom show.. bad idea.. don't do it..
...
 data BalanceAssertion = BalanceAssertion {
       baamount   :: Amount,
+      baflags    :: AssertionFlags,
       baposition :: GenericSourcePos
     } deriving (Eq,Typeable,Data,Generic,Show)
 
 instance NFData BalanceAssertion
 
+data AssertionFlags = AssertionFlags {
+      afexact :: Bool
```
simonmichael: Let's add some docs, eg -- ^ Does this assertion check the full multi-commodity balance or only one commodity ?

simonmichael: Also for BalanceAssertion, eg -- | A statement about the expected account balance after a particular posting.

simonmichael: PS I'd be fine with inlining the flags in BalanceAssertion, your call.

tests/journal/balance-assertions.test
```diff
@@ -308,3 +308,51 @@ hledger -f - stats
...
+# 19. Mix different commodities and exact assignments
+hledger -f - stats
+<<<
+2016/1/1
+    a      $1
+    a      -1 zorkmids
+    b
+
+2016/1/2
+    a     == $1
```
simonmichael: How does this work.. isn't a $1 - 1 zorkmids at this point ?

ag-eitilt: That's an assignment, not an assertion; the `-1 zorkmids` is being removed from the account.  I've updated the `b` posting to make that more clear.

hledger-lib/Hledger/Data/Types.hs
ag-eitilt: Good catch.  The `Types` module in general could use better docs, but that's no reason to compound the issue.  And I'm not entirely sure why I wasn't inlining the flags to begin with, thanks!

hledger-lib/Hledger/Read/Common.hs
ag-eitilt: Mostly, it doesn't change anything about the behaviour, just the code structure.  These particular lines of code move the skipping of spaces  and the `optional` call out to `JournalReader.hs`, to better reflect that the parse function relates to the assertion itself rather than encoding surrounding file structure as well (which brings it in line with some other parsers in the file).  More generally, that commit just changes `BalanceAssertion` from being a typedef'ed tuple to being a standalone datatype.

Yeah, that could probably be merged with behavioural changes, but this way the foundation is separate from the exact assertions; I could create a new branch off of this commit adding sub-account-inclusive assertions without anything related to commodity-exclusive assertions.

ag-eitilt: Actually, which message were you referring to?  I'm happy enough to try to clarify that, but I'm too close to the code to realize which one seems confusing from the outside.

simonmichael:
> Actually, which message were you referring to? I'm happy enough to try to clarify that, but I'm too close to the code to realize which one seems confusing from the outside.

I meant the commit message. Thanks for your comments, I'll respond later.

simonmichael:
I've lost track of this a bit. Can I ask for some review from others ? Go/no-go for merge ?

simonmichael:
(PS and this part is not in the proposed docs yet, unless I've misplaced it.)

simonmichael:
> A discussion I've lost makes me think this also changes
> ```
>   a     = $1
> ```

I misspoke, I think it doesn't change that. It allows new multicommodity/exact assignments:
```
  a    == $1
```

ag-eitilt:
Yeah, if it does change `a  = $1`, it's a bug and needs to be fixed; it should just add the second form.  You are right about me forgetting to add that to the docs, though.  Thanks for the catch!

simonmichael:
Sorry for the slow response. Getting ready to merge this. I will combine "lib: Changes shared between expressions/exact assertions" (2f9db904) into the next commit for clarity.

simonmichael:
Merged, thank you very much!

(I renamed the PR to "multi-commodity assertions" for posterity, I prefer that slightly to as not to suggest that exactness is exceptional for us :)

simonmichael:
Still chewing on what we should call these. "multicommodity" could still be ambiguous it (you could imagine a multicommodity assertion that still allows additional unasserted commodities). Complete assertions ? Total assertions ?

PR #903, opened 2018-10-12
Finish removing journalFinalise

jkr:
This finishes removing `journalFinalise`, and adds a fix in the order of operations.

[Followon from [#893](https://github.com/simonmichael/hledger/pull/893), [#900](https://github.com/simonmichael/hledger/pull/900)]

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -21,7 +21,6 @@ module Hledger.Data.Journal (
   commodityStylesFromAmounts,
   journalCommodityStyles,
   journalConvertAmountsToCost,
-  journalFinalise,
   journalReverse,
   journalSetTime,
   journalSetFilePath,
@@ -522,6 +521,9 @@ filterJournalTransactionsByAccount apats j@Journal{jtxns=ts} = j{jtxns=filter tm
 
 -}
 
+-- | Reverse parsed data to normal order. This is used for post-parse
+-- processing, since data is added to the head of the list during
+-- parsing.
 journalReverse :: Journal -> Journal
 journalReverse j =
   j {jfiles            = reverse $ jfiles j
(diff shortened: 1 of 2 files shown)


hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -532,33 +534,14 @@ journalReverse j =
     ,jmarketprices     = reverse $ jmarketprices j
     }
 
+-- | Set clock time in journal
```
simonmichael: "Set this journal's last read time, ie when its files were last read." ? Rename to journalSetLastReadTIme ?

hledger-lib/Hledger/Data/Journal.hs
```diff
@@ -532,33 +534,14 @@ journalReverse j =
...
 journalSetTime :: ClockTime -> Journal -> Journal
 journalSetTime t j = j{ jlastreadtime = t }
 
+-- | See filepath in journal
```
simonmichael: "Add a file (and its content) to the journal's list of parsed files. (Does not parse the file or change the journal otherwise.)" ? Rename to journalAddFile or similar ? Move next to journalFilePaths ?

jkr:
> Do you want to also remove parseAndFinaliseJournal' ?

Not sure if I understand the full implications of that. I've been keeping it parallel so it would be easy to remove, but I haven't looked at the time-tracking code enough to know why it needed the different behavior.

simonmichael:
Ok no problem. I/we can look at that another time.

hledger-lib/Hledger/Data/Journal.hs
jkr: Sure -- I'll do another renaming/redoc-ing commit for this and the below, and add it to this PR.

hledger-lib/Hledger/Data/Journal.hs
jkr: Actually, there's already a redundant `journalAddFile` in `Read.Common`. The only difference is that it adds to the end. We don't need both, since we can just use the one on Read.Common and put it prior to the reversal.

jkr: Addressed in 0d2f8c4 below.

jkr:
I think this is ready to go, from my end.

simonmichael:
Can't remember why I was waiting on this. I took a chance and squashed it into one commit, which I think worked out ok. Thanks!

Issue #895, opened 2018-10-09
unicode emoji breaks hledger register formatting


tonyxiao:
Test case
```
$ cat ./journals/test.journal
Y2018

1/1 Buy Drinks
  creditcard  -20
  expense


1/1 Buy food 🧀
  creditcard  -20
  expense

1/1 Buy toiletry
  creditcard  -20
  expense

$ hledger -f journals/test.journal register
2018/01/01 Buy Drinks                 creditcard                          -20           -20
                                      expense                              20             0
2018/01/01 Buy food 🧀                 creditcard                          -20           -20
                                      expense                              20             0
2018/01/01 Buy toiletry               creditcard                          -20           -20
                                      expense                              20             0
```

Notice the buy food entry off by 1 character. In an actual console this results in a pretty ugly line break

![image](https://user-images.githubusercontent.com/696842/46691746-7578b280-cbba-11e8-873b-0b52b12a3671.png)

simonmichael:
Thanks for the report. It works here, I think you need to adjust your terminal (or font ?) so that wide characters are displayed as double width. I couldn't find any mention of this in docs, so I've added http://hledger.org/manual.html#unicode-characters .

ony:
@simonmichael if issue would be in wide character being displayed in less than two cells then I would expect to see offset to the left side. But on a screenshot it looks like cheese takes two cells but we still put same amount of spaces as for "Buy Drinks".  
So I assume issue is in formatting that doesn't calculate correctly emoji. I tried [wide characters](https://www.unicode.org/charts/nameslist/n_FF00.html) and they were properly aligned.

![screenshot](http://i.imgur.com/OOqrFU1.png)
Github markdown mis-align wide characters on the other hand:
```plain
zsh% hledger -f x.journal reg
2018/01/01 Green Yellow                     creditcard                                 -20           -20
                                            expense                                     20             0
2018/01/01 Ｗｉｄｅ                         creditcard                                 -20           -20
                                            expense                                     20             0
2018/01/01 Wink! 😉                          creditcard                                 -20           -20
                                            expense                                     20             0
```

Note that even [ANSI escape sequences](https://en.wikipedia.org/wiki/ANSI_escape_code) works fine.

ony:
[...]

PR #905, opened 2018-10-15
lib: use native POSIX wcwidth

ony:
Fixes simonmichael/hledger#895

screenshot

### DIFF
--- a/hledger-lib/Hledger/Utils/String.hs
+++ b/hledger-lib/Hledger/Utils/String.hs
@@ -48,6 +48,7 @@ module Hledger.Utils.String (
 
 
 import Data.Char
+import Data.Char.WCWidth (wcwidth)
 import Data.List
 import Text.Megaparsec
 import Text.Megaparsec.Char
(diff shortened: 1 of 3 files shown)


simonmichael:
Status: blocked on a wcwidth release (https://github.com/simonmichael/hledger/issues/895#issuecomment-430062476, https://github.com/solidsnack/wcwidth/issues/2).

simonmichael:
(Ideally.. or, some other workaround eg #895's suggestions.)

simonmichael:
I think we shouldn't depend on an unmaintained library.

simonmichael:
Pinged: https://github.com/solidsnack/wcwidth/issues/2#issuecomment-511463245

simonmichael:
Update: no change - wcwidth is fixed on github but still unreleased and maintainerless. What to do ?

simonmichael:
@ony it seems not much code, shall we add it to hledger-lib ?

simonmichael:
Correction, it's more code than I thought. And it looks like code that could use maintenance, forever. Ugh. What are your thoughts ? I'd like to close this issue/PR somehow.

simonmichael:
We could also just add the missing emoji codes to our big case statement. Maybe that's the simple way to go ?

ony:
So current approaches that I was able to find:

- [`pandoc` and `hledger` have own hand crafted mappings](https://www.stackage.org/haddock/lts-14.15/pandoc-2.7.3/Text-Pandoc-Pretty.html#v:charWidth)
- [`wcwidth` consults native](https://www.stackage.org/lts-14.15/package/wcwidth-0.0.2) `wcwidth` which is part of POSIX.1-2001 and POSIX.1-2008 as per Linux man. It also can generate Haskell native table.
-  [`neovim` generates table from Unicode files with Lua script](https://github.com/neovim/neovim/blob/master/src/nvim/generators/gen_unicode_tables.lua)
- [Rust version `unicode-rs` generates table with Python script](https://github.com/unicode-rs/unicode-width/blob/master/scripts/unicode.py)
- I think I saw

Maybe Haskell deserves to have own implementation independent of platform...

I.e. understand [how to interpret](http://www.unicode.org/reports/tr11/tr11-19.html) and [`EastAsianWidth.txt`](https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt) (at least it includes cheese and says it is wide).

simonmichael:
Great research. I shared with #ghc but I doubt they have the bandwidth..

Still wondering what happens when you install wcwidth on windows.

simonmichael:
Confirmed by @phadej: installing wcwidth-0.2 on windows fails (`.\Data\Char\WCWidth.o:fake:(.text+0x5a): undefined reference to `wcwidth'`), and installing wcwidth-master succeeds (almost: "needed to add {-# LANGUAGE CPP #-} to one file").

simonmichael:
Closing this in favour of the more portable #1749.

PR #909, opened 2018-10-17
hledger-ui: --flat, --auto, --forecast, hide future txns by default

simonmichael:
Here are most of the changes proposed in
https://groups.google.com/d/topic/hledger/Niv-oS-xO_s/discussion 
(for hledger-ui only).


simonmichael:
This is great, I'm gonna merge it.

PR #916, opened 2018-10-24
Drop undecidable Ord instance for ParseError.

peti:
That code fails to compile with ghc-8.6.1 because the instance is undecidable.
I suppose we could enable the appropriate compiler extension to support it, but
I've found that simply removing the instance causes no problems whatsoever: the
entire repository still compiles fine and it passes all test suites, too.


awjchen:
Oops, I forgot to remove this instance in my last PR, which removed a constructor of `CustomErr` that necessitated this instance. Thanks for catching this.

We should also remove the comment just above the instance ("We require an ...").

simonmichael:
Merged. Thanks!

PR #917, opened 2018-10-24
disregard over-specified version constraints in brick and config-ini

peti:
These packages work fine with the latest version of their dependencies.
They just don't say so on Hackage.


simonmichael:
I'm ignorant of the cabal.project, so trusting your judgement. Thanks!

Issue #816, opened 2018-06-18
wrongly indented literal blocks in web manuals


simonmichael:
This is an old problem that keeps recurring. Somewhere in the doc source/css/toolchain, something is indented which should not be, resulting in some/all literal blocks having all but their first line indented in the rendered HTML. Example http://hledger.org/manual.html#periodic-transactions -> "2 or more spaces" line. We should fix it (again) and figure out a way to stop it recurring.

simonmichael:
Thank you very much! I always do that.

PR #926, opened 2018-11-11
Fix incorrectly indented code blocks in web manuals

ackien:
Fixes incorrectly indented code block issue (#816) by removing all indentation from `body`.

This does also make the generated HTML uglier.


simonmichael:
Thank you very much! I always do that.

PR #929, opened 2018-11-14
Add a custom parse error for "re-parsing"

awjchen:
This PR re-implements a subset of PR #823 on top of the current codebase, specifically for addressing issues raised in PR #807.

In particular, the failing examples in https://github.com/simonmichael/hledger/pull/807#issuecomment-396994403 now work properly.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -93,6 +93,7 @@ module Hledger.Read.Common (
 
   -- ** misc
   singlespacedtextp,
+  singlespacedtextsatisfyingp,
   singlespacep,
 
   -- * tests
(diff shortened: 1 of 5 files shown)


awjchen:
One potential problem in this PR is that I made changes to the shell tests in `tests/budget/forecast.test` and the `periodictransactionp` unit tests in `hledger-lib/Hledger/Read/JournalReader.hs` in order to get them to pass. I believe the changes do not contradict the docs, but I am not confident in my understanding of the periodic expression format. In particular, I am unsure when the comment and description fields should end with newlines. Also, I'm not sure how to satisfy one of the inactive `periodictransactionp` unit tests, so I've left it inactive for now. Any help would be appreciated.

simonmichael:
Sorry, I was delayed by haskell tooling issues. It looks good, thanks! Here are my notes:

- The manual should mention the change to journal syntax: in a periodic transaction, two spaces are now required between the period expression and any following description or comment. I'll mention this in release notes as well, as some journals will break.

- Would it be possible to get a clearer parse error - instead of
```
   |
 3 | ~ 2/1 forecast
   |      ^
 unexpected space
 expecting digit or end of period expression
```
something like: `expecting digit or double space or end of line`

- The test changes look fine, except maybe I would not add the empty status, code, comments, tags and postings fields - they are not important for these tests so just add noise. Also I'd go ahead and enable the second disabled test, adding the extra space to make it pass.

Thanks!

awjchen:
Hi, I have made the following changes to the PR
- elaborated on the necessity for a double space in the periodic transaction section of the journal docs
- reworked the parse error arising from a lack of a double space in a periodic transaction, e.g.
```
  |
1 | ~ every 2 weeks from 2018/6/4 to 2018/9 paycheck
  |                                        ^^^^^^^^^
remainder of period expression cannot be parsed
perhaps you need to terminate the period expression with a double space?
```
- removed the irrelevant fields in the `periodictransactionp` unit tests (though I've kept the comment field for consistency)

However, I wasn't sure where to insert a space in the second disabled test to make it pass without making it identical to the test above it. So, for this revision of the PR I've just removed it.

simonmichael:
Thanks!

PR #944, opened 2019-01-09
cli: if abs(return on investment)<0.01% display it as zero

### LINKED ISSUE #937: roi test failing
Something has happened to cause roi functional test failures:
```
$ make functest
...
:tests/misc/roi.test:4: [Failed]
Command (at line 104):
/Users/simon/src/PLAINTEXTACCOUNTING/hledger/.stack-work/install/x86_64-osx/lts-13.0/8.6.3/bin/hledger  -f- roi --inv investment --pnl pnl -b 2017 -e 2018 -Y
Expected stdout: 
+---++------------+------------++---------------+----------+-------------+-----++--------+--------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) | PnL ||    IRR |    TWR |
+===++============+============++===============+==========+=============+=====++========+========+
| 1 || 2017/01/01 | 2017/12/31 ||             0 |      200 |         220 |  20 || 12.73% | 10.00% |
+---++------------+------------++---------------+----------+-------------+-----++--------+--------+
...(1 more)
Got stdout:      
+---++------------+------------++---------------+----------+-------------+-----++--------+--------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) | PnL ||    IRR |    TWR |
+===++============+============++===============+==========+=============+=====++========+========+
| 1 || 2017/01/01 | 2017/12/31 ||             0 |      200 |         220 |  20 || 12.72% | 10.00% |
+---++------------+------------++---------------+----------+-------------+-----++--------+--------+
...(1 more)
:tests/misc/roi.test:6: [Failed]
Command (at line 150):
/Users/simon/src/PLAINTEXTACCOUNTING/hledger/.stack-work/install/x86_64-osx/lts-13.0/8.6.3/bin/hledger  -f- roi --inv investment --pnl pnl -b 2017 -e 2018 -Q
Expected stdout: 
+---++------------+------------++---------------+----------+-------------+-----++---------+---------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) | PnL ||     IRR |     TWR |
+===++============+============++===============+==========+=============+=====++=========+=========+
| 1 || 2017/01/01 | 2017/03/31 ||             0 |      100 |         100 |   0 ||   0.00% |   0.00% |
| 2 || 2017/04/01 | 2017/06/30 ||           100 |        0 |         110 |  10 ||  46.56% | ...(317 more)
Got stdout:      
+---++------------+------------++---------------+----------+-------------+-----++---------+---------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) | PnL ||     IRR |     TWR |
+===++============+============++===============+==========+=============+=====++=========+=========+
| 1 || 2017/01/01 | 2017/03/31 ||             0 |      100 |         100 |   0 ||  -0.00% |   0.00% |
| 2 || 2017/04/01 | 2017/06/30 ||           100 |        0 |         110 |  10 ||  46.56% | ...(317 more)

         Test Cases    Total        
 Passed  425           425          
 Failed  2             2            
 Total   427           427          
functest FAILED
make: *** [functest] Error 1
```

adept:
Fixes #937 

When return on investment is very small, numerical solvers can jump around zero, sometimes producing very small negative numbers. Let's show everything smaller than 0.01% as 0%

### DIFF
--- a/hledger/Hledger/Cli/Commands/Roi.hs
+++ b/hledger/Hledger/Cli/Commands/Roi.hs
@@ -117,14 +117,15 @@ roi CliOpts{rawopts_=rawopts, reportopts_=ropts} j = do
     irr <- internalRateOfReturn showCashFlow prettyTables thisSpan
     twr <- timeWeightedReturn showCashFlow prettyTables investmentsQuery trans thisSpan
     let cashFlowAmt = negate $ sum $ map snd cashFlow
+    let smallIsZero x = if abs x < 0.01 then 0.0 else x
     return [ showDate spanBegin
            , showDate (addDays (-1) spanEnd)
            , show valueBefore
            , show cashFlowAmt
            , show valueAfter
            , show (valueAfter - (valueBefore + cashFlowAmt))
-           , printf "%0.2f%%" irr
-           , printf "%0.2f%%" twr ]
+           , printf "%0.2f%%" $ smallIsZero irr
+           , printf "%0.2f%%" $ smallIsZero twr ]
 
   let table = Table 
               (Tbl.Group NoLine (map (Header . show) (take (length tableBody) [1..]))) 
--- a/tests/misc/roi.test
+++ b/tests/misc/roi.test
@@ -147,39 +147,38 @@ hledger -f- roi --inv investment --pnl pnl -b 2017 -e 2018 -Y
 >>>=0
 
 # 6. Check that deposits/withdrawals and profit/loss are reflected in the right periods
-# XXX https://github.com/simonmichael/hledger/issues/937, "-0.00%"
-# hledger -f- roi --inv investment --pnl pnl -b 2017 -e 2018 -Q
-# <<<
-# 2017-01-01 investment
-#    assets:cash  -$100
-#    investment
+hledger -f- roi --inv investment --pnl pnl -b 2017 -e 2018 -Q
+<<<
+2017-01-01 investment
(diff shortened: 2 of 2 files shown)


adept:
The number in the test would definitely be stable, so we can just update the test. On the other hand, code could end up producing "-0.00%" for one time period and "0.00%" for another time period in the same report, which is a bit unclean.

On yet another hand, given that finding IRR boils down to solving a system of equations using numeric approximation method, I feel that tweaking answer further when it oscillates about zero is not that bad. What's your call?

simonmichael:
PS I think the immediate #937 issue was already fixed by updating the test, so I guess we are considering this one as a further improvement to the functionality.

adept:
I would be very much in favor of using abs(), as numeric solutions around zero would be inherently unstable and we could easily end up with small negative numbers for periods where there was no growth.

simonmichael:
I will still merge whatever you think best, but returning to this hard-coded 0.01 issue. How well will that work with commodities where 0.01 is quite a significant amount, like bitcoin ? Would something sensitive to the commodity's display precision work better, like isZeroAmount ?

simonmichael:
But maybe this only happens when displaying a percentage, so it's not an issue.

adept:
Yeah, this is percentage value, displayed with two decimal digits, which is why i went for 0.01

Issue #945, opened 2019-01-10
Shell command line completion for hledger


schoettl:
Shell completion of the command line, triggered by the tab key, would be nice.

I started to write one for Bash: https://asciinema.org/a/PdV2PzIU9oDQg1K5FjAX9n3vL

It completes:

- commands and generic options
- comand-specific options
- account names (but not yet for files named by `--file`)
- filenames for options that take a filename as argument

The Bash completion script is generated (GNU make) by parsing output of `hledger`, `hledger -h`, and `hledger <cmd> -h`. The script also uses `hledger accounts` for account name completion. I propose that the Makefile is *not* run at every built but rather manually when the CLI changes.

Can I provide a pull request when it's ready? What is a good place to put the scripts in? Folder `shell-completion`? Where does the code go to install the completion script on a Linux system? The main Makefile?

But first, I have to get it ready for deployment. Currently it only runs in a specific directory and requires some external files.

simonmichael:
> I think it's not too bad to maintain as long as the usage output of hledger does not change fundamentally. Because the important parts of the completion script are auto-generated with simple rules.

It doesn't change often, but it has changed and will likely change again.. 

I find command line handling pretty complicated, with all the permutations of common options, command-specific options, builtin commands and addon commands.

> Do you know some platform package maintainers? E.g. Felix Yan for Arch Linux. I guess he creates the installation scripts and can also install the completion script. It should be one line like of code.
> 

You should be able to find them by following the platform links on the download page.

simonmichael:
We now provide [bash shell completions support](https://github.com/simonmichael/hledger/tree/master/shell-completion thanks to @schoettl . Future work for anyone interested: extend to other popular shells, make more discoverable to hledger users, help packagers to add it to their packages.

schoettl:
I plan to work on your last to points during the next time...

simonmichael:
In case it's useful in future: [cmdargs](https://hackage.haskell.org/package/cmdargs-0.10.15/docs/System-Console-CmdArgs-Explicit.html) has the ability to generate bash completion info from our flag definitions.

simonmichael:
Yes I imagine using it only once and saving the completions, not running hledger every time. I expect we would need to add support for it to expose this feature.

PR #946, opened 2019-01-10
Shell completion

schoettl:
Related to #945 

Comments?

Please don't merge yet, I want to enhance the completion behavior with `--files=`.


schoettl:
Completion for Bash is finished, I think!

I can only provide completion for Bash. Scripts for other Shells (e.g. Fish, zsh) are welcome. I'll add a section in the README on how to use this "framework".

simonmichael:
Nice @schoettl. I just added the (WIP) suffix to remind me not to merge, feel free to remove it now (I'd like to know if you have that access).

Would anyone like to do a little testing before I merge this ?

schoettl:
Another bugfix for filename completion. I think, now it's functioning really good!

Some testers? ... before I remove the WIP in the title...

I'm not completely satisfied with it because I'm not sure if I fully understand Bash's completion system (`complete`, `compgen`, `compopt`).

simonmichael:
Nice. Why version 1.1 ? You probably want to follow http://hledger.org/manual.html#queries

Let me know when you feel ready to have this merged.

schoettl:
Oh, the search engine took me to the old doc. Fixed.

I also added `amt:<=`, `status:!`, and so on – these are not useful for completion but as documentation when the user presses tab twice to show possible completions. The disadvantage is that the number of suggestions is getting unwieldy. But it might be useful.

I'd like to wait a few more days before the merger. Maybe I come up with more ideas :)

schoettl:
TODO for me: install current latest version (AUR, *-git?) of hledger and run make again to get latest option and command completions.

schoettl:
Ready to merge.

simonmichael:
This creates the shell-completions/ subdirectory, which contains a [README](https://github.com/simonmichael/hledger/blob/7644059d550a50fdc666c4a0e48bb2d445a848d9/shell-completion/README.md) for getting started. So far so good. How can we make this more discoverable and get folks likely to use it, using it ? Where could we mention it in the docs ?

simonmichael:
Perhaps a new page in the Cookbook, & linked in the wiki sidebar ?

simonmichael:
And/or.. a new section in the hledger manual ?

schoettl:
I'd rather try to get package maintainers to install it together with hledger. Then, people who use auto completion will quickly find out. (Do we have a changelog?)

I'm not sure if it's worth a section in the `hledger` man page. I'm not aware of other programs that have a description of their shell completion in the man page.

A cookbook entry would be useful if the completion is not installed with the package manager. If it is installed automatically, I would just reference to the README to not duplicate content that is only relevant for developers. What do you think?

simonmichael:
I think it's hard/slow to get new users; I would advertise everywhere that makes sense. And I would guess many/most hledger users aren't installing via system package. For myself trying this out, I'd like to follow some kind of link from the docs accessible from http://hledger.org .  But I'm going to merge the work so far, and we can follow up. Thanks for working on it!

PR #947, opened 2019-01-11
lib: fix parsing of status/code/desc/comments in period txn

### LINKED ISSUE #942: periodic transaction with date, comment line and amountless posting fails to parse
This entry is not parsed correctly:
```
~ 2019/04/01
    ;
    a  1
    b
```
```
$ hledger print --forecast
...
could not balance this transaction - can't have more than one real posting with no amount (remember to put 2 or more spaces before amounts)
2019/04/01
    ; recur: 2019/04/01
    ;
    a               1
    b

```
To reproduce: the date should be within the six months from today, there should be no space after the date, the second line should be a comment line, there should be an amountless posting.

adept:
I dont know why periodic txn parser was rather different from regular txn parser in the part that parses code/status/desc/comments, but if these differences are eliminated, then it fixes #942 .


simonmichael:
I don't either. Thanks!

PR #948, opened 2019-01-15
allow account comments, tags, account types declared with tags

simonmichael:
Account directives can now have comments and tags. Account types can be declared with a `type:` tag, and the account type syntax introduced in hledger 1.12 is deprecated (I guess ?). Docs have been improved. Account tags are not used for anything else (eg queries) yet.

Previous discussion: #877

### DIFF
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -210,15 +210,9 @@ sortAccountTreeByAmount normalsign a
 
 -- | Add extra info for this account derived from the Journal's
 -- account directives, if any (comment, tags, declaration order..).
--- Currently only sets declaration order.
--- Expects that this account is among the Journal's jdeclaredaccounts
--- (otherwise sets declaration order to 0).
 accountSetDeclarationInfo :: Journal -> Account -> Account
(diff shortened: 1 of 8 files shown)


simonmichael:
FWIW, this allows account type declarations to be made without breaking compatibility with hledger < 1.12, ie:
```journal
account assets
       ; type:Asset
```

simonmichael:
I'll merge this soon if there are no objections. To review: 

Previously we had this syntax for declaring account types, and comments were supported as indented lines below:
```
account  assets    A
  ; a comment
```

Now there's an alternate syntax for declaring account type: a tag named `type:`, with values like `A` or `Asset`:
```
account assets
  ; type:A
```

And same-line comments are now supported:
```
account assets  ; type:A
```

For good or ill, this continues the hledger tradition of (ab)using the "comment" field to hold semantically significant information. 

We could also consider using Ledger-style subdirectives, perhaps making them equivalent to comments, or tags. Eg (not yet supported):
```
account assets
  type: A   ; or type A
```

The old syntax is very fast to write. But, it may not be worth keeping. The current plan is to remove it in the release after next.

hpdeifel:
Sorry for not reviewing this. FTR: I have now looked at it and already
find it much nicer than the old syntax.

PR #954, opened 2019-01-20
Fix options in markdown doc, to prevent -- to become – in HTML

schoettl:
I quoted `-f`/`--flag`, to prevent it from becoming –flag in HTML doc.

I haven't compiled the doc to nroff, info, ... but it should work, right?

### DIFF
--- a/hledger-ui/hledger-ui.m4.md
+++ b/hledger-ui/hledger-ui.m4.md
@@ -46,8 +46,8 @@ It is easier than hledger's command-line interface, and
 sometimes quicker and more convenient than the web interface.
 
 Note hledger-ui has some different defaults:
-- it generates rule-based transactions and postings by default (--forecast and --auto are always on). 
-- it hides transactions dated in the future by default (change this with --future or the F key).
+- it generates rule-based transactions and postings by default (`--forecast` and `--auto` are always on). 
+- it hides transactions dated in the future by default (change this with `--future` or the F key).
 Experimental.
 
 Like hledger, it reads _files_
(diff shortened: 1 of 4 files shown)


simonmichael:
@schoettl thanks, I overlooked this. I think a better fix would be to turn off pandoc's "smart typography", which I always try to do. That way we won't have to worry about this.

schoettl:
Hm, most options in the text are already `quoted`, I think. Shouldn't it be consistent?

And why

`--qote`, `-q`    here at this option description – but not `--inline`?

simonmichael:
Perhaps we should be more consistent about quoting options, we can discuss that too. But I'll turn off the (not-so-)smart typography in any case, as it breaks searching and copy/pasting from examples. I'm working on this, it is having some strange effect in the nroff manuals.

simonmichael:
PS I don't understand what "--qote, -q here at this option description – but not --inline?" is referring to

schoettl:
@simonmichael what I mean is: If this was the man page of hledger:

usage: hledger [options]
options:
  `-q`   description of the option
  `-v`   verbose, cannot be used with -q

I think, if we quote the options in the first column describing the option, it should also be quoted when used inline of the description text.

schoettl:
Right, it doesn't look bad per se, but I prefer consistency. Consider the sections Depth limiting and Pivoting – here the options even have parameters and this would look --like THIS if not quoted. Here I clearly, prefer quoting.

simonmichael:
I think I'll close this PR. We can discuss the larger issue (clarify the formatting policy in docs for flags, etc.) more on mail list, irc, or (if specific) a new issue. Thanks!

Issue #955, opened 2019-01-21
Add an option to flip rows and columns in multi-column reports


erikryb:
Multi-column reports will often become too wide when including many periods. It would therefore be nice to be able to flip rows and columns, so that columns become accounts and rows become intervals.

simonmichael:
That could work, if you limit it to show few accounts. Upvoted.

simonmichael:
A workaround I just discovered: generate a report as CSV (hledger is -o is.csv), open it in a spreadsheet, use the spreadsheet's transpose table feature. (In Numbers.app you can do this in the CSV importer's advanced settings, or after import.)

PR #957, opened 2019-01-24
cli: added --transpose to "balance" command

adept:
This implements balance --transpose for text and csv outputs, potentially closing #955 

People braver than me could implement it for HTML as well :)


simonmichael:
Thanks!

simonmichael:
Still to do, if anyone has the urge: 

- balance command's HTML output
- balancesheet, incomestatement, cashflow

adept:
note: balancesheet, incomestatement, cashflow are all implemented via compoundbalancecommand, however transposing its final table in a naive way does not work as it has rather complex internal structure.

erikryb:
Great! I have updated issue #955 to show the progress.

PR #959, opened 2019-01-25
lib: make --auto work on --forecast transactions

### LINKED ISSUE #953: --auto and --forecast should work together
Transaction modifier rules (--auto) should add postings to forecasted transactions generated by --forecast, but they don't.

adept:
This fixes #953 by applying transaction modifiers to generated forecast transactions, if any


adept:
Tests seem to be broken in master at the moment

Issue #307, opened 2015-12-15
register --invert


rstacruz:
Feature request: ledger's `--inverse` option will invert (ie, multiply by -1) values. Here's a use case where it's useful.

Say I want to see what happens in my savings account. That's easily done with `hledger reg`.

```
$ hledger reg Assets:Savings

2015/11/02 ATM withdrawal      Assets:Savings    $-200.00    $1200.00
2015/11/04 ATM withdrawal      Assets:Savings    $-150.00    $1050.00
2015/11/05 Bills payment       Assets:Savings     $-50.00    $1000.00
2015/11/05 Fund transfer       Assets:Savings    $-500.00     $500.00
```

But this isn't that useful because it shows `Assets:Savings` as the account. I want to see where it transfers to. That's easily done using `reg --related`.

```
$ hledger reg Assets:Savings --related

2015/11/02 ATM withdrawal      Assets:Cash        $200.00    $-1200.00
2015/11/04 ATM withdrawal      Assets:Cash        $150.00    $-1050.00
2015/11/05 Bills payment       Expenses:Phone      $50.00    $-1000.00
2015/11/05 Fund transfer       Assets:Checking    $500.00     $-500.00
```

Great! Unfortunately, this makes the running balance negative. In this case, `--inverse` will display it like I expect it.

(This isn't a perfect solution to that problem. A transaction with multiple postings will look strange, but it's a start... what would really make sense is a `--related` that will simply show the accounts where the money was drawn from.)
##

simonmichael:
Does the register view in hledger-web or hledger-ui give what you're looking for ?

rstacruz:
yes, exactly!

rstacruz:
any clues on how i can see it like hledger-web from `hledger`?

simonmichael:
I'm not sure it's possible actually. I think we should add hledger-web/hledger-ui-like account-centric register command to the CLI. I suppose "register" should stay as it is, so maybe "aregister" ?

schoettl:
I recently posted exactly the same question in the hledger IRC. Why don't you like the idea of `--inverse`, Simon?

`hledger-ui --register=.` gives good output for me, but I cannot use the UI output for the report because it's locked in the pager.

Personally, I think it's redundant to add a new subcommand `aregister`, just to invert the amount. It's not an "[orthogonal](https://fishshell.com/docs/current/design.html#ortho)" feature and the subcommand name does not tell much on its purpose or on how it's different to `register`.

[...]

PR #970, opened 2019-02-09
Support register --invert option to negate amounts

schoettl:
[Related to #307.]

Open questions:

- Can we introduce a shorthand for `--invert`? I'd prefer `-v` like in grep. `-i` often stands for in-place or ignore. `-v` would be available in balance and register.

- Do we better apply the `amt:` filter before or after negating the amounts? In balance, `--invert` seems to be applied only the the printed output, so after all filters. I think we should keep the meaning of `--invert` to only apply to the output.

And I don't understand the line
```
newtype MixedAmount = Mixed [Amount] deriving ...
```
  Why are there multiple amounts possible for one posting and what is `Mixed`?

### DIFF
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -116,7 +116,8 @@ matchedPostingsBeforeAndDuring opts q j (DateSpan mstart mend) =
   where
     beforestartq = dbg1 "beforestartq" $ dateqtype $ DateSpan Nothing mstart
     beforeandduringps =
-      dbg1 "ps4" $ sortBy (comparing sortdate) $                               -- sort postings by date or date2
+      dbg1 "ps5" $ sortBy (comparing sortdate) $                               -- sort postings by date or date2
+      dbg1 "ps4" $ (if invert_ opts then map negatePostingAmount else id) $    -- with --invert, invert amounts
       dbg1 "ps3" $ map (filterPostingAmount symq) $                            -- remove amount parts which the query's cur: terms would exclude
       dbg1 "ps2" $ (if related_ opts then concatMap relatedPostings else id) $ -- with -r, replace each with its sibling postings
       dbg1 "ps1" $ filter (beforeandduringq `matchesPosting`) $                -- filter postings by the query, with no start date or depth limit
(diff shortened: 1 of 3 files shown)


schoettl:
And a question: can I refactor these lines
```
,flagNone ["related","r"] (\opts -> setboolopt "related" opts) "show postings' siblings instead"
```
to omit the lambda and write `(setboolopt "related")` instead?

I would open a different pull request for it.

simonmichael:
> Can we introduce a shorthand for --invert? I'd prefer -v like in grep. -i often stands for in-place or ignore. -v would be available in balance and register.
> 
Maybe later, if this is used a lot ? Short flags are valuable.

> Do we better apply the amt: filter before or after negating the amounts? In balance, --invert seems to be applied only the the printed output, so after all filters. I think we should keep the meaning of --invert to only apply to the output.
> 
I agree, it's a final-stage output transformation.
> 
> And I don't understand the line
> 
> newtype MixedAmount = Mixed [Amount] deriving ...
> Why are there multiple amounts possible for one posting and what is Mixed?
> 
Mixed is the constructor for a MixedAmount value. A MixedAmount represents a multi-commodity amount. You're right that postings currently may only post a single-commodity amount, but it was convenient to use a MixedAmount, I forget why. And, we have an open PR that aims to change that.

> And a question: can I refactor these lines
> 
> ,flagNone ["related","r"] (\opts -> setboolopt "related" opts) "show postings' siblings instead"
> to omit the lambda and write (setboolopt "related") instead?
> 
> I would open a different pull request for it.
> 
Sure, good idea. (Sometimes things are written out in a longer way because consistent patterns are easier to work with.)

simonmichael:
That does look useful, when paired with --related, eg: `hledger reg checking --related --invert`.

We haven't mentioned --invert in hledger/Hledger/Cli/Commands/Register.md, but perhaps appearing in the flags list is enough, it's a simple feature.

negateMixedAmount isn't needed, you can just do `pamount = negate $ pamount p`.

schoettl:
The `Register.txt` file is generated from the Markdown file, right? So I should not change it?

simonmichael:
Correct. CONTRIBUTING.md should be up to date on this topic.

simonmichael:
> Correct. CONTRIBUTING.md should be up to date on this topic.

Or maybe not. Commands/README.md is. I'm not sure if it should be merged into CONTRIBUTING.md.

simonmichael:
Thanks!

schoettl:
Thank you, too, for commenting and reviewing!

PR #971, opened 2019-02-09
Reduce simple lambda expressions

schoettl:
As discussed in #970


simonmichael:
Nice cleanup, thanks!

PR #972, opened 2019-02-10
Shell completion update

schoettl:
- Fixed a few minor bugs
- Updated completions (added new options like --invert)
- Added ui, web, api as subcommands, to get option completions for these commands, too. That is, `hledger-ui` does not provide completions but `hledger ui` does.


simonmichael:
I didn't quite understand about hledger-ui vs hledger ui. Both of those get used quite a bit, can we have completions work for both ?

schoettl:
I did not want to do much extra work for the three tools. The completion functions would be a lot simpler (because there are no subcommands). But I thought, I'd have to write three more Makefiles and scripts, m4 macros, installation instruction...

But maybe it's much simpler. Maybe I can re-use the existing completion function and just call it with specific parameters. I'll try that.

simonmichael:
Hopefully there'll be a simpler low-maintenance way. Should I merge the work so far ?

schoettl:
> Hopefully there'll be a simpler low-maintenance way. Should I merge the work so far ?

Not yet, please. I added "WIP" in the title.

schoettl:
Maybe not perfect, because
- all "General options" of hledger are provided as completions for hledger-{ui,web,api}.
- all completions for QUERY are provided for hledger-{ui,web,api}. Because I decided to provide them  for *all* subcommands (as a simplification).

But still pretty nice, I think :)

schoettl:
I updated the wiki and the README with a new ascii demonstration video. And I fixed a bug (long options must start with a letter). I think you can merge now.

simonmichael:
Thanks!

PR #974, opened 2019-02-17
transaction balancing cleanup + inclusive balance assertions

simonmichael:
Reworking transaction balancing code to make it easier to understand, and adding inclusive balance assertions.


simonmichael:
I'll merge this shortly I think. Any last-minute review welcome.

PR #976, opened 2019-02-20
[WIP] add always-convert subdirective of commodity

ninewise:
After discussion in IRC, I started fiddling around with the concept of an "always-convert" subdirective for the commodity directive. It's still far from finished, but I'm opening this PR as a place for discussion.

## Idea

The `always-convert` concept would be implementing by either:
1. add a new subdirective to a commodity, marking it as always-convert; or
2. declare the `%` commodity as special syntax.

Example use of (1), to divide the costs of some expense (un)evenly:

```journal
commoditty parts
    always-convert

2019/02/20 Divide an amount evenly among two accounts
    expense:food       $30
    account1       -1 part
    account2       -1 part
```

Example use of (2), for the same reason:

```journal
2019/02/20 Divide an amount evenly among two accounts
    expense:food       $30
    account1           -50%
    account2           -50%
```

The main difference between (1) and (2) being that you either (1) don't have to calculate the part everyone gets or (2) can have hledger verify the percentages sum to 100%.

## Current state

- a parser for (1)
- a POC of (1) but without the parsed marked, hardcoding `%` as the always-converted commodity.

```journal
2019-02-20 Test
  p1         3%
  p2        97%
  other   -€100

2019/02/20 Divide an amount evenly among two accounts
    expense:food       $30
    account1            -1% ; using 1% here looks weird, but is valid for now
    account2            -1%
```

Questions I have:

- would we prefer (1) or (2)?
- would there be a way to access the complete commodity (not just the symbol) from e.g. `Hledger.Data.Transaction.inferBalancingPrices` without replacing them about everywhere in the library?

### DIFF
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -195,8 +195,9 @@ instance NFData DigitGroupStyle
 type CommoditySymbol = Text
 
 data Commodity = Commodity {
-  csymbol :: CommoditySymbol,
-  cformat :: Maybe AmountStyle
+  csymbol        :: CommoditySymbol,
+  cformat        :: Maybe AmountStyle,
+  calwaysconvert :: Bool
   } deriving (Show,Eq,Data,Generic) --,Ord,Typeable,Data,Generic)
 
 instance NFData Commodity
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -958,15 +958,17 @@ canonicalStyleFrom ss@(first:_) = first {asprecision = prec, asdecimalpoint = md
 --   case ps of (MarketPrice{mpamount=a}:_) -> Just a
 --              _ -> Nothing
 
--- | Convert all this journal's amounts to cost by applying their prices, if any.
-journalConvertAmountsToCost :: Journal -> Journal
-journalConvertAmountsToCost j@Journal{jtxns=ts} = j{jtxns=map fixtransaction ts}
+-- | Convert all this journal's amounts, or just the ones marked as 'always-convert', to cost by applying their prices, if any.
+journalConvertAmountsToCost :: Bool -> Journal -> Journal
+journalConvertAmountsToCost all j@Journal{jtxns=ts, jcommodities=jcs} = j{jtxns=map fixtransaction ts}
     where
       -- similar to journalApplyCommodityStyles
       fixtransaction t@Transaction{tpostings=ps} = t{tpostings=map fixposting ps}
       fixposting p@Posting{pamount=a} = p{pamount=fixmixedamount a}
       fixmixedamount (Mixed as) = Mixed $ map fixamount as
-      fixamount = styleAmount styles . costOfAmount
+      fixamount a@Amount{acommodity=acs} | all || maybe False calwaysconvert (M.lookup acs jcs)
+                                           = styleAmount styles $ costOfAmount a
+                                         | otherwise = a
       styles = journalCommodityStyles j
 
 -- -- | Get this journal's unique, display-preference-canonicalised commodities, by symbol.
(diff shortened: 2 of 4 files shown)


simonmichael:
Thanks! It's hard to tell without some experience, but for me 2 is the more intuitive feature; I wouldn't mind trying 1 as well. I think both should be marked experimental, it will take some time to shake out all the implications.

What do you mean by the complete commodity ?

The #974 cleanup will land soon I hope.

simonmichael:
@ninewise you might also want to propose this on the mail list, for more input. Maybe you can add mockups of sample command output just to make clear what this does.

simonmichael:
And/or a documentation draft.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -196,7 +196,8 @@ type CommoditySymbol = Text
 
 data Commodity = Commodity {
   csymbol :: CommoditySymbol,
-  cformat :: Maybe AmountStyle
+  cformat :: Maybe AmountStyle,
+  ciscost :: Bool
```
simonmichael: or calwaysconvert ? Consistent naming is good

hledger-lib/Hledger/Data/Transaction.hs
```diff
@@ -447,6 +447,17 @@ inferBalancingPrices t@Transaction{tpostings=ps} = t{tpostings=ps'}
...
+convertCosts :: Transaction -> Transaction
+convertCosts t@Transaction{tpostings=ps} = t{tpostings=map convertCostsP ps}
+  where
+    convertCostsP p@Posting{pamount=Mixed [a]} = p{pamount=Mixed [convertCostsA a], porigin=Just $ originalPosting p}
+    convertCostsA a
+      | acommodity a /= "%" = a
+      | otherwise = case aprice a of NoPrice -> a
+                                     UnitPrice a' -> a'{aquantity=aquantity a * aquantity a'}
+                                     TotalPrice a' -> a'
+
```
simonmichael: Possibly related: costOfAmount, costOfMixedAmount, journalConvertAmountsToCost.

ninewise:
> What do you mean by the complete commodity ?

The complete `data Commodity` instead of just `CommoditySymbol`. So I could access the `calwaysconvert`/`ciscost` bool.

> Possibly related: costOfAmount, costOfMixedAmount, journalConvertAmountsToCost.

I did check those, and they are indeed related. Problem is those are only called when `-B` is passed. Another (and possibly better) approach would be to always call journalConvertAmountsToCost, but passing a "convert all amounts to cost" or "convert only always-convert amounts to cost" flag. That'd probably be cleaner than my current attempt.

ninewise:
> you might also want to propose this on the mail list, for more input.

I'm afraid I don't subscribe to google groups; I'll work out some more documentation/samples and collect them here, though.

simonmichael:
> The complete data Commodity instead of just CommoditySymbol. So I could access the calwaysconvert/ciscost bool.

Oh yes I see. That's stored in the Journal, which was last seen several callers above. That's a bit awkward, alright.

> I did check those, and they are indeed related. Problem is those are only called when -B is passed. Another (and possibly better) approach would be to always call journalConvertAmountsToCost, but passing a "convert all amounts to cost" or "convert only always-convert amounts to cost" flag. That'd probably be cleaner than my current attempt.
> 
No worries, I just wondered if there were code sharing possibilities.

> I'm afraid I don't subscribe to google groups; I'll work out some more documentation/samples and collect them here, though.
> 

Fair enough. Thanks.

ninewise:
I've been using this for a while now, and while it's very useful, I've encountered some problems. The way this is now implemented, the "parts" aren't replaced internally, only for showing them. This is what we want, because I want to put the meaning of dividing in parts in the journal. If the values would be replaced internally, using `hledger add` would put the actual values in the journal, not the parts.

Problem is, checks for balance use the internal values, the parts. I've been encountering some failing balances because of this.

Fixing this, replacing parts with costs for checks but not in the ledger, would require a very large patch on the code, which I consider not worth it. I'm therefore closing this PR. If someone else would like to try minimizing the patch, feel free to take over any code.

PR #977, opened 2019-02-20
web: Re-add 'balanced transaction' validation to add form

### LINKED ISSUE #951: web add form can create unbalanced transactions
Lately unbalanced transactions keep showing up at http://demo.hledger.org , which I assume are coming from the add form there. This should not be possible.

zarybnicky:
Fixes #951.

The main purpose of this PR is just adding a `balanceTransaction Nothing txn`
validation, but I've also taken the liberty of cleaning up the mess I'd made in
add form validation half a year ago.

I've tried to rework the basic algorithm two or three times but I had to leave
it at an `unfoldr` as that was the only version that could account for all
corner cases. So instead I've restructured the rest of the code and added
comments and type annotations in hope of making the code in AddForm.hs more
accessible even to casual readers.

### DIFF
--- a/hledger-web/Hledger/Web/Handler/AddR.hs
+++ b/hledger-web/Hledger/Web/Handler/AddR.hs
@@ -9,11 +9,13 @@ module Hledger.Web.Handler.AddR
   , postAddR
   ) where
 
+import qualified Data.Text as T
+import Text.Blaze.Html (preEscapedToHtml)
+
 import Hledger
 import Hledger.Cli.Commands.Add (appendToJournalFileOrStdout)
 import Hledger.Web.Import
 import Hledger.Web.Widget.AddForm (addForm)
-import Hledger.Web.Widget.Common (fromFormSuccess)
 
 getAddR :: Handler ()
 getAddR = postAddR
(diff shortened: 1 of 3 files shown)


simonmichael:
That works much better, thanks. Would you mind force-pushing a rebased version to get rid of the merge commit, and I'll merge. For the future, I wonder if there are some yesod form helpers we should be using ?

zarybnicky:
Yes, merging wasn't what I'd intended - I tried out GitHub's "update branch" button...

As for form helpers... Yesod-form's API isn't a very fluent one, so anytime you need something more advanced you need to work directly with the internals. It doesn't make for pretty code, but that's been my experience with any yesod-form code I've had to work with. The API works well for 80% of web forms, but even [simple cases](https://github.com/yesodweb/yesod-cookbook/blob/master/cookbook/Password-Confirm-Field.md) can start getting hairy.

I don't see any obvious pieces of code that could be simplified in `addForm`. Generating the list of postings would be possible if it was a static list, but it can get extended with JavaScript, so I don't think that would work. Having a better data structure in `validatePostings` would help, but the one I used in the end works and there's minimum of boilerplate for converting between structures (and even that's a lot but it's the best that I could do...).

simonmichael:
I see, good to know. Thanks!

PR #979, opened 2019-02-28
cli, web, ui: Replace withJournalDo* variants

zarybnicky:
This finally removes the duplication between hledger-cli's, hledger-ui's and
hledger-web's entrypoints. I'd wanted to do this a long time ago, but only now
was I annoyed enough by the XXX's around those places to actually do it :)

On another note, this PR conflicts with my other one (#978). I didn't want to
create a PR against another PR, so whichever one you get around to merging
first, I'll just need to rebase the other one (for the three lines of worth of
conflicts, oh well).

There is just one thing I'm not sure about. The original hledger-ui's entrypoint
did some changes to `CLIOpts` that I've pulled out so that they are now sent to
both `withJournalDo` and to `runBrickUI`, which wasn't the case previously as
`runBrickUi` used the unchanged `UIOpts`. The two options affected are `auto_`
and `forecast_`. AFAICT, these pertain just to reading the journal, but the
change might affect a potential `reloadJournalIfChanged`, so I'm mentioning it
here...

### DIFF
--- a/hledger-ui/Hledger/UI/Main.hs
+++ b/hledger-ui/Hledger/UI/Main.hs
@@ -61,28 +61,20 @@ writeChan = BC.writeBChan
 main :: IO ()
 main = do
   opts <- getHledgerUIOpts
+  let copts = cliopts_ opts
+      copts' = copts
+        { inputopts_ = (inputopts_ copts) { auto_ = True }
+        , reportopts_ = (reportopts_ copts) { forecast_ = True }
+        }
+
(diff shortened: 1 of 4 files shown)


simonmichael:
Great, thanks! If you have any time to test that hledger-ui --watch preserves --auto, --forecast, and any startup flags like -R, that would be helpful. I hope to do a release today.

simonmichael:
PS I did that testing, it seems fine.

zarybnicky:
Thanks, I didn't manage to find the time yesterday.

PR #981, opened 2019-03-03
Merge hledger-diff as a cli command

gebner:
This PR merges the external [hledger-diff](https://github.com/gebner/hledger-diff) tool.  See https://github.com/gebner/hledger-diff/issues/8

I hope I've got the `Diff.txt` and `Diff.md` files right.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Diff.md
+++ b/hledger/Hledger/Cli/Commands/Diff.md
@@ -1,0 +1,35 @@
+diff\
+Compares two journal files.  It looks at the transactions of a single
+account and prints out the transactions which are in one journal file but not
+in the other.
+
+This is particularly useful for reconciling existing journals with bank
+statements.  Many banks provide a way to export the transactions between two
+given dates, which can be converted to ledger files using custom scripts or
+read directly as CSV files.  With the diff command you can make sure that these
+transactions from bank match up exactly with the transactions in your ledger
+file, and that the resulting balance is correct.  (One possible concrete
+workflow is to have one ledger file per year and export the transactions for
+the current year, starting on January 1.)
+
+This command compares the postings of a single account (which needs to have the
+same name in both files), and only checks the amount of the postings (not the
+name or the date of the transactions).  Postings are compared (instead of
+transactions) so that you can combine multiple transactions from the bank
+statement in a single transaction in the ledger file.
+
+_FLAGS_
+
+Examples:
+
+```shell
+$ hledger diff assets:bank:giro -f 2014.journal -f bank.journal
+Unmatched transactions in the first journal:
+
+2014/01/01 Opening Balances
+    assets:bank:giro              EUR ...
+    ...
+    equity:opening balances       EUR -...
+
+Unmatched transactions in the second journal:
+```
--- a/hledger/Hledger/Cli/Commands/Diff.txt
+++ b/hledger/Hledger/Cli/Commands/Diff.txt
@@ -1,0 +1,35 @@
+diff
+Compares two journal files. It looks at the transactions of a single
+account and prints out the transactions which are in one journal file
+but not in the other.
+
+This is particularly useful for reconciling existing journals with bank
+statements. Many banks provide a way to export the transactions between
+two given dates, which can be converted to ledger files using custom
+scripts or read directly as CSV files. With the diff command you can
+make sure that these transactions from bank match up exactly with the
+transactions in your ledger file, and that the resulting balance is
+correct. (One possible concrete workflow is to have one ledger file per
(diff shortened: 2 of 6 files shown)


hledger/Hledger/Cli/Commands/Diff.md
```diff
@@ -0,0 +1,21 @@
+diff\
+Compares two journal files.  It looks at the transactions of a single
+account and prints out the transactions which are in one journal file but not
+in the other.  This is particularly useful for reconciling existing journals
+with bank statements.
```
simonmichael: This motivating use case is helpful. Could you add a little more concrete detail here ? Eg do you convert a chunk of CSV between the dates of two bank statements to a journal file, and then diff that with your main file ?

hledger/Hledger/Cli/Commands/Diff.txt
```diff
@@ -0,0 +1,19 @@
+diff
```
simonmichael: ./Shake commandhelp is the intended way to generate this file, maybe you did that already. If there's any differences I can fix.

gebner: Ok, `./Shake commandhelp` fails for me:
```
pandoc -f markdown-smart-tex_math_dollars hledger/Hledger/Cli/Commands/Balancesheetequity.md --lua-filter tools/pandoc-dedent-code-blocks.lua -t plain > hledger/Hledger/Cli/Commands/Balancesheetequity.txt
Error running filter tools/pandoc-dedent-code-blocks.lua:
tools/pandoc-dedent-code-blocks.lua: openBinaryFile: does not exist (No such file or directory)
Error when running Shake build system:
  at src/Development/Shake/Internal/Args.hs:81:69-79:
* Depends on: commandhelp
  at Shake.hs:443:27-43:
* Depends on: hledger/Hledger/Cli/Commands/Balancesheetequity.txt
* Raised the exception:
user error (Development.Shake.cmd, system command failed
Command: /bin/sh -c 'pandoc -f markdown-smart-tex_math_dollars hledger/Hledger/Cli/Commands/Balancesheetequity.md --lua-filter tools/pandoc-dedent-code-blocks.lua -t plain > hledger/Hledger/Cli/Commands/Balancesheetequity.txt'
Exit code: 83
Stderr:
Error running filter tools/pandoc-dedent-code-blocks.lua:
tools/pandoc-dedent-code-blocks.lua: openBinaryFile: does not exist (No such file or directory)
)


```

simonmichael:
> Ok, ./Shake commandhelp fails for me:

Thanks! I had forgotten to commit some files.

hledger/Hledger/Cli/Commands/Diff.md
```diff
@@ -1,8 +1,20 @@
...
+converted to ledger files using custom scripts or read directly as CSV files.
+With the diff command you can make sure that these transactions from the bank
+statement match up exactly with the transactions in your ledger file, and that
+the resulting balance is correct.
+
+This command compares the postings of a single account (which needs to have the
+same name in both files), and only checks the amount of the postings (not the
+name or the date of the transactions).  Postings are compared (instead of
+transactions) so that you can combine multiple transactions from the bank
+statement in a single transaction in the ledger file.
```
simonmichael: Useful extra detail, thanks. We might tweak the wording some day to make it more generally applicable - I've not yet personally encountered a bank that exports separate "statements", instead I tend to see CSV between two arbitrary dates within the last N months.

gebner:
Squashed, rebased, and successfully regenerated the `.txt` file.  I also added the comment to the commit message.

I'll also the deprecation notice to the hledger-diff readme when the next hledger version is released.

Side note: after regenerating the command help files, I now have a new file that is not committed to git, `hledger/Hledger/Cli/Commands/README.txt`.

simonmichael:
Lovely! Thanks @gebner. I've fixed the README.txt.

simonmichael:
A few more little tweaks and snags I noticed:
```
$ hledger diff
Specifiy exactly two journal files
```
Fix typo, be just a little more verbose/polite for consistent tone. "Please specify exactly two journal files with -f, eg hledger diff -f FILE1 -f FILE2" or something ?

```
$ hledger diff -h
diff [OPTIONS] [ACCOUNT] -f [JOURNAL1] -f [JOURNAL2]
```
Should be -f JOURNAL1 -f JOURNAL2 (no brackets). Or maybe -f FILE1 -f FILE2 since they might not be in journal format (also consistent with the above).

The [ACCOUNT] indeed is optional, but I think the command doesn't do anything without it so perhaps it shouldn't be ? And maybe I'd put it after the -f options, more consistent with other helps.

PR #987, opened 2019-03-10
Upgrade to easytest 0.3.

joelburget:
Fixes #983

PR #991, opened 2019-03-15
Constrain easytest dependency to fix cabal build

hpdeifel:
This constraints the easytest dependency to <0.3, because hledger and
hledger-lib currently don't build with 0.3.

Of course the better solution would be to fix the build errors, but easytest-0.3
is not even in stackage nightly yet and I just need it to build right now :-)

### DIFF
--- a/hledger-lib/hledger-lib.cabal
+++ b/hledger-lib/hledger-lib.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: e2c444e055dd9ef61f9cbddeeddf69ce3274e6c2eef4b64301b7107144d7d84b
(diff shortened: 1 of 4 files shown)


hpdeifel:
PS: Could you please also make a hackage revision for `hledger-lib` since currently this prevents `hledger-iadd` from being `cabal new-install`able.

Also, would you mind a PR putting the dependency behind a test flag?

simonmichael:
Thanks! I'll use this if #987 doesn't land shortly. I made revisions already.

hpdeifel:
Oh, I didn't even see #987; was just pushing my fix because I'm trying to setup a CI pipeline with hledger-lib HEAD for hledger-iadd.

simonmichael:
I prefer to avoid build flags. I think once this issue is resolved, depending on easytest shouldn't cause much hassle  ?

hpdeifel:
I'm just still trying to keep that dependency tree somewhat in check :-)

PR #993, opened 2019-03-15
CSV parser: balance1 and balance2

### LINKED ISSUE #992: CSV parser favors all things "1" and neglects all things "2"
There are several pieces of code in CSV parser that heavily favor things with "1" in them:

1. The amount with cost (2 EUR @@ 1 USD) can only appear as `amount1`. `Amount2` is computed via `costOfMixedAmount` and will therefore never have cost

2. `balance` generates balance assertion for `account1`. There is no way to generate balance assertion for `account2`

If you want balance assertion on one account and amount with cost on the other one, you are out of luck. If you want to give it a try, here is a gist with sample files: https://gist.github.com/adept/f97058809feb7b23b9a6ee8f2c85b356

I propose to improve the situation, at least a bit. I envision doing one of the following:

1. Renaming `balance` to `balance1` and introducing `balance2`, with obvious semantics
2. Introducing `balance1` and `balance2` in addition to existing `balance`
3. Introducing `balance1` and `balance2` (leaving or renaming `balance`) and also introducing `amount2`, `amount2-in`, `amount2-out`

I am currently in favor of "1", mostly because it does not complicate things.

I do not like "2" that much because having `balance` and `balance1` looks weird.

And "3" introduces too many new moving parts. 

Can all in favor say "aye" ?:)

adept:
I claim that this closes #992

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -563,6 +563,8 @@ journalfieldnames = [
   ,"amount-in"
   ,"amount-out"
   ,"amount"
+  ,"balance1"
+  ,"balance2"
   ,"balance"
   ,"code"
   ,"comment"
@@ -712,7 +714,7 @@ transactionFromCsvRecord sourcepos rules record = t
       ]
     amount1 = case maybeamount of
                 Just a -> a
-                Nothing | balance /= Nothing -> nullmixedamt
+                Nothing | balance1 /= Nothing || balance2 /= Nothing -> nullmixedamt
                 Nothing -> error' $ "amount and balance have no value\n"++showRecord record
     -- convert balancing amount to cost like hledger print, so eg if
     -- amount1 is "10 GBP @@ 15 USD", amount2 will be "-15 USD".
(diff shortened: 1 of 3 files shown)


hledger-lib/Hledger/Read/CsvReader.hs
```diff
@@ -724,14 +726,26 @@ transactionFromCsvRecord sourcepos rules record = t
...
+    balance1     =
+      case (balance', balance1') of
+        (Nothing, Nothing)  -> Nothing
+        (balance, Nothing)  -> balance
+        (Nothing, balance1) -> balance1
+        (balance, balance1) ->
+          error' $    "both balance and balance1 have a value\n"
+                   ++ "    balance: "  ++ show balance ++ "\n"
+                   ++ "    balance1: " ++ show balance1 ++ "\n"
+                   ++ "    record: "     ++ showRecord record
```
simonmichael: I think this allows both balance and balance 1 to be used simultaneously as long as they don't both have a non-empty value in the same record. Ideally perhaps we could be more strict here, failing if both balance and balance1 are seen when parsing the rules.

Issue #367, opened 2016-07-13
ledger-mode setup for hledger needs documenting


simonmichael:

taksuyu:
Since I'm working on getting this working, I don't mind writing something up once I got it working reasonably well.

simonmichael:
FYI:
```elisp
; Simon's ledger-mode tweaks

(custom-set-variables

 ; neutralise some ledgerisms
 '(ledger-binary-path (expand-file-name "~/.local/bin/hledger"))
 '(ledger-mode-should-check-version nil)
 '(ledger-init-file-name " ")

 ; move default amount position right allowing longer account names
 '(ledger-post-amount-alignment-column 64)

 ; disable distracting highlight
 '(ledger-highlight-xact-under-point nil)

 )

; automatically show new transactions from hledger add or hledger-web
(add-hook 'ledger-mode-hook 'auto-revert-tail-mode)

; M-1 to collapse all transactions to one line, M-0 to reset. Useful for quickly scanning
(global-set-key "\M-0" (lambda () (interactive) (set-selective-display (* tab-width 0))))
(global-set-key "\M-1" (lambda () (interactive) (set-selective-display (* tab-width 1))))
(add-hook 'ledger-mode-hook (lambda () (setq tab-width 4)))

; enable some highlighting for CSV rules files
(add-to-list 'auto-mode-alist '("\\.rules$" . conf-mode))

; useful when running reports in a shell buffer
(defun highlight-negative-amounts nil (interactive)
  (highlight-regexp "\\(\\$-\\|-\\$\\)[.,0-9]+" (quote hi-red-b))
)
```

simonmichael:
```elisp
; recognise more file suffixes
(add-to-list 'auto-mode-alist '("\\.\\(h?ledger\\|journal\\|j\\)$" . ledger-mode))

; enable orgstruct minor mode, TAB while on a * comment expands/collapses org node
(add-hook 'ledger-mode-hook 'orgstruct-mode)
```

simonmichael:
https://github.com/narendraj9/dotfiles/tree/master/.emacs.d/packages/rest/hledger-mode should be investigated too, I haven't tried it.

simonmichael:
Other editor modes like https://github.com/ledger/vim-ledger would also be nice to know about.

taksuyu:
I seem to be comfortable using this configuration. I imagine there is more to look into and fixing little things.

```lisp
(setq ledger-binary-path "hledger")
[...]

PR #1002, opened 2019-04-06
add --uncleared option for Ledger compatibility ?

hackeryarn:
@simonmichael is this what you had in mind as the fix for #367?

I am not sure if you want this option documented on the website since it's primarily there to support `ledger-mode`, but I will be happy to add it.


simonmichael:
Thanks! Yes, that's what I was talking about. Do you think it's a worthwhile addition, or not worth the cruft ? If the former, yes we should probably make sure it's documented. Maybe I'd mention that it's for compatibility, something like "equivalent to --unmarked --pending, for Ledger compatibility".

hackeryarn:
I personally don't think it's worth the cruft. There is already a good explanation of how to combine `-UP` to get the same results. Adding two ways to do the same thing may confuse people more than it will help.

I do like the idea of adding the compatibility note to the cli options and text file. 

I would also like to add unit tests for this feature under `rawOptsToReportOpts`. But I am not sure how format the test if the output of the function under test is wrapped in `IO` (e.g. `IO ReportOpts`). Sorry, I am a bit new to Haskell and never used `easytest` before.

simonmichael:
That sounds great. A test would be fine but doesn't seem essential ? I'm not sure what there is there that's likely to break.

hackeryarn:
@simonmichael I updated the text to include a note about compatibility. Please let me know if there is anything else you see missing.

simonmichael:
@hackeryarn sorry if I was unclear. Considering what you said, I agree with you ("I personally don't think it's worth the cruft").

simonmichael:
So I guess we'll close this ? Can reopen if someone finds it worthwhile in future.

hackeryarn:
Sorry, why are we closing this? The of this was to bring the API in line with the ledger API so the Emacs ledger-mode could run all operations. At least I would very much love to see this merged.

simonmichael:
Ok I misunderstood, I thought we both felt this was not worth the added UI cruft and confusion. So you feel adding the `--uncleared` option is worthwhile to work more smoothly with ledger-mode ? Could we improve ledger-mode instead ?

hackeryarn:
That's a good idea! I just checked, and it's only used in one place inside `ledger` mode. I will try to go that route since the same effect can be achieved by combining two flags in `hledger` and `ledger`. 

Let's close this issue for now unless `ledger-mode` pushes back on the PR.

simonmichael:
Great. Feel free to link it here.

PR #1004, opened 2019-04-06
Rename hledger-makeitso to hledger-flow

apauley:
I've finally gotten round to giving `hledger-makeitso` a more serious name (as mentioned on the mailing list).

This PR just changes the name and the link to `hledger-flow`.

I see there is a similar link on https://plaintextaccounting.org/
I'll submit another PR on https://github.com/plaintextaccounting/plaintextaccounting.github.io

Are there any other places that needs updating?


simonmichael:
Actually, this home page content is generated automatically from https://github.com/simonmichael/hledger/wiki/Home - you should change that one as well.

apauley:
Thanks, I've updated it on the wiki as well

PR #1009, opened 2019-04-16
Fix #1008

### LINKED ISSUE #1008: The order of -M and -p options can produce different reports
Entering this issue from the discussion list so it doesn't get lost:
https://groups.google.com/d/topic/hledger/ehFfd6nQ1y4/discussion

On the forum, Simon states that "-p 2019 -M, -M -p 2019,  -p 'monthly in 2019' should all be equivalent"

These 2 examples show the different outputs when the order of -M and -p are reversed.

**Example 1: -p followed by -M**
```
$ hleger b fuel -p 2019 -M -A
             $460.49  expenses:auto:fuel
--------------------
             $460.49
```

**Example 2: -M followed by -p**
```
$ hledger b fuel -M -p 2019 -A
Balance changes in 2019q1:

                    ||     Jan      Feb      Mar  Average
====================++====================================
 expenses:auto:fuel || $164.37  $178.24  $117.88  $153.50
--------------------++------------------------------------
                    || $164.37  $178.24  $117.88  $153.50
```

schoettl:
This fixes #1008 

@simonmichael , I can remove the last commit (argument "undefined"). But I think it make sense and it is good coding style.

```
parsePeriodExpr undefined (stripquotes $ T.pack v) -- reference date does not affect the interval
-- before, undefined was nulldate (= `fromGregorian 0 1 1`)
```
Consider this case: due to a bug, the first arg *does* affect the interval. Then the program crashes (without a usefull error message) but at least we know that there is a bug. If we pass `nulldate` instead, the bug might not be obvious but produces wrong reports.

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -274,19 +274,28 @@ endDatesFromRawOpts d = catMaybes . map (enddatefromrawopt d)
 
 -- | Get the report interval, if any, specified by the last of -p/--period,
 -- -D/--daily, -W/--weekly, -M/--monthly etc. options.
+-- An interval from --period counts only if it is explicitly defined.
 intervalFromRawOpts :: RawOpts -> Interval
 intervalFromRawOpts = lastDef NoInterval . catMaybes . map intervalfromrawopt
   where
     intervalfromrawopt (n,v)
       | n == "period" =
-          either (\e -> usageError $ "could not parse period option: "++customErrorBundlePretty e) (Just . fst) $
-          parsePeriodExpr nulldate (stripquotes $ T.pack v) -- reference date does not affect the interval
+          either
+            (\e -> usageError $ "could not parse period option: "++customErrorBundlePretty e)
+            extractIntervalOrNothing $
+            parsePeriodExpr nulldate (stripquotes $ T.pack v) -- reference date does not affect the interval
       | n == "daily"     = Just $ Days 1
       | n == "weekly"    = Just $ Weeks 1
       | n == "monthly"   = Just $ Months 1
       | n == "quarterly" = Just $ Quarters 1
       | n == "yearly"    = Just $ Years 1
       | otherwise = Nothing
+
+-- | Extract the interval from the parsed -p/--period expression.
+-- Return Nothing if an interval is not explicitly defined.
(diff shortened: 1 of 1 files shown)


schoettl:
Rebase on current master

simonmichael:
Sorry @schoettl, working through backlog here. Thanks for the pull request. 

I would feel more confident that about this if we had some tests. What about eg tests/cli/date-options.test.

simonmichael:
I don't fully understand the bug/undefined issue yet. I agree that an error is preferable to silent misbehaviour, but if we can't fix more thoroughly let's use `error "some sensible message"` instead of `undefined`.

simonmichael:
Related, we should mention this somewhere in the date option docs starting around https://hledger.org/manual.html#report-start-end-date . (And those docs are pretty complicated, cleanup needed.)

schoettl:
Hi Simon, I'll look into it.

I'm just working on some CLI tests. Great testing infrastructure!

Can you help me with this questions?

- What is the second and the third `>>>` per test? Is the third one the exit status?
- How can I run the tests?

simonmichael:
Great. These docs should make things clear: https://hledger.org/CONTRIBUTING.html#tests, https://github.com/simonmichael/shelltestrunner#readme . https://github.com/simonmichael/shelltestrunner#test-formats describes the three formats, we prefer format 3 nowadays. You'll see some examples of it under tests/

schoettl:
Wow, your shelltest tool is awesome. Such a neat test syntax and so easy to use.

I added tests for the report interval CLI and the bug #1008 .

The undefined thing was not a bug fix nor was it related to the #1008 . It was only refactoring. Currently, passing `nulldate` or `undefined` to the function makes no difference.
Currently, the `undefined` argument is not used in combination with the second argument.

But if some day the function is changed and uses the argument, then it should raise an error instead of silently using `nulldate` (which is just a normal date) and producing a wrong report.

schoettl:
I just read the documentation:
https://hledger.org/manual.html#report-start-end-date
and the the sections Report intervals and Period expressions.

I think it is pretty good and clear. It does not describe the behavior of #1008 but I would rather not make the documentation more comprehensive (and complicated) at this point. Because I think, the new behavior is what one would expect anyway i.e. good usability.

simonmichael:
Thanks! Ledger's regression tests gave me the idea.
More docs added at https://github.com/simonmichael/hledger/blob/master/tests/README.md.

PR #1011, opened 2019-04-19
Bugfix, CLI tests and doc for date options (begin/end of time period)

schoettl:
Date options for begin and end of time period have problems (see below).

This PR provides tests and aims to fix the problems.

### DIFF
--- a/doc/common.m4
+++ b/doc/common.m4
@@ -138,7 +138,7 @@ m4_define({{_reportingoptions_}}, {{
 : multiperiod/multicolumn report by year
 
 `-p --period=PERIODEXP`
-: set start date, end date, and/or reporting interval all at once using [period expressions](manual.html#period-expressions) syntax (overrides the flags above)
+: set start date, end date, and/or reporting interval all at once using [period expressions](manual.html#period-expressions) syntax
 
 `--date2`
 : match the secondary date instead (see command help for other effects)
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -232,19 +232,19 @@ balancetypeopt rawopts =
 -- -p/--period options, using the given date to interpret relative date expressions.
 periodFromRawOpts :: Day -> RawOpts -> Period
 periodFromRawOpts d rawopts =
-  case (mearliestb, mlateste) of
+  case (mlastb, mlaste) of
     (Nothing, Nothing) -> PeriodAll
     (Just b, Nothing)  -> PeriodFrom b
     (Nothing, Just e)  -> PeriodTo e
     (Just b, Just e)   -> simplifyPeriod $
                           PeriodBetween b e
   where
-    mearliestb = case beginDatesFromRawOpts d rawopts of
+    mlastb = case beginDatesFromRawOpts d rawopts of
                    [] -> Nothing
-                   bs -> Just $ minimum bs
-    mlateste   = case endDatesFromRawOpts d rawopts of
+                   bs -> Just $ last bs
+    mlaste = case endDatesFromRawOpts d rawopts of
                    [] -> Nothing
-                   es -> Just $ maximum es
+                   es -> Just $ last es
 
 -- Get all begin dates specified by -b/--begin or -p/--period options, in order,
 -- using the given date to interpret relative date expressions.
--- a/hledger/Hledger/Cli/CliOptions.hs
+++ b/hledger/Hledger/Cli/CliOptions.hs
@@ -141,7 +141,7 @@ reportflags = [
  ,flagNone ["monthly","M"]   (setboolopt "monthly") "multiperiod/multicolumn report by month"
  ,flagNone ["quarterly","Q"] (setboolopt "quarterly") "multiperiod/multicolumn report by quarter"
  ,flagNone ["yearly","Y"]    (setboolopt "yearly") "multiperiod/multicolumn report by year"
- ,flagReq  ["period","p"]    (\s opts -> Right $ setopt "period" s opts) "PERIODEXP" "set start date, end date, and/or report interval all at once (overrides the flags above)"
+ ,flagReq  ["period","p"]    (\s opts -> Right $ setopt "period" s opts) "PERIODEXP" "set start date, end date, and/or report interval all at once"
  ,flagNone ["date2"]         (setboolopt "date2") "match the secondary date instead. See command help for other effects. (--effective, --aux-date also accepted)"  -- see also hiddenflags
 
   -- status/realness/depth/zero filters
--- a/hledger/Hledger/Cli/Main.hs
+++ b/hledger/Hledger/Cli/Main.hs
@@ -43,7 +43,7 @@ module Hledger.Cli.Main where
 import Data.Char (isDigit)
 import Data.List
 import Safe
-import System.Console.CmdArgs.Explicit as C
+import qualified System.Console.CmdArgs.Explicit as C
 import System.Environment
 import System.Exit
 import System.FilePath
@@ -204,7 +204,7 @@ argsToCliOpts :: [String] -> [String] -> IO CliOpts
 argsToCliOpts args addons = do
   let
     args'        = moveFlagsAfterCommand $ replaceNumericFlags args
-    cmdargsopts  = either usageError id $ process (mainmode addons) args'
+    cmdargsopts  = either usageError id $ C.process (mainmode addons) args'
     cmdargsopts' = decodeRawOpts cmdargsopts
   rawOptsToCliOpts cmdargsopts'
 
--- a/hledger/hledger_options.m4.md
+++ b/hledger/hledger_options.m4.md
@@ -234,8 +234,18 @@ You can specify a start and/or end date using
 [`-p/--period`](#period-expressions) 
 or a [`date:` query](#queries) (described below).
 All of these accept the [smart date](#smart-dates) syntax.
-One important thing to be aware of when specifying end dates: as in Ledger, end dates are exclusive, 
-so you need to write the date *after* the last day you want to include.
+
+Some notes:
+
+- As in Ledger, end dates are exclusive, so you need to write the date *after*
+  the last day you want to include.
+- As noted in [reporting options](#general-options):
+  among start/end dates specified with *options*, the last (i.e. right-most)
+  option takes precedence.
+- The effective report start and end dates are the intersection of the
+  start/end dates from options and that from `date:` queries.
+  That is, `date:2019-01 date:2019 -p'2000 to 2030'` yields January 2019, the
+  smallest common time span.
 
 Examples:
 
@@ -258,7 +268,7 @@ reports into multiple subperiods.  The basic intervals can be
 selected with one of `-D/--daily`, `-W/--weekly`, `-M/--monthly`,
 `-Q/--quarterly`, or `-Y/--yearly`.  More complex intervals may be
 specified with a [period expression](#period-expressions). 
-Report intervals can not be specified with a [query](#queries), currently. 
+Report intervals can not be specified with a [query](#queries).
 
 ## Period expressions
 
--- a/shell-completion/hledger-completion.bash
+++ b/shell-completion/hledger-completion.bash
@@ -118,6 +118,7 @@ import
 check-dates
 check-dupes
 close
+diff
 rewrite
 balancesheet
 balancesheetequity
@@ -127,6 +128,7 @@ roi
 accounts
 activity
 balance
+commodities
 files
 prices
 print
@@ -176,6 +178,7 @@ cat <<TEXT > "$_HLEDGER_COMPLETION_TEMPDIR/generic-options.txt"
 --alias
 --anon
 --auto
+--aux-date
 --begin
 --cleared
 --cost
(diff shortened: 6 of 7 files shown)


tests/cli/report-interval.test
```diff
@@ -0,0 +1,93 @@
...
+Balance changes in 2019/01/28-2019/02/10:
+
+                 || 2019/01/28w05  2019/02/04w06 
+=================++==============================
+ assets:checking ||          2000             -6 
+-----------------++------------------------------
+                 ||          2000             -6 
+>= 0
+
+# The last report interval option takes presedence.
```
simonmichael: precedence

schoettl:
Thanks for the tipps.

I cannot find documentation on the "input file inline" feature and `-f -` in the shelltestrunner README.

simonmichael:
Here you go:
https://github.com/simonmichael/shelltestrunner#format-2 (and format-3)
https://hledger.org/manual.html#input-files

tests/cli/report-interval.test
```diff
@@ -0,0 +1,60 @@
...
+$ hledger -f- register --weekly -p 'monthly in 2019'
+>
+2019/01                 a                                        2             2
+2019/02                 a                                        1             3
+
+
+
+# The next three tests test the bugfix for issue #1008:
+# The report interval from an --period expression only counts if it is
+# explicitly specified (e.g. -p 'monthly in ...').
```
simonmichael: Actually it should work like:
"start date, end date, and report interval are calculated from the union of -b/-e/-p/-D/-W/-M/-Q/-Y flags and date: arguments, with those appearing later on the command line taking precedence".

simonmichael: PS I don't mean that we should merge multiple date ranges. Here are some examples that hopefully show what I mean:
```
-b 1/1 -p apr    # same as -b 4/1 -e 5/1 (the -p's start date overrides the -b)
-p apr -b 1/1    # same as -b 1/1 -e 5/1 (-b overrides the -p's start date)
```

tests/cli/report-interval.test
```diff
@@ -0,0 +1,60 @@
...
+2019-02-01
+    (a)   1
+
+
+
+
+# The last report interval option takes precedence.
+
+$ hledger -f- register --weekly --monthly
+>
```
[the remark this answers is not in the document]
schoettl: Oh, now I understand why you decided to define the order to be `< $ > >2 >=` (vs. `$ < > >2 >=`).

I first preferred the latter form because the command is the most important line an should be on top (in my opinion).

But switching the `<` and `$` lines seems to be OK anyway, shelltest does not throw an error. Is this intended?

tests/cli/report-interval.test
schoettl: OK, I can add some tests for this. But they would be not directly related to #1008 und this test file `report-interval.test`. So I would add a new file `date-options.test` as you suggested previously.

Regarding your first comment: --period also sets the `ReportInterval` to `NoInterval` if no interval is specified. And this `NoInterval` would override a previous -W/-M/... which does not make sense for me.

simonmichael:
A -p not specifying a report interval should not override -M etc.

schoettl:
> A -p not specifying a report interval should not override -M etc.

OK, then we are on the same page :)

schoettl:
@simonmichael seems like the documentation you quoted does not match the real behavior:

> start date, end date, and report interval are calculated from the union of -b/-e/-p/-D/-W/-M/-Q/-Y flags and date: arguments, with those appearing later on the command line taking precedence

I added a few tests in `date-options.test` which fail.

I have not yet checked the code but it seems like a bug - or at least an inconsistency because -D/-W/... *do* override the report interval from a previous option.

simonmichael:
Which test, for example ?

schoettl:
number 3-5

simonmichael:
So I don't get confused and waste your time, could you be more specific - link/file/line numbers ?

If it helps, my "it should work like" is describing where we want to get to. My belief is it has been working like that except for -p options which have been overriding everything.

schoettl:
https://github.com/simonmichael/hledger/blob/aceec7becf4e11f7143d54ac3f576e08efd9ca2f/tests/cli/date-options.test#L23

this test and the following two tests fail.

I think I pretty much used your examples and "it should work like". -b/-e/-p seem to not override anything.

simonmichael:
Thanks. You're right! The later -b is not overriding the earlier one. I'm surprised by this.

schoettl:
I found the function:
```
periodFromRawOpts :: Day -> RawOpts -> Period
```
It's comment says
> Get the period specified by the intersection of -b/--begin, -e/--end and/or
> -- -p/--period options, using the given date to interpret relative date expressions.

*Intersection* actually means, the smallest time period that all begin/end values have in common, right?

*Union* would be the opposite.

The function however works against the comment, it produces the *union* period from the options. The tests I wrote earlier confirm this.

So the question is, how should hledger behave?

- union of all period options (gives the "widest" period)?
- intersection of all period options (gives the "narrowest" period)?
- last period options taking precedence?
- throw error on over-determined periods?

The manual I quoted earlier, talks about *union* which is correct if you take it as mathematical union. But it also mentions the precedence of later options.

I think neither union nor intersection are intuitive. Last option taking precedence seems like a work-around (not cleanly handling the user error). Throwing an error is not easy to implement because `-p2019-01 -e2019-01-15` (strictly speaking, over-determined) seems OK for me.

simonmichael:
@schoettl good question. Generally, when options are overlapping in some way, or repeated, hledger aims to apply them in order, as they appear on the command line. That's intuititive, and means eg when you paste a hledger command line, or have it defined as a shell alias, you can always customise it by adding more flags. .

With periodFromRawOpts and (beginDates/endDates/interval)FromRawOpts, I think the idea was to gather the individual flags according to the above rule - eg find the latest -b flag, the latest -e flag, the latest -p flag etc. Then apply all of those date span restrictions, to get their intersection.

Questions: 1. Does that goal sound reasonable and optimal for usability ? 2. Is that what we're testing for currently ? 3. Where did I go wrong implementing it ? I'm a little unclear on that right now

simonmichael:
More precisely, the idea was to gather the right-most begin date from a -b or -p, the right-most end date from a -e or -p, the right-most interval from a -D/W/M/Q/Y or -p, and use those. And it looks like that's basically what the code does. (Ideally it would also consider date: arguments). So I may be confused about something.

schoettl:
> Does that goal sound reasonable and optimal for usability ?

I understand the reason for the right-most option taking precedence. That is a good point! (Aliases and overriding options...)

> Is that what we're testing for currently ?

Yes, that is what I tested (both, interval and period options).

> Where did I go wrong implementing it ?

The implemention I mentioned above is wrong: It does not use the last i.e. right-most -b/-e/-p options but it uses the **earliest date for -b/-p** and the **latest date for -e/-p**. It's the `minimum` and `maximum` functions in `periodFromRawOpts`.

Doing it this way (mathematical union of the dates) is not intuitive. You could as well use the mathematical intersection. But I don't see how these approaches could be useful.

simonmichael:
> The implemention I mentioned above is wrong: It does not use the last i.e. right-most -b/-e/-p options but it uses the earliest date for -b/-p and the latest date for -e/-p. It's the minimum and maximum functions in periodFromRawOpts.
> 

Doh! Now I see it. You're right.

schoettl:
I fixed the bug and added one more test.

I still want to
- tweak the manual a little bit (man for -p is not quite correct)
- checkout how it behaves with `date:` queries in combination with options

Reagarding `date:` queries: Here too, the last query or option on the command line takes precedence, right?

simonmichael:
Cool, I'll look later. What would be the most intuitive works-like-you'd-guess behaviour for `date:` do you think ? Give options a higher precedence I think ? So the order of processing would be: date arguments left to right, then options left to right ?

schoettl:
I tried to find the place where the (`date:`) query is parsed and processed. But I cannot find it.

In Cli/Main.hs, there is `argsToCliOpts` but here I'm lost. How are the `args :: [String]` converted to the `RawOpts`? It seems to be the `process` function but I cannot find its definition or documentation.

Does the `process` function drop everything but command-line options?

Current status: a `date:` query *always* overrides date/period options! Independently of argument order.

schoettl:
A different question: who is supposed to re-generate the documentation, when the documentation source was changed (previous commits)?

simonmichael:
> A different question: who is supposed to re-generate the documentation, when the documentation source was changed (previous commits)?

Typically I do that, every so often, and always before release. I often postpone it to reduce distraction.

simonmichael:
> I tried to find the place where the (date:) query is parsed and processed. But I cannot find it.

You're on the right track. process is from cmdargs, which we call to do the heavy lifting:
http://hackage.haskell.org/package/cmdargs-0.10.20/docs/System-Console-CmdArgs-Explicit.html#v:process <http://hackage.haskell.org/package/cmdargs-0.10.20/docs/System-Console-CmdArgs-Explicit.html#v:process>

simonmichael:
PS and if you're wondering, yes our command line processing really is quite involved. Since we care about usability, I think this is somewhat unavoidable, alas,

schoettl:
I found the place where date options and `date:` queries are merged.
```haskell
-- | Convert report options and arguments to a query.
queryFromOpts :: Day -> ReportOpts -> Query
queryFromOpts d ReportOpts{..} = simplifyQuery $ And $ [flagsq, argsq]
   ... -- from options_::String and period_::Period
```

The `And` produces the mathematical intersection of all date filters (options and query args). Example: `date:'from 2018' -b2019` -> from 2019

It is independent of the order of args/options or precedence of option over args...

Look, I added some successful tests to demonstrate this: https://github.com/simonmichael/hledger/blob/fix/1008/tests/cli/date-options.test#L43

1. I also think that options should override date queries.
2. The report result is made up of the intersection of all queries (`And [queries]`). Last date query taking precedence would be a special case only for date filters. That is not good. All query args should be handled the same way, i.e. producing the intersection.

Do you agree with 1. and 2., @simonmichael ?

simonmichael:
Sorry - on phone - but I think that you’re agreeing with what I said/meant earlier:

- process arguments left to right 
- then options left to right
- statefully mutating begin date/end date/interval
- use their last state

schoettl:
> I think that you’re agreeing with what I said/meant earlier:

not quite

> - process arguments left to right

Right, query arguments are processed left to right. However, later date queries *do not* override previous ones. Instead, they narrow down the search (intersection, i.e. logical *and*).
And as I said, this make sense for query arguments. Although it's different from date *options*.

> - then options left to right

Right, options *should* override query arguments. But that is not easy to fix currently.

simonmichael:
@schoettl ah I understand you better now. Yes you're quite right. My suggestion of giving the right-most `date:` arguments precedence was silly. I'll try to write down the desired spec again:

1. combine `date:` arguments: as described in the query docs, ie by AND'ing them to get their intersection
2. combine begin date/end date/interval-specifying options: the right-most of each takes precedence
3. the intersection of 1 and 2 is the report period

Correct I hope ?

simonmichael:
PS this might still not be what you're thinking, but 

- intersecting 1 and 2 this way is the current behaviour. It's useful eg where the person managing a hledger-web instance can use command line options to set hard date limits and users can search anywhere within that period.
- it's reasonably simple and intuitive, I think.

simonmichael:
@schoettl checking in.. I think you are still working on this, or should I be merging these ? (If so let's clean up the history a bit, or I can just squash them). I guess our goals include 
1. clarifying current behaviour with both tests and docs, 
2. specifying the desired behaviour more clearly
3. updating implementation/tests/docs to meet that spec

schoettl:
I agree with your spec from yesterday.

I still work on this. I will check out and write tests to see, if it meets our new spec. And I will update docs.

Then I can squash the commits.

simonmichael:
Awesome, thank you.

schoettl:
A little side discussion:

> Report intervals can not be specified with a query, currently.

-- manual on Report intervals

I would like remove the word "currently". I think `date:` (as all queries) should not involve complexity beside filtering. I think it's better to have only `--period` for that. What do you think?

schoettl:
Comments on the change in doc:

https://github.com/simonmichael/hledger/pull/1011/files?short_path=585e4ff#diff-585e4ffb133b08e30c0c48236e78fff7

I think it should be clear now. If you approve, I would squash the commits a bit... and merge.

simonmichael:
> I would like remove the word "currently". I think date: (as all queries) should not involve complexity beside filtering. I think it's better to have only --period for that.

You could be right. I first introduced query arguments for use in hledger-web's search field (and urls). Since they overlap with -b/-e and part of -p (begin/end date) I figured some day perhaps they could also specify report interval like -p/D/W/M/Q/Y, for consistency and to provide a way to specify that in hledger-web. It's not something I'm eagerly working on, and I haven't thought about how it could work.

simonmichael:
The changes look good to me.

simonmichael:
> I haven't thought about how it could work.

Now I have, and I think if we wanted to control this via query argument we'd add a new one, `interval:` or something. So I agree.

schoettl:
Sorry for the very long delay on this issue. I squashed and rebased the commits, just testing based on latest master. Then I'll push and we can finally merge...

schoettl:
The multiple-file shell tests currently fail (even before my commits).

My new tests work. I would "Rebase and merge" this branch tomorrow.

simonmichael:
I just now got a chance to review this. It looks good, I also updated one of the function docs. @schoettl, thanks for your persistence on this.

Issue #131, opened 2013-07-28
Allow reporting in an arbitrary commodity (Ledger's -X, or something like it)


tphyahoo:
[Original title: **implement Ledger's -V (& -X ?) feature, for converting to a desired currency and showing market value**. Part of #1015. 2019 discussion starts [here](#issuecomment-493772892).]

Hledger should have some notion of exchange rates over time, as sketched out in

https://github.com/simonmichael/hledger/issues/115#issuecomment-21672724

The currency reported should be an allowed "view" that changes if currency rates over time data changes. To make it clear that this is a view and not core data, computed values should be reported in parenthesis, or otherwise distinguished somehow.

This is a feature intended for use in --require-balanced=true mode. Context: https://github.com/simonmichael/hledger/issues/115

Example of usage: 
  hledger print --currency-numeraire=USD --currency-conversions=currency-conversions.txt

where currency-conversions.txt file is something like

2013/01/01 USD1=CAD1.31
2013/01/02 USD1=CAD1.32

This is similar use case as print --cost, but requires --require-balanced=true.
For backwards compatibility, I think we should keep current behavior for print --cost/-B where require-balanced=false.

For concreteness, consider this transaction 

```
    2013/01/01  Canadian in the USA Buys food with US dollars (1 USD = 1.30 CAD)
       assets:cash          -USD40
       equity:forex         USD40
       equity:expense:food  CAD52 
       equity:forex         -CAD52 
```

  If conversions file (as above) specifies 1USD=1.31 CAD, contradicting recorded ledger rate of 1.30,
  then for the food purchase above
    print --numeraire=USD --numeraire-file=currency-conversions=conversions.txt
  would yield something like

```
    2013/01/01  Canadian in the USA Buys food with US dollars (1 USD = 1.30 CAD)
       assets:cash          -USD40
       equity:forex         USD40
       equity:expense:food  CAD52 (USD39.69)
       equity:forex         -CAD52 (-USD39.69)

Q Does print --currency-numeraire=USD imply --currency-conversions specified somewhere, or can it work without the numeraire file?
A There are two reasonable behaviors
1) require the flag
2) don't require the flag, but have very boring behavior: compute the conversions from information available in file, and then reports numbers in parenthesis that agree everywhere like:

    2013/01/01  Canadian in the USA Buys food with US dollars (1 USD = 1.30 CAD)
       assets:cash          -USD40
       equity:forex         USD40
       equity:expense:food  CAD52 (USD40)
[...]

PR #1044, opened 2019-06-11
allow specifying valuation commodity, add -X

simonmichael:
Seems to work, more testing welcome. I'll merge this soon unless anyone has suggestions.

This is for #131. The docs can be viewed [here](https://github.com/simonmichael/hledger/blob/8d665ecc9f98c22ae8ec9a4db9f408c137877121/hledger/hledger_options.m4.md#valuation).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -414,7 +414,7 @@ Posting -- MixedAmount
 MixedAmount *-- "*" Amount
 Amount -- CommoditySymbol
 Amount -- Quantity
-Amount -- Price
+Amount -- AmountPrice
 Amount -- AmountStyle
 </uml>
 -->
--- a/hledger-api/hledger-api.hs
+++ b/hledger-api/hledger-api.hs
@@ -152,7 +152,7 @@ hledgerApiApp staticdir j = Servant.serve api server
       where
         accountnamesH = return $ journalAccountNames j
         transactionsH = return $ jtxns j
-        pricesH       = return $ jmarketprices j
+        pricesH       = return $ jpricedirectives j
         commoditiesH  = return $ (M.keys . jinferredcommodities) j
         accountsH     = return $ ledgerTopAccounts $ ledgerFromJournal Hledger.Query.Any j
         accounttransactionsH (a::AccountName) = do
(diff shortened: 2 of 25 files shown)


alerque:
I'm brand new to the Ledger ecosystem and have just been importing various finances using hledger since last week. I came here after scrubbing the docs looking for just this feature! Thanks for your work on this (and the project in general). I'll be eager for it to land in a release.

danielverdugo:
Great. The "-X" is long waited feature of mine.

@simonmichael Does it show the exchange rate effective on the date of transaction?

simonmichael:
@danielverdugo, I like -X but so far I've always found -V was enough. Can you share a use case or two ? 

At one point it had a mode that used exchange rates on the date of each transaction, but it was complicated and I didn't have any clear use case for it. Currently it uses exchange rates at the end of each report subperiod instead. If you see a real-world need for it, please say more (and specify the desired report in detail).

danielverdugo:
@simonmichael In my accounting system I use multiple currencies regularly. Primarily EUR and RUB. 

I use them in different occasions, both in Balance and Income statement accounts.

RUB is, unfortunately, not the stablest currency in the world, being the currency of the developing economy. So, many times I want to see my Balance or Income statement converted to other, more solid currencies, like EUR and USD.

More than that, with ledger I can use -X to track my investments.

I do it this way. For example, I have 10 stocks of Apple, Inc., which I bought for 199 USD each. I would record it as follows:
```
01-05 *
   Assets:Investments  10 AAPL
   Equity:Unrealised investment gain  -10 AAPL
   Equity:Unrealised investment gain  1990 USD
   Assets:Cash  -1990 USD
```

Now, just using the price directive (`P 2019-01-05 AAPL 210 USD`) I may monitor market vs. cost difference (in other words, unrealized investment gain) on the equity account using "-X USD".

I realize, my use cases may be rare. However, I find them quite handy.

simonmichael:
> many times I want to see my Balance or Income statement converted to other, more solid currencies

Thanks @danielverdugo, that makes sense. While -V would convert to the currency you normally use, but you need -X for occasionally viewing in alternate currencies (without having to swap price files). 

> Now, just using the price directive (P 2019-01-05 AAPL 210 USD) I may monitor market vs. cost difference

-V seems enough for this one, but maybe it's a similar case to above. Thanks for the input.

Issue #1045, opened 2019-06-11
Add option to print with normalized output width


alerque:
I use `hledger` itself to process and normalize all my ledgers whether imported from other sources or manually composed. The `print` command makes sure the values for every posting in a transaction are aligned.

```ledger
2016/01/01 Example transaction
	Expense:Foo  $6
	Expense:Foo:Baz  $3
	Cash
```
Becomes the much nicer:
```ledger
2016/01/01 Example transaction
    Expense:Foo                  $6
    Expense:Foo:Baz              $3
    Cash                        $-9
```
However the column at which this alignment happens varies from transaction to transaction. Consider this case:
```ledger
2016/01/01 Example transaction
	Expense:Foo  $6
	Cash

2016/01/01 Another transaction
	Expense:Bar:Baz  $6
	Cash
```
Normalizing this using `print -x` will return each transaction aligned at a a different place:
```
2016/01/01 Example transaction
    Expense:Foo              $6
    Cash                    $-6

2016/01/01 Another transaction
    Expense:Bar:Baz              $6
    Cash                        $-6
```
In this case `ledger print` is superior in that all transactions are aligned together:
```ledger
2016/01/01 Example transaction
    Expense:Foo                                   $6
    Cash

2016/01/01 Another transaction
    Expense:Bar:Baz                               $6
    Cash
```
I considered normalizing through `ledger`, but there are many other ways in which `hledger print` is preferable, such as alignment of postings with commodity costs:

```ledger
# Source
2016/01/01 Unit prices
	Expense:Foo  56 @ $6.00
	Cash  $-336

# Output of hledger print
2016/01/01 Unit prices
    Expense:Foo      56 @ $6.00
    Cash                  $-336

# Output of ledger print
2016/01/01 Unit prices
    Expense:Foo                                   56 @ $6
    Cash                                       $-336
```
Personally I really like having the final cost of the posting right aligned as in the `hledger` output.

What I would like is some kind of format option that specifies the right edge to align to (if possible without truncation of course).

The reason this would be preferable to a fully automatic width guesser that spans multiple transactions is that I could standardize on a width such that future changes such as renaming an account or moving transactions from one ledger to another would not cause churn in my Git commit history just because the length of an account string changed. Right now running `git blame` on a Ledger file shows quite a bit of useless results as every posting in a transaction can change when just one posting is edited (because the column amounts are aligned to changes).

alerque:
[...]

PR #1048, opened 2019-06-12
print: differentiate line indent (4spc) from pre-comment space (2spc)

### LINKED ISSUE #1047: Number of spaces used before comment separator in print should be configurable
By default `ledger` outputs comments with 2 spaces before the separator `;`

```ledger
2019/06/11 (TXNID) Office Building Complex
    Expenses:Building:Dues                 ₺1.020,00  ; date: 2019/04/01
    Expenses:Building:Dues                 ₺1.020,00  ; date: 2019/05/01
    Expenses:Building:Dues                 ₺1.020,00  ; date: 2019/06/01
    Expenses:Fees:Banking                      ₺6,75  ; payee: Bank Name
    Assets:Bankaccount                    ₺-3.066,75
```
On the other hand `hledger` uses four:
```ledger
2019/06/11 (TXNID) Office Building Complex
    Expenses:Building:Dues                 ₺1.020,00    ; date: 2019/04/01
    Expenses:Building:Dues                 ₺1.020,00    ; date: 2019/05/01
    Expenses:Building:Dues                 ₺1.020,00    ; date: 2019/06/01
    Expenses:Fees:Banking                      ₺6,75    ; payee: Bank Name
    Assets:Bankaccount                    ₺-3.066,75
```

I find the four spaces somewhat excessive (and in more verbose examples such as with tags the extra space can cause unnecessary wrapping or scrolling) and would like to configure it to output 2 instead. I don't see any option to configure this preference.

Note `ledger` requires at least one space to parse comments, `hledger` doesn't need any at all, so this isn't a parsing issue, it is just a preferred output format issue.

alerque:
Fixes #1047.

- [x] Split `indent` value so comment separation spaces can be different from posting indents
- [x] Update examples in documentation
- [x] Update output of tests

Optionally...

- [ ] Make number of spaces user configurable

### DIFF
--- a/doc/mockups/entry-macros.txt
+++ b/doc/mockups/entry-macros.txt
@@ -10,7 +10,7 @@ $ ls *.entry
 laundry.entry  date.entry
 
 $ hledger entry laundry 20
-2015-05-06 laundry    ;  household:$20
+2015-05-06 laundry  ;  household:$20
     expenses:personal:home:laundry                            $10
     liabilities:personal:accounts payable:household           $10
     assets:personal:cash:wallet                              $-20
--- a/hledger-lib/Hledger/Data/Transaction.hs
+++ b/hledger-lib/Hledger/Data/Transaction.hs
@@ -169,11 +169,16 @@ showTransactionHelper elide onelineamounts t =
                                                 c:cs -> (c,cs)
 
 -- | Render a transaction or posting's comment as indented, semicolon-prefixed comment lines.
+-- The first line (unless empty) will have leading space, subsequent lines will have a larger indent.
 renderCommentLines :: Text -> [String]
-renderCommentLines t  = case lines $ T.unpack t of ("":ls) -> "":map commentprefix ls
-                                                   ls      -> map commentprefix ls
-    where
-      commentprefix = indent . ("; "++)
+renderCommentLines t =
+  case lines $ T.unpack t of
+    []      -> []
+    [l]     -> [(commentSpace . comment) l]        -- single-line comment
+    ("":ls) -> "" : map (lineIndent . comment) ls  -- multi-line comment with empty first line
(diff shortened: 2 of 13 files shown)


alerque:
I might be stuck on this one. I got part of the way there in that line level comments properly indent with 4spcs and inline comments are preceded by 2spcs, but the test cases include a multi line comment that gets split on newlines, and I don't know how to handle that. Right now I'm flunking the tests suite because its inserting 2spcs for the wrapped comment newlines instead of the expected 4.

https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Data/Transaction.hs#L571

simonmichael:
Thanks! I'll take a look later if you don't beat me to it.

alerque:
The branch is open to your pushes ;-)

Interestingly the test case that's failing cannot be replicated from normal input, it is only even possible from Haskell programmatically generating the transaction object.

alerque:
I was looking into making this an option, I'm not sure where the option should go. It would go the same place as the new flag(s) for #1045 go, but I can't figure out where that would go to affect `print`, `print-unique`, and any other places affected.

On even further review, I'm not convinced making this configurable is even such a great idea. Two spaces seems to be pretty universal already, and allowing less breaks some things (`ledger` for example would parse that okay in postings, but not in the transaction description line).

I guess the question is: is there actually any demand for MORE than 2 spaces?

simonmichael:
Making what an option.. the number of spaces before a same-line comment ? I don't think we need to. I suppose the few folks who commit and compare report output to detect changes, will notice diffs and maybe ask why we are bothering them with this. Hopefully not.

alerque:
In that case I think this is ready to go.

simonmichael:
Thanks!

PR #1052, opened 2019-06-13
print: allow decimal point aligning, like ledger-mode (WIP)

simonmichael:
Part of #1045. I created showTransaction2 and helpers, with the immediate goal of (a) clarifying this code and (b) implementing output like ledger-mode's :decimal setting.

What I had in mind: for postingAsLines to do decimal mark alignment, it needs to know which part of each rendered amount is to left and which to the right of the decimal mark. The decimal mark can be period or comma, and that character can also appear in a (double quoted) commodity symbol, so parsing the rendered amount seems a bit fiddly with potential to get out of sync with rendering code. I thought instead to have helpers like showMixedAmountLR & showMixedAmountOneLineLR which return the left and right parts; possibly showMixedAmount[OneLine] would use these also. These would need a showAmountLR :: Amount -> (String,String).

### DIFF
--- a/hledger-lib/Hledger/Data/Transaction.hs
+++ b/hledger-lib/Hledger/Data/Transaction.hs
@@ -12,6 +12,7 @@ tags.
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE RecordWildCards #-}
+-- {-# LANGUAGE NamedFieldPuns #-}
 
 module Hledger.Data.Transaction (
   -- * Transaction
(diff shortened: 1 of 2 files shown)


alerque:
I've tampered with this a couple times but my Haskell isn't really up to delivering anything useful. On the other hand I do wish for it to be done quite frequently. My current reliance on the [vim plugin](https://github.com/ledger/vim-ledger/) to normalize my journal files is somewhat brittle and, while it gets closer to this output than `hledger` does now, leaves a lot to be desired.

simonmichael:
I accidentally closed this by cleaning up old branches. I have restored the https://github.com/simonmichael/hledger/tree/1045-print branch, but the PR can't be reopened it seems.

simonmichael:
Closed because of branch rename (#2574).

PR #1068, opened 2019-07-15
Strip extraneous trailing whitespaces

alerque:
Partly as a result of my editor highlighting these, partly because I have git configured to warn on them — trailing white space in languages where it is not normally semantically drives me crazy. Is it a serious problem? No, but it _can_ lead to small discrepancies here and there. Most tooling works just a little bit better with this sort of thing cleaned up.

Note this should _NOT_ be merged if there are any other major contributions pending that are likely to be be merge candidates in their current form because it is quite  likely to introduce merge conflicts. Instead it should be merged at some juncture when everything else is quiet on the merge front.

For reference this work can be easily rebased/updated using this command (Neovim with [vim-better-whitespace](https://github.com/ntpeters/vim-better-whitespace) plugin):

```sh
git ls-files '*.hs' | xargs -n1 nvim --headless +StripWhitespace +x
```


simonmichael:
Oh, I marked it ready for review. Looks good! Can I merge ?

alerque:
I just looked at pending PRs and don't see anything that's going to me much of a hickup to rebase, so sure...

simonmichael:
Thanks for the cleanup!

PR #1074, opened 2019-07-18
[WIP] web: Add option --socket to use UNIX socket file

Amarandus:
# Motivation and Use-Cases

I wanted to run multiple instances of `hledger-web` behind an nginx reverse proxy, but wanted to get around an arbitrary `user -> port`-mapping.
Thus, my idea was to use unix socket files in a scheme like `/tmp/hledger-$username.socket`. This would allow to use an nginx directive like the following:
```
  proxy_pass http://unix:/tmp/hledger-${remote_user}.socket;
```

Other usecases were also given in the IRC channel:
 - `hpd_> maybe because unix sockets allow for proper access control`
 - `alerque[m]> A socket would also be useful for local GUI wrappers.`

# Changes

The main changes are:
 - The code now requires Network.Socket (and thus `network` as a dependency)
   - This gives us the functions to create a socket
 - Introduces `--socket=/path/to/socket` option
   - This implies `--serve`, as most browsers probably won't handle socket files that well
 - Conditionally use the warp handler `runSettingsSocket` instead of `runSettings`

# Problems

While testing, I also found small problems where I do not know how to approach it.

Sockets are not cleaned up properly when quitting. So restarting directly after quitting results in:
```
hledger-web: Network.Socket.bind: resource busy (Address already in use)
```
As far as I can see, this can be solved by just deleting the socket file when quitting. One code to solve this might be here: https://stackoverflow.com/a/36421958 but I did not want to introduce too many requirements (unsure regarding the code organization and style).

As an obvious workaround, one can just `rm` the socket file.

Another problem is that sockets are not supported on windows, so there is probably some conditional inclusion of this code required, so that the option is not available at all on windows systems.

Also, more documentation besides the option description is probably missing

### DIFF
--- a/hledger-web/Hledger/Web/Main.hs
+++ b/hledger-web/Hledger/Web/Main.hs
@@ -9,15 +9,19 @@ Released under GPL version 3 or later.
 
 module Hledger.Web.Main where
 
+import Control.Exception (bracket)
 import Control.Monad (when)
 import Data.String (fromString)
 import qualified Data.Text as T
+import Network.Socket
 import Network.Wai (Application)
-import Network.Wai.Handler.Warp (runSettings, defaultSettings, setHost, setPort)
+import Network.Wai.Handler.Warp (runSettings, runSettingsSocket, defaultSettings, setHost, setPort)
 import Network.Wai.Handler.Launch (runHostPortUrl)
 import Prelude hiding (putStrLn)
-import System.Exit (exitSuccess)
+import System.Directory (removeFile)
+import System.Exit (exitSuccess, exitFailure)
 import System.IO (hFlush, stdout)
+import System.PosixCompat.Files (getFileStatus, isSocket)
 import Text.Printf (printf)
 import Yesod.Default.Config
 import Yesod.Default.Main (defaultDevelApp)
@@ -76,7 +80,27 @@ web opts j = do
       putStrLn "Press ctrl-c to quit"
       hFlush stdout
       let warpsettings = setHost (fromString h) (setPort p defaultSettings)
-      Network.Wai.Handler.Warp.runSettings warpsettings app
+      case socket_ opts of
+        Just s -> do
+          if isUnixDomainSocketAvailable then
+            bracket
+              (do
+                  sock <- socket AF_UNIX Stream 0
+                  setSocketOption sock ReuseAddr 1
+                  bind sock $ SockAddrUnix s
+                  listen sock maxListenQueue
+                  return sock
+              )
+              (\_ -> do
+                  sockstat <-  getFileStatus s
+                  when (isSocket sockstat) $ removeFile s
+              )
+              (\sock -> Network.Wai.Handler.Warp.runSettingsSocket warpsettings sock app)
+            else do
+              putStrLn "Unix domain sockets are not available on your operating system"
+              putStrLn "Please try again without --socket"
+              exitFailure
+        Nothing -> Network.Wai.Handler.Warp.runSettings warpsettings app
     else do
       putStrLn "This server will exit after 2m with no browser windows open (or press ctrl-c)"
       putStrLn "Opening web browser..."
--- a/hledger-web/Hledger/Web/WebOptions.hs
+++ b/hledger-web/Hledger/Web/WebOptions.hs
@@ -44,6 +44,11 @@ webflags =
       "ORIGIN"
       ("allow cross-origin requests from the specified origin; setting ORIGIN to \"*\" allows requests from any origin")
   , flagReq
+      ["socket"]
+      (\s opts -> Right $ setopt "socket" s opts)
+      "SOCKET"
+      "use the given socket instead of the given IP and port (implies --serve)"
+  , flagReq
       ["host"]
       (\s opts -> Right $ setopt "host" s opts)
       "IPADDR"
@@ -110,10 +115,11 @@ data WebOpts = WebOpts
   , capabilities_ :: [Capability]
   , capabilitiesHeader_ :: Maybe (CI ByteString)
   , cliopts_ :: CliOpts
+  , socket_ :: Maybe String
   } deriving (Show)
 
 defwebopts :: WebOpts
-defwebopts = WebOpts def def Nothing def def def def [CapView, CapAdd] Nothing def
+defwebopts = WebOpts def def Nothing def def def def [CapView, CapAdd] Nothing def Nothing
 
 instance Default WebOpts where def = defwebopts
 
@@ -131,9 +137,12 @@ rawOptsToWebOpts rawopts =
           Left e -> error' ("Unknown capability: " ++ T.unpack e)
           Right [] -> [CapView, CapAdd]
           Right xs -> xs
+        sock = stripTrailingSlash <$> maybestringopt "socket" rawopts
     return
       defwebopts
-      { serve_ = boolopt "serve" rawopts
+      { serve_ = case sock of
+          Just _ -> True
+          Nothing -> boolopt "serve" rawopts
       , serve_api_ = boolopt "serve-api" rawopts
       , cors_ = maybestringopt "cors" rawopts
       , host_ = h
@@ -143,6 +152,7 @@ rawOptsToWebOpts rawopts =
       , capabilities_ = caps
       , capabilitiesHeader_ = mk . BC.pack <$> maybestringopt "capabilities-header" rawopts
       , cliopts_ = cliopts
+      , socket_ = sock
       }
   where
     stripTrailingSlash = reverse . dropWhile (== '/') . reverse -- yesod don't like it
--- a/hledger-web/hledger-web.cabal
+++ b/hledger-web/hledger-web.cabal
@@ -1,10 +1,10 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.31.2.
+-- This file has been generated from package.yaml by hpack version 0.32.0.
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 194a16a30440803cd2bef5d5df0b6115ad66076c44407ed31efea9c4602e5e95
+-- hash: a9a6dea39ea5c963970cda9f595e7c4251332954aacd137fb6638a1e7640ee59
 
 name:           hledger-web
 version:        1.16.99
@@ -176,12 +176,14 @@ library
     , http-types
     , megaparsec >=7.0.0 && <8
     , mtl >=2.2.1
+    , network
     , semigroups
     , shakespeare >=2.0.2.2
     , template-haskell
     , text >=1.2
     , time >=1.5
     , transformers
+    , unix-compat
     , utf8-string
     , wai
     , wai-cors
--- a/hledger-web/hledger-web.m4.md
+++ b/hledger-web/hledger-web.m4.md
@@ -72,6 +72,10 @@ as shown in the synopsis above.
 `--port=PORT`
 : listen on this TCP port (default: 5000)
 
+`--socket=SOCKETFILE`
+: use a unix domain socket file to listen for requests instead of a TCP socket. Implies
+`--serve`. It can only be used if the operating system can provide this type of socket.
+
 `--base-url=URL`
 : set the base url (default: http://IPADDR:PORT).
 You would change this when sharing over the network, or integrating within a larger website.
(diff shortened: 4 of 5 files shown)


Amarandus:
I hope that at least the documentation is covered so far with the last commit.

simonmichael:
@Amarandus, this sounds good, thanks for the PR. 

I think we need some solution for the cleanup problem before merging. Are you aware of any other socket-using apps that we could follow ? How do they handle it ? If we can reliably clean up on exit, that sounds ideal. In case that fails somehow, we should also try to make the startup failure message clear enough that the user could resolve if manually, eg naming the socket file they should delete. 

A little more explanation of this might also be needed, to clarify when to use this. I've heard about socket files for a long time but I'm not used to using them. The advantages mentioned sound good, but I'm still wondering

- How do sockets improve on ports for your use case, exactly ?  Is it that you can name them according to usernames rather than just by number ? 
- Where are those usernames coming from ? 
- Could we also document an equivalent config for Apache users ?
- How do they help with access control ? Does that mean relying on unix users and file permissions, rather than configuring authentication in your web server ? What would be a simple example ?
- How do they help local GUI wrappers ?

Amarandus:
> Are you aware of any other socket-using apps that we could follow ? How do they handle it ?

In general, most aspects of socket files are decribed in the man page `unix(7)` http://man7.org/linux/man-pages/man7/unix.7.html

In the section "Notes", it describes that exact cleanup:

> Binding to a socket with a filename creates a socket in the
> filesystem that must be deleted by the caller when it is no longer
> needed (using unlink(2)).

So in the worst case, one solution might be to just delete the file at the given path. This might introduce race conditions, as the socket file can be renamed or moved while it is in use, so blindly deleting the file can lead to problems (although I can not imagine any non-harmful usecase where moving a different file in place of the socket might be desired).

In the best case, it would make sense to use the socket file descriptor (maybe with `withFdSocket`) and run `unlink` on that file descriptor (which seems to be available as `removeLink` in `System.Posix.Files`), but I was not able to find a way to do this in Haskell so far (but as mentioned already, I'm still a haskell novice).


> How do sockets improve on ports for your use case, exactly ? Is it that you can name them according to usernames rather than just by number ?

Yes, exactly this is the case. I am using NixOS and have a list of usernames that I want to be able to use hledger-web. For each user, I just want to run hledger-web with a socket at `/tmp/hledger-${username}.socket`. This allows to use the nginx config directive I mentioned earlier. `${remote_user}` holds the username provided by HTTP basic authentication.

> Where are those usernames coming from ?

In my setup, I'm using one list of usernames to generate the users for the system.
For each user of that list, I also create a systemd unit that executes `hledger-web --socket /tmp/hledger-${username}.socket`. My `nginx` is configured to use PAM to allow authentication for these users.

> Could we also document an equivalent config for Apache users ?

The appropriate directive for this is `ProxyPass` and is documented here: https://httpd.apache.org/docs/trunk/mod/mod_proxy.html#proxypass

I am not able to test this (generally not used to apache), but it should be something similar to:
```
ProxyPass / unix:/tmp/hledger-web.socket|http://localhost/
ProxyPassReverse / unix:/tmp/hledger-web.socket|http://localhost/
```
Sadly, I did not find a way to integrate the authenticated username in that path for apache (but as I mentioned earlier, I do not really know much about apache)

 > How do they help with access control ? Does that mean relying on unix users and file permissions, rather than configuring authentication in your web server ? What would be a simple example ?

Socket files behave basically just like files (or more like bidirectional named pipes). For my usecase (a server where multiple users have access to it), it helps reducing the exposure of each instance, as I want the socket files to only be readable by the nginx user and the original "owning" user. All other users won't have access to the file, which means that they can't reach the hledger-web instance.
User authentication on its own is still handled by the webserver.

> How do they help local GUI wrappers ?
To be honest, I don't know. Pinging @alerque as he wrote that in the IRC channel.

alerque:
> How do they help local GUI wrappers?

An app that wanted to wrap the web UI or alternatively one that just wanted to query the API, would not need to setup a port and go through the motions required to properly secure such. Securing ports is hard and almost never done properly. Localhost only bindings are usefully for dedicated single service boxes, but don't provide access control. That is fine for things meant to run as a service or with their own authentication mechanisms, but not fine for my financial data on a shared box.

In other words simplicity plus access control.

simonmichael:
Hi @Amarandus. Is this still something you're using ? Would you have time to look at the cleanup issue again ?

simonmichael:
Maybe https://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Exception.html#v:bracket is the way ?

Amarandus:
Hi @simonmichael, thanks for the ping - life got in the way a bit, so I did not have time to continue working on it.

I am pretty sure that `bracket` should be the way to go, but I don't know if just deleting the socket file is the correct approach (although I'd guess that the problem here should be marginally important, but I don't want to introduce any problems by this).
I'll rebase my PR to the newest code base and try to get it running with `bracket` in the next days.

I am do not know how to limit this feature to unix systems (windows does not support socket files, as far as I know).

Amarandus:
Picking this PR finally up again. I just rebased to make a merge easier in the end (hope I did not break anything by doing this).

Amarandus:
So, while the last commit switches to using `bracket` to delete the socket file after use, this definitely introduces something similar to a Time-of-check to time-of-use race condition [1], as we can just switch a different file in the place of the socket. This causes `hledger-web` to delete the file after termination.

To test/show this, I've written a small bash script:

```sh
#!/bin/sh

set -x

# Clean old test files
rm bak.socket testsock.socket

# Generate TOCTOU test
echo "This is an important file, please don't delete me" > toctou-testfile

# Start hledger-web
./hledger-web/.stack-work/dist/x86_64-linux-nix/Cabal-3.0.0.0/build/hledger-web/hledger-web --socket testsock.socket &
child_pid=$!
echo "Hledger running on PID ${child_pid}"

# Wait for it to start
sleep 5;

# Play some file switcheroo
mv ./testsock.socket ./bak.socket
mv toctou-testfile ./testsock.socket

sleep 2;

curl --silent --output /dev/null -XGET --write-out "Response Code: %{http_code}\n" --unix-socket ./bak.socket http://localhost/journal

sleep 2;

# Terminate hledger
kill -2 "${child_pid}"

sleep 3;

# Check if the file still exists
if [ -f ./testsock.socket ];
then
    echo "File still exists";
    cat ./testsock.socket
else
    echo "File got deleted";
fi
```

The saner way to do that would basically be to use the file descriptor (available with [`withFdSocket`)](https://hackage.haskell.org/package/network-3.1.1.1/docs/Network-Socket.html#v:withFdSocket) to get the current path, and then remove that socket, but I did not find a way to get that path (and looking into `/proc/$pid/fd`, it seems like this is not even possible as the fd of the socket does not link to a valid file path).

To be honest, this might be a really constructed problem, but it is real and I do not see myself in a position to decide whether this is fine and can be ignored. Moving a file into that position requires write permissions on that directory, but this would allow a malicious person to delete the file directly [2].
I also do not know how other programs solve this problem.

As far as I see, the following things need to be done to remove that WIP-Tag:
 - Decide whether that socket file deletion can be a problem
 - Disable this option on systems that do not have unix file sockets (i.e. windows)

[1] https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use
[2] https://unix.stackexchange.com/questions/272225/minimum-file-permissions-to-rename-file

Amarandus:
Easier stuff first: 9315ecd  causes `--socket` to be ignored if unix domain sockets are not available. This should be the best/easiest solution to solve the portability issue.

It might be desirable to print a warning if it is supplied but not available, but that might be harder than simply ignoring the flag.

---

In 504b175 I tried limiting the problem with the file switching as much as possible. Before deleting the file, it checks if the file type is in fact still a socket.
We still have that small time window for that race condition, but I want to be honest: I do not think this is a problem, as it allows the user to delete a file in a complex way, but he was already able to delete the file beforehand (elsewise he would not be able to move it into place).
The other thing that still works is moving a different socket in place, but this won't do any harm, as sockets are more or less volatile nonetheless (I just want to say here that the code does not check if it is "our" socket, it only checks if it *is* a socket).

I just did not want to make that decision without at least making other people aware that there _might_ be a problem, even if the impact is probably pretty negligible (as I said - it allows deleting a file that the user *could* have deleted by a simple `rm /path/to/file`).

In my opinion, this PR is now ready for review.

hledger-web/hledger-web.m4.md
```diff
@@ -119,6 +123,12 @@ You can use `--host` to change this, eg `--host 0.0.0.0` to listen on all config
 Similarly, use `--port` to set a TCP port other than 5000, eg if you are
 running multiple hledger-web instances.
 
+Both of these options are ignored when `--socket` is used. In this case, it creates
+an `AF_UNIX` socket file at the supplied path and uses that for communication.
+This can be used to run multiple hledger-web instances on the same host, eg
+behind a reverse proxy.
+This option also implies `--serve`.
+
```
simonmichael: I think just the first sentence is enough here; the explanation of --socket should be at that flag.

hledger-web/hledger-web.m4.md
```diff
@@ -72,6 +72,10 @@ as shown in the synopsis above.
 `--port=PORT`
 : listen on this TCP port (default: 5000)
 
+`--socket=SOCKETFILE`
+: use a socket file to listen for requests instead of a TCP socket. Implies 
+`--serve`. It can only be used on systems that are capable of this type of socket.
```
simonmichael: Let's be more explicit:  "...capable of this type of socket (eg, not Windows)." ?

hledger-web/hledger-web.cabal
```diff
@@ -183,6 +183,7 @@ library
     , text >=1.2
     , time >=1.5
     , transformers
+    , unix
```
simonmichael: I haven't tested, but I think this will prevent finding an install plan on windows, and perhaps the solution is to replace that with `unix-compat` (and import from System.PosixCompat.Files).

hledger-web/Hledger/Web/WebOptions.hs
```diff
@@ -47,7 +48,7 @@ webflags =
       ["socket"]
       (\s opts -> Right $ setopt "socket" s opts)
       "SOCKET"
-      "use the given socket instead of the given IP and port (implies --serve)"
+      "use the given socket instead of the given IP and port (implies --serve, ignored it not available)"
   , flagReq
```
simonmichael: it not available -> if not supported by OS ?

But actually, if it's not supported and they try to use it, we should give an error rather than silently running in a different mode.

hledger-web/Hledger/Web/Main.hs
```diff
@@ -79,11 +81,16 @@ web opts j = do
...
-          Network.Wai.Handler.Warp.runSettingsSocket warpsettings sock app
+          bracket
+            (do
+                sock <- socket AF_UNIX Stream 0
+                setSocketOption sock ReuseAddr 1
+                bind sock $ SockAddrUnix s
+                listen sock maxListenQueue
+                return sock
+            )
+            (\_ -> removeFile s)
```
simonmichael: Could use bracket_ here since the arg is ignored

Amarandus: It is only ignored for the final operation, but still required for the in-between-computation.
`bracket` is `IO a -> (a -> IO b) -> (a -> IO c) -> IO c`, while something like `IO a -> IO b -> (a -> IO c) -> IO c` would be sufficient (so `bracket_` is not sufficient).

hledger-web/Hledger/Web/WebOptions.hs
Amarandus: That would make more sense. I'll add an error.

hledger-web/hledger-web.m4.md
Amarandus: I just wanted to confirm it, and it seems like Windows *might* be able to provide `AF_UNIX` sockets now [1]. I'd say that explicitly calling windows out might be wrong, but I reworded it now so that the importance of the operating system becomes a bit more clear here.

[1] https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

Amarandus:
Squashed and hopefully all problems addressed :)

simonmichael:
Looks good! Thanks for pushing this through to completion!

simonmichael:
Just for the record, a related solution for multiple private instances, from stack overflow:

> Loopback addresses are 127.0.0.0/8 That is 127.x.x.x not just 127.0.0.1.
> Therefore use a different IP address for each server. E.g. 127.0.0.2, 127.0.0.3 ...

simonmichael:
Also for the record, serving via unix socket can be tested like so:
```
curl -v --unix-socket SOCKETFILE http://anything/PATH
```

PR #1080, opened 2019-08-21
;doc: Excise quick links to domains that no longer exist

alerque:
Found an old commit I forgot to submit...


simonmichael:
Thanks!

PR #1086, opened 2019-09-11
Fix periodictxn generation

### LINKED ISSUE #1085: "every 12 months from <past date>" starts counting from current date/month
I want to have a transaction that runs in April every year (so every 12 months from 2019/04). However, if the specified from date is past the current date, the current month is used:

```
$ cat t
~ every 12 months from 2019/04
        a  -5 EUR
        b  5 EUR
$ hledger -f t  p --forecast -p 2020 
2020/09/01
    ; generated-transaction:~ every 12 months from 2019/04
    a          -5 EUR
    b           5 EUR
$ hledger --version   # from nix
hledger 1.15.2
```

adept:
This fixes #1085 

runPeriodicTransaction tried really hard to avoid generating all instances of periodic transaction since specified start date, trying instead to figure out closest start date before the requested span. Turns out that this is hard, and all attempts done so far (two of them by me) failed on some corner case or other.

Lets make it right, and then see if we need to make it fast :) If we generate all instances of transaction from its start date and up to the end of reporting interval, and then only take those that fall inside reporting interval, code becomes refreshingly simple :)

I added some tests from #1085 and commit 129f6e6839db91fb17c13aa9ce0718f168b5a193

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -60,7 +60,6 @@ module Hledger.Data.Dates (
   spansSpan,
   spanIntersect,
   spansIntersect,
-  spanIntervalIntersect,
   spanDefaultsFrom,
   spanUnion,
   spansUnion,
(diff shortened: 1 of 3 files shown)


simonmichael:
That sounds alright! And seems to fix the bug. Looks good to me.

PR #1093, opened 2019-09-28
Allow CORS to be enabled in hledger-web

### LINKED ISSUE #1031: Allow Cross-Origin in hledger-web
I am at the point where I'm connecting to hledger-web JSON API endpoints from [Cashier](https://cashier.alensiljak.ml/). At the Sync page, I'm using `http://localhost:5000` server URL.
As expected, the error pops up:

> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://127.0.0.1:5000/accounts. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

Would it be possible to add the CORS header, so that a PWA can connect to a locally-run server (on the same machine)?
I would suggest to use the '*' spec, allowing all sources to access the API:
`Access-Control-Allow-Origin: *`
Open to discussion.

Some references 
- [Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin)

agarciamontoro:
This PR adds a `--cors` option to the hledger-web command, which enables a simple CORS policy: it returns the `Access-Control-Allow-Origin: *` HTTP header.

The implementation was proposed by @MisterY in #1031. I just added the management of the CLI option.

Please, review thoroughly as this is my first ever PR in a Haskell project.

Fixes #1031.

### DIFF
--- a/hledger-web/Hledger/Web/Application.hs
+++ b/hledger-web/Hledger/Web/Application.hs
@@ -10,7 +10,6 @@ module Hledger.Web.Application
 
 import Data.IORef (newIORef, writeIORef)
 import Network.Wai.Middleware.RequestLogger (logStdoutDev, logStdout)
-import Network.Wai.Middleware.Cors
 import Network.HTTP.Client (defaultManagerSettings)
 import Network.HTTP.Conduit (newManager)
 import Yesod.Default.Config
@@ -24,7 +23,7 @@ import Hledger.Web.Handler.UploadR
 import Hledger.Web.Handler.JournalR
 import Hledger.Web.Handler.RegisterR
 import Hledger.Web.Import
-import Hledger.Web.WebOptions (WebOpts(serve_,serve_api_,cors_))
+import Hledger.Web.WebOptions (WebOpts(serve_,serve_api_), corsPolicy)
 
 -- This line actually creates our YesodDispatch instance. It is the second half
 -- of the call to mkYesodData which occurs in Foundation.hs. Please see the
@@ -39,13 +38,11 @@ makeApplication :: WebOpts -> Journal -> AppConfig DefaultEnv Extra -> IO Applic
 makeApplication opts' j' conf' = do
     foundation <- makeFoundation conf' opts'
     writeIORef (appJournal foundation) j'
-    (logWare . corsWare) <$> toWaiApp foundation
+    (logWare . (corsPolicy opts')) <$> toWaiApp foundation
   where
     logWare | development  = logStdoutDev
             | serve_ opts' || serve_api_ opts' = logStdout
             | otherwise    = id
-    corsWare | cors_ opts' = simpleCors
-             | otherwise = id
 
 makeFoundation :: AppConfig DefaultEnv Extra -> WebOpts -> IO App
 makeFoundation conf opts' = do
--- a/hledger-web/Hledger/Web/Application.hs
+++ b/hledger-web/Hledger/Web/Application.hs
@@ -10,6 +10,7 @@ module Hledger.Web.Application
 
 import Data.IORef (newIORef, writeIORef)
 import Network.Wai.Middleware.RequestLogger (logStdoutDev, logStdout)
+import Network.Wai.Middleware.Cors
 import Network.HTTP.Client (defaultManagerSettings)
 import Network.HTTP.Conduit (newManager)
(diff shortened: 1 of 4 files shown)


simonmichael:
@agarciamontoro great, thank you!

Would it be fine to have this always on ? Then we wouldn't need to bother the user with a flag.

agarciamontoro:
@simonmichael 
> Would it be fine to have this always on ? Then we wouldn't need to bother the user with a flag.
Should we add the option to disable it? Or do you rather just make it non-configurable?

simonmichael:
Fewer flags is best, all else being equal. Are there downsides to always serving it ? I don't know much about this topic.

agarciamontoro:
For security concerns, I'd rather have it disabled by default: for example, it makes CSRF attacks a bit easier (although not having CORS enabled is not enough to prevent them, of course).

alerque:
I kind of agree. Going with the more secure defaults out of the box and leaving it up to people that know they are taking a risk and may mitigate it to enable the flag is a better idea than having risky defaults when they won't be needed by many people.

simonmichael:
Could someone remind us of some real scenarios where you'd need this feature, and some threat vectors it opens if you have it on. And, are there cases where it would be reasonable to have it on by default, eg if your url is "localhost" ? (in support of hledger's usual "just works" goal).

simonmichael:
Are there any extra steps we should take in light of the warnings at https://github.com/larskuhtz/wai-cors#note-on-security ?

agarciamontoro:
**Note**: I am not an expert in this _at all_, so please add any more information here that you consider necessary.

## Use case

### Local development of a web app

A developer is building a web application that makes requests from the client to the hledger-web API to interact with the journal.

For example, I am developing a web client in Elm that shows the info from the journal in a friendly way for users that are not used to the terminal.

In my specific development workflow, I launch the hledger-web API, which is listening in port 5000, and I launch my Elm application from a different server using elm-live, which listens in port 8000. As the hledger-web API only accepts requests from the same origin (i.e., localhost:5000), my Elm application, served from localhost:8000 cannot access the API if CORS is not enabled.

### Isolation of services

Once I finish my application I plan to set up everything in the same server, and I will probably make the requests to the hledger-web API from the server itself, not the browser.

However, the previous setup may be still necessary if the requests are made from the client side. This could come handy if the developer wants the services isolated, as in the server serving the application does not have any contact with the server serving the hledger-web API.

## New threat vectors

Enabling CORS means "breaking" (in a controlled way) the Same Origin Policy enforced by all browsers. Sending the `Access-Control-Allow-Origin: *` header tells the browsers to allow any request to the server, whichever the origin of the request is.

This may make it easier for an attacker to set up malicious Javascript code that, e.g., makes a request to the hledger-web API to change the journal without explicit knowledge of the user.

However, this threat vector is in place regardless of the CORS setup, as another common case is an email that has a link that again makes a request to the hledger-web API to change some info. If the user is contacting directly the API server, the CORS setup is irrelevant here: this should be addressed independently by using CSRF tokens, for example.

## On-by-default cases

I don't think it makes sense to enable it without explicit knowledge of the user because of the former discussion.

simonmichael:
Thanks @agarciamontoro. Do you think it should be an option with a value, so you can specify the value of the header, eg to restrict it to just the appopriate domain.

agarciamontoro:
@simonmichael 
> Do you think it should be an option with a value

It would be also useful. I opted for the current implementation because it was simpler and it only added a boolean flag, but I guess that adding the option to explicitly specify the origin is also good. Maybe we could add two flags:

- `--cors <explicit origin>`
- `--simplecors`, an alias of `--cors "*"`

I would need to check how to specify the origin with `wai`, but if you think this is the best option I can try!

simonmichael:
Yes I think we should offer that control. Also https://en.wikipedia.org/wiki/Cross-origin_resource_sharing#Simple_example says something about `*` not allowing authentication or cookies. Let's make it `--cors ARG`, and include `--cors "*"` as an example in the help. I don't think the second no-argument flag is needed.

agarciamontoro:
My latest commit address the previous discussion. Sorry for the delay on this one.

simonmichael:
Sounds good - thanks @agarciamontoro !

Issue #1076, opened 2019-07-26
Skip Footer Lines in CSV Import


rkelly:
TD Ameritrade adds an end-of-file message in their transaction download file. Here's an example:

```
DATE,TRANSACTION ID,DESCRIPTION,QUANTITY,SYMBOL,PRICE,COMMISSION,AMOUNT,NET CASH BALANCE,REG FEE,SHORT-TERM RDM FEE,FUND REDEMPTION FEE, DEFERRED SALES CHARGE
01/01/2019,99999999997,RECURRING ELECTRONIC FUNDING,,,,,999.00,999.00,,,,
..
01/03/2019,99999999999,CASH ALTERNATIVES PURCHASE,,,,,-999.00,0.00,,,,
***END OF FILE***
```

When processing this, hledger-csv errors out:
```
hledger: user error (CSV record ["***END OF FILE***"] has less than two fields)
```

The hledger-csv [Skip function](https://hledger.org/csv.html#skip) only handles headers.

Would be great if we could skip from the footer as well.

simonmichael:
I think adding this to hledger sounds reasonable. What would you call it, `skipend` ?

I assume your footer is a fixed size. For cases where the number of lines can vary, perhaps we'd want something like an `end` directive (`if END OF FILE\n  end`).

adept:
If one could write:
```
if REFEX
  skip 3
```
this would've solved it, right?

simonmichael:
True, and that's something we want. I imagine some cases where there's no reliable REGEX though.

simonmichael:
..but we could start with that, and see.. ?

rkelly:
> I think adding this to hledger sounds reasonable. What would you call it, `skipend` ?
> 
> I assume your footer is a fixed size. For cases where the number of lines can vary, perhaps we'd want something like an `end` directive (`if END OF FILE\n end`).

This one does appear to be just one line, and the same text every time.

`skipend` seems OK. Or `skipfooter`. An `end` or `skip` conditional directive would be OK too.

This is the most problematic download file I've seen (having gotten several others to work fine with `hledger-csv`). Footers aren't even in the RFC 4180 spec. I will bring this up with the bank's tech support.

In the meantime, I'm preprocessing the file like this:
`$ head -n -1 transactions.csv | hledger --file=- --rules-file=transactions.csv.rules print`

simonmichael:
[...]

Issue #1001, opened 2019-04-03
empty csv field assignment consumes next line


simonmichael:
This should assign "" to amount, but instead it assigns "balance %5":
```
amount
balance %5
```

simonmichael:
Consuming of next line is fixed by #1095.

Issue #1000, opened 2019-04-03
should be possible to use balance assertions/assignments when reading csv


simonmichael:
From apo on IRC: [balance assertions](https://hledger.org/journal.html#balance-assertions)/[balance assignments](https://hledger.org/journal.html#balance-assignments) are not [being applied](https://hledger.org/csv.html#csv-balance-assertionsassignments) when reading CSV. Also, it should let you omit the `amount` field assignment if there's a `balance` field assignment.
```
2018-10-15,100
2018-10-16,200
2018-10-17,300
```
```
fields date,bal
# Currently an amount field is required, but we can leave it blank.
# The following blank line is required to avoid a parsing bug.
amount 

balance EUR %bal
date-format %Y-%m-%d
description Assets Update
account1 assets
account2 income
```
```
$ hledger -f a.csv reg assets
2018/10/15 Assets Update           assets                             0             0
2018/10/16 Assets Update           assets                             0             0
2018/10/17 Assets Update           assets                             0             0
```
```
$ hledger -f a.csv print -x
2018/10/15 Assets Update
    assets     = EUR 100
    income

2018/10/16 Assets Update
    assets     = EUR 200
    income

2018/10/17 Assets Update
    assets     = EUR 300
    income

```
We should see:
```
$ hledger -f a.csv print -x
2018/10/15 Assets Update
    assets         EUR 100 = EUR 100
    income        EUR -100

2018/10/16 Assets Update
    assets         EUR 100 = EUR 200
    income        EUR -100

2018/10/17 Assets Update
    assets         EUR 100 = EUR 300
    income        EUR -100

```
You can work around by parsing a second time:
```
$ hledger -f a.csv print | hledger -f- print -x
```

simonmichael:
Related (tested with 61e8ccb):

- `currency` should not add a currency symbol when the amount field is empty

- amount*/balance* rule validation should be more robust, eg these malformed rules:
       ```
       amount-in %3
       balance %4
       ```
       give the following error:
       ```
       hledger: found values for amount and for amount-in/amount-out
       please use either amount or amount-in/amount-out
       ```
       which should be more precise, and possibly should reuse the existing
       ```
       Please specify (as a top level CSV rule) either the amount field,
       both the amount-in and amount-out fields, or the balance field. Eg:
       amount %2
       ```

simonmichael:
@adept: this is really #1095 followup, but I'll post here: this issue
(CSV rules that set balance assignments) provides an example of how
[...]

Issue #841, opened 2018-07-16
newlines in CSV data aren't handled right


simonmichael:
Our CSV parser does allow newlines in data if the field is double-quoted, as can happen eg in Coinbase's `notes` field, but assigning this to a transaction field like `description` will produce a malformed journal entry. It should probably convert multi-line values to one line.

PR #1095, opened 2019-10-12
Csv fixes mega pack 2019

adept:
Let's fix some outstanding CSV issues. This PR fixes:

- #841 and #416 : newlines in description/comment/precomment/... are not breaking output
- #570 : zero amount-in and non-zero amount-out (or vice versa) is no longer an error
- #500 : recursive interpolation works for all fields, including amount
- #1000: it is possible to use just balance assertions when reading csv
- #1001: empty field assignments are possible now
- #1076: it is now possible to write "if" expressions with "skip" in the body that will skip records
- #627 : csv rules could produce up to 9 postings, with different comments, balance assertions or currencies for each.

Obviously, the biggest change is multi-postings generation. I tried to streamline the code, so hopefully this bit is not too hard to read. 

Parser now recognizes "account1", "amount1", "amount1-in", "amount1-out", "comment1", "balance1", "currency1" and likewise with "2", "3", ... up to "9" and will create postings for all sets of values that contain sensible data. Legacy names "account", "amount-in", "amount-out", "balance" are recognized as well, and parser tries to resolve conflicts between those and "first" set of fields. I've added some tests to demonstrate this

Current code tries really hard to be backwards-compatible, so, for example, it uses "currency" as currency for all amounts unless overwritten by more specific "currencyX" and still tries to assign account "income:unknown" or "expenses:unknown" to posting 2 (even if there are more than two postings in the transaction), which might be bad/surprising. I am not sure what is the best way to proceed here. It would be nice to just drop backwards compatibility and aim for "least surprise", but I am not sure how feasible this is.

For example, one option could be to set account to "unknown" for any posting that has amount or balance assertion defined, but no account. This will break backwards compatibility wrt "income:unknown", but will be more resilient. What do you think?

Docs are not updated yet, but I am opening this up for an early review nevertheless.

### DIFF
--- a/hledger-lib/hledger_csv.m4.md
+++ b/hledger-lib/hledger_csv.m4.md
@@ -253,11 +253,8 @@ The order of same-day entries will be preserved
 
 ## CSV accounts
 
-Each journal entry will have at least two [postings](journal.html#postings), usually to `account1` and `account2`.
-
-It is entirely up to you which account field to assign to the account whose CSV we are reading, however you might
-consider assigning it to the last posting (`account2` in the simplest case). This way, if the CSV contrains transactions in multiple currencies,
-and you don't provide posting [prices](journal.html#transaction-prices) explicitly they will be priced in the currency of the last posting.
+Each journal entry will have at least two [postings](journal.html#postings), to `account1` and some other account (usually `account2`).
+It's conventional and recommended to use `account1` for the account whose CSV we are reading.
 
 ## CSV amounts
 
--- a/hledger-lib/hledger_csv.m4.md
+++ b/hledger-lib/hledger_csv.m4.md
@@ -38,7 +38,7 @@ At minimum, the rules file must identify the date and amount fields.
 It's often necessary to specify the date format, and the number of header lines to skip, also.
 Eg:
 ```
-fields date, _, _, amount
+fields date, _, _, amount1
 date-format  %d/%m/%Y
 skip 1
 ```
@@ -55,7 +55,7 @@ A more complete example:
 skip 1
 
 # name the csv fields (and assign the transaction's date, amount and code)
-fields date, _, toorfrom, name, amzstatus, amount, fees, code
+fields date, _, toorfrom, name, amzstatus, amount1, fees, code
 
 # how to parse the date
 date-format %b %-d, %Y
@@ -64,7 +64,7 @@ date-format %b %-d, %Y
 description %toorfrom %name
 
 # save these fields as tags
-comment     status:%amzstatus, fees:%fees
+comment     status:%amzstatus
 
 # set the base account for all transactions
 account1    assets:amazon
@@ -72,6 +72,9 @@ account1    assets:amazon
 # flip the sign on the amount
 amount      -%amount
 
+# Put fees in a separate posting
+amount3     %fees
+comment3    fees
 ```
 
 For more examples, see [Convert CSV files](https://github.com/simonmichael/hledger/wiki/Convert-CSV-files).
@@ -140,7 +143,7 @@ An assignment to any of `accountN`, `amountN`, `amountN-in`, `amountN-out`, `bal
 
 Fields `amount`, `amount-in`, `amount-out`, `currency`, `balance` and `comment` are treated as aliases for `amount1`, and so on. If your rules file leads to both aliased fields having different values, `hledger` will raise an error.
 
-You need to provide enough information to create at least two postings.
+You need to provide enough information to create at least one posting.
 
 Eg:
 ```rules
@@ -150,7 +153,7 @@ Eg:
 # CSV field:
 #      1     2            3 4       5 6 7          8
 # entry field:
-fields date, description, , amount, , , somefield, anotherfield
+fields date, description, , amount1, , , somefield, anotherfield
 ```
 
 ## field assignment
@@ -178,12 +181,20 @@ Note, interpolation strips any outer whitespace, so a CSV value like
 ## conditional block
 
 `if` *`PATTERN`*\
-&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS` or `skip N` or `skip end`*...
+&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS`*...
 
 `if`\
 *`PATTERN`*\
 *`PATTERN`*...\
-&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS` or `skip N` or `skip end`*...
+&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS`*...
+
+`if` *`PATTERN`*\
+*`PATTERN`*...\
+&nbsp;&nbsp;&nbsp;&nbsp;*`skip N`*...
+
+`if` *`PATTERN`*\
+*`PATTERN`*...\
+&nbsp;&nbsp;&nbsp;&nbsp;*`skip end`*...
 
 This applies one or more field assignments, only to those CSV records matched by one of the PATTERNs.
 The patterns are case-insensitive regular expressions which match anywhere
@@ -192,7 +203,7 @@ specific field).  When there are multiple patterns they can be written
 on separate lines, unindented.
 The field assignments are on separate lines indented by at least one space.
 
-Instead of field assignments you can specify `skip N` to skip the next N records (including the one that matchied). Special form `skip end` will cause the rest of the file to be skipped.
+Instead of field assignments you can specify `skip N` to skip the next N records (including the one that matchied) or `skip end` to skip the rest of the file.
 
 Examples:
 ```rules
@@ -242,22 +253,21 @@ The order of same-day entries will be preserved
 
 ## CSV accounts
 
-Each journal entry will have two [postings](journal.html#postings), to `account1` and `account2` respectively.
-It's not yet possible to generate entries with more than two postings.
+Each journal entry will have at least two [postings](journal.html#postings), to `account1` and some other account (usually `account2`).
 It's conventional and recommended to use `account1` for the account whose CSV we are reading.
 
 ## CSV amounts
 
-A transaction [amount](journal.html#amounts) must be set, in one of these ways:
+A posting [amount](journal.html#amounts) could be set in one of these ways:
 
-- with an `amount` or `amount1` field assignment, which sets the first posting's amount
+- with an `amountN` field assignment, which sets the Nth posting's amount
 
 - (When the CSV has debit and credit amounts in separate fields:)\
-  with field assignments for the `amount-in` and `amount-out` (or `amount1-in` and `amount1-out`) pseudo
+  with field assignments for the `amountN-in` and `amountN-out` pseudo
   fields (both of them). Whichever one has a value will be used, with
   appropriate sign. If both contain a value, it might not work so well.
 
-- or implicitly by means of a [balance assignment](journal.html#balance-assignments) (see below).
+- with `balanceN` field assignment that creates a [balance assignment](journal.html#balance-assignments) (see below).
 
 There is some special handling for sign in amounts:
 
@@ -265,13 +275,13 @@ There is some special handling for sign in amounts:
 - If an amount value begins with a double minus sign, those will cancel out and be removed.
 
 If the currency/commodity symbol is provided as a separate CSV field,
-assign it to the `currency` pseudo field; the symbol will be prepended
+assign it to the `currency` pseudo field (applicable to the whole transaction) or `currencyN` (applicable to Nth posting only); the symbol will be prepended
 to the amount 
 (TODO: <s>when there is an amount</s>).
-Or, you can use an `amount` [field assignment](#field-assignment) for more control, eg:
+Or, you can use an `amountN` [field assignment](#field-assignment) for more control, eg:
 ```
-fields date,description,currency,amount
-amount %amount %currency
+fields date,description,currency,amount_
+amount1 %amount_ %currency
 ```
 
 ## CSV balance assertions/assignments
--- a/hledger-lib/hledger_csv.m4.md
+++ b/hledger-lib/hledger_csv.m4.md
@@ -92,7 +92,7 @@ You'll need this whenever your CSV data contains header lines. Eg:
 <!-- XXX -->
 <!-- hledger tries to skip initial CSV header lines automatically. -->
 <!-- If it guesses wrong, use this directive to skip exactly N lines. -->
-This can also be used in a conditional block (without numeric argument) to ignore certain CSV records.
+This can also be used in a conditional block to ignore certain CSV records.
 ```rules
 # ignore the first CSV line
 skip 1
@@ -178,12 +178,12 @@ Note, interpolation strips any outer whitespace, so a CSV value like
 ## conditional block
 
 `if` *`PATTERN`*\
-&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS or skip`*...
+&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS or skip N`*...
 
 `if`\
 *`PATTERN`*\
 *`PATTERN`*...\
-&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS or skip`*...
+&nbsp;&nbsp;&nbsp;&nbsp;*`FIELDASSIGNMENTS or skip N`*...
 
 This applies one or more field assignments, only to those CSV records matched by one of the PATTERNs.
 The patterns are case-insensitive regular expressions which match anywhere
@@ -192,7 +192,7 @@ specific field).  When there are multiple patterns they can be written
 on separate lines, unindented.
 The field assignments are on separate lines indented by at least one space.
 
-Instead of field assignments you can specify `skip` to skip the matching record.
+Instead of field assignments you can specify `skip N` to skip the next N records (including the one that matchied).
 
 Examples:
 ```rules
--- a/hledger-lib/hledger_csv.m4.md
+++ b/hledger-lib/hledger_csv.m4.md
@@ -253,8 +253,11 @@ The order of same-day entries will be preserved
 
 ## CSV accounts
 
-Each journal entry will have at least two [postings](journal.html#postings), to `account1` and some other account (usually `account2`).
-It's conventional and recommended to use `account1` for the account whose CSV we are reading.
+Each journal entry will have at least two [postings](journal.html#postings), usually to `account1` and `account2`.
+
+It is entirely up to you which account field to assign to the account whose CSV we are reading, however you might
+consider assigning it to the last posting (`account2` in the simplest case). This way, if the CSV contrains transactions in multiple currencies,
+and you don't provide posting [prices](journal.html#transaction-prices) explicitly they will be priced in the currency of the last posting.
 
 ## CSV amounts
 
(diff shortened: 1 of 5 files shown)


tests/csv.test
```diff
@@ -111,6 +111,25 @@ $  printf 'fields date, description, amount, balance1, balance2\ndate-format %%d
...
+# 12. More than two postings and different currencies 
+$  printf 'fields date, description, amount, balance1, balance2, currency3, amount3,comment3\ndate-format %%d/%%Y/%%m\ncurrency $\naccount1 assets:myacct\naccount3   expenses:tax\n' >t.$$.csv.rules; printf '10/2009/09,Flubber Co,50,321,123,£,0.234,VAT\n' | hledger -f csv:- --rules-file t.$$.csv.rules print && rm -rf t.$$.csv.rules
+2009/09/10 Flubber Co
+    assets:myacct               $50 = $321
+    expense:unknown     = $123
+    expenses:tax             £0.234  ; VAT
+
+>=0
+
+
```
simonmichael: Looking good!

I wonder how we can write these tests so that the other input files' content is more visible. They are hard to read.

How well does this handle the various validity checks I (hope I) mentioned on the original issue ? Eg requiring the numbers to be consecutive, starting at 1, etc. etc. ?

simonmichael: https://github.com/simonmichael/hledger/issues/627#issuecomment-342315800 -> “Some things to check..”
Maybe total validity checking isn’t important, I’m just hoping for good error messages and easy troubleshooting so this doesn’t cause confusion.

adept:
> Is it always safe for us to be ignoring a 0 value like this, do you think ? What happens if both amounts are 0, or both are empty, or both are non zero ?

I claim that it is safe. We have four cases:

1. in/=0, out/=0. Impossible to interpret, and we will error out
2. in=0, out/=0. We take "out", and this seems the only possible logical interpretation
3. in/=0, out=0. We take "in", and this seems the only possible logical interpretation
4. in=0, out=0. The only interpretation is "no amount change", and this is what the code will do - treat it as if the amount is unassigned as there was no amount change. If there was a balance assertion on same line, we will take it and try and make a posting of it all, but if not - this line is effectively no-op and will be ignored. I am open to explanations how it could be anything else but no-op :)

tests/csv.test
adept: > I wonder how we can write these tests so that the other input files' content is more visible. They are hard to read.

I've pushed test 16 that does away with the use of printf and allows one to lay out csv file and rules file on several lines with no extra quoting/escaping. What do you think?

> How well does this handle the various validity checks I (hope I) mentioned on the original issue ? Eg requiring the numbers to be consecutive, starting at 1, etc. etc. ?

I think that there is no need for this. Csv parser should not attempt to be holier than thou - if input+rules produces output that is impossible to balance, then it is on the user to fix the rules or fix the data. The parser itself should be as small as possible - any validity checks would just introduce opportunities to make subtle bugs that will prevent csv parsing. Plus validity check would have to always be in line with what Journal parser actually allows (or not), and as Journal evolves I see no easy way to ensure this.

Eventually, validity checks will get out of sync with journal parser and will start to produce false rejects. 

If users want to immediately validate the output of csv parser, they could just pipe it into "hledger -f - print", and we could mention this in the docs and that is it.

adept:
> Would this make a double space with windows line endings ? And multiple spaces when there’s consecutive newlines ? Could limit it to one space ?

Removing consequtive breaks and replacing with a single space seems sensible. I amended the test and committed the fix.

tests/csv.test
adept: > I wonder how we can write these tests so that the other input files' content is more visible. They are hard to read.

I've pushed test 16 that does away with the use of printf and allows one to lay out csv file and rules file on several lines with no extra quoting/escaping. What do you think? If you like it, i will update the rest of the tests to match

simonmichael:
I was going to ask if case 4 (both zero) should produce an explicit zero but I’m less clear after seeing the comment edit. But I believe you’re clear so that’s fine.

simonmichael:
I’ll do a little real world testing soon.

simonmichael:
Maybe no checking will be less painful overall - we’ll find out. Ok on not validating for now.

simonmichael:
Here's a first experience report, for further study. I'm using the new feature to generate a third, fees posting for a paypal donation. (It's miscategorised as an expense, ignore that.)

test command, shows generated entry and if it is parseable and if so the same entry made explicit:

    $ ls a.csv* | entr bash -c 'hledger -f a.csv print && (hledger -f a.csv print | hledger -f- print -I -x 2>/dev/null) && echo ok || echo not ok'

I initially had:

rules (fragment):

    fields date, time, timezone, description, type, status_, currency, grossamount, feeamount, netamount, fromemail, toemail, code, itemtitle, itemid, referencetxnid, receiptid, balance, note
    amount %netamount

result:

    2018/12/22 (12U43407AR385363D) Someone for Joyful Systems
        sm:assets:online:paypal           $9.41 = $57.60
        sm:expenses:unknown

    2018/12/22 (12U43407AR385363D) Someone for Joyful Systems
        sm:assets:online:paypal           $9.41 = $57.60
        sm:expenses:unknown              $-9.41

    ok

I added amount2:

rules:

    amount %netamount
    amount2 %feeamount

result:

    2018/12/22 (12U43407AR385363D) Someone for Joyful Systems
        sm:assets:online:paypal           $9.41 = $57.60
        sm:expenses:unknown              $-0.59

    not ok

I changed that to amount3, defined account3:

rules:

    amount3 %feeamount
    account3 JS:expenses:banking:paypal

result:

    2018/12/22 (12U43407AR385363D) Someone for Joyful Systems
        sm:assets:online:paypal              $9.41 = $57.60
        sm:expenses:unknown
        JS:expenses:banking:paypal          $-0.59

    2018/12/22 (12U43407AR385363D) Someone for Joyful Systems
        sm:assets:online:paypal              $9.41 = $57.60
        sm:expenses:unknown                 $-8.82
        JS:expenses:banking:paypal          $-0.59

    ok

I renamed amount to amount1:

rules:

    amount1 %netamount

result:

    2018/12/22 (12U43407AR385363D) Someone for Joyful Systems
    hledger: amount/amount-in/amount-out and amount1/amount1-in/amount1-out produced conflicting values
    the CSV record is:       "12/22/2018","06:22:50","PST","Someone","Subscription Payment","Completed","USD","10.00","-0.59","9.41","someone@some.where","simon@joyful.com","12U43407AR385363D","Joyful Systems","","9KCXBV2DBXXAX","","57.60",""
    the account1 rule is: sm:assets:online:paypal
    the amount1 rule is: %netamount
    the account2 rule is: sm:expenses:unknown
    the account3 rule is: JS:expenses:banking:paypal
    the amount3 rule is: %feeamount
    the balance rule is: %18
    the code rule is: %13
    the currency rule is: $
    the date rule is: %1
    the description rule is: %description for %itemtitle 

    amount/amount-in/amount-out is Mixed []
    amount1/amount1-in/amount1-out isMixed [Amount {acommodity = "$", aquantity = 9.41, aismultiplier = False, astyle = AmountStylePP "L False 2 Just '.' Nothing..", aprice = Nothing}]
                                     ^
                                     TODO: add space
    not ok

This last one stumped me, can you tell how it's going wrong ?

simonmichael:
(PS and maybe you can fix that missing space (TODO:), I didn't want to conflict)

adept:
> This last one stumped me, can you tell how it's going wrong ?
Yep, thats a bug. Fixed, and fixed error formatting as well

adept:
Added fix for #1076 and #1000 got fixed as a side-effect of introducing muti-postings

simonmichael:
> Added fix for #1076 and #1000 got fixed as a side-effect of introducing muti-postings

Oh, excellent.

About doc updates.. would it be fair to say that "an assignment to any of accountN, amountN, amount-inN, amount-outN or currencyN will generate a posting (though it's your responsibility to ensure it is a well formed one). Normally the N's are consecutive starting from 1 (but it's not required). One posting will be generated for each unique N."

adept:
> About doc updates.. would it be fair to say that "an assignment to any of accountN, amountN, amount-inN, amount-outN or currencyN will generate a posting (though it's your responsibility to ensure it is a well formed one). Normally the N's are consecutive starting from 1 (but it's not required). One posting will be generated for each unique N."

Very sensible. Applied with slight rewording

simonmichael:
I'm using conditional skip successfully to ignore trailing junk. Here, ignoring everything after the first record with an empty value:
```
if ^,
 skip 999999
```
Issue: it only seems to skip one line, ignoring the argument.
Is it worth adding a "end" directive to provide certainty that it works even with an extra-long csv file ?

simonmichael:
Hmm.. I think it currently requires `amount` (no number) to always be assigned:
```
amount1  0
```
```
Please specify (as a top level CSV rule) either the amount field,
both the amount-in and amount-out fields, or the balance field. Eg:
amount %2
```

adept:
> Issue: it only seems to skip one line, ignoring the argument.

That's because my initial idea for `skip` implementation made implementing `skip N` to be hard, so I did not do it. But my subsequent idea actually made it easier, so i just pushed implementation for `skip N`, and your example should work.

adept:
> Hmm.. I think it currently requires amount (no number) to always be assigned:

Yep, that's a bug. Fixed!

simonmichael:
- Nice. Oh, I wasn't thinking `skip end`.. could it be just `end` ?

- Now one of `amount` or `amount1` is required, which doesn't match "Normally the N's are consecutive starting from 1 (but it's not required).". Probably any amount[N] should be accepted here, right ? But actually just checking for one amount isn't as useful now. We should probably just scrap the amount checking (or do more thorough checking), eh ?

adept:
> Nice. Oh, I wasn't thinking skip end.. could it be just end ?

I'd rather keep "skip end", otherwise you need to handle a case when both "skip N" and "end" are in effect, possibly through two different "if" blocks. With "skip end", ambiguous cases like those are made impossible :)

adept:
> Now one of amount or amount1 is required, which doesn't match "Normally the N's are consecutive starting from 1 (but it's not required).". Probably any amount[N] should be accepted here, right ? But actually just checking for one amount isn't as useful now. We should probably just scrap the amount checking (or do more thorough checking), eh ?

I part-wrote/part-kept a piece of code that forces backwards-compatible behavior when just two postings are generated: if there are just two postings, and `account2` is not defined, then its name would be income:unknown or expense:unknown, and it would be driven by `amount1`. If we are happy with generic name like "unknown" when amount is not known, then this special case could be removed, and requirement to always define amount1 could be lifted.

Worth noting that docs currently says that you always need first posting, but after that you can go wild...

simonmichael:
> I'd rather keep "skip end", otherwise you need to handle a case when both "skip N" and "end" are in effect, possibly through two different "if" blocks. With "skip end", ambiguous cases like those are made impossible :)

Good point, but consider a csv record matched by two or more conditionals, each doing a skip N. That's already a bit ambiguous, right ? I guess it will skip the sum of the Ns, which probably isn't useful, but never mind. 

Isn't "end" a little clearer ? It's a simple "stop reading further input now" command, if reached the rest of the rules don't matter.

simonmichael:
> If we are happy with generic name like "unknown" when amount is not known, then this special case could be removed, and requirement to always define amount1 could be lifted.

I think that would be fine, if it makes things simpler overall. Eg:
"If a posting's account is not assigned, it defaults to 'income:unknown' or 'expenses:unknown' depending on the sign of the posting's amount (or if the posting's amount is not assigned, it will be just 'unknown')."

However things still seem a bit fragile. 
hledger tests the amount of posting 1,
and assumes that  account 1 is the base/source account for the CSV (I think we encourage this in the docs),
and that it is an asset account,
...so that if amount1 is negative, it knows to choose "income". (NB: inconsistent; we mostly use "revenue" elsewhere.)
With multiple amounts now and more freedom, it's going to be easy to confuse this.
Should it just always pick "expenses:unknown" ?

simonmichael:
I think recursive field assignments are not quite there yet ?
I believe you made it possible for a field assignment to reference a
field of the same name assigned via field list:

```
2019-01-01,a,1
```
```
fields   date, account1, amount
account1 %account1 x
```
```
2019/01/01
    a x                          1
    income:unknown
```

But field assignments still can't reference each other, eg to
incrementally build up a value; all but the last one are ignored:

```
fields   date, account1, amount
account1 %account1 x
account1 %account1 y
account1 %account1 z
```
```
2019/01/01
    a z                          1
    income:unknown
```

A separate issue: I think "account" (no number) has regressed, eg this is no longer parsed:
```
account  a
```

adept:
> Good point, but consider a csv record matched by two or more conditionals, each doing a skip N. That's already a bit ambiguous, right ? I guess it will skip the sum of the Ns, which probably isn't useful, but never mind.

> But field assignments still can't reference each other, eg to
> incrementally build up a value; all but the last one are ignored:
> 
> fields   date, account1, amount
> account1 %account1 x
> account1 %account1 y
> account1 %account1 z

Let me try and address both of these points at the same time.
The code effectively has two separate parts:

1. Field assignment selection: how do we choose which fee assignments to apply
2. Field assignment value resolution: how do we compute the value to be assigned to the field

This PR changes (2), but does not change (1) at all. Furthermore, change for (2) in this PR is very limited: code already allowed field value to recursively refer to the field name, so you get limited one-step recursion. This PR simply fixed a small issue where this recursive application mistakenly assumed that grammar for field names are more limited that it really is.

(1) still operates the same way it did before: if there are multiple field assignments in the rule file at the top level, subsequent field assignment override earlier ones. If there are multiple matching "if", field assignment in the subsequent ones override earlier ones. Only when (1) is complete, at the very end, step (2) kicks in and evaluates *values* for selected assignment.

This covers both "skip" and "account1" examples from above. For "skip" the last assignment will win, and for "account1" the last assignment wins.

Changing (1) will probably require separate PR, as we will move from declarative field assigment language to imperative field assignment language. Suddenly order of field declarations start to matter way more than it currently is. I do not immediately see a clear model to do what you want in your `account1` example. Consider a couple of puzzles:
```
fields   date, account1, account2, amount
if SOME COND
   account1 %account2
   account2 %account1
```

Currently this swaps two accounts which seems like a nice thing to have. In the proposed model this will require third field, something like:
```
fields   date, account1, account2, amount
if SOME COND
   account9 %account2
   account2 %account1
   account1 %account9
   account9
```

Or this:
```
fields   date, account1, amount
account1 %account1 x
account1 %account1 y

if SOME COND
  account1 %account1 z
  
account1 %account1 t
```
For row that does not match "if" you expect `account1` to be "<...> x y z". What about the row that matches though?  Do you expect it to be "<...> x y z t" or "<...> x y t z"?
Now lets consider include files which could have their own top-level field assignment statements and "if" statements....

Are you sure that a change to introduce this level of complexity is warranted?
Even if it is, I kinda feel that it should be in its own PR.

adept:
Regarding "income:expense" and "account" regression - i will check this later today

simonmichael:
> This PR changes (2), but does not change (1) at all. Furthermore, change for (2) in this PR is very limited: code already allowed field value to recursively refer to the field name, so you get limited one-step recursion. This PR simply fixed a small issue where this recursive application mistakenly assumed that grammar for field names are more limited that it really is.
> 
> (1) still operates the same way it did before: if there are multiple field assignments in the rule file at the top level, subsequent field assignment override earlier ones. If there are multiple matching "if", field assignment in the subsequent ones override earlier ones. Only when (1) is complete, at the very end, step (2) kicks in and evaluates values for selected assignment.
> 
> This covers both "skip" and "account1" examples from above. For "skip" the last assignment will win, and for "account1" the last assignment wins.
> 
> Changing (1) will probably require separate PR, as we will move from declarative field assigment language to imperative field assignment language. Suddenly order of field declarations start to matter way more than it currently is. I do not immediately see a clear model to do what you want in your account1 example. Consider a couple of puzzles:
> 
Ok.. if so I had a wrong mental model of csv rule evaluation. My mental model has always been the imperative one, where rules are applied in the order they appears in the rules file, if they can be applied, modifying the assignments as they go. In your examples:
> fields   date, account1, account2, amount
> if SOME COND
>    account1 %account2
>    account2 %account1
> Currently this swaps two accounts which seems like a nice thing to have.
> 
I would not have expected that to work, I would have expected to get %account2's original value in both.
> In the proposed model this will require third field, something like:
> 
> fields   date, account1, account2, amount
> if SOME COND
>    account9 %account2
>    account2 %account1
>    account1 %account9
>    account9
Yes, or 

    fields date, account1_, account2_, amount
    if COND
      account1 account2_
      account2 account1_

> Or this:
> 
> fields   date, account1, amount
> account1 %account1 x
> account1 %account1 y
> 
> if SOME COND
>   account1 %account1 z
>   
> account1 %account1 t
> For row that does not match "if" you expect account1 to be "<...> x y z". What about the row that matches though? Do you expect it to be "<...> x y z t" or "<...> x y t z"?
> 
I would have expected ... x y z t.
> Now lets consider include files which could have their own top-level field assignment statements and "if" statements....
> 
I think my mental model covers that do - rules are processed in order, just as if the include files were inlined (I think they are).
> Are you sure that a change to introduce this level of complexity is warranted?
> 
Not entirely.

Csv rules are supposed to be simple, intuitive, non-surprising, fairly hard to screw up, not turing complete. As we enhance them there's a risk of compromising that, perhaps in ways I don't notice until later. So I appreciate your questioning.

I'm still inclined to think the imperative model would be ideal, because

1. it's what most users will find natural.
2. it's consistent with how we process journals.
3. I have wished for it multiple times in practice, eg to let multiple conditional rules add things to the comments field.
4. I feel it does handle all examples so far, unambiguously.

I just tried to describe your model again, for 1.15, or 1.15 + this PR, and actually I'm not clear enough to write it down with confidence. 

Doesn't this PR currently add some "imperativity" ? Since a standalone field assignment can override a field list assignment. But it's only partial - users can't generally reassign things. It will probably be hard to remember when you can reassign, or assign to self, and when you can't. It might be ok if we document it as a special case.

Whichever model we are using, we should make sure it's clear to users.

> Even if it is, I kinda feel that it should be in its own PR.
> 
That's totally fair. Whatever this PR ends up doing, could you do me a favour and check that the manual describes that behaviour accurately, and also draft the changelog entries for this PR. (Rough bullet points is fine, in hledger/CHANGES or here.)

adept:
> A separate issue: I think "account" (no number) has regressed, eg this is no longer parsed:
> ```
> account  a
> ```

Current master does not support `account`, only `account1` and `account2`. Is it worth adding it for consistency?/////

adept:
> Doesn't this PR currently add some "imperativity" ? Since a standalone field assignment can override a field list assignment.

No, I don't think so. As I said, recursive interpolation already worked before this PR, but just for fields with names conforming to regexp `%[a-zA-Z]+`. This PR simply extends this to all possible fields (with dashes, underscores, etc).

hledger-lib/hledger_csv.m4.md
```diff
@@ -279,8 +279,8 @@ to the amount
 (TODO: <s>when there is an amount</s>).
 Or, you can use an `amountN` [field assignment](#field-assignment) for more control, eg:
 ```
-fields date,description,currency,amount_
-amount1 %amount_ %currency
+fields date,description,currency,amount1
+amount1 %amount1 %currency
```
simonmichael: Seems ok, but now we are showing them that in a field assignment you can reference the same field name previously assigned by a field list, and add to its value. This would lead me (did lead me) to think that I can do the same thing with two field assignments, which is not the case. This is the "special case" I think we need to mention somehow.

simonmichael: But actually this whole manual is in need of a cleanup, so let's not worry about it right now.

simonmichael:
Following our extensive IRC chats.. #1095 has landed! Thank you very much @adept ! \o/

PR #1102, opened 2019-10-18
Make resolver in hledger-install.sh consistent with that in stack.yaml

bwignall:
All three results in "git grep resolver | grep 14" now say 14.4.


simonmichael:
Thanks!

PR #1103, opened 2019-10-18
Fix internal Markdown links in CONTRIBUTING.md

bwignall:
https://github.com/simonmichael/hledger/blob/master/CONTRIBUTING.md and https://hledger.org/CONTRIBUTING.html were displaying [[...]] literals


simonmichael:
Thanks!

Issue #901, opened 2018-10-11
accounts --anon doesn't anonymise account names declared with account directives


simonmichael:
```
account a
```
```
$ hledger accounts
a
$ hledger accounts --anon
a
```

ony:
Sounds like we anonymize journal entries only.

ony:
To fix cashflow/balancesheet/etc we need to re-factor our current reporting in favor building default account type mappings from regexps instead of building queries directly.

simonmichael:
That looks nice. Do you think there's any advantage to putting it in hledger-lib ?

ony:
As an advantage, I see that we can define instances close to data types and thus less chances that they will go out of sync.

My initial idea was to use default constructors and if new field appears we just ignore it instead of copying it without anonymization. But this require more effort.

PR #1105, opened 2019-10-19
Fix some missing anonymization

ony:
This fixes account anonymization ( #901 ) and transaction code.

_(here was a big banner)_


simonmichael:
Thanks!

PR #1106, opened 2019-10-19
Tiny re-factor RawOpts to hide internal representation

ony:
This was triggered by hardships of addressing #1104 .

- Hide `RawOpts` so no one can mess with it other than using `RawOptions` module.
- Remove obsolete `decodeRawOpts` and `SystemString`.
- Factor out few common ways of accessing `RawOpts` to `choiceopt` and `collectopts` in `RawOptions`.

With these changes confirmed that we use quotes only to build query string.

_(here was a big banner)_

### DIFF
--- a/hledger-lib/Hledger/Data/RawOptions.hs
+++ b/hledger-lib/Hledger/Data/RawOptions.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DeriveDataTypeable #-}
+
 {-|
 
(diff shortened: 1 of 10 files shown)


simonmichael:
Looks nice. Just needs a bit of cleanup to avoid warnings. `make buildtest`/`make help-buildtest` can help.

simonmichael:
Hooray! Thanks a lot.

PR #1119, opened 2019-11-12
lib: csv reader gets balance-style directive

adept:
This implements a suggestion from #1000: balance assertion style in CSV reader is now fully configurable


simonmichael:
This is great. Although "balance" is an overloaded term in our world, I agree that "balance-*" is good here because it relates to "balance", and it covers both "balance assertions" and "balance assignments". But would you consider changing "style" to "type" ? So far we have only used "style" for cosmetic/display-type things, and this is more than that.

adept:
Yeah, this makes sense. Renamed.

PR #1123, opened 2019-11-16
Drop amount elision in showing last posting

ony:
This behavior is highly depends on journal. If we want to re-introduce
    it we'd better re-consider how transaction entry can be "simplified".
    I.e. besides dropping last amount we may drop prices that can be assumed
    implicitly.
    Note that there is no need to knit it into showTransaction since it
    easily achievable with pre-processing (similar to implicit balances
    etc).


simonmichael:
Thanks for the cleanup! Maybe somebody will want to do this in future, but let's apply YAGNI.

Issue #319, opened 2016-01-29
Support for ledger's -%/--percent flag


matze:
To determine the relative allocation of certain accounts I often use ledger-cli's `-%`/`--percent` flag which is used to

> Calculate the percentage value of each account in balance reports. Only works for accounts that have a single commodity.

It would be a really nice if this could find its way into hledger.
##

simonmichael:
Agreed!

narendraj9:
Are there any plans to implement this? This is very important because looking at the numbers doesn't really tell much if we don't know the relative allocation. And it seems that there is no support for sorting the balances as well. Currently, I sort them once I have the output.

simonmichael:
These are not on my short list, pull requests welcome. For sorting balances, hledger bal | sort -n +1 perhaps ?

narendraj9:
I have written a major mode for it, I do the sorting inside Emacs sort-numeric-fields function.

https://github.com/narendraj9/dotfiles/blob/master/.emacs.d/packages/rest/hledger-mode/hledger-reports.el#L191

narendraj9:
What is the short list? I might be missing things that are important and your answer might help me understand managing things better.

simonmichael:
Oh, nice.

My priority in hledger right now is to deliver the next release, so my list is pretty much this: https://github.com/simonmichael/hledger/milestone/4

narendraj9:
Okay. Thanks. I might want to try to implement this. Which files should I start looking into?

simonmichael:
Let's assume for now this flag would be for the balance command only, then [hledger/Hledger/Cli/Balance.hs](https://github.com/simonmichael/hledger/blob/master/hledger/Hledger/Cli/Balance.hs) is the place to start. It calls singleBalanceReport or multiBalanceReport depending on mode. I don't know if the percentage-reporting mode should be done inside those or outside them as is done for -V (see line 317).

simonmichael:
This was added in 1.16.

PR #1124, opened 2019-11-16
cli: Add -% to compound balance commands

kaini:
This commit introduces the commandline argument -%/--percent to show
percentages of the column's total instead of the absolute amounts for
each account in reports. The signs of the values are preserved.

This option is especially useful for the balance and incomestatement
commands.

If there are multiple commodities involved in a report hledger bails
with an error message. This can be avoided by using --cost. Also note
that if one uses -% with the balance command the chances are high that
all numbers are 0. This is due to the fact that by default balance sums
up to zero. If one wants to use -% in a meaningful way with balance one
has to add a query.

In order to keep the implementation as simple as possible --tree has no
influence over how the percentages are calculated, i.e., the percentages
always represent the fraction of the columns total. If one wants to know
the percentages relative to a parent account, one has to use a query to
narrow down the accounts.

---

Relates to #319.

Here an example for incomestatement:

```text
$ stack run -- is -p "quarterly this year" -% --depth 2
Income Statement 2019

                     ||  2019q1   2019q2   2019q3   2019q4
=====================++====================================
 Revenues            ||
---------------------++------------------------------------
 EK:Xkasldfsl        ||   1.4 %        0        0        0
 EK:Zasdfa           ||   0.2 %    7.9 %    0.9 %        0
 EK:Gasdfasdf        ||  98.4 %   92.1 %   99.1 %  100.0 %
---------------------++------------------------------------
                     || 100.0 %  100.0 %  100.0 %  100.0 %
=====================++====================================
 Expenses            ||
---------------------++------------------------------------
 AK:Iasdfcaswefndkls ||       0        0    1.3 %        0
 AK:Dassd            ||  27.3 %   45.4 %   61.9 %   34.2 %
 AK:Vdsda            ||  72.7 %   54.6 %   36.8 %   65.8 %
---------------------++------------------------------------
                     || 100.0 %  100.0 %  100.0 %  100.0 %
=====================++====================================
 Net:                ||       0        0        0        0
```

And here an example for balance. Notice how even in --tree only the column's total is considered as base for the percentages. This was done in order to keep the implementation as simple as possible. I would suggest that if a "relative-to-parent-percentage" feature is needed as well, that this should be implemented with an additional (or different) commandline flag.

```
$ stack run -- balance -p "quarterly this year" -% hobby --tree
Balance changes in 2019:

                ||  2019q1   2019q2   2019q3   2019q4
================++====================================
 AK             || 100.0 %  100.0 %  100.0 %  100.0 %
   Hobby        || 100.0 %  100.0 %  100.0 %  100.0 %
     Aasdf      ||  34.0 %   18.7 %   15.5 %   57.6 %
     Fdsddf     ||   3.7 %    3.7 %    1.7 %    3.9 %
     VGS        ||  31.6 %        0        0        0
     Ds         ||  30.8 %   77.6 %   82.7 %   38.5 %
       Nnkjsjkj ||       0   63.4 %   81.0 %        0
       Bdskls   ||  27.0 %    6.6 %    1.8 %        0
       Utsadfds ||       0        0        0   38.5 %
       Dsdasd   ||   3.7 %    7.6 %        0        0
----------------++------------------------------------
                || 100.0 %  100.0 %  100.0 %  100.0 %
```

The flag is also available for balancesheet, balancesheetequity and cashflow.

Of course there are many queryies where converting the amounts into percentages leads to useless results. For example: balance -V -% expresses each account's balance as percentage of the currently unrealized gains.


hledger-lib/Hledger/Reports/BalanceReport.hs
```diff
@@ -185,6 +195,28 @@ brNegate (is, tot) = (map brItemNegate is, -tot)
...
+unifyMixedAmount :: MixedAmount -> Amount
+unifyMixedAmount mixedAmount = foldl combine (num 0) (amounts mixedAmount)
+  where
+    combine amount result =
+      if isReallyZeroAmount amount
+      then result
+      else if isReallyZeroAmount result
+        then amount
+        else if acommodity amount == acommodity result
+          then amount + result
```
simonmichael: Nice PR, thanks!

Does this duplicate [normaliseMixedAmount](https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Data/Amount.hs#L489) (or "normaliseMixedAmountSquashPricesForDisplay") ?

simonmichael:
PS also a test or two in tests/* ?

hledger-lib/Hledger/Reports/BalanceReport.hs
kaini: I do not think so: normaliseMixedAmount simplifies an MixedAmount as much as possible, while unifyMixedAmount forces the simplification to a single commodity and bails with an user-visible error message if this is not possible.

kaini:
I added documentation and some tests.

I just noticed that balancesheet, balancesheetequity and incomestatement have a grand-total sum row labeled "Net:". I believe that this column should be hidden when -% is used, as it is always going to be useless.

Also I believe a found a bug in my implementation (hledger bs -% -V sometimes bails with an error message about mixed commidities).

I am going to fix these two things later today.

kaini:
> I am going to fix these two things later today.

Done.

simonmichael:
Looks great! I'm merging this, thank you.

PR #1125, opened 2019-11-16
re-enable doctests and easytests

ony:
Use a work-around described in https://github.com/sol/hpack/issues/188 .

Unfortunately for `easytests` this produces ugly warning `-Wmissing-home-modules` ~~that I simply don't understand since the only difference between `doctests` and `easytests` is that last one include `hledger-lib` in dependencies to build simple CLI that is equivalent to `hledger test`~~.  
I had to use work-around to ensure that Cabal understand that we want to get `tests_Hledger` via library that we built.

### DIFF
--- a/hledger-lib/Hledger/Data/RawOptions.hs
+++ b/hledger-lib/Hledger/Data/RawOptions.hs
@@ -55,15 +55,19 @@ inRawOpts name = isJust . lookup name . unRawOpts
 boolopt :: String -> RawOpts -> Bool
 boolopt = inRawOpts
 
--- | Get latests successfully parsed flag
+-- | Get latests successfully parsed flag.
+--
+-- Expected to be used for exclusive choice flags like "--json" vs "--csv".
 --
 -- >>> choiceopt Just (RawOpts [("a",""), ("b",""), ("c","")])
 -- Just "c"
 -- >>> choiceopt (const Nothing) (RawOpts [("a","")])
 -- Nothing
--- >>> choiceopt (listToMaybe . filter (`elem` ["a","b"])) (RawOpts [("a",""), ("b",""), ("c","")])
--- Just "b"
-choiceopt :: (String -> Maybe a) -> RawOpts -> Maybe a
+-- >>> choiceopt readMay (RawOpts [("LT",""),("EQ",""),("Neither","")]) :: Maybe Ordering
+-- Just EQ
+choiceopt :: (String -> Maybe a) -- ^ "parser" that returns 'Just' value for valid choice
+          -> RawOpts             -- ^ actual options where to look for flag
+          -> Maybe a             -- ^ exclusive choice among those returned as 'Just' from "parser"
 choiceopt f = lastMay . collectopts (f . fst)
 
 -- | Collects processed and filtered list of options preserving their order


simonmichael:
@virkony, what does the https://github.com/sol/hpack/issues/188 workaround fix ? Unnecessary recompilation when switching between stack build and stack test ? So building test suites becomes much faster ?

simonmichael:
Also do you see much benefit in leaving the doctest and easytest suites enabled ? After having them enabled for years, I feel like they provide value to hledger developers when invoked as needed, such as before release to check build plans/deps; but otherwise just waste cpu cycles and slow down every CI build (ours, hackage's, stackage's..)

ony:
> @virkony, what does the [sol/hpack#188](https://github.com/sol/hpack/issues/188) workaround fix ? Unnecessary recompilation when switching between stack build and stack test ? So building test suites becomes much faster ?

Yes, it doesn't build sources of `hledger-lib` anymore for each of them. `hpack` automatically puts all modules from "home" to `other-modules`. This behavior might be considered useful for tests that do not link with code they test (e.g. when there is no library exposed like application), but not applicable for all tests. Figuring out actually used modules in module specified as `main-is` should have same effect for doctest and EasyTest.  
I put link there so people can learn about it and will not remove that until that issue resolved.


> Also do you see much benefit in leaving the doctest and easytest suites enabled ? After having them enabled for years, I feel like they provide value to hledger developers when invoked as needed, such as before release to check build plans/deps; but otherwise just waste cpu cycles and slow down every CI build (ours, hackage's, stackage's..)

My usual development environment:
- Run in yet another terminal neovim to edit source.
- Run in one terminal `stack test --fast --file-watch hledger-lib`. Every time I save I get test run. It would be nice if doctests had some kind of "cache" mode when they test only changed files and their dependencies, but it is not yet there.
- Run in another terminal `stack ghci`. Whenever I stuck I consult types here and do simple try out.

doctests require sources only at run-time and they use "interpret" mode as I do in GHCi. And they do not re-compile during development since sources almost never change. 

easytests require `hledger-lib` to be linked. And require no re-compile during development as well.

So as you can see I invoke/re-invoke during development. I run `make functest buildtest` when I did something bigger. And I see it is important to have them running during CI, otherwise they are not that far from "try that thing out in GHCi before pushing".

*Note: surprisingly, when I re-enabled doctests I found one issue. But that in*

simonmichael:
I see how it fixes double compilation now - in master, "modify FILE, stack build hledger-lib:test:easytests --no-run-tests" compiles FILE twice; with this fix, it compiles it just once. Great fix, that was really annoying. sol/hpack#188 also mentions a compilation error with newer cabal, which I didn't see.

I'm still unclear on the actual benefits of enabling the test suites by default. Apart from the fact that "stack test/cabal test" is the most obvious way a hledger developer might try to run tests but it doesn't work. That is a bit unfortunate.

Did you know about these faster ways to rerun doctests, easytests, or just one test or test group quickly on file change ?
```
make help-ghcid
make ghcid-doctest
make ghcid-test
make ghcid-test-Hledger.Data.Transaction.showTransaction
```

simonmichael:
Well, on my mac hledger-lib hunit tests take 1s and doctests take 22s. I'd love to speed up the doctests, but it's ok. I applied the same workaround to hledger's test/benchmark suites, just in case. Merged, thank you!

I'll leave it to you to fix the failing doctest. Perhaps you could document choiceopt more, it wasn't obvious enough for me to fix quickly.

ony:
> I'll leave it to you to fix the failing doctest. Perhaps you could document choiceopt more, it wasn't obvious enough for me to fix quickly.

Done. It was just wrong. Correct version could look like:
```haskell
choiceopt (\flag -> headMay $ filter (flag==) ["a","b"]) ...
```
But it becomes unreadable. So I just replaced it with `readMay` which is kinda "parser".

PR #1135, opened 2019-11-23
Add exit status code to check-dates

amitaibu:
Currently check-dates doesn't send a proper exist status.

This means we cannot rely on the status for automatic checks. For example
image

PR wraps the output of the command with the needed status codes.


simonmichael:
Thank you very much @amitaibu!

PR #1136, opened 2019-11-24
WIP: Adopt pretty rendering

ony:
## Purpose of PR
Unify how we render colorful and plain reports in `hledger` so there is no need to have them implemented twice.

## Libraries considered
This is only my quick look and opinion. Maybe someone have more to add.

* [`pretty`](http://hackage.haskell.org/package/pretty). Quiet old and no recent changes, but used by Cabal. Colors need to be introduced via annotations like in [`wl-pprint-terminfo`](http://hackage.haskell.org/package/wl-pprint-terminfo).
* [`ansi-wl-pprint`](http://hackage.haskell.org/package/ansi-wl-pprint). Have support for colors out of the box via [`ansi-terminal`](http://hackage.haskell.org/package/ansi-terminal) that we use right now.
* [`prettyprinter`](http://hackage.haskell.org/package/prettyprinter). Seems to be fresh. As in `pretty` colors need to be supported via annotations, but there is already [`prettyprinter-ansi-terminal`](http://hackage.haskell.org/package/prettyprinter-ansi-terminal) that use [`ansi-terminal`](http://hackage.haskell.org/package/ansi-terminal).
* [`annotated-wl-pprint`](http://hackage.haskell.org/package/annotated-wl-pprint). Apparently stripped version `ansi-wl-pprint`. Used by [`rio-prettyprint`](http://hackage.haskell.org/package/rio-prettyprint) (authored by Michael Snoyman).
* [`boxes`](http://hackage.haskell.org/package/boxes). Highly oriented on layout management. Nothing 
* [`table-layout`](hackage.haskell.org/package/table-layout). Looks to be highly oriented on how to render tables. No clear composition of primitive objects.

## Progress

- [x] `Hledger.Data.Amount`.
- [ ] `Hledger.Commands.Balance`.
- [ ] `Hledger.Commands.Register`.
- [ ] `Hledger.Reports.BalanceReport`.
- [ ] Benchmark added overhead for rendering.
- [ ] Review `hledger-ui` for potential re-use.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -71,6 +71,7 @@ module Hledger.Data.Amount (
   showAmount,
   cshowAmount,
   showAmountWithZeroCommodity,
+  prettyAmountWithZeroCommodity,
   showAmountDebug,
   showAmountWithoutPrice,
   maxprecision,
@@ -85,6 +86,8 @@ module Hledger.Data.Amount (
   setAmountDecimalPoint,
   withDecimalPoint,
   canonicaliseAmount,
+  simplifyZeroAmount,
+  amountStripPrice,
   -- * MixedAmount
   nullmixedamt,
   missingmixedamt,
@@ -94,6 +97,7 @@ module Hledger.Data.Amount (
   filterMixedAmountByCommodity,
   normaliseMixedAmountSquashPricesForDisplay,
   normaliseMixedAmount,
+  mixedAmountStripPrices,
   -- ** arithmetic
   costOfMixedAmount,
   mixedAmountToCost,
@@ -169,9 +173,10 @@ instance Num Amount where
     (*)                          = similarAmountsOp (*)
 
 instance Pretty Amount where
+    pretty a | a == missingamt = empty
     pretty a =
         (if isNegativeAmount a then dullred else id) $
-        text $ showAmountHelper False a
+        text $ showAmount a
 
 -- | The empty simple amount.
 amount, nullamt :: Amount
@@ -273,16 +278,11 @@ digits = "123456789" :: String
 -- | Does this amount appear to be zero when displayed with its given precision ?
 isZeroAmount :: Amount -> Bool
 isZeroAmount --  a==missingamt = False
-  = not . any (`elem` digits) . showAmountWithoutPriceOrCommodity
+  = not . any (`elem` digits) . showamountquantity
 
 -- | Is this amount "really" zero, regardless of the display precision ?
 isReallyZeroAmount :: Amount -> Bool
 isReallyZeroAmount Amount{aquantity=q} = q == 0
-
--- | Get the string representation of an amount, based on its commodity's
--- display settings except using the specified precision.
-showAmountWithPrecision :: Int -> Amount -> String
-showAmountWithPrecision p = showAmount . setAmountPrecision p
 
 -- | Set an amount's display precision.
 setAmountPrecision :: Int -> Amount -> Amount
@@ -352,10 +352,6 @@ setAmountDecimalPoint mc a@Amount{ astyle=s } = a{ astyle=s{asdecimalpoint=mc} }
 withDecimalPoint :: Amount -> Maybe Char -> Amount
 withDecimalPoint = flip setAmountDecimalPoint
 
--- | Get the string representation of an amount, without any price or commodity symbol.
-showAmountWithoutPriceOrCommodity :: Amount -> String
-showAmountWithoutPriceOrCommodity a = showAmount a{acommodity="", aprice=Nothing}
-
 showAmountPrice :: Maybe AmountPrice -> String
 showAmountPrice Nothing                = ""
 showAmountPrice (Just (UnitPrice pa))  = " @ "  ++ showAmount pa
@@ -386,7 +382,7 @@ styleAmountExceptPrecision styles a@Amount{astyle=AmountStyle{asprecision=origp}
 -- zero are converted to just \"0\". The special "missing" amount is
 -- displayed as the empty string.
 showAmount :: Amount -> String
-showAmount = showPretty
+showAmount = showAmountWithZeroCommodity . simplifyZeroAmount
 
 -- | Colour version. For a negative amount, adds ANSI codes to change the colour,
 -- currently to hard-coded red.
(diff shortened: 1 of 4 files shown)


simonmichael:
I've read the commits so far.. could you give a little more explanation of the goals and benefits of the proposed approach ? I'm not sure if there's a clear example yet of how it helps, or if you are just laying the groundwork so far.

ony:
> I've read the commits so far.. could you give a little more explanation of the goals and benefits of the proposed approach ?

Instead of having 4 implementations:

```haskell
showMixedAmountWithoutPrice m = intercalate "\n" $ map showamt as
  where
    Mixed as = normaliseMixedAmountSquashPricesForDisplay $ mixedAmountStripPrices m
    showamt = printf (printf "%%%ds" width) . showAmountWithoutPrice
      where
        width = maximumDef 0 $ map (length . showAmount) as

cshowMixedAmountWithoutPrice m = intercalate "\n" $ map showamt as
  where
    Mixed as = normaliseMixedAmountSquashPricesForDisplay $ mixedAmountStripPrices m
    showamt a =
      (if isNegativeAmount a then color Dull Red else id) $
      printf (printf "%%%ds" width) $ showAmountWithoutPrice a
      where
        width = maximumDef 0 $ map (length . showAmount) as

showMixedAmountOneLineWithoutPrice m = intercalate ", " $ map showAmountWithoutPrice as
    where
      (Mixed as) = normaliseMixedAmountSquashPricesForDisplay $ stripPrices m
      stripPrices (Mixed as) = Mixed $ map stripprice as where stripprice a = a{aprice=Nothing}

cshowMixedAmountOneLineWithoutPrice m = intercalate ", " $ map cshowAmountWithoutPrice as
    where
      (Mixed as) = normaliseMixedAmountSquashPricesForDisplay $ stripPrices m
      stripPrices (Mixed as) = Mixed $ map stripprice as where stripprice a = a{aprice=Nothing}
```

We have kinda one

```haskell
instance Pretty MixedAmount where
    pretty = prettyMixedAmountHelper pretty

prettyMixedAmountHelper prettyAmount m = verical `flatAlt` flat where
    Mixed as = normaliseMixedAmountSquashPricesForDisplay m
    docs = map prettyAmount as
    verical = vcatRight docs
    flat = encloseSep mempty mempty ", " docs
```

From which we can infer them all

```haskell
showMixedAmountWithoutPrice = showWide . plain . pretty . mixedAmountStripPrices
cshowMixedAmountWithoutPrice = showWide . pretty . mixedAmountStripPrices
showMixedAmountOneLineWithoutPrice = showWide . plain . hgroup . pretty . mixedAmountStripPrices
cshowMixedAmountOneLineWithoutPrice = showWide . hgroup . pretty . mixedAmountStripPrices
```

- `plain` - strips color
- `hgroup` - switches to horizontal layout

After that we can start using `Doc AnsiStyle` (or whatever we choose) in CLIs and benefit from flexibility to wrap text when we need or make it compact by putting in single line.

Another thing is to get rid of those helpers that can be replaced with adding `mixedAmountStripPrices` and/or `simplifyZeroAmount`.

> I'm not sure if there's a clear example yet of how it helps, or if you are just laying the groundwork so far.

It supposed to be clear already. Further work might be about table layout abstraction with text re-flow inside of cells (e.g. for multi-commodity balances `register` looks kinda empty on the left side).

simonmichael:
@ony, thanks. I'm getting clearer on this. I think:

Currently for debug info we use the Show typeclass augmented by pretty-show, 
and for user-facing output we use various custom functions, often with "show" somewhere in their name.

You propose to replace the user output functions with ansi-wl-pprint's Pretty typeclass. 
This might bring more consistency and clarity, and provide useful features and combinators.

It's probably a great idea but I propose we don't try to get it into 1.16 release. 
FYI, in case it would be conflictful: there's a slight chance that tests will be ported to tasty for 1.16.

Some issues that come to mind:

- We would render everything in (ANSI) colour by default, and use ansi-wl-pprint's plain helper
  to strip colours when needed. But I imagine we'll want to be able to customise colours,
  as hledger-ui does. This can probably be done with Pretty, but it may complicate things.

- Does ansi-wl-pprint have any support for wide characters ?

ony:
> It's probably a great idea but I propose we don't try to get it into 1.16 release.

Sure. I use `hledger`  in rolling release mode and don't care much about releases since I don't dedicate effort for them :stuck_out_tongue: .

I'd think this PR more like a branch on the review at this moment. We should drop if during this exploration we'll not benefit or will suffer significantly on performance degradation.

> FYI, in case it would be conflictful: there's a slight chance that tests will be ported to tasty for 1.16.

Not sure how this will affect. As I understand [tasty](http://hackage.haskell.org/package/tasty) is just organizer for tests. For shell tests I think it will be easier to write adapter similar to [`tasty-golden`](https://hackage.haskell.org/package/tasty-golden) rather than move them all into code on top of some framework.

> 
> Some issues that come to mind:
> 
>     * We would render everything in (ANSI) colour by default, and use ansi-wl-pprint's `plain` helper
>       to strip colours when needed. But I imagine we'll want to be able to customise colours,
>       as hledger-ui does. This can probably be done with Pretty, but it may complicate things.

Indeed. I switched to `prettyprinter` that allows to easily switching to something else beside `AnsiStyle` and do styling right before rendering to terminal. But before reaching that point we need to get intermediate `String`/`Text` that we currently have.

>     * Does ansi-wl-pprint have any support for wide characters ?
I don't think so :speak_no_evil: ( quchen/prettyprinter#103 ). We can push this problem to them :laughing: .  

For `pandoc` it can be somehow justified since they are directly related with rendering to various formats.  
But for `hledger` I really think it is inappropriate to how embedded `charWidth`. We should either delegate this to somewhere or we should chop off that part of `hledger` into self-sufficient library that lives through its own release cycles. Making it available for consumption by other projects we increase pairs of eyes that are looking for better solution.

simonmichael:
Great. I'm all for extracting/outsourcing reusable functionality like wide chars, as long as we keep international characters working in hledger.

Xitian9:
What's the status here? I think this has mostly been unified with the `AmountDisplayOpts`. Is there more to do?

ony:
@Xitian9 , if I understand correctly idea of `AmountDisplayOpts` is that we pass in configuration how to render to render function.

https://github.com/simonmichael/hledger/blob/ff76927ee8cfd1710ca2d85b9ea28bcc78a3d331/hledger-lib/Hledger/Data/Amount.hs#L468

https://github.com/simonmichael/hledger/blob/ff76927ee8cfd1710ca2d85b9ea28bcc78a3d331/hledger-lib/Hledger/Data/Amount.hs#L460-L461

In this proposal idea was a bit different. Instead of passing in options color/no-color that goes through all renderers we return value that can be both colorful and non-colorful.

https://github.com/simonmichael/hledger/blob/4c27916e6f15a3401e4a337fa558dd2a620ee91e/hledger-lib/Hledger/Utils/Pretty.hs#L64-L66

This way you can get rid of `displayColour` in https://github.com/simonmichael/hledger/blob/ff76927ee8cfd1710ca2d85b9ea28bcc78a3d331/hledger-lib/Hledger/Data/Amount.hs#L203

I'm not familiar with new code-base, so I'm not sure how right now that information reaches `showAmountB`, but I would expect that it had to be passed around tossing from some `TransactionDisplayOpts` to `AmountDisplayOpts` to have global effect from `--colour` or detection of terminal.

Beside that idea was to get rid of unique names for every data-type and use one overloaded `showPretty` instead.  

If I would try to re-based this on top of `AmountDisplayOpts` it might look like:
```haskell
class Render a where
    type DisplayOpts a
    renderWith :: DisplayOpts a -> a -> Display b
    render :: a -> Display (DisplayOpts a)

-- | Some "markup" maybe in some library
data Display opts where

-- | That can be arranged in plain text accounting specifics of output device (e.g. terminal width)
foldDisplayTo :: OutputInfo -> Display Void -> Text

class LiftDisplay a b where
    liftDisplay :: Display b -> Display a
    unpackDisplayOpts :: a -> b

instance LiftDisplay TransationOpts AmountOpts where
    unpackDisplayOpts = amountOpts

instance Render Transaction where
    type DisplayOpts Transaction = TransactionOpts
    render txn = {- ... -} liftDisplay (render amt) {- ... -}
```

It was also an attempt to adopt use of frameworks to construct text output that does all the re-flowing for us.

I don't mind closing this draft PR since I don't have plans to invest into it in near future.

Xitian9:
Ah, I understand. That makes a lot of sense.

I'm pushing to move out table rendering infrastructure over to `table-layout`, in which case their `Formatted` class would be a natural fit for this. Once that code is ready perhaps we can implement this within that context.

simonmichael:
Closing for now, we can reactivate this later if needed.

PR #1138, opened 2019-11-29
Simplify date comparison code.

Xitian9:
`compareMaybeDates` is equivalent to `compare` due to the Ord instance for Maybe a. Finding the smallest/largest elements of a list is probably best done with `minimum` and `maximum` rather than sorting and taking the head.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -59,7 +59,7 @@ import Data.Monoid ((<>))
 #endif
 import qualified Data.Text as T
 import Data.Time.Calendar
-import Safe (readDef, headDef)
+import Safe (readDef, maximumByDef, maximumDef, minimumDef)
 import Text.Megaparsec
 import Text.Megaparsec.Char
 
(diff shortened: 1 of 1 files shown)


Xitian9:
This sholud be equivalent to the current code, but the comment for `earliestMaybeDate` suggests maybe you want this to be equivalent to `earliestMaybeDate'`. Either the code should be changed or the comment should be clarified.

Xitian9:
I think this should now give correct results. Maybe some tests should be written to specify the correct behaviour.

simonmichael:
Looks good, would you mind re-pushing these as one clean patch ?

simonmichael:
PS how is it a bug fix - could you show an example of buggy behaviour or doc ?

Xitian9:
Previously the code failed the last test I've added for queryEndDate.
`queryEndDate $ Or [Date start (Just end), Date start Nothing]` would give a result of `Just end`.

simonmichael:
Is that not the correct result, given the doc: "What end date .. does this query specify, if any ? For OR expressions, use the latest of the dates." ? Ie one of the OR alternatives specifies an end date, so that's the one we should return.

Xitian9:
That may be the correct behaviour, but then it would suggest a bug with queryStartDate, as
`queryStartDate $ Or [Date $ DateSpan Nothing end, Date $ DateSpan (Just start) end]` returns Nothing, both before and after the change.

The question to resolve this is ‘Does a Nothing in DateSpan indicate that the start/end has not been specified, or that it extends to the start/beginning of time?’

simonmichael:
Agreed, that does seem wrong. 

I think “not specified” and “extends to the end of time” are generally the same thing in hledger date spans.

Xitian9:
In that case, I would expect `queryEndDate $ Or [Date $ DateSpan start (Just end), Date $ DateSpan start Nothing]` to give a value of Nothing. Something which matches either being between start and end, or between start and the end of time, would be the same as matching between start and the end of time.

Can you confirm that the tests I wrote give the behaviour that you would expect too? If so, I believe the implementation in my patch is correct.

simonmichael:
> In that case, I would expect queryEndDate $ Or [Date $ DateSpan start (Just end), Date $ DateSpan start Nothing] to give a value of Nothing.

Perhaps I'm missing your point, but I disagree, because the docs say:
```
-- | What end date (or secondary date) does this query specify, if any ?
-- For OR expressions, use the latest of the dates. NOT is ignored.
```

If that does not clear it up, I'll look more at the tests another time.

Xitian9:
Maybe I'm misinterpretting the documentation, but that seems to me to say that it should give Nothing, as ‘the end of time’ is the latest possible date. I interpreted ‘Not is ignored’ to mean that `queryEndDate True (Not . Date $ DateSpan start end)` would return Nothing, which is consistent with hledger returning the smallest contiguous `DateSpan` containing the union of the `DateSpan`s in the Or expression.

Xitian9:
Perhaps I should explain further why I feel that the function should return Nothing.

Morally, there should be no difference between `compound = Or [Date $ DateSpan (Just late) Nothing), Date $ DateSpan (Just early) (Just later)]` and `simple = Date $ DateSpan (Just early) Nothing`, where `early < late < later`. In plain English, something which lies either between early and later, or between late and the end of time, is logically equivalent to lying between early and the end of time.

Now, `queryEndDate True simple` returns Nothing, and I believe this is uncontroversial (what else could it return?). So it would be strange if `queryEndDate True compound` didn't also return `Nothing`. As currently coded, it will return `Just later`. In the patched version, it will return `Nothing`.

This has probably not been an issue so far because hledger's simplified query handler will never create a query of the form `Or [Date …, Date …]`, so no buggy behaviour is currently accessible unless you're creating the Queries yourself in code.

simonmichael:
Sorry for the delay on this. I agree with your comment above, and tests pass, so I'm merging this. Thanks a lot!

PR #1140, opened 2019-11-30
fix parsing of < and > in add-on commands (#457)

### LINKED ISSUE #457: hledger ui 'amt:>200' fails unexpectedly
~~~
$ hledger ui "amt:>200"
hledger-ui: could not parse as '=', '<', or '>' (optional) followed by a (optionally signed) numeric quantity: 
~~~

I suppose there is some kind of quoting error when `hledger` forwards the command-line to `hledger-ui`?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/40525920-hledger-ui-amt-200-fails-unexpectedly?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jacg:
This PR attemts to fix #457.

It contains 2 commits:

+ The first commit adds four shelltests demonstrating the problem. I am not sure what the best place is for these tests to live. Please advise.

+ The second commit solves the problem by using existing machinery (`String.quoteIfNeeded`) to quote arguments which contain the characters which seem to be causing the problem: `<` and `>`. While ad-hoc experiments on the command line suggest that this solution fixes the problem, the tests added in the first commit continue to fail. I'm guessing that this might be related to the extra level of quoting that `shelltest` injects.

### DIFF
--- a/hledger-lib/Hledger/Utils/String.hs
+++ b/hledger-lib/Hledger/Utils/String.hs
@@ -111,7 +111,7 @@ underline s = s' ++ replicate (length s) '-' ++ "\n"
 -- | Double-quote this string if it contains whitespace, single quotes
 -- or double-quotes, escaping the quotes as needed.
 quoteIfNeeded :: String -> String
-quoteIfNeeded s | any (`elem` s) (quotechars++whitespacechars) = "\"" ++ escapeDoubleQuotes s ++ "\""
+quoteIfNeeded s | any (`elem` s) (quotechars++whitespacechars++redirectchars) = "\"" ++ escapeDoubleQuotes s ++ "\""
                 | otherwise = s
 -- | Single-quote this string if it contains whitespace or double-quotes.
 -- No good for strings containing single quotes.
@@ -119,9 +119,10 @@ singleQuoteIfNeeded :: String -> String
 singleQuoteIfNeeded s | any (`elem` s) whitespacechars = "'"++s++"'"
                       | otherwise = s
 
-quotechars, whitespacechars :: [Char]
+quotechars, whitespacechars, redirectchars :: [Char]
 quotechars      = "'\""
 whitespacechars = " \t\n\r"
+redirectchars   = "<>"
 
 escapeDoubleQuotes :: String -> String
 escapeDoubleQuotes = regexReplace "\"" "\""
--- a/tests/addons/addons.test
+++ b/tests/addons/addons.test
@@ -23,26 +23,30 @@
 # PATH= ~/.local/bin/stack exec -- hledger
 #>>>=0
 
-############################    issue 457    #####################################
-# 5. test for `>` in https://github.com/simonmichael/hledger/issues/457          #
-# note a shelltest already provides one level of quoting, so amt:>0 not 'amt:>0' #
-#                                                                                #
-$ hledger -f - ui amt:>0                                                         #
->2 !/could not parse/                                                            #
->=1                                                                              #
-                                                                                 #
-# 6. test for `<` in https://github.com/simonmichael/hledger/issues/457          #
-$ hledger -f - ui amt:<0                                                         #
->2 !/could not parse/                                                            #
->=1                                                                              #
-                                                                                 #
-# 7. test for `>=` in https://github.com/simonmichael/hledger/issues/457         #
-$ hledger -f - ui amt:>=0                                                        #
->2 !/could not parse/                                                            #
->=1                                                                              #
-                                                                                 #
-# 8. test for `<=` in https://github.com/simonmichael/hledger/issues/457         #
-$ hledger -f - ui amt:<=0                                                        #
->2 !/could not parse/                                                            #
->=1                                                                              #
-##################################################################################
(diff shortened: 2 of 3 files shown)


simonmichael:
Thank you very much! I would welcome help reviewing this. Otherwise, I'll get to it after the 1.16 release.

tests/amt.test
```diff
@@ -0,0 +1,21 @@
...
+
+
+$ hledger -f - ui amt:>=0
+>2 !/could not parse/
+>=1
+
+
+$ hledger -f - ui amt:<=0
+>2 !/could not parse/
+>=1
```
simonmichael: +1 for tests. There's tests/cli/cli.test, but I think tests/addons/addons.test would be the place for these.

jacg: I have moved the tests to `tests/addons/addons.test`.

There's still the problem that the tests actually fail (even though the solution appears to work on the CLI!)

jacg:
There is a further problem in that the solution causes two existing tests to fail, both like this:
````
:tests/rewrite.test:5: [Failed]
Command:
hledger rewrite -f- assets:bank and 'amt:<0' --add-posting 'expenses:fee  $5'  --add-posting 'assets:bank  $-5'
Expected stdout: 
   [...]
   expenses:fee              $5  ; generated-posting: = assets:bank and amt:<0
   [...]
Got stdout:      
   [...]
   expenses:fee              $5  ; generated-posting: = assets:bank and "amt:<0"
````

that is, quotes appear around in the *generated-posting* comments. Arguably, those quotes *should* be there: they are needed in order to generate the posting in the first place. If so, should I just modify these two tests to expect the quotes?

**Comment by jacg** [comment]
Also, it seems that `tests/addons` is excluded explicitly on CI, so the new tests don't run at all on CI (while failing locally).

**Comment by jacg** [comment]
@simonmichael 

1. I have added the quotes to the newly-failing tests, as it seems that should really be there. Do you agree?

2. Any hints on how to make the new tests pass? (Everything works fine on the command line, but the test fails nonetheless: my guess it's something to do with the automatic quoting done by `shelltest`)

3. As it stands the new tests are not run by CI because the `addons` directory (where you suggested that I put them) is explicitly excluded on Travis. Is this OK?

**Comment by simonmichael** [comment]
Hi @jacg.. 

1. I agree with the fix for the rewrite tests, that sounds good.
2. Hmm. Yes it might be a shelltest issue. I see that adding quotes in the test makes it pass.. but then it also passes with hledger 1.15.
3. Yes I guess there was some unsolved problem with the addon tests on CI, I'll look at that another time unless you want to dig into it.

Because of 2 I guess we can forget about testing this. But since you wrote nice tests let's keep them, commented out. Also could you also convert the initial tests to the new test format, since we can't mix two formats in one test file. Eg:
```
# 1. flags after an add-command are handled by the add-on
$  PATH=$PATH:. hledger addon --help
> /hledger-addon/
>=0

# 2. add-on flags which are not also defined in the main executable are a problem
$  PATH=$PATH:. hledger addon --addonflag
>2 /Unknown flag: --addonflag/
>=1

# 3. hledger main executable ignores anything after -- (and hides the -- from the add-on)
$  PATH=$PATH:. hledger addon --help -- --addonflag
> /hledger-addon/
>=0

# TODO how to reliably ensure no addons but still find the hledger executable ?
# . having no addons shouldn't break the commands list
# $  PATH= ~/.local/bin/stack exec -- hledger
# >=0
```

simonmichael:
And finally, if you wouldn't mind please rebase and rewrite into clean commit(s), and I'll merge.

jacg:
@simonmichael I think I have done all that you asked for. Please check whether it meets your standards/expectations.

simonmichael:
That looks fine, thanks @jacq. Could you mark the PR non-draft so I can merge.

simonmichael:
Thanks a lot!

jacg:
@simonmichael Unmarked as draft.

(BTW, this isn't the first time that poor jacQ gets spammed with stuff destined for me (jacG) :-)

simonmichael:
Oops! My apologies.

Jacq:
Nice work need doing it @jacg, will see you around in the next spam 😁

jacg:
@simonmichael 

> I guess there was some unsolved problem with the addon tests on CI, I'll look at that another time unless you want to dig into it.

Maybe I could look into this. Do you have any recollection of what the problem might have been, or any hints on where to start?

simonmichael:
@Jacq you are welcome here too, then I could spam jacg instead.

simonmichael:
@jacg, yes I would look at https://github.com/simonmichael/hledger/blob/master/.travis.yml, and dig back in time using git blame - maybe there's a clue.

simonmichael:
@jacg: I just found https://github.com/simonmichael/shelltestrunner/issues/15 (and PR). I wonder if they help.

PR #1144, opened 2019-12-03
Only calculate posting date once for each posting.

Xitian9:
Previously, `multiBalanceReportWith` had been calculating the date of each posting once for every column. This resulted in quite significant slowdown when calculating for many postings and many `DateSpans`. This commit calculates the posting date of each posting only once and re-uses it. It also calculates the columns for each posting rather than searching through the whole colspan list every time.

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -63,6 +63,8 @@ module Hledger.Data.Dates (
   spanDefaultsFrom,
   spanUnion,
   spansUnion,
+  daysSpan,
+  latestSpanContaining,
   smartdate,
   splitSpan,
   fixSmartDate,
@@ -79,10 +81,11 @@ import Prelude ()
 import "base-compat-batteries" Prelude.Compat hiding (fail)
 import qualified "base-compat-batteries" Control.Monad.Fail.Compat as Fail (MonadFail, fail)
 import Control.Applicative.Permutations
-import Control.Monad (unless)
+import Control.Monad (guard, unless)
 import "base-compat-batteries" Data.List.Compat
 import Data.Default
 import Data.Maybe
+import qualified Data.Set as Set
 import Data.Text (Text)
 import qualified Data.Text as T
 #if MIN_VERSION_time(1,5,0)
@@ -95,7 +98,7 @@ import Data.Time.Calendar
 import Data.Time.Calendar.OrdinalDate
 import Data.Time.Clock
 import Data.Time.LocalTime
-import Safe (headMay, lastMay, readMay)
+import Safe (headMay, lastMay, readMay, maximumMay, minimumMay)
 import Text.Megaparsec
 import Text.Megaparsec.Char
 import Text.Megaparsec.Custom
@@ -231,9 +234,8 @@ daysInSpan _ = Nothing
 
 -- | Is this an empty span, ie closed with the end date on or before the start date ?
 isEmptySpan :: DateSpan -> Bool
-isEmptySpan s = case daysInSpan s of
-                  Just n  -> n < 1
-                  Nothing -> False
+isEmptySpan (DateSpan (Just s) (Just e)) = e <= s
+isEmptySpan _                            = False
 
 -- | Does the span include the given date ?
 spanContainsDate :: DateSpan -> Day -> Bool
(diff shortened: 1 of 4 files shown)


simonmichael:
Sorry for the delay on this and your other PRs, it's a busy month. I'll get to them, if noone beats me to it.

Xitian9:
Not a problem at all. Let me know what I can do to help.

simonmichael:
Thanks for this optimisation and cleanup. Is it possible to observe any speedup, even informally ? (eg make quickbench)

Xitian9:
No observed difference in quickbench, but it results in the following speedups in my program compiled against hledger-lib, which makes repeated calls to multiBalanceReport with daily report frequency. Time reports:

```
cabal new-run exe:hledger-report  33.45s user 5.00s system 137% cpu 28.055 total
cabal new-run exe:hledger-report  27.06s user 5.31s system 143% cpu 22.570 total
```

Interestingly I didn't see any speedup by adding the following to bench.sh, as I would have expected.
```
hledger -f examples/10000x1000x10.journal balance -D
```

I'll work to find a benchmark which shows the difference.

hledger-lib/Hledger/Data/Posting.hs
```diff
@@ -246,25 +253,23 @@ relatedPostings _ = []
...
 -- | Does this posting fall within the given date span ?
 isPostingInDateSpan :: DateSpan -> Posting -> Bool
-isPostingInDateSpan s = spanContainsDate s . postingDate
+isPostingInDateSpan = isPostingInDateSpan' PrimaryDate
 
 -- --date2-sensitive version, separate for now to avoid disturbing multiBalanceReport.
 isPostingInDateSpan' :: WhichDate -> DateSpan -> Posting -> Bool
-isPostingInDateSpan' PrimaryDate   s = spanContainsDate s . postingDate
-isPostingInDateSpan' SecondaryDate s = spanContainsDate s . postingDate2
+isPostingInDateSpan' w s = spanContainsDate s . postingDate' w
```
simonmichael: I'm ok with this if you think it's better, but I do think that secondary is a misfeature and I'd be in favour of not letting it spread even further through the code.

hledger-lib/Hledger/Data/Dates.hs
```diff
@@ -292,6 +293,30 @@ earliest (Just d1) (Just d2) = Just $ min d1 d2
 daysSpan :: [Day] -> DateSpan
 daysSpan ds = DateSpan (minimumMay ds) (addDays 1 <$> maximumMay ds)
 
+-- | Calculate the DateSpan containing a given Day, if any, from a given list of
+-- DateSpans.
```
simonmichael: Select instead of Calculate, perhaps ?

How about calling it latestSpanContaining ?

hledger-lib/Hledger/Data/Posting.hs
```diff
@@ -69,6 +70,8 @@ module Hledger.Data.Posting (
   tests_Posting
 )
 where
+
+import Data.Foldable (asum)
```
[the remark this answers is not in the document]
simonmichael: Yes, this one is fine.

simonmichael:
Overall: I like your cleanups here, but multiBalanceReportWith seems less clear than before. Probably I'm not understanding what exactly it optimises yet. A little more comment at that section of code might help.

hledger-lib/Hledger/Data/Posting.hs
Xitian9: I'm happy to get rid of that. Out of curiosity, what would your vision be for getting rid of SecondaryDate?

hledger-lib/Hledger/Reports/MultiBalanceReport.hs
```diff
@@ -187,17 +190,32 @@ multiBalanceReportWith ropts@ReportOpts{..} q j@Journal{..} priceoracle =
...
+          [ (posts, spanEnd dateSpan) | (dateSpan, posts) <- M.toList colMap ]
+        where
+          colMap :: Map DateSpan [Posting]
+          colMap = foldr (\(p, d) -> M.adjust (p:) $ getSpan d)
+                     (M.fromList . zip colspans $ repeat []) ps
+          colSet = Set.fromList colspans
+          getSpan d = fromMaybe (DateSpan Nothing Nothing) $
+                        Set.lookupLT datespan colSet
+            where
+              datespan = DateSpan (Just $ addDays 1 d) Nothing
```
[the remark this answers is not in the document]
Xitian9: I pushed another commit (6a5c977) which breaks apart this logic into a separate function, which I believe makes the logic clearer.

hledger-lib/Hledger/Data/Dates.hs
```diff
@@ -287,6 +288,11 @@ earliest d Nothing = d
 earliest Nothing d = d
 earliest (Just d1) (Just d2) = Just $ min d1 d2
 
+-- | Calculate a DateSpan from a list of days
```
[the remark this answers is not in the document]
Xitian9: Sounds good. Will change.

hledger-lib/Hledger/Data/Dates.hs
Xitian9: Again, sounds reasonable.

hledger-lib/Hledger/Reports/MultiBalanceReport.hs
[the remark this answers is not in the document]
Xitian9: I'm actually not 100% sure where the speedup comes from either. I came to this point by profiling my program, and seeing that 16.9% of the time came from calling `postingDate` in `isPostingInDateSpan'`. In fact, it was being called 59 184 544 times, despite `isPostingInDateSpan'` only being called 28 604 times, and `multiBalanceReportWith` only being called 30 times.

Given this, I would have also expected to see a speedup in `hledger balance -D`, but that didn't  actually happen, so I'm a bit confused.

hledger-lib/Hledger/Data/Posting.hs
simonmichael: Since it's working and helps Ledger compatibility I think we should keep it. But I'd be glad to keep it's impact to the minimum.

simonmichael:
Hi @Xitian9.. The reduction in call count certainly sounds worthwhile. I'll wait for your additional cleanups.

Xitian9:
I've made the suggested changes. Let me know if you'd like me to squash all the commits.

One thing to consider. `postingsDateSpan` is now only called in one place, and `postingsDateSpan'` not at all. Since most of the logic has been broken out into `daysSpan`, we could consider removing the functions altogether. The only required change would be a substitution at the following point:

https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Data/Ledger.hs#L102-L103

```
ledgerDateSpan = daysSpan . map postingDate . ledgerPostings
```

simonmichael:
Yes that removal sounds good, and squashing sounds good.

simonmichael:
Thanks!

PR #1146, opened 2019-12-05
Add a basic example of rule parsing for the output of csb2format.

evilham:
csb2format deals with the CSB43/AEB43 format, which all banks operating in
Spain must support.
Having these example rules enables easens bootstraping for users with a
Spanish bank account.

### DIFF
--- a/examples/csv/csb43.csv.rules
+++ b/examples/csv/csb43.csv.rules
@@ -1,0 +1,15 @@
+# csb43.csv.rules
+# Rules for csb2format -f csv "${CSV_FILE}" -
+# Using https://pypi.org/project/csb43/
+
+# Skip header
+skip 1
+
+# Change to your taste
+account1 assets:bank:csb43
+
+# These appear to be the useful columns for hledger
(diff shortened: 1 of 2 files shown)


examples/csv/csb43.csv.rules
```diff
@@ -0,0 +1,15 @@
...
+# Using https://pypi.org/project/csb43/
+
+# Skip header
+skip 1
+
+# Change to your taste
+account1 assets:bank:csb43
+
+# These appear to be the useful columns for hledger
+fields _,_,_,_,_,_,currency,_,_,_,_,_,_,_,_,_,_,_,_,item1,item2,reference1,reference2,date,_,amount,_,_
```
adept: I think (for 1.16 and later) it is best to call columns currency1 and amount1. Plus from the csv sample it seems that you can add balance1 as well

simonmichael: I'm fine with currency/amount too, they are affecting both postings after all. Your call OP.

Nice example!

PR #1151, opened 2019-12-17
Change the '<' command to go back one step during addition

### LINKED ISSUE #577: `hledger add` should remember previous inputs
When we enter '<' in prompt to restart transaction it is often tiresome to enter the same details again, because we've made mistake in the last line. There's currently implemented a default answer, which is taken from previous similar transaction. What I would like to see (and probably implement) is additional '<e' option in prompt to set this default value to answers we typed in the amended transactions without looking for similar transactions.

If this feature is welcome, I would be happy to implement it, but I need some guidance to start working on hledger.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46631434-hledger-add-should-remember-previous-inputs?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F536505&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ghallak:
This PR changes the `confirmedTransactionWizard` function to accept a list that is used as a stack to store the stages of the addition command so that the back command can go back to the right stage.

Is this a good approach towards implementing a back command for `add`?

Closes #577

### DIFF
--- a/hledger/Hledger/Cli/Commands/Add.hs
+++ b/hledger/Hledger/Cli/Commands/Add.hs
@@ -222,8 +222,7 @@ confirmedTransactionWizard prevInput es@EntryState{..} stack@(currentStage : _) 
             EnterAmountAndComment _ _ -> False
             EnterDescAndComment _ -> False
             _ -> True
-          prevAccount' = take (length esPostings) (prevAccount prevInput)
-      confirmedTransactionWizard prevInput{prevAccount=prevAccount'} es{esPostings=init esPostings} (dropWhile notPrevAmountAndNotEnterDesc stack)
+      confirmedTransactionWizard prevInput es{esPostings=init esPostings} (dropWhile notPrevAmountAndNotEnterDesc stack)
 
   EnterAmountAndComment txnParams account -> amountAndCommentWizard prevInput es >>= \case
     Just (amount, comment) -> do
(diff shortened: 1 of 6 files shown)


simonmichael:
Nice!

From some quick testing: I find that most of the prompts, if I go back one step, don't offer the previous input as their default. (Date did).

simonmichael:
By the way, if you see how and would prefer it, I think showing the previous values in the input area (on the right of the colon rather than on the left) might be even better.

ghallak:
After this commit https://github.com/simonmichael/hledger/pull/1151/commits/ee18227a126e31cd8645b51795c571e1128bdaed the last input should show up in the input area (rather than in the default value area). The default value will stay unaffected by the input except for the date default value, which is affected by the entered date (should the be changed?).

simonmichael:
Great! Works for me. I'll merge this now if you're ready. 

Or, more brainstorming:

- Currently to keep stepping back you have to delete each old input and type <. Should there be a way to step back without deleting ? Eg, I suppose, by just adding < at the end and pressing enter.
- Once you have changed an old input, you also have to re-enter all the following inputs. Should it remember the old inputs as you are stepping forward also ?

ghallak:
>Eg, I suppose, by just adding < at the end and pressing enter.

This sounds ok, and it's easily doable by checking the last character of the input string, but if this is done, no comments after descriptions or amounts will be allowed to end with `<`. Do you think that will cause any problems?

>Should it remember the old inputs as you are stepping forward also ?

I have done that at first, then I decided to undo it (but I forgot to do so for the description field, which still remembers when stepping forward). I guess I can bring it back.

simonmichael:
> This sounds ok, and it's easily doable by checking the last character of the input string, but if this is done, no comments after descriptions or amounts will be allowed to end with <. Do you think that will cause any problems?

It's kind of ugly. Maybe it's not needed, especially if the input that you removed while stepping back is replaced as you step forward..

ghallak:
>Once you have changed an old input, you also have to re-enter all the following inputs. Should it remember the old inputs as you are stepping forward also ?

I have added this commit https://github.com/simonmichael/hledger/pull/1151/commits/e1bbefbd1b4ec5d7b7462b4a79666d7e0a655d0d to remember the previously entered values after stepping back and then stepping forward again.

I've also updated updated the instructions that shows when running the `add` command here  https://github.com/simonmichael/hledger/pull/1151/commits/9b6652b93f6418710b8742229d81959474262378 to reflect the changes.

simonmichael:
Thanks @ghallak !

PR #1159, opened 2020-01-02
Fix dead pointer in contributing

adimit:
Fix link in contributing

Since the git history indicated that the file was factored out into its own repository, I opted to just use the external link to the image.


adimit:
The failing test assumes it's 2019 still. I've created a PR for that: #1160

PR #1160, opened 2020-01-02
Fix auto postings test.

adimit:
Happy new year!

Since my PR #1159 failed (pure documentation) I checked what the problem was. Unfortunately, I don't really have a way of interpolating the current year in the result, so I figured adjusting the year manually would at least make the tests pass again!


simonmichael:
Thanks!

Issue #868, opened 2018-09-11
move --separator flag into the CSV rules file


simonmichael:
Follow-on from #176. We'd like to replace the --separator flag with a `separator CHAR` directive in the rules file, since it belongs more to the file than the command invocation.

schoettl:
What if CHAR is a white-space char, tab or space?

The config file would look like
```
separator 	
```
which, I think, is still better than `hledger --separator=$'\t'`.

Should the config file parser `parse "separator" token, one " ", and then one character`?

Or better (and more consistent) `parse "separator" token, one or more " ", and then Maybe one character`? If it parses Nothing for CHAR, it could default to space as separator.

simonmichael:
Good question. Till now, spaces surrounding csv rule values have always been ignored. I think we should require that a non-empty value must be visible to the eye in some way. 

Eg, by writing an escape sequence like `\t` instead of a literal tab. Unfortunately there's no [escape sequence](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences) for space. 

Or, by enclosing the separator in either single or double quotes:

    separator  ' '      # a single space
    separator  "  \t"   # two spaces and a tab

In this case, we'd need to decide how thoroughly to support quotes. Would they be allowed/ignored/useful in other csv rules ? Would they support quote escaping, like `"\""` ? Would there be any different behaviour of escape sequences (like `\t` or `\n`) in unquoted, single-quoted, double-quoted values ? Would an unbalanced quote be an error ? 

Or, we could use some made-up marker:

  separator <SPACE>
  separator \s\s\t

cfellinger:
On Sun, Feb 10, 2019 at 09:33:24AM -0800, Simon Michael wrote:

> Good question. Till now, spaces surrounding csv rule values have
> always been ignored. I think we should require that a non-empty
> value must be visible to the eye in some way.

> Eg, by writing an escape sequence like `\t` instead of a literal
> tab. Unfortunately there's no [escape sequence](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences)
> for space.

One might use a visible space "␣" (utf8: U2423, OPEN BOX) for those
places where an actual explicit space is needed.  We at home use it
nowadays to write our CSV harvesting rules and have a shortcut for
it in our ~/.XCompose files:

  <Multi_key> <space> <space>            : "␣"   U2423 # OPEN BOX

As an aside, I remember clearly the first time I used lowercase
[...]

PR #1161, opened 2020-01-02
Add separator to config

adimit:
Hi @simonmichael, I've added a new config directive `separator`, with its behaviour outlined in my comment in #868. The command line flag `--separator` can be used to override the rules directive `separator`. If neither is specified, the separator defaults to `,`.

I have *not* yet added the documentation, (this is a draft PR) as I first wanted to check back with you on the semantics of the new feature. Please try it out and send me your comments, and I'll add the docs.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -50,6 +50,7 @@ module Hledger.Read.Common (
   getAccountAliases,
   clearAccountAliases,
   journalAddFile,
+  getSpecialSeparators,
 
   -- * parsers
   -- ** transaction bits
@@ -162,7 +163,7 @@ data InputOpts = InputOpts {
      mformat_           :: Maybe StorageFormat  -- ^ a file/storage format to try, unless overridden
                                                 --   by a filename prefix. Nothing means try all.
     ,mrules_file_       :: Maybe FilePath       -- ^ a conversion rules file to use (when reading CSV)
-    ,separator_         :: Char                 -- ^ the separator to use (when reading CSV)
+    ,separator_         :: Maybe Char           -- ^ the separator to use (when reading CSV)
     ,aliases_           :: [String]             -- ^ account name aliases to apply
     ,anon_              :: Bool                 -- ^ do light anonymisation/obfuscation of the data
     ,ignore_assertions_ :: Bool                 -- ^ don't check balance assertions
@@ -175,14 +176,14 @@ data InputOpts = InputOpts {
 instance Default InputOpts where def = definputopts
 
 definputopts :: InputOpts
-definputopts = InputOpts def def ',' def def def def True def def
+definputopts = InputOpts def def def def def def def True def def
 
 rawOptsToInputOpts :: RawOpts -> InputOpts
 rawOptsToInputOpts rawopts = InputOpts{
    -- files_             = listofstringopt "file" rawopts
    mformat_           = Nothing
   ,mrules_file_       = maybestringopt "rules-file" rawopts
-  ,separator_         = fromMaybe ',' (maybecharopt "separator" rawopts)
+  ,separator_         = maybestringopt "separator" rawopts >>= getSpecialSeparators
   ,aliases_           = listofstringopt "alias" rawopts
   ,anon_              = boolopt "anon" rawopts
   ,ignore_assertions_ = boolopt "ignore-assertions" rawopts
@@ -193,6 +194,14 @@ rawOptsToInputOpts rawopts = InputOpts{
   }
 
 --- * parsing utilities
+
+-- | Parse special separator names TAB and SPACE, or return the first
+-- character. Return Nothing on empty string
+getSpecialSeparators :: String -> Maybe Char
+getSpecialSeparators "SPACE" = Just ' '
+getSpecialSeparators "TAB" = Just '\t'
+getSpecialSeparators (x:_) = Just x
+getSpecialSeparators [] = Nothing
 
 -- | Run a text parser in the identity monad. See also: parseWithState.
 runTextParser, rtp
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -50,7 +50,6 @@ module Hledger.Read.Common (
   getAccountAliases,
   clearAccountAliases,
   journalAddFile,
-  getSpecialSeparators,
 
   -- * parsers
   -- ** transaction bits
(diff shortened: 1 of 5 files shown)


adimit:
One deviation: multi-char separators that are neither `TAB` nor `SPACE` split on the first character. This behaviour is a bit contentious, but I thought it might be a good thing to avoid erroring too eagerly. A matter of preference. I could change it to error out on anything but `x:[]`, `"SPACE"` or `"TAB"`.

simonmichael:
Thanks @adimit! I seem to have missed your 12/2 comment. I will review asap, but it might be a day or two. If someone else can give feedback, please jump in.

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -175,14 +176,14 @@ data InputOpts = InputOpts {
...
-definputopts = InputOpts def def ',' def def def def True def def
+definputopts = InputOpts def def def def def def def True def def
 
 rawOptsToInputOpts :: RawOpts -> InputOpts
 rawOptsToInputOpts rawopts = InputOpts{
    -- files_             = listofstringopt "file" rawopts
    mformat_           = Nothing
   ,mrules_file_       = maybestringopt "rules-file" rawopts
-  ,separator_         = fromMaybe ',' (maybecharopt "separator" rawopts)
+  ,separator_         = maybestringopt "separator" rawopts >>= getSpecialSeparators
```
adimit: Instead of calling `getSpecialSeparators` in two places, we could declare `separator_` to be `Maybe String` and do all the parsing in `CsvReader`.

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -194,6 +195,14 @@ rawOptsToInputOpts rawopts = InputOpts{
 
 --- * parsing utilities
 
+-- | Parse special separator names TAB and SPACE, or return the first
+-- character. Return Nothing on empty string
+getSpecialSeparators :: String -> Maybe Char
+getSpecialSeparators "SPACE" = Just ' '
+getSpecialSeparators "TAB" = Just '\t'
+getSpecialSeparators (x:_) = Just x
```
adimit: We could do error handling here and split off this case from the `x:[]`. But the return signature would have to be IO (Maybe Char) because we're throwing exceptions.

simonmichael:
The semantics sound great. Some questions:

- which reserved words will be allowed ? Just SPACE and TAB ?
- will they be english words only ?
- what's the best thing for us to do if the user writes them with lower case letters ?
- which single characters will be allowed ? Or: can you think of any restrictions we might want to impose on that ?
- will single character separators be case sensitive ?

I intended to remove --separator, but since it's already there I guess there's no hurry. How useful do you think it will be ?

simonmichael:
PS, here's a test case for --separator that was reported to not work on windows, in case you have that platform handy and in case it might influence you: 
https://hastebin.com/hizevenari.pl

adimit:
Good questions all around. Let me try to answer:

- The purpose of the reserved words is to render otherwise invisible characters in the config in a user-friendly manner. I think whitespace characters qualify. I'm not sure `SPACE` is needed: I've never seen a SPACE-separated value file…
- I think English words only, yes. All other declarations in the config file are English-only afaict.
- I'm OK with either rejecting lowercase, or allowing promiscuous case (`taB` would work)
- At the moment, it's single characters from the latin-1 charset. Something (either in the testing setup, or in the parsing library) seems to choke on Unicode.
- I do not think case-sensitivity for single-char separators is necessary. We don't expect users to use cased characters anyway. Those that are… creative enough to use them anyway, will probably expect them to only split on the exact separator codepoint they specified.

I unfortunately don't have Windows handy. The idea behind keeping `--separator` is to make it overridable on the command line. I don't think that's an outstandingly common use case, so I wouldn't mind ditching it.

simonmichael:
>     * The purpose of the reserved words is to render otherwise invisible characters in the config in a user-friendly manner. I think whitespace characters qualify. I'm not sure `SPACE` is needed: I've never seen a SPACE-separated value file… I cannot think of anything useful but TAB, really, but SPACE could be included for completion's sake.

I know someone is going to want SPACES (one or more spaces), TABS (one or more tabs), and WHITESPACE (one or more whitespace chars)... but as you said, let's call this out of scope for now. I think we might as well include both TAB and SPACE.

>     * I'm OK with either rejecting lowercase, or allowing promiscuous case (`taB` would work)

Your call. Requiring all uppercase feels a little unusual for hledger, but it's simple.

>     * At the moment, it's single characters from the latin-1 charset. Something (either in the testing setup, or in the parsing library) seems to choke on Unicode.

Strange..

>     * I do not think case-sensitivity for single-char separators is necessary. We don't expect users to use cased characters anyway. Those that are… creative enough to use them anyway, will probably expect them to only split on the exact separator codepoint they specified.

So.. `separator X` will see 2 fields in `axbXc` ? Sounds fine.

> I unfortunately don't have Windows handy. The idea behind keeping `--separator` is to make it overridable on the command line. I don't think that's an outstandingly common use case, so I wouldn't mind ditching it.

I say let's drop it.

simonmichael:
> I say let's drop it.

Which might annoy somebody who just recently started using it, but it's marked "(experimental)" so I think it's ok.

adimit:
I do think the rules file `separator` directive is superior UX to the command line flag, so I hope people will not get *too* annoyed.

I think adding additional keywords will be simple, so let's do it when someone says they want it. I do think I'll change the keyword detection to be case-insensitive, since you say it's more typical for hledger to be case-insesitive. The added complication in the code is trivial.

I'll add the following to this PR:

- [ ] Remove command line flag
- [ ] Make keyword detection case insensitive
- [ ] investigate why Unicode separators are currently unavailable
- [ ] Add documentation on the new behaviour
- [ ] Delete old documentation on `--separator`

Thanks for your feedback!

simonmichael:
It's ok to keep --separator for now if you prefer. But as #868 mentions, it's a property of the CSV file, just like the number and identity of the fields, and we don't provide a command flag for those..

adimit:
I've debugged the Unicode-Problem: Cassava [separators are `Word8`](https://hackage.haskell.org/package/cassava-0.5.2.0/docs/Data-Csv.html#t:DecodeOptions), which means that nothing outside of one byte can be a separator in cassava.

adimit:
I've finished the modifications so far, and updated the documentation. I hope that's all the docs, I'm not sure if the separator is mentioned elsewhere.

adimit:
@simonmichael It seems Travis is running into timeouts :confused:

simonmichael:
Indeed, I restarted but no luck. Merging now, thanks for this!

adimit:
Great! Thanks for writing hledger :)

PR #1163, opened 2020-01-04
Use nubSort instead of nub . sort.

Xitian9:
A small performance improvement, but the change is trivial.


simonmichael:
#haskell seems to think this has the same efficiency as nub . sort, but it's at least a nice cleanup. I merged it locally and it is pushed to master, even though this issue's state hasn't updated for some reason. Thanks!

PR #1166, opened 2020-01-05
Calculate account leaf name at point of consumption.

Xitian9:
I think this cleans up a lot of clutter, and I doubt it results in much of a performance hit at all. It also removes any ambiguity about what is the proper label for a row in a MultiBalanceReport.


simonmichael:
Nice cleanup!

PR #1167, opened 2020-01-05
Improve font display on different OS

crispgm:
I opened this PR to improve the font on different operating system. The new css chooses OS-specific font at first and fallback to Helvetic Neue, Helvetic, and Arial. I am new hledger new and this is my first contribution and I am not familiar with Haskell, so user interface at first.

Here is the difference, based on hledger-web demo on macOS:

Before

before

After

after

I Windows and Linux screenshots are needed, I will post later. Thanks.

### DIFF
--- a/hledger-web/static/hledger.css
+++ b/hledger-web/static/hledger.css
@@ -18,6 +18,12 @@
 }
 
 /*------------------------------------------------------------------------------------------*/
+/* 2. fonts */
+body {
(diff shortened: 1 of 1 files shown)


simonmichael:
Nice! I see the kerning is better in numbers, eg.
Are you able to confirm that it makes things no worse on non-mac platforms ?

crispgm:
Windows 10:
windows10

crispgm:
@simonmichael Windows version is uploaded. Literally, it would not be worse, because the default font now "Helvetica Neue" is still a fallback.

simonmichael:
I trust you. Thanks for the contribution!

PR #1173, opened 2020-01-12
Bugfix in timeclock format parsing

### LINKED ISSUE #1169: timeclock format can contain an error and is silently parsed wrong
I noticed a bug when using hledger with the timeclock format: If the input file has a slight error, the output is very surprising and in certain reports I get wrong results but it is not at all obvious that they are wrong.

Reproduce:

```
$ hledger --version
hledger 1.14.99

$ cat > test.timeclock
i 2020-01-05 11:00 a
i 2020-01-05 17:00 a
o 2020-01-05 22:00

$ hledger -f test.timeclock register
2020/01/05 11:00-17:00                       (a)                      6.00h         6.00h
2020/01/05 22:00-23:59                       ()                       2.00h         8.00h
2020/01/06 00:00-15:33                       ()                      15.56h        23.56h
```

schoettl:
This should close #1169

Handle cases in timeclock format where a `i` line comes after an `i` line or an `o` line comes after an `o` line. See the shelltests which I added for details.

### DIFF
--- a/hledger-lib/Hledger/Data/Timeclock.hs
+++ b/hledger-lib/Hledger/Data/Timeclock.hs
@@ -57,6 +57,7 @@ instance Read TimeclockCode where
 timeclockEntriesToTransactions :: LocalTime -> [TimeclockEntry] -> [Transaction]
 timeclockEntriesToTransactions _ [] = []
 timeclockEntriesToTransactions now [i]
+    | tlcode i /= In = errorExpectedCodeButGot In i
     | odate > idate = entryFromTimeclockInOut i o' : timeclockEntriesToTransactions now [i',o]
     | otherwise = [entryFromTimeclockInOut i o]
     where
@@ -67,6 +68,8 @@ timeclockEntriesToTransactions now [i]
       o' = o{tldatetime=itime{localDay=idate, localTimeOfDay=TimeOfDay 23 59 59}}
       i' = i{tldatetime=itime{localDay=addDays 1 idate, localTimeOfDay=midnight}}
 timeclockEntriesToTransactions now (i:o:rest)
+    | tlcode i /= In = errorExpectedCodeButGot In i
+    | tlcode o /= Out =errorExpectedCodeButGot Out o
     | odate > idate = entryFromTimeclockInOut i o' : timeclockEntriesToTransactions now (i':o:rest)
     | otherwise = entryFromTimeclockInOut i o : timeclockEntriesToTransactions now rest
     where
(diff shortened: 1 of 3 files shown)


schoettl:
Hm, I cannot see why tests fail.

```
make test
stack test
```

both succeed on my system. Only, I get warnings:

```
Stack has not been tested with GHC versions above 8.6, and using 8.8.1, this may fail
Stack has not been tested with Cabal versions above 2.4, but version 3.0.0.0 was found, this may fail
```

schoettl:
Oh, I see, I can click on the trevis log... OK, I'll fix it.

tests/timeclock.test
```diff
@@ -18,11 +18,41 @@ o 2009/1/3 09:00:00
...
+>= 0
+
+# For a missing clock-out, now is implied
+<
+i 2020/1/1 08:00
+$ hledger -f - balance
+> /./
+>= 0
+
+# For a log starting not with clock-out, print error
```
simonmichael: "not starting with clock-in" ?

tests/timeclock.test
```diff
@@ -18,11 +18,41 @@ o 2009/1/3 09:00:00
...
+o 2020/1/1 08:00
+$ hledger -f - balance
+>2 /line 1: expected timeclock code i/
+>= !0
+
+# For a different log starting not with clock-out, print error
+<
+o 2020/1/1 08:00
+o 2020/1/1 09:00
+$ hledger -f - balance
```
simonmichael: How is this one different from the previous ? Should it be testing i o o instead ?

tests/timeclock.test
```diff
@@ -70,7 +100,6 @@ o 2009/1/3 09:00:00
 #>>>2
 #>>>=
 
-# 4. 
 
 # ledger timeclock example from #ledger
 # ==== consulting.timeclock
```
simonmichael: Great tests.

hledger-lib/Hledger/Data/Timeclock.hs
```diff
@@ -79,9 +79,13 @@ timeclockEntriesToTransactions now (i:o:rest)
...
 
-errorExpectedCodeButGot expected actual = errorWithSourceLine (tlsourcepos actual) $ "expected timeclock code " ++ (show expected) ++ " but got " ++ show (tlcode actual)
+errorExpectedCodeButGot expected actual = errorWithSourceLine line $ "expected timeclock code " ++ (show expected) ++ " but got " ++ show (tlcode actual)
+    where
+        line = case tlsourcepos actual of
+                  GenericSourcePos _ l _ -> l
+                  JournalSourcePos _ (l, _) -> l
 
-errorWithSourceLine (GenericSourcePos _ l _) msg = error $ "line " ++ show l ++ ": " ++ msg
+errorWithSourceLine line msg = error $ "line " ++ show line ++ ": " ++ msg
```
simonmichael: Would you mind combining this with the previous commit as a fixup.

tests/timeclock.test
schoettl: The previous test is different because a log with only one entry is a special case.

schoettl: We could test i o o or i o o o, but I would provide a minimal example as input.

simonmichael:
Thanks!

PR #1174, opened 2020-01-13
;doc: update shell output in hledger*/*.m4.md to use ISO date formatting

bwignall:
Note that this PR does not change the bodies of these files to use yyyy-mm-dd formatting; some yyyy/mm/dd remain there, subject to discussion.


simonmichael:
Thanks

PR #1176, opened 2020-01-14
web: Add IPv6 support

### LINKED ISSUE #1145: [hledger-web] Lacking IPv6 support
Hello, I'm trying out hledger (considering migrating from GnuCash) and found lack of IPv6 support in hledger-web's server to be an issue.

It's possible that the changes are minima, but I understand that everything is work :-). Just opening to document it.

Cheers and thank you.

Further information:

```
> uname -oprb
FreeBSD 13.0-CURRENT amd64 5b8f656016fa216bb99a3853141b8d4d9eb9b6de
> hledger-web --version
hledger-web 1.16.1
> hledger-web --serve --host '::1' --port 8080
hledger-web: --host requires an IP address, not "::1" (use -h to see usage)
```

CRTified:
# Motivation

This PR closes #1145. It allows `hledger-web` to be reachable over IPv6, e.g. by calling it like
```
hledger-web --host ::1
```

# Changes

 - Modify the `--host` validation in `hledger-web/Hledger/Web/WebOptions.hs` to allow IPv6 addresses
 - Modifies the default base URL to surround the host in `[` and `]` if an IPv6 address is used (required for proper IPv6-URLs in the browser)

# Problems

 - `Network.Wai.Handler.Launch.runHostPortUrl` uses the hardcoded host `127.0.0.1` to launch the browser. **Until there is a solution for https://github.com/yesodweb/wai-handlers/issues/1, this PR should probably not be merged, as the server is not reachable over the URL that is opened**
   - This problem is not relevant if `--serve` is used.
 - Not tested with [IPv4-mapped addresses](https://en.wikipedia.org/wiki/IPv6#IPv4-mapped_IPv6_addresses)

### DIFF
--- a/hledger-web/Hledger/Web/Main.hs
+++ b/hledger-web/Hledger/Web/Main.hs
@@ -16,7 +16,7 @@ import qualified Data.Text as T
 import Network.Socket
 import Network.Wai (Application)
 import Network.Wai.Handler.Warp (runSettings, runSettingsSocket, defaultSettings, setHost, setPort)
-import Network.Wai.Handler.Launch (runHostPortUrl)
+import Network.Wai.Handler.Launch (runHostPortFullUrl)
 import Prelude hiding (putStrLn)
 import System.Directory (removeFile)
 import System.Exit (exitSuccess, exitFailure)
@@ -106,5 +106,5 @@ web opts j = do
       putStrLn "Opening web browser..."
       hFlush stdout
       -- exits after 2m of inactivity (hardcoded)
-      Network.Wai.Handler.Launch.runHostPortUrl h p "" app
+      Network.Wai.Handler.Launch.runHostPortFullUrl h p u app
 
--- a/hledger-web/Hledger/Web/Settings.hs
+++ b/hledger-web/Hledger/Web/Settings.hs
@@ -45,7 +45,10 @@ defport = 5000
 
 defbaseurl :: String -> Int -> String
 defbaseurl host port =
-  "http://" ++ host ++ if port /= 80 then ":" ++ show port else ""
+  if ':' `elem` host then
+    "http://[" ++ host ++ "]" ++ if port /= 80 then ":" ++ show port else ""
+  else
+    "http://" ++ host ++ if port /= 80 then ":" ++ show port else ""
 
 -- Static setting below. Changing these requires a recompile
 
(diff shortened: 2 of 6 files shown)


Amarandus:
As soon as https://github.com/yesodweb/wai-handlers/pull/2 merges and is available, we can switch to `Network.Wai.Handler.Launch.runHostPortFullUrl` and solve this problem.
With a local copy of my changes, it already works successful.

@simonmichael I'm not used to the tooling. Do I need to add a constraint for `wai-handler-launch` to require at least version `3.0.2.5` (the one I PR'd)?

simonmichael:
@Amarandus yes, you'll change `wai-handler-launch >=1.3` in hledger-web/package.yaml to `wai-handler-launch >=3.0.3` or whatever. Your next `stack build hledger-web` will update hledger-web.cabal accordingly. I'll have to double check that won't cause a problem for the older GHC versions/package sets we support, but it looks like it'll be fine.

CRTified:
As https://github.com/yesodweb/wai-handlers/pull/2 is now merged, I tried updating the `package.yaml` accordingly. 
If I understand it correctly, we have to wait until the next LTS snapshot on stackage is available (which should be anytime soon).

---

In the meantime, I checked whether IPv4-mapped addresses work, and this is the case:

```
$ stack exec -- hledger-web --host ::ffff:127.0.0.1
Stack has not been tested with GHC versions above 8.6, and using 8.8.1, this may fail
Stack has not been tested with Cabal versions above 2.4, but version 3.0.0.0 was found, this may fail
17/Jan/2020:20:43:53 +0100 [Info#yesod-core] Application launched @(yesod-core-1.6.17-8khiymix0izHkaDEW7detY:Yesod.Core.Dispatch src/Yesod/Core/Dispatch.hs:163:11)
Serving web UI and API on ::ffff:127.0.0.1:5000 with base url http://[::ffff:127.0.0.1]:5000
This server will exit after 2m with no browser windows open (or press ctrl-c)
```

simonmichael:
> we have to wait until the next LTS snapshot on stackage

Hopefully not. stack will pull newer versions of deps from hackage when it needs to. If you're seeing an error while testing stack install in master, maybe I can help.

simonmichael:
Thanks!

Issue #933, opened 2018-12-07
Replace YYYY/MM/DD with YYYY-MM-DD in all documentation and in all printing commands


ad-si:
This has already been mentioned in https://github.com/simonmichael/hledger/issues/210, but there has been no further progress since then, so I wanted to pick it up again.

Is there still any discussion needed that this is the right way to go?
Otherwise I'd skip that part and gather a TODO list 😅:

Change output of
- [x] `hledger print`
- [x] `hledger rewrite`
- [x] `hledger add`
- [x] `hledger balancesheet`
- [x] `hledger close`
- [x] `hledger register`
- [x] `hledger print-unique`
- [x] `hledger register-match`

Did I miss something?

`hledger stats` already uses the ISO format.

If somebody is using rewrite in a automated fashion or similar, this would be a severe breaking change.
So maybe introduce the feature via a config flag first and issue a warning for all commands, that the default will change in the future and that in order to keep the current formatting one needs to add the config flag (e.g. `hledger --output-date-format yyyy/mm/dd rewrite`)?

I'd be happy to create a PR as soon as we're in agreement how this should be handled.

Looking forward to your feedback!

swalladge:
I too prefer yyyy-mm-dd format and would like to see this. I'd even be happy with a non-breaking change for this, such as a configuration option (does hledger support or plan to support rc files?).

simonmichael:
I've added a [comment](https://github.com/simonmichael/hledger/issues/210#issuecomment-451670931) to #210. With regard to ouput from all hledger tools, and all docs, I agree that we should switch to ISO YYYY-MM-DD as the default/preferred date format. There's still the question of whether to always show leading zeros, eg 0001-01-01 not 1-1-1, I'm not sure if the latter would ever be preferable; probably not. I feel adding options or a config file are a last resort, to be avoided unless really needed.

simonmichael:
Trying to focus this issue more on the output side, with #210 is covering input. "Documentation" of course involves both input and output examples.

What does "using rewrite in a automated fashion" mean - parsing hledger's output ? That's not a supported API (unlike hledger-api) and I'm not expecting this issue to cause too much compatibility pain.

simonmichael:
From #210:
> What I miss, is --output-date-format for hledger output (e.g. the register command's output). I think that would be useful to produce printed reports that are not awkward for "normal" clerks in Germany.

I suppose you're right. If ISO dates in output are really a problem for people ? Ledger calls it --date-format.

[...]

PR #1186, opened 2020-02-08
Fix documentation, make ISO date format standard in some places

schoettl:
For #933

### DIFF
--- a/hledger-lib/Hledger/Data/Transaction.hs
+++ b/hledger-lib/Hledger/Data/Transaction.hs
@@ -127,12 +127,12 @@ payeeAndNoteFromDescription t
     (p, n) = T.span (/= '|') t
 
 {-|
-Render a journal transaction as text in the style of Ledger's print command.
+Render a journal transaction as text similar to the style of Ledger's print command.
 
-Ledger 2.x's standard format looks like this:
+Adapted from Ledger 2.x and 3.x standard format:
 
 @
-yyyy/mm/dd[ *][ CODE] description.........          [  ; comment...............]
+yyyy-mm-dd[ *][ CODE] description.........          [  ; comment...............]
     account name 1.....................  ...$amount1[  ; comment...............]
     account name 2.....................  ..$-amount1[  ; comment...............]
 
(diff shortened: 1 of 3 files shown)


hledger-lib/Hledger/Data/Transaction.hs
```diff
@@ -132,7 +132,7 @@ Render a journal transaction as text in the style of Ledger's print command.
 Ledger 2.x's standard format looks like this:
 
 @
-yyyy/mm/dd[ *][ CODE] description.........          [  ; comment...............]
+yyyy-mm-dd[ *][ CODE] description.........          [  ; comment...............]
```
simonmichael: Not correct I think, both Ledger 2 and 3 print dates with slashes.

hledger-lib/Hledger/Read/TimeclockReader.hs
```diff
@@ -14,7 +14,7 @@ From timeclock.el 2.6:
 A timeclock contains data in the form of a single entry per line.
 Each entry has the form:
 
-  CODE YYYY/MM/DD HH:MM:SS [COMMENT]
+  CODE YYYY-MM-DD HH:MM:SS [COMMENT]
```
simonmichael: https://www.ledger-cli.org/3.0/doc/ledger3.html#Time-Keeping says it contains slashes, but Ledger reads slash or dash or dot, like hledger. Emacs' timeclock.el requires slashes. I tweak this comment later maybe.

hledger-lib/Hledger/Data/Transaction.hs
schoettl: Oh, right. But this is the documentation for the showTransaction function, so we also have to fix the sentence above my change.

hledger-lib/Hledger/Read/TimeclockReader.hs
schoettl: I reverted this change for now, to keep the compatibility to the original doc and Emacs.

hledger-lib/Hledger/Data/Transaction.hs
schoettl: I changed the above part of comment.

PR #1193, opened 2020-02-17
ui: don't force --auto/--forecast on; merge --future with --forecast

simonmichael:
Periodic transactions were always on, which meant that periodic
transactions occurring today were always shown, in future or present
mode.

Now, both periodic transactions and display of future transactions are
controlled by --forecast, and toggleable by the F key ("forecast
mode"). The --future flag has been dropped (it still works as a hidden
alias for --forecast, but is deprecated).

It seemed to also make sense to leave auto postings off by default, like hledger.


simonmichael:
An alternative solution could be: always start periodic transactions from tomorrow, not today.

simonmichael:
Simplified to avoid reloading the journal, and merged.

PR #1200, opened 2020-02-28
lib: Replace some utility functions with library functions.

Xitian9:
Minor clean-up.

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -84,6 +84,7 @@ import Control.Applicative.Permutations
 import Control.Monad (guard, unless)
 import "base-compat-batteries" Data.List.Compat
 import Data.Default
+import Data.Foldable (asum)
 import Data.Maybe
 import qualified Data.Set as Set
 import Data.Text (Text)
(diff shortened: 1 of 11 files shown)


simonmichael:
Thanks @Xitian9, I'll review soon. That CI failure is my fault.

Xitian9:
Actually, looking at the definition of `Journal`, I think this is not actually correct. It looks like `mempty` is only a left unit for `mappend`, not a right unit, i.e. `mappend` is right-biased in some fields.

I'll remove the relevant parts from this pull request, but this seems like something which should be changed in general. I can think of two options:
(1) Journal is just a Semigroup, not a Monoid (possibly a member of Semigroup and Default).
(2) We change the definition of `mappend` to prefer non-empty fields.

(1) Seems like the most straightforward solution. What do you think?

simonmichael:
@Xitian9, thanks for the cleanups. We currently support GHC 8.0 and up, and should build warning-free; for this it looks like more work is needed. You can use the "buildtest" make rules to check this, eg:
```
make help-buildtest
make buildtest-all
make buildtest-stack-8.0.yaml
```

Xitian9:
It should now build without error on stack 8.0 and above. I'm not sure why CI isn't running.

simonmichael:
Looking good. Thank you!

PR #1209, opened 2020-03-15
simpler, more flexible generation of postings from csv

simonmichael:
A rewrite and simplification of the posting-generating code. The
"special handling for pre 1.17 rules" should now be less noticeable.
amount1/amount2 no longer force a second posting or explicit amounts
on both postings. (Only amount/amount-in/amount-out do that.)
Error messages and handling of corner cases may be more robust, also.

@adept, does it sound ok ? This scheme seems like one we would have considered, but I think maybe we didn't, perhaps because the old code really didn't allow it.

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -457,6 +457,8 @@ journalfieldnamep = do
   lift (dbgparse 2 "trying journalfieldnamep")
   T.unpack <$> choiceInState (map (lift . string . T.pack) journalfieldnames)
 
+maxpostings = 9
+
 -- Transaction fields and pseudo fields for CSV conversion.
 -- Names must precede any other name they contain, for the parser
 -- (amount-in before amount; date2 before date). TODO: fix
@@ -468,7 +470,7 @@ journalfieldnames =
           ,"balance" ++ i
           ,"comment" ++ i
           ,"currency" ++ i
-          ] | x <- [1..9], let i = show x]
+          ] | x <- [1..maxpostings], let i = show x]
   ++
   ["amount-in"
   ,"amount-out"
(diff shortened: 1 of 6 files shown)


simonmichael:
PS here's the docs change: https://github.com/simonmichael/hledger/pull/1209/commits/feef6fdc4838a5fdae8261d33e77b195477251de#diff-9e1414f093c6a11d19cc91eb4cc88665

I broke up mkPostings completely into its helpers (getAmount, getBalance, getAccount). I thought I had reduced code size more, but since I've been wrestling with this code for days I'm just happy to pass tests and make the rules a little more flexible. Probably we'll find more corner cases later but I think it'll be easier to handle them.

hledger-lib/Hledger/Read/CsvReader.hs
```diff
@@ -899,98 +864,98 @@ transactionFromCsvRecord sourcepos rules record = t
...
+    fieldnames = map (("amount"++show n)++) ["","-in","-out"]
+                 -- For posting 1, also recognise the old field names.
+                 -- For posting 2, the same but only if posting 1 needs balancing.
+                 ++ if n==1 || n==2 && not p1IsVirtual then unnumberedfieldnames else []
+    nonzeroamounts = [(f,a') | f <- fieldnames
+                     , Just v@(_:_) <- [strip . renderTemplate rules record <$> hledgerField rules record f]
+                     , let a = parseAmount rules record currency v
+                     , not $ isZeroMixedAmount a
+                       -- With an old field name, flip the sign in posting 2
+                     , let a' = if f `elem` unnumberedfieldnames && n==2 then -a else a
```
adept: I think that for better compatibility with pre-1.17 this line should be:
```
let a' = if f `elem` unnumberedfieldnames && n==2 then costOfMixedAmount(-a) else a

```

simonmichael: Thanks!

PR #1210, opened 2020-03-16
Improve code readability

ysangkok:
Use more type signatures and avoid dense list comprehension.


simonmichael:
@ysangkok thanks for your input on #haskell, and for taking the time to make a pull request. I'm inclined not to merge this one, as I think it goes too far with type signatures and whitespace and it isn't worth more than doubling the amount of code.

PR #1211, opened 2020-03-16
Add rules example for the new YNAB Web csv export format

erplsf:
Was looking around examples and saw an outdated example for YNAB, decided to add a rules file that works for the latest export.

### DIFF
--- a/examples/csv/ynab-web-register.csv.rules
+++ b/examples/csv/ynab-web-register.csv.rules
@@ -1,0 +1,13 @@
+skip 1
+
+account1     assets:personal:ynab
+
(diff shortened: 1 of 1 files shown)


simonmichael:
Great! Thanks for that. Does the outdated example need removing ?

Thanks for using `ci skip` - it should be no longer necessary, we are now using only Github CI and automatically ignoring doc commits.

erplsf:
I'd leave it, as some people may still use YNAB 4, because it was a radical change in product, and some folks dislike new web version. But feel free to ditch it as there's enough examples of rules for csv files in there still.

simonmichael:
Ok cool, let's keep both. The old one has a comment heading describing it, maybe you could add that to the new one to reduce confusion ?

erplsf:
Sure, will see to it.

PR #1213, opened 2020-03-22
lib: More consistent period formatting.

Xitian9:
Since date formatting was switched to ISO format, some of the results of showPeriod are inconsistent, or a bit silly. For example
```
showPeriod (PeriodBetween (fromGregorian 2000 01 01) (fromGregorian 2000 02 01)) == "2000-01-01-2000-01-31"
showPeriod (MonthPeriod 2000 01) == "2000/01"
```
I've updated this to use ISO formatting whenever possible, including the ISO date range separator, so these are now
```
showPeriod (PeriodBetween (fromGregorian 2000 01 01) (fromGregorian 2000 02 01)) == "2000-01-01/2000-01-31"
showPeriod (MonthPeriod 2000 01) == "2000-01"
```

There are still a few ambiguities to deal with. I think technically, since these are periods, we should be representing DayPeriod, MonthPeriod, etc. using durations, so
```
showPeriod (MonthPeriod 2000 01) == "2000-01/P1M"
showPeriod (DayPeriod (fromGregorian 2000 01 01)) == "2000-01/P1D"
```
That said, it's probably reasonable to omit these for the sake of brevity.

On the other hand, the week period here is a bit strange. The ISO-recommended answer would be
```
showPeriod (WeekPeriod (fromGregorian 2000 01 01)) == "2000-01-01/P1W"
```
but it's not clear this is what we want here. The existing representation shows the ISO week number, in which case we might actually want
```
showPeriod (WeekPeriod (fromGregorian 2000 01 03)) == "2000-W01"
```
This presupposes that the WeekPeriod starts on a Monday. The existing representation allows the start date to be an arbitrary day, but I believe the reports insist of starting on a Monday in any case. I'm unsure what people would prefer here: whether to leave it as is, or change it to one of the above.


simonmichael:
Just a quick reply - yes, this is a can of worms. :/

As someone used to using slash to separate parts of a date, "Budget performance in 2016-12-01/2016-12-03" looks really weird. Maybe "2016-12-01..2016-12-03" could work.

Xitian9:
Given that it's an agreed standard I would advocate for using a `/`, but I agree it takes some getting used to. I'm happy with any solution that gets rid of `2000-01-01-2000-01-31`. The double dot `2000-01-01..2000-01-31` looks reasonable to me.

Maybe I could ask for some clarification on what the WeekPeriod is supposed to represent. Is it a single week from an arbitrary given date, or is it a calendar week which contains a given arbitrary date?

Xitian9:
Here are some options I can think of for representing `WeekPeriod (fromGregorian 2000 01 01)`:
```
"2000-01-01w52"    (existing representation)
"2000-01-01/P1W"   (ISO standard for durations)
"1999-W52-6"       (ISO date format for year-week-day of week)
"1999-W52-6/P1W"   (ISO date format for year-week-day of week, plus duration)
```

simonmichael:
Hi @Xitian9. Re changing `-` to `/`, do you have a good link for the relevant part of the ISO standard ? Or might there be a wikipedia page discussion international usage, like the very useful one for decimal separators ?

There may also be an impact on UI, in date queries like `date:FROM-TO`, `date:FROM-` and `date:-TO`. These are mentioned (briefly) at https://hledger.org/hledger.html#report-start-end-date and https://hledger.org/hledger.html#queries (source: hledger/hledger.m4.md). I wonder what we should do with these. Breaking existing queries and scripts would be bad.

Re weeks, yes a WeekPeriod is a type of Period, representing the seven days starting at the given date, which in current hledger is always a monday.

I think we must be a little careful to minimise "gratuitous" output changes unless a lot of folks are on board. Would you be able to mention the changes being proposed here on the mail list, so we can get more input ?

simonmichael:
PS we could also discuss the rendering of week periods as a separate issue, to keep things moving.

Xitian9:
Wikipedia discusses the ISO standard [here](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals). Unfortunately I haven't been able to find a page discussing common date interval formats: I suspect that it's relatively informal outside of the ISO standard.

I suspect that allowing either `-` or `/` for `to` in queries is probably doable. Leading or trailing `-` or `/` can be unambiguously interpreted, and a four-digit year will also make it unambiguous. The only potential issue is if the year is omitted, giving something like `1-4-2-4`. I think that since years need to be 4-digits, even this can be unambiguously parsed. But I'm curious how much of an issue you think this actually is, as I have a hard time imagining somebody typing something like this in and expecting it to work.

I'll mention it on the mailing list.

simonmichael:
The wikipedia page mentions double hyphen (`--`), too.

Xitian9:
My impression is that the double-hyphen is basically just to get around `/` being a reserved character in filenames. I'm actually not sure whether I'd prefer `--` or `..`: the former is in the standard, but the latter has greater visual distinction from a hyphen.

alerque:
The double dot to indicate ranges is used by `git` and other geeky tools and in some math programs I've seen. That would probably be my personal preference.

A double dash is frequently used as an ASCII stand-in for an em-dash (–). Uses for an em-dash include ranges of figures. It's probably more widely recognizable to use this.

Xitian9:
We seem to have an overall preference for `..` over `/`. I'm happy to go with that. I'll update the commit.

Xitian9:
Just a quick poke to see if there are any problems left to be addressed here.

simonmichael:
Thanks for the ping. I think this PR makes two changes:

1. capitalise quarter notation, 2008q1 -> 2008Q1. 

Should `w` also be capitalised then , 2020-01-06w02 -> 2020-01-06W02 ?

2. use `..` instead of `-` to denote date ranges, avoiding clash with new YYYY-MM-DD format. 

`hledger bal assets -1 -ME date:feb..` gave a parse error, adding the new syntax to doubledatespanp & todatespan will probably take care of it.

Would you mind rebasing against latest master. I think there's one new test to update, in forecast.test.

simonmichael:
Let's mention the `..` syntax in addition to `-` at: https://hledger.org/hledger.html#report-start-end-date and https://hledger.org/hledger.html#period-expressions (in hledger.m4.md). Or I can do that if you prefer.

simonmichael:
Also just FYI, balancesheet's title shows `,,` in an attempt to distinguish a series of end dates from a continuous date span. I think we'll just show `..` here too, if you have no objection.

Xitian9:
It also changes the format of the month period, 2020/01 -> 2020-01, for consistency.

I didn't touch the week period formatting because that was separated out into #1216. If you're happy to make a call on that, I can incorporate that into this pull request too.

Those changes sound reasonable to me. I'll include them.

simonmichael:
Oh thanks. I'll reply there. My feeling is for the coming release we should just capitalise W.

Xitian9:
Changes made. There's a build failure with stack8.0, which seems to relate to a dependency issue with base-compat and aeson-pretty.

simonmichael:
Thank you very much! Merged. The CI failure is my fault.

PR #1217, opened 2020-04-01
;lib: use lexical ordering for subaccounts

legrostdg:
Attempt to fix https://github.com/simonmichael/hledger/issues/1218

### DIFF
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -243,7 +243,7 @@ sortAccountTreeByDeclaration a
   | null $ asubs a = a
   | otherwise      = a{asubs=
       sortOn accountDeclarationOrderAndName $
-      map sortAccountTreeByDeclaration $ asubs a
+      asubs a
       }
 
 accountDeclarationOrderAndName :: Account -> (Int, AccountName)


legrostdg:
It does fix my example, now `hledger b --file test.journal` returns:

```
$ stack exec -- hledger b --file ~/test.journal
                   3  2.a
                   1    1.c:0.d
                   1    2.e
                   1    3.f
                  -3  1.b
--------------------
                   0
```

however the declaration order is not respected anymore for subaccounts... So, given this `test2.journal`:

```
account 2.a
account 2.a:1.c:0.d
account 2.a:3.f
account 2.a:2.e
account 1.b

2020/01/01 bla
    1.b          -3
    2.a:2.e      +1
    2.a:1.c:0.d  +1
    2.a:3.f      +1
```

On `master`:

```
$ hledger b --file ~/test2.journal 
                   3  2.a
                   1    3.f
                   1    2.e
                   1    1.c:0.d
                  -3  1.b
--------------------
                   0
```

On this branch:

```
$ stack exec -- hledger b --file ~/test2.journal
                   3  2.a
                   1    1.c:0.d
                   1    2.e
                   1    3.f
                  -3  1.b
--------------------
                   0
```

legrostdg:
I guess it's better than before (it does fix my use case where accounts are declared in lexical order :)), but it's a complete fix: declaration order is only kept at the top level... I tried to understand why subaccounts with subaccounts are always sorted after subaccounts without subaccounts, but I got lost in the source :).

Issue #1025, opened 2019-05-14
Support customization/localization of reports


mildred:
Reports (such as balance sheet or income statement) should be able to be customized so they can be localized and translated.

Some simple words like Assets or Liabilities must be customized (in French, it corresponds to *Actif* and *Passif*)

Split from #1006

simonmichael:
Agreed! Thanks for the issue. We need someone to research and/or build a small prototype, eg localising just the title of the balancesheet report. Also, to think about the hledger as a whole and clarify which parts can/should be localised. Could you help ?

simonmichael:
Related: #1219

simonmichael:
Related: #1219

simonmichael:
Perhaps implementing the flexible `balancereport` command would be a good start. Help welcome, including a better name for it.

simonmichael:
Open questions:

- in current compound balance reports, each subsection shows an account type. In the mockups above they are based on arbitrary queries, and account types aren't mentioned. What would account types be used for ?
- the above doesn't really address how to localise headings of the standard builtin reports, unless we allowed custom commands to override builtin commands, and it would still be a bit over-complex and error-prone.
- it would be useful to be able to localise the standard report titles once eg in a user config file, and not have to do it in every journal or on every command line.
- it could be useful to start doing traditional l10n, building in translations of the standard report titles and selecting them based on system locale with no configuration needed. There would need to be a way to override this.

alerque:
Also things like month names should follow the output locale:

```
$ LANG=tr_TR.UTF-8 hledger -f=maliye.ledger bal Daire --transpose --no-total --empty -M
     ||  Daire 1   Daire 2   Daire 3   Daire 4   Daire 5
=====++==================================================
 Jan || ₺-250,00         0         0  ₺-250,00         0
 Feb || ₺-250,00         0         0  ₺-250,00         0
 Mar || ₺-250,00         0         0  ₺-250,00         0
 Apr || ₺-250,00  ₺-250,00  ₺-250,00  ₺-250,00  ₺-250,00
```

Should be:

```
     ||  Daire 1   Daire 2   Daire 3   Daire 4   Daire 5
=====++==================================================
 Oca || ₺-250,00         0         0  ₺-250,00         0
 Şub || ₺-250,00         0         0  ₺-250,00         0
 Mar || ₺-250,00         0         0  ₺-250,00         0
[...]

PR #1219, opened 2020-04-01
use Revenue and Expense account names in incomestatement

legrostdg:
fixes https://github.com/simonmichael/hledger/issues/1025 for incomestatement

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -302,7 +302,7 @@ journalAccountTypeQuery :: AccountType -> Regexp -> Journal -> Query
 journalAccountTypeQuery atype fallbackregex j =
   case M.lookup atype (jdeclaredaccounttypes j) of
     Nothing -> Acct fallbackregex
-    Just as ->
+    Just (as,_) ->
       -- XXX Query isn't able to match account type since that requires extra info from the journal.
       -- So we do a hacky search by name instead.
       And [
@@ -311,7 +311,7 @@ journalAccountTypeQuery atype fallbackregex j =
         ]
       where
         differentlytypedsubs = concat
-          [subs | (t,bs) <- M.toList (jdeclaredaccounttypes j)
+          [subs | (t,(bs, _)) <- M.toList (jdeclaredaccounttypes j)
               , t /= atype
               , let subs = [b | b <- bs, any (`isAccountNamePrefixOf` b) as]
           ]
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -468,7 +468,7 @@ data Journal = Journal {
   ,jincludefilestack      :: [FilePath]
   -- principal data
   ,jdeclaredaccounts      :: [(AccountName,AccountDeclarationInfo)] -- ^ Accounts declared by account directives, in parse order (after journal finalisation)
-  ,jdeclaredaccounttypes  :: M.Map AccountType [AccountName]        -- ^ Accounts whose type has been declared in account directives (usually 5 top-level accounts)
+  ,jdeclaredaccounttypes  :: M.Map AccountType ([AccountName], Maybe Text) -- ^ Accounts whose type has been declared in account directives (usually 5 top-level accounts)
   ,jcommodities           :: M.Map CommoditySymbol Commodity        -- ^ commodities and formats declared by commodity directives
   ,jinferredcommodities   :: M.Map CommoditySymbol AmountStyle      -- ^ commodities and formats inferred from journal amounts  TODO misnamed - jusedstyles
   ,jpricedirectives       :: [PriceDirective]                       -- ^ All market price declarations (P directives), in parse order (after journal finalisation).
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -60,7 +60,7 @@ module Hledger.Read.Common (
   getAccountAliases,
   clearAccountAliases,
   journalAddFile,
-  getAccountTypeName,
+  getAccountTypeTitle,
 
   -- * parsers
   -- ** transaction bits
@@ -399,10 +399,10 @@ journalAddFile f j@Journal{jfiles=fs} = j{jfiles=fs++[f]}
   -- append, unlike the other fields, even though we do a final reverse,
   -- to compensate for additional reversal due to including/monoid-concatting
 
-getAccountTypeName :: AccountType -> Text -> Journal -> Text
-getAccountTypeName actype defaultatypename j = case M.lookup actype (jdeclaredaccounttypes j) of
-    Nothing -> defaultatypename
-    Just (_,Nothing) -> defaultatypename
+getAccountTypeTitle :: AccountType -> Text -> Journal -> Text
+getAccountTypeTitle actype defaultatypetitle j = case M.lookup actype (jdeclaredaccounttypes j) of
+    Nothing -> defaultatypetitle
+    Just (_,Nothing) -> defaultatypetitle
     Just (_,Just atname) -> atname
 
 -- A version of `match` that is strict in the returned text
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -60,6 +60,7 @@ module Hledger.Read.Common (
   getAccountAliases,
   clearAccountAliases,
   journalAddFile,
+  getAccountTypeName,
 
   -- * parsers
   -- ** transaction bits
@@ -353,9 +354,11 @@ getAmountStyle commodity = do
     let effectiveStyle = listToMaybe $ catMaybes [specificStyle, defaultStyle]
     return effectiveStyle
 
-addDeclaredAccountType :: AccountName -> AccountType -> JournalParser m ()
-addDeclaredAccountType acct atype =
-  modify' (\j -> j{jdeclaredaccounttypes = M.insertWith (++) atype [acct] (jdeclaredaccounttypes j)})
+addDeclaredAccountType :: AccountName -> AccountType -> Maybe Text -> JournalParser m ()
+addDeclaredAccountType acct atype matname =
+  modify' (\j -> j{jdeclaredaccounttypes = M.insertWith insertacctype atype ([acct],matname) (jdeclaredaccounttypes j)})
+  where
+    insertacctype (accts,matname) (accts',matname') = (accts ++ accts',matname' <> matname)
 
 pushParentAccount :: AccountName -> JournalParser m ()
 pushParentAccount acct = modify' (\j -> j{jparseparentaccounts = acct : jparseparentaccounts j})
@@ -395,6 +398,12 @@ journalAddFile :: (FilePath,Text) -> Journal -> Journal
 journalAddFile f j@Journal{jfiles=fs} = j{jfiles=fs++[f]}
   -- append, unlike the other fields, even though we do a final reverse,
   -- to compensate for additional reversal due to including/monoid-concatting
+
+getAccountTypeName :: AccountType -> Text -> Journal -> Text
+getAccountTypeName actype defaultatypename j = case M.lookup actype (jdeclaredaccounttypes j) of
+    Nothing -> defaultatypename
+    Just (_,Nothing) -> defaultatypename
+    Just (_,Just atname) -> atname
 
 -- A version of `match` that is strict in the returned text
 match' :: TextParser m a -> TextParser m (Text, a)
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -351,13 +351,18 @@ accountdirectivep = do
       Just
       $ lookup accountTypeTagName tags
     metype = parseAccountTypeCode <$> mtypecode'
+    mtypename = lookup accountTypeNameTagName tags
 
   -- update the journal
   addAccountDeclaration (acct, cmt, tags)
   case metype of
     Nothing         -> return ()
-    Just (Right t)  -> addDeclaredAccountType acct t
+    Just (Right t)  -> addDeclaredAccountType acct t mtypename
     Just (Left err) -> customFailure $ parseErrorAt off err
+
+-- The special tag used for declaring account type name. Used for display in
+-- incomestatement, balancestatement, ...
+accountTypeNameTagName = "typename"
 
 -- The special tag used for declaring account type. XXX change to "class" ?
 accountTypeTagName = "type"
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -351,18 +351,18 @@ accountdirectivep = do
       Just
       $ lookup accountTypeTagName tags
     metype = parseAccountTypeCode <$> mtypecode'
-    mtypename = lookup accountTypeNameTagName tags
+    mtypetitle = lookup accountTypeTitleTagName tags
 
   -- update the journal
   addAccountDeclaration (acct, cmt, tags)
   case metype of
     Nothing         -> return ()
-    Just (Right t)  -> addDeclaredAccountType acct t mtypename
+    Just (Right t)  -> addDeclaredAccountType acct t mtypetitle
     Just (Left err) -> customFailure $ parseErrorAt off err
 
 -- The special tag used for declaring account type name. Used for display in
 -- incomestatement, balancestatement, ...
-accountTypeNameTagName = "typename"
+accountTypeTitleTagName = "type-title"
 
 -- The special tag used for declaring account type. XXX change to "class" ?
 accountTypeTagName = "type"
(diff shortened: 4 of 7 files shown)


legrostdg:
also fixes #1025 for balancesheet and balancesheetequity

legrostdg:
note that it keeps the title intact (not localized), but this would need to add a custom syntax...

simonmichael:
Thanks for the PR. I see CI is failing because of a GHC warning - `make ghcid` is a good way to see these locally I believe.

The idea is neat, but I'm not sure it's quite robust enough. Eg in my setup, which I admit is a bit unusual, assets/liabilities/etc. are under top-level `JS` and `sm` accounts, and this change picks `sm` for all section titles.

legrostdg:
The warnings seem unrelated to this PR:

```
hledger-lib/Text/Megaparsec/Custom.hs:181:0: error:
     error: missing binary operator before token "("
      181 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
181 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^

hledger-lib/Text/Megaparsec/Custom.hs:196:0: error:
     error: missing binary operator before token "("
      196 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
196 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^

hledger/Hledger/Cli/Commands/Roi.hs:210:0: error:
     error: missing binary operator before token "("
      210 | #if MIN_VERSION_math_functions(0,3,0)
          | 
    |
210 | #if MIN_VERSION_math_functions(0,3,0)
    | ^

hledger-lib/Text/Megaparsec/Custom.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE BangPatterns #-}
  | ^

hledger/Hledger/Cli/Commands/Roi.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE ParallelListComp, CPP #-}
  | ^
Failed, no modules loaded.

hledger/Hledger/Cli/Commands/Roi.hs:210:0: error:
     error: missing binary operator before token "("
      210 | #if MIN_VERSION_math_functions(0,3,0)
          | 
    |
210 | #if MIN_VERSION_math_functions(0,3,0)
    | ^
hledger/Hledger/Cli/Commands/Roi.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE ParallelListComp, CPP #-}
  | ^
hledger-lib/Text/Megaparsec/Custom.hs:181:0: error:
     error: missing binary operator before token "("
      181 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
181 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^
hledger-lib/Text/Megaparsec/Custom.hs:196:0: error:
     error: missing binary operator before token "("
      196 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
196 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^
hledger-lib/Text/Megaparsec/Custom.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE BangPatterns #-}
  | ^
```

legrostdg:
Other try: I added a tag named "typename" to change the display name of account types.

legrostdg:
@simonmichael is this solution, based on a new tag, would suit you? `typename` is maybe not the best name ever, though...

simonmichael:
> The warnings generated by make ghcid seem unrelated

You're right, I think those are fixed by doing `stack build hledger` once before `make ghcid`.

> would this solution, based on a new tag, suit you?

I'm not sure, could you show me draft documentation ?

legrostdg:
```
account basic  ; type:Liability
  ; typename:Basic account
```

```
$ hledger accounts
Balance Sheet 2019/01/01-2020/04/06

                                                                       ||  2020/04/06 
=======================================================================++=============
 Assets                                                                ||             
-----------------------------------------------------------------------++-------------
...
-----------------------------------------------------------------------++-------------
                                                                       || 
=======================================================================++=============
 Basic account                                                         ||             
-----------------------------------------------------------------------++-------------
 basic                                                                 || 
...
=======================================================================++=============
 Net:                                                                  ||
```

legrostdg:
The warnings are still there if I run `stack build hledger` before `make ghcid`.

legrostdg:
@simonmichael did you have time to have a look?

simonmichael:
@legrostdg thanks for the ping. I'll investigate that warning today. I'm not too keen on adding configuration options unless they're necessary and easy to understand, could you show how you'd explain the proposed feature to users in the manual ?

legrostdg:
I can do that, but I would prefer if you said the feature is OK for you before. I don't have much time to invest in coding, so I can't write doc etc. for something which will not get merged anyway.

legrostdg:
I think that this is necessary, at least for non-English speakers. I guess you got the idea of this branch from my previous comment: the user can define new names for account types through this `typename` (could be changed to something else if needed), and this new names will be used to generate bs, etc.

simonmichael:
PS as always, pasting it here would be fine - no need to learn the doc toolchain..

legrostdg:
Maybe this is the best way for you, as the one that wrote the code and is a native English speaker (I assume). To me (and I guess to others), writing docs in English is not the easiest, and writing a spec without knowing the implied cost in coding seems like a waste of time.

legrostdg:
The rationale:

> I think that this is necessary, at least for non-English speakers.

The "spec":

> I guess you got the idea of this branch from my previous comment: the user can define new names for account types through this `typename` (could be changed to something else if needed), and this new names will be used to generate bs, etc.

https://github.com/simonmichael/hledger/pull/1219#issuecomment-610779497

legrostdg:
Of course I can try to write a better documentation, but I think you have enough there to get the idea...

simonmichael:
I hear that. It's not the easiest thing for me either. I think when you try to write the doc you often will see that it's hard to describe the thing proposed, in an intuitive way.

We can say the doc would go between https://hledger.org/journal.html#account-types and https://hledger.org/journal.html#account-display-order .

I'm not sure of the title. "Typename" isn't immediately descriptive to me. I don't know yet if we can think of a better name.

Which accounts would this tag be legal on ? All, top level, accounts with explicit type tags, accounts whose type is implicitly recognised based on name, some combination of these ? If used on some other account, should it be ignored or give an error ?

The effect of the tag will be to customise the report section headings in the bs, cf and is reports. But not the main title, right ? I expect that will feel like a missing feature unless we support that too.

simonmichael:
As described so far, you would have to define these headings in every journal file you want to report on. This could get tedious, also you won't always want or be able to write to every journal file. It seems more properly a user configuration option.

simonmichael:
Would command line options work better ? We have common options that modify report appearance.

simonmichael:
Options mockup:
```
bs --help
--title=TITLE
--asset-title=TITLE
--liability-title=TITLE

bse --help
--title=TITLE
--asset-title=TITLE
--liability-title=TITLE
--equity-title=TITLE

is --help
--title=TITLE
--revenue-title=TITLE
--expense-title=TITLE
```

legrostdg:
> As described so far, you would have to define these headings in every journal file you want to report on. This could get tedious, also you won't always want or be able to write to every journal file. It seems more properly a user configuration option.

I think this is the right way to define this: a user configuration option to define default names could also be nice to change the default everywhere, but I think we should have a way to change these names for each journal. Note that we already have a sane default for this account type names: "Assets", "Liabilities", etc.

> Would command line options work better ? We have common options that modify report appearance.

Command line options would have to be reminded by the user, with possibly different options for each journal file. But they also could be nice to override the `typename` (or whatever we chose to name it).

simonmichael:
And since you wouldn't want to type those every time, and since we have so far chosen not to have a dedicated config file, we would need to suggest to users how to use a shell alias and/or arguments file to make this easier (and this would add motivation for some kind of easier config file support).

Why do you say these would need to vary per journal ?

legrostdg:
> I hear that. It's not the easiest thing for me either. I think when you try to write the doc you often will see that it's hard to describe the thing proposed, in an intuitive way.

OK.

> We can say the doc would go between https://hledger.org/journal.html#account-types and https://hledger.org/journal.html#account-display-order .

OK, that helped.

> I'm not sure of the title. "Typename" isn't immediately descriptive to me. I don't know yet if we can think of a better name.
> 
> Which accounts would this tag be legal on ? All, top level, accounts with explicit type tags, accounts whose type is implicitly recognised based on name, some combination of these ?

Maybe "type-title"? This way we get that this is intend for account types only. The other accounts get their names directly from the account definition. My first solution was to also get the account type names from there, but I think that we agreed that it was not a good solution.

> If used on some other account, should it be ignored or give an error ?

Maybe "type-title"? This way we get that this is intend for account types. For the rest

> 
> The effect of the tag will be to customise the report section headings in the bs, cf and is reports. But not the main title, right ? I expect that will feel like a missing feature unless we support that too.

Sure, for this I don't have a solution yet. Can we define something like "bs-title" if the journal?

legrostdg:
> And since you wouldn't want to type those every time, and since we have so far chosen not to have a dedicated config file, we would need to suggest to users how to use a shell alias and/or arguments file to make this easier (and this would add motivation for some kind of easier config file support).
> 
> Why do you say these would need to vary per journal ?

For example I may have a journal that is shared with others concerning an international charity (English) accounting, another for my farm (French) accounting, ...

legrostdg:
# Overriding account type names in balancesheet, balancesheetequity and incomestatement

If you ever want to override the default English account type names (Assets, Liabilities, Equity, Revenues, Expenses) used in balancesheet, balancesheetequity and incomestatement, you can use a special `type-name:` tag after defining an account type. Eg:

    account assets       ; type:Asset
        ; type-name:Actif
    account liabilities       ; type:Liability
        ; type-name:Passif

simonmichael:
Re the CI warnings: you can see them in the CI logs, it looks like a redundant import needs to be removed. (Next time you push; no hurry.)

Re the make ghcid warnings: I see discussion of these at https://github.com/fpco/http-reverse-proxy/issues/3 . I'm not able to reproduce them here for some reason. What if you try `stack build; make ghcid` ? (building all packages, not just hledger-lib & hledger).

legrostdg:
OK, I did not have the warnings locally (sometimes stack does not display warnings...), I got them from the CI.

```
$ stack build ; make ghcid
Stack has not been tested with GHC versions above 8.6, and using 8.8.3, this may fail
Stack has not been tested with Cabal versions above 2.4, but version 3.0.1.0 was found, this may fail
ghcid -c 'make ghci'
Loading make ghci ...
make[1]: Entering directory '/home/gueux/src/hledger'
stack --stack-yaml=stack-8.10.yaml exec -- ghci  -rtsopts -Wall -fno-warn-unused-do-bind -fno-warn-name-shadowing -fno-warn-missing-signatures -fno-warn-orphans -fno-warn-type-defaults  -ihledger-lib -ihledger-lib/other/ledger-parse -ihledger -ihledger-ui -ihledger-web -ihledger-web/app    -DPATCHLEVEL=197 -DDEVELOPMENT -DVERSION="\"1.17.99\""  hledger/Hledger/Cli/Main.hs
Stack has not been tested with GHC versions above 8.6, and using 8.10.0.20200123, this may fail
Stack has not been tested with Cabal versions above 2.4, but version 3.2.0.0 was found, this may fail
GHCi, version 8.10.0.20200123: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/gueux/.ghci
Loaded GHCi configuration from /home/gueux/src/hledger/.ghci

hledger-lib/Text/Megaparsec/Custom.hs:181:0: error:
     error: missing binary operator before token "("
      181 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
181 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^

hledger-lib/Text/Megaparsec/Custom.hs:196:0: error:
     error: missing binary operator before token "("
      196 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
196 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^

hledger/Hledger/Cli/Commands/Roi.hs:210:0: error:
     error: missing binary operator before token "("
      210 | #if MIN_VERSION_math_functions(0,3,0)
          | 
    |
210 | #if MIN_VERSION_math_functions(0,3,0)
    | ^

hledger-lib/Text/Megaparsec/Custom.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE BangPatterns #-}
  | ^

hledger/Hledger/Cli/Commands/Roi.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE ParallelListComp, CPP #-}
  | ^
Failed, no modules loaded.

hledger/Hledger/Cli/Commands/Roi.hs:210:0: error:
     error: missing binary operator before token "("
      210 | #if MIN_VERSION_math_functions(0,3,0)
          | 
    |
210 | #if MIN_VERSION_math_functions(0,3,0)
    | ^
hledger/Hledger/Cli/Commands/Roi.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE ParallelListComp, CPP #-}
  | ^
hledger-lib/Text/Megaparsec/Custom.hs:181:0: error:
     error: missing binary operator before token "("
      181 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
181 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^
hledger-lib/Text/Megaparsec/Custom.hs:196:0: error:
     error: missing binary operator before token "("
      196 | #if MIN_VERSION_megaparsec(8,0,0)
          | 
    |
196 | #if MIN_VERSION_megaparsec(8,0,0)
    | ^
hledger-lib/Text/Megaparsec/Custom.hs:1:1: error:
    `gcc' failed in phase `C pre-processor'. (Exit code: 1)
  |
1 | {-# LANGUAGE BangPatterns #-}
  | ^

**Comment by legrostdg** [comment]
What do you think about the small doc https://github.com/simonmichael/hledger/pull/1219#issuecomment-616636095 ?

**Comment by simonmichael** [comment]
Ah, that output is helpful. For a while`make ghcid` was using GHC 8.10, while `stack build` generates files for GHC 8.8. This was fixed a few weeks ago, if you rebase on latest master then `stack build --fast && make ghcid` should work.

The latest CI failure is disk space related, which I hope to fix soon. 

Thanks for the doc draft, I'll give more thought and feedback asap.

**Comment by legrostdg** [comment]
https://gist.github.com/legrostdg/e5f45e262bd86fbc08ade553c3eecad5

**Comment by simonmichael** [comment]
> https://gist.github.com/legrostdg/e5f45e262bd86fbc08ade553c3eecad5

I can reproduce this error only if I add `{-# LANGUAGE OverloadedStrings #-}` to `hledger-lib/Hledger/Utils/String.hs`. Nothing in my setup turns on OverloadedStrings for this file. Could it be some change in your working copy, or a .ghci config file ?

**Comment by simonmichael** [comment]
Aside: I notice your CI jobs are still taking ~30 minutes, when I would expect more like 3 minutes. I'm not sure exactly why, but rebasing your branch against latest master might help it with caching ?

**Comment by simonmichael** [comment]
I think these report headings don't really fit as an attribute of accounts. An attribute of account types seems a better fit. 

Eg, there could be a new top-level directive like this:
```
account-type ID NAME
```
where ID is the short letter code, one of ALERX (defining new account types would not be allowed, for now), and NAME is the human readable name & report heading. For a fairly smooth upgrade path, the [account type: tag](https://hledger.org/journal.html#account-types) would accept either ID or NAME, and the following would be the built-in defaults:
```
account-type A Asset
account-type L Liability
account-type E Equity
account-type R Revenue
account-type X Expense
```
This would keep existing type: declarations working, while changing the default report headings just slightly ("Liabilities" -> "Liability").

However, that's still not the best fit - what if at some point we have reports whose sections are not based on the standard 5 account types ? Eg compound reports with arbitrary user-specified sections/grouping. Report headings ultimately are a property of the report.

It might be ok to add the account-type directive anyhow, since since it would solve the report customisation need for now, it wouldn't preclude another custom report heading mechanism later, and as a bonus it allows use of native-language long account type names in account type: declarations. (Though that last is a pretty weak benefit, adding more complexity and more variance across hledger journal files.)

legrostdg:
> > https://gist.github.com/legrostdg/e5f45e262bd86fbc08ade553c3eecad5
> 
> I can reproduce this error only if I add `{-# LANGUAGE OverloadedStrings #-}` to `hledger-lib/Hledger/Utils/String.hs`. Nothing in my setup turns on OverloadedStrings for this file. Could it be some change in your working copy, or a .ghci config file ?

Good catch, this is because of a ~/.ghci config file. Maybe we could add a .ghci config to the repo, to override user config?

simonmichael:
Will that negate the user's other GHCI config, or is it additive ?

legrostdg:
Maybe there is a way to just make "make ghcid" ignore the user ~/.ghci file?

legrostdg:
The best would probably be to have a dsl to be able to customize all of the reporting commands (and be able to add new ones)...

simonmichael:
> The best would probably be to have a dsl to be able to customize all of the reporting commands (and be able to add new ones)...

I agree, I think that is what would solve this well. 

I propose we close this PR and continue design discussions on #1025.

PR #1228, opened 2020-04-20
web: avoid using sessions (partly fixes #1081)

### LINKED ISSUE #1081: Please follow the XDG Base Dir Spec
Although hledger allows specifying any file for configuration, it would be nice to have sane defaults that follow the [spec](https://wiki.archlinux.org/index.php/XDG_Base_Directory) and falls back to the legacy default locations.

There is a nice [haskell library](https://hackage.haskell.org/package/xdg-basedir) that implements this.

legrostdg:
do not use ~/.hledger-web_client_session_key.aes anymore (as seen in https://github.com/simonmichael/hledger/issues/1081#issuecomment-530955978)

### DIFF
--- a/hledger-web/Hledger/Web/Foundation.hs
+++ b/hledger-web/Hledger/Web/Foundation.hs
@@ -99,9 +99,7 @@ type Form x = Html -> MForm (HandlerT App IO) (FormResult x, Widget)
 instance Yesod App where
   approot = ApprootMaster $ appRoot . settings
 
-  makeSessionBackend _ =
-    let sessionexpirysecs = 120
-    in  Just <$> defaultClientSessionBackend sessionexpirysecs ".hledger-web_client_session_key.aes"
+  makeSessionBackend _ = return Nothing
 
   -- defaultLayout :: WidgetFor site () -> HandlerFor site Html
   defaultLayout widget = do


simonmichael:
Nice! Any user-visible consequences of this ?

simonmichael:
#yesod says:
```
If you're using alerts, that would probably break them (i.e. {get|set}Message)
ultimate destinations will probably suffer as well
```
Could you test alerts (eg after a failed or successful add form submission), and check that redirecting is working as usual  ?

legrostdg:
I tried to submit an invalid transaction and got a blue alert with:

    "" (lines 1-1)
    could not balance this transaction (real postings are off by 23124)
    2020-04-03 test
    actif:2.Comptes d'immobilisations:1.Immobilisations corporelles:1.Terrains 23123
    auieauie 1

legrostdg:
not an alert in fact, a modal

legrostdg:
... but a successful add form submission did not generate a modal anymore :(

legrostdg:
So, can we conclude web session are needed after all? I don't get why it works for errors and not for success, though...

legrostdg:
If web session are needed, I'll close this and open a new pull request using XDG_RUNTIME_DIR for this file.

simonmichael:
Without building this myself.. it sounds like it breaks some existing functionality that we want to keep; also it seems quite likely that sooner or later we'll add authentication to this app. So perhaps we need web sessions..

Meanwhile.. that seemingly-useless file has always been an annoyance. I wonder if it's possible, reliable on all platforms, and safe with multiple instances or users, to write it to /tmp .. ?

simonmichael:
(I'm not sure what happens right now if you run multiple hledger-web instances in the same working directory. Do they disrupt each others alerts ?)

legrostdg:
I don't know for other OSes, but, on Linux, it seems suboptimal and not secure to write to /tmp... We have $XDG_RUNTIME_DIR exactly for this purpose, so I think you should use it.

simonmichael:
> We have $XDG_RUNTIME_DIR exactly for this purpose, so I think you should use it.

That sounds fine. You'll need to pick a fallback location (on all platforms) when that variable is not defined. 

It sounds like contention from multiple hledger-web instances isn't a problem for us ("It won't make it so the sessions will get messed up or anything like that, it's just the service (single or multiple instances) will use the same key to decrypt session for all instances, and that's ok... unless you have some unusual scenario going on"), so the current filename is fine.

legrostdg:
Closing in favour of #1344

PR #1232, opened 2020-04-26
csv: Increase maximum number of postings in CSV reader from 9 to 99.

vsorokin:
The ordering of journalfieldnames is changed in order to comply with the requirement stated in the comment:
"Names must precede any other name they contain, for the parser".
If left unchanged, "account1" would precede "account11", "account12", and so on, which would break the parsing.
With the new ordering, "account11" precedes "account1".


simonmichael:
Great! I needed this already for a complex entry. Thank you.

vsorokin:
@simonmichael Thank you Simon!

PR #1240, opened 2020-05-23
lib: multiBalanceReport derives query from ReportOpts

adept:
All call sites of multiBalanceReport do `multiBalanceReport opts (queryFromOpts d opts)`.

I think it is best to move derivation of query inside `multiBalanceReport` to reduce the "gotcha" factor, as passing a query that is "wider" than period requested in opts (like `Any`) effectively renders period selection in opts null and void -- I discovered that a painful way while playing with multiBalanceReport in ghci


adept:
Extra flexibility is needed in couple of places, but it could (still) be had via `multiBalanceReportWith`

simonmichael:
Oh I see. Looks good to me. Unless you think it's worth making q the first argument of multiBalanceReportWith, so it's "doWith THING something" ?

adept:
There are two "with" things there - query and price oracle, and it is not obvious why one of them deserves to be moved forward and not the other. Perhaps leave it as it is?

Issue #1239, opened 2020-05-19
(most recent) Basis as valuation if no other data are present


jblachly:
For periodic stock purchases, the most recent purchase price serves as a relatively current proxy of valuation.  For many instruments, including mutual funds etc. it can be inconvenient to set up automatic ticker updates and at a minimum will require the end-user to write custom software.

Instead, I propose that if no more recent ticker data are available, the most recent basis price be exposed as the valuation (`-V`). 

This was actually a "gotcha" when I first started with hledger (i.e., even when using the `-V` flag I still saw security symbols until entering the `P... lines`) when I first started with hledger because
(a) I believe ledger operates this way and
(b) this is how I would have expected the software to operate ("principle of least surprise")

simonmichael:
@jblachly, thanks for the real-world feedback. It's somewhat persuasive. Here are some objections:

Your least surprise as an experienced user differs from mine when I was learning PTA I think. It took me a long time to understand that Ledger worked this way. That was partly a documentation failing, but nobody reads docs so I think it's still an issue.

Using hledger terminology here for convenience: I feel that having transaction prices also imply market prices adds unnecessary complexity to an already complicated system, making it harder to learn, harder to be certain it's calculating the way you intend, and harder to troubleshoot when things go wrong. Providing a simpler conceptual model, at the modest cost of having to write a P declaration or two, seemed better.

Also I feel that it's an approximation that can potentially be misleading. Transacted price and market price are different things, even if usually similar, and it seems wrong that the tool could show you values based on a "market price" X based on a possibly very non-market-rate transaction.

I do agree that fetching and declaring market prices is a troublesome hurdle; I still don't have a satisfactory workflow for it myself. And I agree that just maybe, Ledger-style behaviour might be more convenient for a user that already knows how it works: without any P directives, they can still get some reasonable numbers from -V.

Perhaps there's something else we could do, that brings the benefits/avoids the pitfalls of both ?

There's one absolute prerequisite for this if we decide we want it: to write down a full specification, including all the corner cases.

simonmichael:
G'day all. I hope you are healthy and solvent, wherever you may be.

From the issue tracker: if you have an opinion on whether hledger should infer market prices from transaction prices, as Ledger does, you may want to comment on https://github.com/simonmichael/hledger/issues/1239 .

adept:
[...]

PR #1241, opened 2020-05-24
lib: also infer market prices from transactions, like Ledger

simonmichael:
WIP for #1239, untested and may not work yet. Draft docs:
```hs
-- When the valuation commodity is specified, this looks for an
-- exchange rate (market price) calculated in any of the following
-- ways, in order of preference:
--
-- 1. a declared market price (DMP) - a P directive giving the
--    exchange rate from source commodity to valuation commodity
--
-- 2. a transaction-implied market price (TMP) - a market price
--    equivalent to the transaction price used in the latest
--    transaction from source commodity to valuation commodity
--    (on or before the valuation date)
--
-- 3. a reverse declared market price (RDMP) - calculated by inverting
--    a DMP
--
-- 4. a reverse transaction-implied market price (RTMP) - calculated
--    by inverting a TMP
--
-- 5. an indirect market price (IMP) - calculated by combining the
--    shortest chain of market prices (any of the above types) leading
--    from source commodity to valuation commodity.
```
Any testing/problem reports welcome.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -21,6 +21,7 @@ module Hledger.Data.Journal (
   addPeriodicTransaction,
   addTransaction,
   journalBalanceTransactions,
+  journalInferMarketPricesFromTransactions,
   journalApplyCommodityStyles,
   commodityStylesFromAmounts,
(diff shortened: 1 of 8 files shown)


simonmichael:
Updated, lightly tested now.

PR #1244, opened 2020-05-25
;web: maintain query when choosing a different 'inacct' account on the sidebar

thielema:
This is an experimental feature
that I proposed already last year in googlegroups.
I am annoyed that date restrictions get lost
when I traverse through the accounts on the left sidebar.
Thus I added the non-inacct parts of part to the account links.
I could also add these to the links in the journal and the register.
It works only if the query was already sent to the server.
I hope I got the quotation right.
I had to export Query.prefixes.
Alternatively I could move replaceInacct to the Query module.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -49,6 +49,7 @@ module Hledger.Query (
   matchesCommodity,
   matchesPriceDirective,
   words'',
+  prefixes,
   -- * tests
   tests_Query
 )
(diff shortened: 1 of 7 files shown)


simonmichael:
I can see how that is annoying.

I probably chose the current behaviour to support hledger-web's goal of being reliably simple and non-confusing for non-techies, free of UI modes etc. So if you click a hyperlinked account name, you get that account's register, end of story. 

If we do make this change, I'm not sure whether it should also be applied to the other links.

I'll try using this for a bit.

simonmichael:
I haven't had a chance to use this, but I expect we'll want it. Thanks!

PR #1248, opened 2020-05-25
cli: ability to pass arguments to getHledgerCliOpts + two example scr…

adept:
Two scripts that combine balances into a single report + small change to commandline parsing code


simonmichael:
Thanks!

PR #1249, opened 2020-06-01
lib: test for issue with sign checking code + fix

adept:
The issue is that `isNegativeMixedAmount` does not return `Just False` for _positive_ mixed amount, so new sign-checking code ignores any such mixed amount completely and falsely thinks that all remaining amounts are of the same sign.


simonmichael:
Good catch, thank you!

Issue #1236, opened 2020-05-06
Allow overlap of future transactions for forecasts


makdisse:
### Context

The journal documentation about [periodic transactions says](https://hledger.org/journal.html#periodic-transactions): `Forecasted transactions will begin only after the last non-forecasted transaction’s date.` I believe it would be useful to remove this limitation.

In developing countries it is very common to purchase for things and pay in instalments. One can easily have multiple future transactions in his journal defined for the next 6-8 months. Currently if you do this to an account that account will be excluded from the forecast.

 ### Request

- Add a flag (eg: `--allow-forecast-overlap`) that will allow the forecast to overlap dates where there are existing future transactions in the journal file
- The command should be smart enough to avoid duplicated transactions by matching the date, accounts used and amount
  - Example: 

**Given the following journal file `test.ledger`**:

```
~ monthly to 2021-05
    Expenses:Travel   1 GBP
    Assets:Bank

~ monthly from 2020-05 to 2020-09
    Expenses:Other   1 GBP
    Assets:Bank

2020-04-01 Spend travel
    Expenses:Travel   1 GBP
    Assets:Bank

2020-05-01 Spend travel
    Expenses:Travel   1 GBP
    Assets:Bank

2020-06-01 Spend other
    Expenses:Other   1 GBP
    Assets:Bank
```

**When I run the command**: `hledger bal -b 2020-03 -e 2021-03 --forecast -M -f test.ledger --allow-forecast-overlap`
**I will see the following output**:

```
Balance changes in 2020-04-01-2021-02-28:

                 ||    Apr     May     Jun     Jul     Aug     Sep     Oct     Nov     Dec     Jan     Feb
=================++========================================================================================
 Assets:Bank     || -1 GBP  -2 GBP  -2 GBP  -2 GBP  -2 GBP  -1 GBP  -1 GBP  -1 GBP  -1 GBP  -1 GBP  -1 GBP
 Expenses:Other  ||      0   1 GBP   1 GBP   1 GBP   1 GBP       0       0       0       0       0       0
 Expenses:Travel ||  1 GBP   1 GBP   1 GBP   1 GBP   1 GBP   1 GBP   1 GBP   1 GBP   1 GBP   1 GBP   1 GBP
-----------------++----------------------------------------------------------------------------------------
                 ||      0       0       0       0       0       0       0       0       0       0       0
```
- The feature should also allow for the following parameters:
  - `--begin-forecast` sets forecast start date, overriding what's in the periodic rule
[...]

PR #1250, opened 2020-06-02
lib: --forecast=PERIODICEXPR. Fixes #835, #1236

### LINKED ISSUE #835: start forecast transactions before today ? clarify forecast start/end rules
We currently have some functional tests failing due to --forecast being current-date-dependent. I think the [manual](http://hledger.org/manual.html#forecasting-with-periodic-transactions) is not quite right about forecast transactions' start date. It seems not possible to generate forecast transactions before today's date. Perhaps it should be.

adept:
Given that we already have a nice single place where "forecast begin" and "forecast end" is defined, plugging in options that allow overriding one or both of the values is pretty straightforward. I added a test nevertheless

### DIFF
--- a/hledger/Hledger/Cli/CliOptions.hs
+++ b/hledger/Hledger/Cli/CliOptions.hs
@@ -181,8 +181,12 @@ reportflags = [
 
   -- generated postings/transactions
  ,flagNone ["auto"]          (setboolopt "auto") "apply automated posting rules to modify transactions"
- ,flagNone ["forecast"]      (setboolopt "forecast") "generate future transactions from periodic transaction rules, for the next 6 months or till report end date. In hledger-ui, also make ordinary future transactions visible."
-
+ ,flagOpt "" ["forecast"]    (\s opts -> Right $ setopt "forecast" s opts) "PERIODEXP" 
+  (unlines
+   [ "Generate periodic transactions (from periodic transaction rules). By default these begin after the latest recorded transaction, and end 6 months from today, or at the report end date."
+   , "Also, in hledger-ui, make future transactions visible."
+   , "Note that = (and not a space) is required before PERIODEXP if you wish to supply it."
+   ])
  ]
 
 -- | Common flags that are accepted but not shown in --help,
--- a/tests/forecast.test
+++ b/tests/forecast.test
@@ -156,3 +156,48 @@ Y 2000
 
 >>>2
 >>>=0
+
+
+# 8. A balance report with forecast-begin enabling transaction before report end
+hledger bal -M -b 2016-10 -e 2017-02 -f - --forecast=20160801-
+<<<
+2016/12/31
+    expenses:housing  $600
+    assets:cash
+
+~ monthly from 2016/1  salary
+    income  $-1000
+    assets:cash
+>>>
+Balance changes in 2016-10-01-2017-01-31:
+
+                  ||    Oct     Nov     Dec     Jan 
+==================++================================
+ assets:cash      ||  $1000   $1000    $400   $1000 
+ expenses:housing ||      0       0    $600       0 
+ income           || $-1000  $-1000  $-1000  $-1000 
+------------------++--------------------------------
+                  ||      0       0       0       0 
+>>>2
+>>>=0
+
+# 9. Parse error in malformed forecast period expression
+hledger bal -M -b 2016-10 -e 2017-02 -f - --forecast=20160801-foobar
+<<<
+2016/12/31
+    expenses:housing  $600
+    assets:cash
+
+~ monthly from 2016/1  salary
+    income  $-1000
+    assets:cash
+>>>
+>>>2
+hledger: could not parse forecast period : 1:10:
+  |
+1 | 20160801-foobar
+  |          ^
+unexpected 'f'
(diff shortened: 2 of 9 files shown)


hledger/Hledger/Cli/CliOptions.hs
```diff
@@ -181,7 +181,9 @@ reportflags = [
 
   -- generated postings/transactions
  ,flagNone ["auto"]          (setboolopt "auto") "apply automated posting rules to modify transactions"
- ,flagNone ["forecast"]      (setboolopt "forecast") "generate future transactions from periodic transaction rules, for the next 6 months or till report end date. In hledger-ui, also make ordinary future transactions visible."
+ ,flagNone ["forecast"]      (setboolopt "forecast") "generate future transactions from periodic transaction rules, for the next 6 months or till report end date (could be overriden by --forecast-begin/--forecast-end). In hledger-ui, also make ordinary future transactions visible."
+ ,flagReq  ["forecast-begin"]     (\s opts -> Right $ setopt "forecast-begin" s opts) "DATE" "include forecast postings/txns on or after this date"
+ ,flagReq  ["forecast-end"]       (\s opts -> Right $ setopt "forecast-end" s opts) "DATE" "include forecast postings/txns before this date"
```
simonmichael, line 186: I like that you put "forecast" first, grouping these flags.

I wanted to improve the help somehow. What do you think of:
```
--forecast        generate periodic transactions (from periodic transaction rules).
                  By default these begin after the latest recorded transaction,
                  and end 6 months from today, or at the report end date.
                  Also, in hledger-ui, make future transactions visible.

--forecast-begin  generate periodic transactions from this date

--forecast-end    generate periodic transactions up to this date
```

These seem slightly redundant and verbose right now
(hledger --forecast --forecast-begin=...), I wonder if we can simplify.

tests/forecast.test
```diff
@@ -156,3 +156,27 @@ Y 2000
...
+
+                  ||    Oct     Nov     Dec     Jan 
+==================++================================
+ assets:cash      ||  $1000   $1000    $400   $1000 
+ expenses:housing ||      0       0    $600       0 
+ income           || $-1000  $-1000  $-1000  $-1000 
+------------------++--------------------------------
+                  ||      0       0       0       0 
+>>>2
+>>>=0
```
simonmichael, line 182: I wonder where in the manual this should be documented. I guess
https://hledger.org/hledger.html#general-options and 
https://hledger.org/journal.html#periodic-transactions.

hledger/Hledger/Cli/CliOptions.hs
simonmichael: Eg (brainstorming):
```
--forecast | --forecast-in PERIODEXPR
--forecast [forecastdate:PERIODEXPR]  
```

simonmichael: Or, just say that using any of the three flags enables periodic transactions ? (hledger --forecast-begin ...  is enough).

simonmichael:
I'll add more thoughts here, hope it's not too much.

**"forecast" or something else ?**
As you said [last year](https://github.com/simonmichael/hledger/issues/835#issuecomment-530413670), "forecast" implies future. Now we're allowing periodic transactions to be generated in the past as well, though that will be less common. We can keep calling it "forecast", I think it's ok. Or we could look for a new name, eg `--auto-transactions` (and make `--auto` short for `--auto-transactions --auto-postings`), or something else, though also with `--forecast` as an alias.

**begin/end or from/to ?**
Whatever we X choose above, should it be --X-begin/--X-end, or --X-from/--X-to ? The first is consistent with --begin and --end, helping to show that --begin/--end/--forecast-begin/--forecast-end are the four dates affecting a report. The second reads a little more naturally and grammatically. What's your preference ?

**showing future transactions in hledger-ui ?**
--forecast has a second function of revealing future-dated transactions in hledger-ui (hidden by default). If we allow any of --forecast, --forecast-begin, --forecast-end to enable periodic transactions, would all of them also do the thing in hledger-ui ? If the "forecast" period is in the past that would be unintuitive.

hledger/Hledger/Cli/CliOptions.hs
simonmichael: Yet another:
```
--forecast-period PERIODEXPR
```

adept:
I'd like to keep `--forecast` because supplying PERIODEXPR does not necessary mean that we are talking about the past -- some accounts could be more in the future than the others, and implicitly taking date of the last transaction in the journal as the start of the forecast might imply that we are "going to the future". In this case, using PERIODEXPR means that we simply restore the status quo.

I've just pushed the rework that makes `--forecast` accept optional PERIODEXPR. When omitted, it will default to current defaults

hledger/Hledger/Cli/CliOptions.hs
adept: For posterity: i took the help text and tried to shoot for `--forecast[=PERIODICEXPR]`

tests/forecast.test
adept: I added help to the section about `--forecast`

simonmichael:
Thanks!

Issue #373, opened 2016-07-18
balance -Y --tree indenting accounts wrongly [2]


simonmichael:
balance should not indent these accounts in multicolumn tree mode:

```
$ hledger -f data/ascii.journal balance -Y --tree
Balance changes in 2014:

           ||  2014
===========++=======
     3     ||     1
         5 ||     1
-----------++-------
           ||
```
##

simonmichael:
```
<sm> I'm comparing hledger/ledger, tree/flat mode, with/without -E and -Y.. hm, this is deceptively tricky
<sm> and a nice functional test candidate
<sm> hledger bal -Y --tree is clearly wrongn

$ hledger bal -Y --tree
Balance changes in 2015:

           ||  2015 
===========++=======
     3     ||     1 
         5 ||     1 
-----------++-------
           ||       

<sm> one of hledger bal --flat and ledger bal --flat is wrong

$ hledger bal --flat
                  -1  1:2
                   1  1:2:3
                  -1  1:2:3:4
                   1  1:2:3:4:5
--------------------
                   0
$ ledger bal --flat
                   1  1:2:3
                   1  1:2:3:4:5
--------------------
                   0


<sm> hledger bal is arguably wrong, should be eliding all zero accounts like ledger bal

$ hledger bal
                   0  1:2
                   1    3
                   0      4
                   1        5
--------------------
                   0
$ ledger bal
                   1  1:2:3
                   1    4:5
--------------------
                   0

<sm> suggestion: let's work up a complete functional test and decide the output there first to minimise wasted work
```

Xitian9:
I think this issue can probably be closed now; we've made some calls on what the correct output is for the above tests.

simonmichael:
Seems fixed 👍🏻

PR #1256, opened 2020-06-13
Use multiBalanceReport for balanceReport

Xitian9:
This PR aims to accomplish two major goals:

-  Get boring parent ellision working for `multiBalanceReport`
-  Remove the special `BalanceReport` code, and just use `multiBalanceReport`

I believe it does both, with the following additional benefits:

- A refactor of `multiBalanceReportWith`, to make the structure easier to follow, and with a clearer division of responsibilities
- All decisions for how an account name is to be displayed are now made in `multiBalanceReport`, rather than scattered around the code base
- Some miscellaneous improvements in account name rendering, including `--drop` now working with `MultiBalanceReport`s, and addressing some of #373

### DIFF
--- a/hledger-lib/Hledger/Reports/BalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/BalanceReport.hs
@@ -11,18 +11,12 @@ module Hledger.Reports.BalanceReport (
   BalanceReportItem,
   balanceReport,
   flatShowsExclusiveBalance,
-  sortAccountItemsLike,
-  unifyMixedAmount,
-  perdivide,
 
   -- * Tests
   tests_BalanceReport
 )
 where
 
-import Data.List
-import Data.Ord
-import Data.Maybe
 import Data.Time.Calendar
 
 import Hledger.Data
@@ -30,6 +24,7 @@ import Hledger.Read (mamountp')
 import Hledger.Query
 import Hledger.Utils
 import Hledger.Reports.ReportOptions
+import Hledger.Reports.MultiBalanceReport (balanceReportFromMultiBalanceReport)
 
 
 -- | A simple balance report. It has:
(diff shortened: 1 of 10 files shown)


Xitian9:
Bah, it looks like there's a new test which has been added. If you have an idea where that valuation error is coming from, let me know.

simonmichael:
@Xitian9, thank you for tackling this! 

Sorry for the test failure, you can ignore it.

Does this address the old obstacle for dropping balanceReport: its ability to compress boring parent accounts into a single report line ?

simonmichael:
Doh! "two major goals ... Get boring parent ellision working for multiBalanceReport". I will review asap.

adept:
I haven't tried to build and run it, but this looks awesome.

Xitian9:
Come to think of it, I'm not sure why account name dropping never worked in tree mode. I don't see a reason not to implement it now.

What is the expected behaviour when you ask to drop more account levels than there are? Currently it gives you an empty string. Would it be better to elide it?

`accountNameDrop 1 "expenses" == ""` -> `accountNameDrop 1 "expenses" == "..."`

Xitian9:
It also looks like we now have a policy for debug levels. I can adhere to that.

simonmichael:
Nice, @Xitian9. I hope to look at this PR after laying 1.18 issues to rest.

Xitian9:
A couple of issues for discussion.

When writing this code, I had to squish a number of annoying bugs which were of the form `accountlistmode_` is set to `ALDefault`, and so both `flat_` and `tree_` return `False`. I ended up explicitly removing `ALDefault` as a possibility using `setDefaultAccountListMode`, but this feels like a bit of a hack.

In an ideal world, I feel that there should be two versions of `AccountListMode` – one with and one without `ALDefault`, so we never run into this issue, but this is probably too much boilerplate. Next best would be to decide where the responsibility for getting rid of `ALDefault` lies. Here it's done at the beginning of both `multiBalanceReportWith` and `balanceReport`, setting the defaults to `ALFlat` and `ALTree` respectively. Is this acceptable?

Speaking of which, the default for `multiBalanceReportWith` is `ALFlat`. Now that boring parent ellision works, is this still the desired behaviour?

Also, should we get rid of `balanceReportFromMultiBalanceReport` entirely, and just define `balanceReport` directly?

simonmichael:
@Xitian9, sorry for the delay on this. When you have a chance, could you rebase against latest master.

simonmichael:
> What is the expected behaviour when you ask to drop more account levels than there are? Currently it gives you an empty string. Would it be better to elide it?

That sounds good to me.

The need for a three-way flat/tree/default option, and the different defaults for different reports, is a bit of a pain. Was it causing a user problem, or just confusing for devs ? Yes I think we should review and maybe tweak the report flat/tree defaults (I think bs is different from is, eg.) But see below..

> should we get rid of balanceReportFromMultiBalanceReport entirely, and just define balanceReport directly?

Quite likely, but let's work through the changes pushed so far and hold off pushing new ones till I catch up a little. More comments on the way.

Xitian9:
Rebased.

> The need for a three-way flat/tree/default option, and the different defaults for different reports, is a bit of a pain. Was it causing a user problem, or just confusing for devs?

I'm not aware of user-facing problems, but it was definitely confusing for this dev (me). The crux of the confusion was that every instance of `if flat_ then a else b` would give `b` when `ALDefault` was set, even though `ALFlat` was the default for `MultiBalanceReport`. An explicit pattern match would avoid that.

simonmichael:
@Xitian9, looking good. 

Can you say more about the data type & algorithmic changes ? I think HashMap is new here, isn't it ?

simonmichael:
It seems to be giving a nice speedup for large balance reports:
```
Best times:
+-----------------------------------------++---------+--------------+
|                                         || hledger | hledger.1256 |
+=========================================++=========+==============+
| -f examples/100x100x10.journal bal      ||    0.10 |         0.09 |
| -f examples/1000x1000x10.journal bal    ||    0.36 |         0.37 |
| -f examples/2000x1000x10.journal bal    ||    0.52 |         0.55 |
| -f examples/100x100x10.journal bal -M   ||    0.10 |         0.11 |
| -f examples/1000x1000x10.journal bal -M ||    1.25 |         1.04 |
| -f examples/2000x1000x10.journal bal -M ||    2.22 |         1.49 |
| -f examples/100x100x10.journal bal -D   ||    0.27 |         0.26 |
| -f examples/1000x1000x10.journal bal -D ||   28.57 |        19.18 |
| -f examples/2000x1000x10.journal bal -D ||   56.74 |        34.49 |
+-----------------------------------------++---------+--------------+
```

simonmichael:
Testing the boring account eliding - 
Here's the traditional eliding of boring parent accounts (bank, liabilities):
```
$ hledger -f examples/sample.journal bal  assets liabilities
                 $-1  assets
                  $1    bank:saving
                 $-2    cash
                  $1  liabilities:debts
--------------------
                   0
```
And the lack of eliding in a periodic report (using multiBalanceReport):
```
$ hledger -f examples/sample.journal bal  assets liabilities -Y --tree
Balance changes in 2008:

             || 2008 
=============++======
 assets      ||  $-1 
   bank      ||   $1 
     saving  ||   $1 
   cash      ||  $-2 
 liabilities ||   $1 
   debts     ||   $1 
-------------++------
             ||    0 
```
Which we can see is fixed by your code - periodic balance reports, too,
now elide boring parents:
```
$ hledger.1256 -f examples/sample.journal bal  assets liabilities -Y --tree
Balance changes in 2008:

                   || 2008 
===================++======
 assets            ||  $-1 
   bank:saving     ||   $1 
   cash            ||  $-2 
 liabilities:debts ||   $1 
-------------------++------
                   ||    0 
```
And the --no-elide flag prevents it, as it should:
```
$ hledger.1256 -f examples/sample.journal bal  assets liabilities -Y --tree --no-elide
Balance changes in 2008:

              || 2008 
==============++======
 assets       ||  $-1 
   bank       ||   $1 
     checking ||    0 
     saving   ||   $1 
   cash       ||  $-2 
 liabilities  ||   $1 
   debts      ||   $1 
--------------++------
              ||    0 
```
The bs/cf/is reports could elide boring parents now too, but currently they don't. Should they ?
```
$ hledger.1256 -f examples/sample.journal bs --tree
Balance Sheet 2008-12-31

             || 2008-12-31 
=============++============
 Assets      ||            
-------------++------------
 assets      ||        $-1 
   bank      ||         $1 
     saving  ||         $1 
   cash      ||        $-2 
-------------++------------
             ||        $-1 
=============++============
 Liabilities ||            
-------------++------------
 liabilities ||        $-1 
   debts     ||        $-1 
-------------++------------
             ||        $-1 
=============++============
 Net:        ||          0 
```

simonmichael:
@Xitian9, I'm ready to merge this as it stands. How about you ?

Xitian9:
Algorithmic changes:
- Using `HashMap AccountName (Map DateSpan Account)` instead of `[[MixedAmount]]` is new. I admit I didn't profile this change (though given the `nub`s and `lookup`s, I thought it was appropriate), so I'm glad it produces a speedup.
- Producing the starting balances no longer calls the whole `balanceReport`, just the first few functions to get what it needs.
- `displayedAccounts` is completely rewritten. Perhaps one subtle thing to note is that in tree mode it no longer excludes nodes with zero inclusive balance unless they also have zero exclusive balance.
- The compound balance reports don't yet have boring parent ellision working, but they're an obvious next step. I didn't want to dive into that until these changes were merged. The issue lies here, where some black magic involving setting `empty_` is done to ensure there are the same columns in each report. This precludes boring parent ellision. It should be straightforward to do another workaround, especially now that multiBalanceReportWith has been broken down into smaller functions.

https://github.com/simonmichael/hledger/blob/b804167c9fdf0868d7e2e692a60719b49a91367c/hledger/Hledger/Cli/CompoundBalanceCommand.hs#L262-L263

I'm also happy to merge as it stands.

Xitian9:
Interesting that in your fourth example its not omitting assets:bank:checking. I'll have a look at that.

Xitian9:
Okay, I think I've caught that case. I'm ready for this to be merged if you are.

simonmichael:
> Interesting that in your fourth example it's not omitting assets:bank:checking

Good catch, I missed that.

Thanks for the info. I'll include this in a merge commit message.

I'll mark the passing of the old multiBalanceReport, into which I poured many an hour :). It is in a way the heart (brain ?) of hledger - the key feature of ledgerlikes (balance report) and a key improvement introduced by hledger (tabular multiperiod balance reports). You have split that 300-line though well documented function into modular parts, which could be a little harder to understand in detail but are easier to understand in the large and more amenable to further refactoring. Then you fixed some old limitations (boring parent eliding in multi period balance reports, --drop with tree mode reports), allowing us to drop the old balanceReport and focus on just the new multiBalanceReport. And for representing the tabular data you replaced the semantically correct but inefficient list of lists with a map of maps, speeding up many-columned balance reports significantly (~40%). Last and not least you made it really easy to review. Thanks @Xitian, great work.

PR #1258, opened 2020-06-13
cli: changes to "hledger print -O csv" output

adept:
Couple of changes to csv output of "hledger print":

1. digit group separators suppressed in amounts
2. prices suppressed in amounts
3. added posting idx

Result of (1) + (2) is that amount are now properly numeric (and thus CSV output could be used in number-munching applications like python or R)

Effect of (3) is nice to have if your transactions are always built is predictable way, as it allows you to choose/filter for first/second/last posting, etc

### DIFF
--- a/hledger/Hledger/Cli/Commands/Print.hs
+++ b/hledger/Hledger/Cli/Commands/Print.hs
@@ -148,7 +148,9 @@ transactionToCSV t =
 postingToCSV :: Posting -> CSV
 postingToCSV p =
   map (\(a@(Amount {aquantity=q,acommodity=c})) ->
-    let a_ = a{acommodity=""} in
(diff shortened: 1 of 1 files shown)


simonmichael:
1 and 2 sound good to me. I'm unsure about the use case for 3, do we really have one ? In the past I removed other "nice to have" fields from the CSV in favour of simple CSV that didn't require edits to make it human friendly when pasted into a spreadsheet.

adept:
> 1 and 2 sound good to me. I'm unsure about the use case for 3, do we really have one ? In the past I removed other "nice to have" fields in favour of simple CSV that doesn't require edits to make it human friendly when pasted into a spreadsheet.

Ok, lets drop the (3) :) Pushed

simonmichael:
Thanks!

PR #1261, opened 2020-06-17
Fix typos

tbm:


tbm:
BTW, it wasn't completely obvious to me what the source files are and whether I should regenerate the generated files, but I guess you can take it from here.

simonmichael:
Thank you @tbm! You changed just the right files.

PR #1262, opened 2020-06-18
lib: condition tables in csv rules + tests

### LINKED ISSUE #1252: Proposal: CSV rules in a file
I am thinking about adding a new syntax for CSV rules to supplement (not replace) the existing one.

Rules will be kept in a delimited file where the header line defines which fields are assigned here:
```
regexp,account2,comment
FOO,expenses:foo,this is a foo
BAR,expenses:bar,
BAZ,,what a BAZ
```

This file will be semantically equivalent to this rules file:
```
if
FOO
  account2   expenses:foo
  comment   this is a foo

if
BAR
   account2  expenses:bar

if
BAZ
  comment   what a BAZ
```

Header `regexp` is mandatory, all other headers are optional and could be any of the assignable CSV fields.

Files in this format will have to be included in another (top-level) rules file, as they have no facility to define date format, `fields`, etc:
```
include delimiter "," rules.csv
```

They will not allow any further includes inside themselves

### Rationale

I'm actually using a simplified version of this format, where columns are set to be "regexp,accounts,comment" and file is being translated into hledger rules syntax by [awk script](https://github.com/adept/full-fledged-hledger/tree/master/06-maintaining-lots-of-csv-rules/import/psv-to-rules.awk).

I've had several reasons to create this system:
1)Speed. When I am making rules for unclassified entries, I have one window with something like "hledger print acct:expenses:unknown" and another window with rules file, and I want to write down rules for the transactions as quickly as possible. Default hledger syntax is not exactly made for speed: "if", Enter, <regexp>, Enter, space, space,"account2",space,space,type out account name - you get the idea. Counting newlines and white space, the amount of boilerplate is almost equal to the amount of actual information. 

Contrast this with "<regexp>|<account name>|<optional comment>" - format overhead is really low (just two separators). The difference becomes more pronounce the more rules you write out in a single sitting.

Arguably, this could be solved by copy-pasting or editor macro. But then we get to the second issue

2)The simplicity of keeping rules organized
In the current syntax, you could combine several rules that assign the same values to transaction fields via:
```
if
regex1
regex2
...
regex2
   field   assignment
   field    assignment
```

This is nice, neat, and compact. However, if you want to keep your rules file neatly organized as you are adding new rules, you will have to either: a)search in the rules file for the right "if" block to modify whenever you add new regex or b)add each new rule in its own "if" block and sort/combine later.

Both of these things are very time-consuming and not something that you really want to do if you plan to add 20+ new rules.

On the other hand, in the tabular format, you could just keep adding new rules (one line == one rule), and when you are done, you sort your file by the second column - and there you are, everything is nice, neat and organized. 

If you edit your delimited file in excel or OpenOffice, you also get poor man's autocomplete as you add new rules, since your accounts are in a single column, so as you type them, spreadsheet editor will suggest values to you.

What do you think?

adept:
This closes #1252  and fixes seemingly long-standing issue in csv rules parser where excessive backtracking prevented proper error reporting. 

I've also found out that `blocksAssigning` occupy 50-60% of the run time, and memorized it. This yields ~50% speedup on my files.

I've added tests to illustrate error messages + a bunch of test cases for #1252  + docs

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -569,7 +569,7 @@ conditionaltablep = do
   lift $ dbgparse 8 "trying conditionaltablep"
   start <- getOffset
   string "if" 
-  sep <- lift $ satisfy (not.isAlphaNum)
+  sep <- lift $ satisfy (\c -> not (isAlphaNum c || isSpace c))
   fields <- journalfieldnamep `sepBy1` (char sep)
   newline
   body <- flip manyTill (lift eolof) $ do
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -42,16 +42,16 @@ where
 --- ** imports
 import Prelude ()
 import "base-compat-batteries" Prelude.Compat hiding (fail)
-import qualified "base-compat-batteries" Control.Monad.Fail.Compat as Fail (fail)
 import Control.Exception          (IOException, handle, throw)
 import Control.Monad              (liftM, unless, when)
 import Control.Monad.Except       (ExceptT, throwError)
 import Control.Monad.IO.Class     (MonadIO, liftIO)
 import Control.Monad.State.Strict (StateT, get, modify', evalStateT)
 import Control.Monad.Trans.Class  (lift)
-import Data.Char                  (toLower, isDigit, isSpace, ord)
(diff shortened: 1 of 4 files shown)


simonmichael:
This PR
- adds "if tables" to csv rules, a more compact format for conditional rules
- fixes a bug in the csv rules parser that was causing vague parse error messages
- speeds up csv conversion performance when there are a lot of conditional rules

Nice work, thanks!

PR #1263, opened 2020-06-18
lib: Add fortnightly as a synonym for biweekly.

Xitian9:


simonmichael:
I'm not sure about this one:

- it adds another special case, anglocentric, never-to-be-removed term to the period expression language, which arguably should be moving in the other direction
- biweekly exists, is easier to type, and generalises to other periods. Does anyone really want to type fortnightly instead ?

Xitian9:
I'm happy to either take or leave this one.
- I'm Canadian, where using the term fortnightly would be either archaic or a joke.
- I live in Australia, where it is the universally preferred term, and is roughly as common as the word monthly.
Given this, I was just noticing its absence.

That said, if the desire is simplification, biweekly is probably the one to go with. Let me know if I should close this.

simonmichael:
In that case, I suppose why not. Thanks!

PR #1269, opened 2020-06-24
Get boring parent ellision working in CompoundBalanceReport.

Xitian9:
This PR rewrites compound balance report generation to use the newly refactored multiBalanceReport framework. In the process, it gets boring parent ellision working by restricting the setting of `empty_` in `ReportOpts`.

It also re-uses the postings filtered by the user query amongst the subreports. This might produce a speedup, but I haven't checked.

I believe this accomplishes its stated goals, but at this point its worth asking if we want to change the behaviour with respect to empty columns. There are three types of columns in the requested date range you might want included in your report.

1. Columns with entries
2. Empty columns in the middle of full columns
3. Empty columns which lead or trail all full columns

All reports will obviously include (1), and in fact all reports currently also include (2). Compound Balance Reports currently include (3) in all cases, and `multiBalanceReport` includes (3) only if `empty_` is set.

There is an opportunity to rewrite this so `compoundBalanceReport` has the same behaviour as `multiBalanceReport`, but there's also the question of whether we ever want to exclude columns of type (2) in either `multiBalanceReport` or `compoundBalanceReport`. This might be desired with sparse data. What do you think?


simonmichael:
Great. This may be the right move, but since simplicity is worthwhile for these reports intended to support non-technical users, and it changes familiar output, I've [polled the mail list](https://groups.google.com/g/hledger/c/he93V46nk6M/m/BG1zREVxCAAJ).

Let's discuss the handling of empty columns elsewhere.  (If it would save you time, feel free to drop by #hledger too.) There are a lot of corner cases.. hopefully our tests cover them reasonably well. I think excluding 2 isn't a priority unless it becomes a real need. (A more common request is to show Average [& Total] column only.)  I didn't know 3 was handled differently, it sounds like we should be more consistent.

simonmichael:
@Xitian9, if you get a chance, please rebase this and I'll merge.

Xitian9:
Done.

PR #1271, opened 2020-06-25
lib: Fix interaction of --drop, --tree, and boring parent ellision.

Xitian9:
When using --drop with --tree and boring parent ellision, previously all parents were mistakenly considered boring. Fix this, and add some tests so this doesn't happen again.

### DIFF
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -344,7 +344,7 @@ displayedAccounts :: ReportOpts -> Query
                   -> HashMap AccountName [Account]
                   -> HashMap AccountName DisplayName
 displayedAccounts ropts q valuedaccts
-    | depth == 0 = HM.singleton "..." $ DisplayName "..." "..." 0
+    | depth == 0 = HM.singleton "..." $ DisplayName "..." "..." 1
     | otherwise  = HM.mapWithKey (\a _ -> displayedName a) displayedAccts
   where
     -- Accounts which are to be displayed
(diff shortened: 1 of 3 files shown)


Xitian9:
Actually, hold off on merging this. I think it can be made simpler.

Xitian9:
Okay, this is better. I'm not sure what that build failure is, but I don't think it's me.

simonmichael:
Looks great, thank you.

I've disabled artifact uploading in the linux (& PR) CI for now, that action seems unreliable. (And we don't need artifacts for every PR & GHC version.)

simonmichael:
Oh here's a little bit of fun, found by "chance":
```
~/src/hledger$ hledger bal --tree --drop 999999999999999999
                   1  ...
                   1  ...
                   2  ...
                   1  ...
                   1  ...
--------------------
                   5
~/src/hledger$ hledger bal --tree --drop 9999999999999999999   # add one more 9, it wraps around
                   1  a:k
                   3  b
                   1  i:p
                   2  j
                   1  q
                   1  c
--------------------
                   5
~/src/hledger$ hledger bal --tree --drop 99999999999999999999  # add one more 9, it swiftly consumes all memory
  C-c C-c
```

Xitian9:
Ow, tricksy. I was actually considering putting in a check to make sure drop_ would never be a negative number (which would have the effect of adding extra indentation). I'm guessing this is because drop_ and take both use `Int` instead of `Integer`. Is this something that needs solving (is it a security risk?), or just a quirk?

Xitian9:
So I see two issues:
- accountNameDrop will wrap-around, and will not drop anything if given a high enough number. This is perhaps unexpected, but not a security issue (it just drops nothing from a list)
- if drop wraps around, the account depth in `DisplayName` is set very high. When rendering, this then puts an arbitrary `Int` worth of spaces on the terminal, resulting in memory problems. This is a security risk.

Changing drop_ to an Integer would solve this in a principled way (at the cost of introducing some genericTake or genericDrop). Or we could just do an ad-hoc check in one or both of these functions. What do you think?

simonmichael:
I think this might be a general problem affecting more than just this option (all Ints ?), and since it could accidentally (typo, mispaste..) or maliciously lock up your machine (macs, for one) yes we had better do something (or at least, open an issue to track it). I suppose, review options to find similar cases, then maybe change them to Integer, but wouldn't that have the same problem ? Perhaps we need a helper that ensures option arguments are within bounds of Int or whatever it may be.

simonmichael:
PS I don't know if you saw, I edited my comment: in the wrap around case it didn't just do nothing, but removed all indentation.

Xitian9:
At least in this case Integer wouldn't have the same problem. The issue is that `level = accountNameLevel a - drop_ ropts` becomes a large positive number when `drop_ ropts` wraps-around to become negative. If `drop_` is an integer, this can only result in a large negative number, which is treated as zero since it's only used in `drop`, `take`, and printing a positive number of things.

That said, the same problem would arise if drop were allowed to be set to a negative number, but this is a general problem with the use of Integer over Natural, and the command line interface currently won't let it be set to a negative number.

The only ReportOpt Ints are `drop_` and `depth_`. I'll look into making them Integers?

simonmichael:
I suppose it would help. Still it feels a bit unsafe to rely on such arguments. I’d be interested to know how exactly the runaway leak is triggered. Perhaps there are other ways to trigger it. But I’d understand if you don’t have time to pursue that.

PR #1272, opened 2020-06-26
balcmds: elide amounts with 3 or more commodities, unless --no-elide

simonmichael:
Multicolumn balance reports showing many commodities tend to become
unreadably wide, especially in tree mode. Now by default we show at
most two commodities, and a count of the rest if there are more than
two. This should help keep reports somewhat readable by default.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -121,8 +121,10 @@ module Hledger.Data.Amount (
   showMixedAmountDebug,
   showMixedAmountWithoutPrice,
   showMixedAmountOneLineWithoutPrice,
+  showMixedAmountElided,
   cshowMixedAmountWithoutPrice,
   cshowMixedAmountOneLineWithoutPrice,
+  cshowMixedAmountElided,
   showMixedAmountWithZeroCommodity,
   showMixedAmountWithPrecision,
   setMixedAmountPrecision,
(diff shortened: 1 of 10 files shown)


simonmichael:
I got tired of multicolumn balance reports going offscreen even with `-N` hiding totals, when in tree mode. What do you think ?

hledger balance --budget and hledger-ui register screen are also affected. For the latter it would be nice to show more commodities if there's width available, but I haven't done that.

Xitian9:
This sounds like a reasonable change to me (I haven't dowloaded the code myself yet).

simonmichael:
[Polled](https://groups.google.com/g/hledger/c/fO7ewJh10vU/m/9rdfug51BwAJ) the mail list also.

adept:
I gave this a try. My impressions:

1. I just saw this diff in the cash-flow report:
```
-                                        || BTCXXXXXXXXXX, EUR 34.00, LTCXXXXXXX 
+                                        || BTCXXXXXXXXXX, EUR 34.00, 1 more.. 
```
In this case I think that common consensus would be "I would rather have three of the three commodities for the extra price of two characters"

2. I really love "and X more" elision of commodities. My reports are now as narrow as they should be! However, I found out that my main commodity (£) was inevitably sorted towards the end of the commodity list and thus elided. This is a bit annoying, as I would rather just see it and nothing else. Is it possible to decide which commodities would be elided base on the number of postings, so that least-used will be omitted?

simonmichael:
We discussed some ideas like:

- display commodities in the order they are declared by commodity directives
- elide to specified/available width, not just to two commodities
- progressive eliding as width allows: no eliding -> all the commodity symbols -> the first few symbols and a count of the rest

I haven't had time to work more on it. Is it worth merging this as is, or not ?

simonmichael:
Guess I need to use it a bit more myself. Installed.

simonmichael:
More ideas:

- show the numerically largest amounts first
- show the most valuable amounts first (based on their -V value)

These would mean commodities could appear in a different order in different report rows/cells.

simonmichael:
Seems an improvement in my usage, merged.

simonmichael:
A year later: this has been only a "mixed success", possibly a failure. New users don't know how to turn this off and see everything.

PR #1274, opened 2020-06-28
ROADMAP: One more achievements

DamienCassou:
A few more than 100 people talking on IRC:


DamienCassou:
We can't rely on Matrix to count IRC users. https://webchat.freenode.net says we are only 92.

PR #1281, opened 2020-07-05
Remove references to old limitations of boring parent elision / Make balance reports flat by default

Xitian9:
Perhaps this is also the place to discuss default mode for each command.
Current:
- `hledger balance`: tree mode
- `hledger balance -Y`: flat mode
- `hledger balancesheet`: tree mode
- `hledger balancesheet --change`: flat mode
- `hledger balancesheet -Y`: flat mode
- `hledger incomestatement`: flat mode
- `hledger incomestatement --historical`: tree mode
- `hledger cashflow` as incomestatement

Extracting from this, I believe the rule is:
- For `hledger balance`, the default is tree mode, unless it's called in multiperiod mode, in which case it's flat
- For the compound balance reports, the default is tree mode unless called with `--change` or in multiperiod mode, in which case it's flat

My impression is that these rules were all created to get around the limitations of boring parent ellision. If that's true, we could just set the default mode for all reports to be tree mode. Less invasively, we could get rid of the effect of multiperiod mode on the defaults. An advantage of having a single default for all reports is that we could get rid of `ALDefault` entirely.

Thoughts?


simonmichael:
Thanks for updating docs.

Sure! I went looking for this old inconsistency the other day and couldn't find it. There it is. 

I think it happened this way: the original balance command is tree mode by default, inherited from Ledger. After adding multibalancereport, I made those flat by default because I found myself always adding --flat to clarify tabular reports. I don't remember why balancesheet and is/cf --historical are tree mode by default.

Options:

- tree mode by default for everything - the simplest change; preserves Ledger compatibility and traditional behaviour of (the simplest form of) the balance command.
- flat mode by default for everything - arguably a more readable and useful default:
  - reduces the new concepts new users must master on day 1 (account hierarchy, hierarchical balances, boring parent eliding)
  - less redundancy & visual noise
  - more compact, both vertically and sometimes horizontally as well (account names are wider but multicurrency amounts need to be displayed less often)

Xitian9:
Given that either behaviour can be explicitly selected, I'm perfectly happy with either outcome here. Some further comments.

- Flat mode seems better for a report showing period change, due to the use of exclusive rather than inclusive balances. It's much easier to see where there's money moving around in the first example here rather than the second.
```
assets:cash        50 XXX
assets:cash:a    2531 XXX
assets:cash:b    1128 XXX
```
```
assets:cash      3709 XXX
  a              2531 XXX
  b              1128 XXX
```

 - Tree mode is more convenient for historical or cumulative balances, as there you care much more about the overall numbers and less about the exact change, so having the subtotals displayed is convenient.

simonmichael:
Good point. I guess that partly explains the current defaults. Let's think a bit about the relative merits of a predictable and easy to learn consistent default, versus varying defaults tuned to make each report more useful out of the box.

simonmichael:
I meant to build a few variants (1. always tree, 2. always flat, 3. tree for end balances & flat for changes) and try using them for a while, but haven't had time. But after watching myself use hledger intensively the last couple of days, I'm leaning towards having everything flat by default, even end balance reports. If you're not opposed, let's go ahead with that.

I expect the disturbance to long-standing defaults might disturb some folks here and there over the coming months/years, but I think it's worth doing. And perhaps this would help a little: #1286.

simonmichael:
PS I can merge this PR and let the display mode changes happen in a new one, or we can keep it all here - let me know.

simonmichael:
Disregard, let's just continue here.

Xitian9:
This now sets flat as the default mode for balance commands, addesses #1282, #1283, and #1287.

As a side effect, there was previously a bit of inconsistency about whether empty subreports of a CompoundBalanceReport displayed nothing for the subtotals row, or a 0. It now always displays nothing (see the changes to incomestatement.test).

simonmichael:
Thanks!

There's too many separatable changes going on in this PR. This one is ok, but let's prefer smaller, more focussed PRs. I think it'll make review and merging go faster, even though many issues and PRs is sometimes a pain.

I have a new concern about sorting by absolute amount. It's no good when there are amounts of both signs, right ?

simonmichael:
And yes I sometimes give conflicting advice. :) Sorry.

simonmichael:
Latest request: If it's easy to do, lets move the sorting changes and future discussion of that to #1283, focussing on the doc updates and flat-mode-by-default here. If not, no worries, I'll just hold off merging this till all we're happy with all of it.

Xitian9:
I've separated out the doc changes and the switch to --flat as default. The other changes all depend on this one, so I'll wait until this is merged to submit the others.

simonmichael:
Wonderful. Thanks @Xitian9.

PR #1285, opened 2020-07-06
csv: Add support for & operators in conditional blocks

msanders:
This adds support for the `&` operator in conditional blocks for CSV rule files. A more general alternative would be nested if statements, but that seems to be a bit more involved from what I can tell.

No prior Haskell experience, so any feedback is much appreciated.

### DIFF
--- a/hledger-lib/hledger_csv.m4.md
+++ b/hledger-lib/hledger_csv.m4.md
@@ -564,8 +564,8 @@ CSVFIELD is a percent sign followed by the field's name or column number, like `
 
 A single matcher can be written on the same line as the "if";
 or multiple matchers can be written on the following lines, non-indented.
-Multiple matchers are OR'd (any one of them can match), unless one includes 
-an `&` symbol, in which case it is combined with the previous matcher.
+Multiple matchers are OR'd (any one of them can match), unless one begins with
+an `&` symbol, in which case it is AND'ed with the previous matcher.
 
 ```rules
 if
--- a/hledger-lib/hledger_csv.m4.md
+++ b/hledger-lib/hledger_csv.m4.md
@@ -564,7 +564,15 @@ CSVFIELD is a percent sign followed by the field's name or column number, like `
 
 A single matcher can be written on the same line as the "if";
 or multiple matchers can be written on the following lines, non-indented.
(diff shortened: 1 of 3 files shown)


simonmichael:
@msanders: thank you, it looks pretty good, from a quick look! Would you mind adding docs for the new features in [this section](https://hledger.org/csv.html#if-block) (hledger-lib/csv.m4.md; no need to generate the other doc files). 

Have you had much opportunity to use this in practice yourself ? I will test and I invite others to do the same.

msanders:
Thanks, will do. I've used it with some of my own files and it seems to work as expected. Also included some additional unit tests, although it may be prudent to add an entry to [csv.test](https://github.com/simonmichael/hledger/blob/81d3ac7d66d0aec1e3cd33754a968e80ee71246a/tests/csv.test) as well.

msanders:
@simonmichael: added docs and integration test

hledger-lib/hledger_csv.m4.md
```diff
@@ -564,7 +564,15 @@ CSVFIELD is a percent sign followed by the field's name or column number, like `
 
 A single matcher can be written on the same line as the "if";
 or multiple matchers can be written on the following lines, non-indented.
-Multiple matchers are OR'd (any one of them can match).
+Multiple matchers are OR'd (any one of them can match), unless one includes
```
simonmichael, line 567: includes -> begins with ?

hledger-lib/hledger_csv.m4.md
```diff
@@ -564,7 +564,15 @@ CSVFIELD is a percent sign followed by the field's name or column number, like `
 
 A single matcher can be written on the same line as the "if";
 or multiple matchers can be written on the following lines, non-indented.
-Multiple matchers are OR'd (any one of them can match).
+Multiple matchers are OR'd (any one of them can match), unless one includes 
+an `&` symbol, in which case it is combined with the previous matcher.
```
simonmichael, line 568: combined -> AND'ed ?

msanders:
Done, thanks!

simonmichael:
@msanders, thank you very much and welcome to the project (and Haskell coding!)

PR #1290, opened 2020-07-08
lib: Refactor multiBalanceReport and budgetReport

Xitian9:
This refactoring has a couple of effects:
- Display name and pruning for multiBalanceReport rows is done earlier on, rather than at the end
- As a result, calculating column totals now doesn't include pruned rows (this wasn't a problem before, as pruned rows always contributed nothing to the total, but could always have become a problem later)
- sortRows is now knows how to calculate the inclusive balances for missing rows, and so it can be applied after pruning. This means it can be used to re-sort the rows of a MultiBalanceReport after it is already generated.
- Budget report now takes advantage of this to outsource its old sorting functions to MultiBalanceReport. This gets rid of a bunch of code duplication.
- A side effect/consequence of all of this is that empty subreports from CompoundBalanceReports will now display an empty totals line, rather than a zero. Previously, this behaviour was inconsistent.


simonmichael:
Looks good to me! You're on a roll!

Issue #1283, opened 2020-07-06
bal -S won't show large revenues first, even with --invert


simonmichael:
Since revenues are negative, `hledger bal revenues -S` shows the smallest ones at the top. I think maybe it should reverse them, showing the largest absolute magnitudes at the top (with or without `--invert`).

Xitian9:
I agree. I think anything fancier will be more trouble than it's worth

I also think the sorting code can be simplified given the new multiBalanceReport code. I'll look into this.

Xitian9:
A related issue:
`hledger bal assets -S --historical` will sort alphabetically (as will `--cumulative`). This is due to the totals column being set to zero for these kinds of reports. Perhaps a more meaningful number for the total is the last column for cumulative and historical reports.

simonmichael:
> hledger bal assets -S --historical will sort alphabetically (as will --cumulative). 

Wow, good find. I didn't reproduce that exactly, but I think I did reproduce a lack of amount-sorting even in a non-tabular balance report. We should probably open a new issue.

Xitian9:
A minor issue. MixedAmount does not currently support `abs`. I presume `map abs` is not used because there is not a sensible definition of `signum` which would respect the law that `signum x * abs x == x`. I imagine it's fine to just define `abs` as above and leave `signum` undefined. Correct me if you feel otherwise.

Xitian9:
Perhaps an option is, when sorting with -S, to partition each level of accounts into positive and negative amounts. Then place the positive accounts first, in decreasing order, and after that the negative accounts, in decreasing order of magnitude. Thoughts?

simonmichael:
Example:
```
1/1
 (a)   100
 (b)  -100
 (c)   200
 (d)  -200
```
Unsorted:
```
$ hledger bal -N
                 100  a
                -100  b
                 200  c
                -200  d
```
Current amount sorting:
```
$ hledger bal -N -S
                 200  c
                 100  a
                -100  b
                -200  d
```
#1291 amount sorting
```
$ hledger bal -N -S
                 200  c
                 100  a
                -200  d
                -100  b
```
"Most positive first, followed by most negative first". It feels a bit weird. Am I just looking at it wrong ?

simonmichael:
With more realistic account names:
```
$ hledger bal -f a.j -N 
                 100  expenses:a
                 200  expenses:c
                -100  revenues:b
                -200  revenues:d
$ hledger bal -f a.j -N -S
[...]

PR #1291, opened 2020-07-08
lib: Sort accounts: positive first, negative second, higher magnitudes first within each.

Xitian9:
What do you think of this as a solution for #1283?

### DIFF
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -9,10 +9,10 @@ account, and subaccounting-excluding and -including balances.
 
 module Hledger.Data.Account
 where
-import Data.List
+import Data.List (find, partition, sortOn)
 import Data.List.Extra (groupSort, groupOn)
-import Data.Maybe
-import Data.Ord
+import Data.Maybe (fromMaybe)
+import Data.Ord (Down(..))
 import qualified Data.Map as M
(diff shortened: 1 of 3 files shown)


Xitian9:
A tangential comment:
There seems to be a bit of overlap between the `normalbalance_` and `invert_` options. Perhaps they could be unified?
- `invert_` has the effect of multiplying all amounts in a PostingReport or BalanceReport by (-1), and no other effect (in particular, no effect on sorting. It can be toggled by a command flag.
- `normalbalance_` will affect the sorting order of BalanceReports, but it cannot be set by a command line toggle, only by being set in code. This is done in CompoundBalanceReport, where it also multiplies amounts by (-1).

simonmichael:
Aren't those semantically different, and liable to get things muddled if combined ? 

normalbalance_ is a hard coded property of the account type (debit or credit), and comes from the laws of accounting. 

invert_ is a last-minute cosmetic sign flipper, intended for (a) getting the signs you want when using bal rather than is/bs and (b) general utility, purpose unknown.

Xitian9:
Okay. I'm happy to drop this pull request. Another option is to make it so that `invert_` gets applied before sorting, which means it will affect sorting too. Let me know how you'd like to proceed.

simonmichael:
Let's leave it.

PR #1293, opened 2020-07-10
lib: Speed up amountLooksZero.

Xitian9:
I'd appreciate a second set of eyes on this. I believe the rewriting of amountLooksZero is equivalent to the previous one, but is much simpler and faster. Since amountLooksZero is called whenever you normalise a mixed amount, which happens quite a lot, this actually seems to have a small but noticable effect on performance. However, the original definition was a bit convoluted, so I may have missed some edge cases.

I've also cleaned up normaliseHelper a bit. Previously it would check to see if there were any zero amounts with empty commodity, and if there were it would return the last zero amount, even if it had an empty commodity. It now returns the last zero amount with nonempty commodity, which I believe was the intended result.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -263,12 +263,11 @@ multiplyAmountAndPrice n a@Amount{aquantity=q,aprice=p} = a{aquantity=q*n, apric
 isNegativeAmount :: Amount -> Bool
 isNegativeAmount Amount{aquantity=q} = q < 0
 
-digits = "123456789" :: String
-
 -- | Does mixed amount appear to be zero when rendered with its
 -- display precision ?
 amountLooksZero :: Amount -> Bool
-amountLooksZero = not . any (`elem` digits) . (showAmountWithoutPriceOrCommodity False)
+amountLooksZero Amount{aquantity=q, astyle=AmountStyle{asprecision=p}} =
+    roundTo (fromIntegral p) q == 0
 
 -- | Is this amount exactly zero, ignoring its display precision ?
 amountIsZero :: Amount -> Bool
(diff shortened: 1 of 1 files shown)


simonmichael:
That looks very much simpler indeed, and nothing broke. In Tests We Trust!

Also, I added `make save*` as a step towards easier benchmarking, saved binaries built before and after this PR, adjusted what to run in bench.sh, and saw these results on my 2013 macbook air:
```
$ quickbench -n2 -N2 -w hledger-1.18-146-gf518da74,hledger-1.18-148-g00d08f8d
Running 6 tests 2 times with 2 executables at 2020-07-10 12:10:26 PDT:

Best times 1:
+--------------------------------------------++----------------------------+----------------------------+
|                                            || hledger-1.18-146-gf518da74 | hledger-1.18-148-g00d08f8d |
+============================================++============================+============================+
| -f examples/1000x1000x10.journal print     ||                       0.66 |                       0.29 |
| -f examples/10000x1000x10.journal print    ||                       6.18 |                       2.41 |
| -f examples/1000x1000x10.journal register  ||                       0.80 |                       0.34 |
| -f examples/10000x1000x10.journal register ||                       7.49 |                       2.90 |
| -f examples/1000x1000x10.journal balance   ||                       0.64 |                       0.27 |
| -f examples/10000x1000x10.journal balance  ||                       4.36 |                       1.21 |
+--------------------------------------------++----------------------------+----------------------------+

Best times 2:
+--------------------------------------------++----------------------------+----------------------------+
|                                            || hledger-1.18-146-gf518da74 | hledger-1.18-148-g00d08f8d |
+============================================++============================+============================+
| -f examples/1000x1000x10.journal print     ||                       0.67 |                       0.26 |
| -f examples/10000x1000x10.journal print    ||                       6.41 |                       2.68 |
| -f examples/1000x1000x10.journal register  ||                       0.78 |                       0.31 |
| -f examples/10000x1000x10.journal register ||                       7.69 |                       3.02 |
| -f examples/1000x1000x10.journal balance   ||                       0.69 |                       0.27 |
| -f examples/10000x1000x10.journal balance  ||                       4.72 |                       1.26 |
+--------------------------------------------++----------------------------+----------------------------+
```
If I did it right, this looks rather terrific ! Thanks !

simonmichael:
Well, after merging/building/benching more, I can't reproduce this. I think we'd better assume I made an unoptimised build or something. :(

PR #1299, opened 2020-07-17
lib: account name clipping takes Maybe Int as depth.

Xitian9:
This just makes the treatment of depth as an option more uniform, and cleans up some of the function calls.


simonmichael:
I don't love a function that may do nothing (sometimes I add "maybe" to their name but that'd be too unwieldy here), but it does look more convenient overall - thanks.

PR #1300, opened 2020-07-21
lib: Speed up space parsing

Xitian9:
This replaces all uses of `skipMany spacenonewline` with `takeWhileP isNonNewlineSpace`, and similarly for `skipSome` and `takeWhile1P`.

Profiling suggested this would give some performance improvements. Evidence for actually getting this improvement is mixed:
- quickbench is showing no significant difference in run time.
- running `hledger balance assets +RTS -P` on my system suggests a roughly 8% decrease in total time and total memeory allocation.

In any case, I think this change probably gives improvements, and is unlikely to cause any harm.


simonmichael:
Thanks!

PR #1304, opened 2020-07-25
doc: clarify virtual prices

tbm:


simonmichael:
What's going on here ? Adding a hyperlink to "Like" ?

tbm:
I'm just removing an empty bullet point in the list.  See 4 in the pic

hledger

Hmm, now that I look at it again, I think it's intentional.  4 being (@) and 5 being (@@).  I'll propose an alternative patch tomorrow to make it clearer.

tbm:
Huh, and I have no idea where that link on Like came from...

tbm:
Is this patch better?

simonmichael:
Thank you very much!

PR #1313, opened 2020-07-31
implement parsing of "q1" and "2020q1"

thielema:
Unfortunately I was not able to run the tests.
Nevertheless, I added some doctest examples.


simonmichael:
Great, thanks!

Issue #203, opened 2014-09-05
boolean query expressions for more flexible queries, OR support


simonmichael:
From the list:

> > Is there a way to get `hledger bal` to show balances of transactions that have either TAG1 or TAG2? The docs seem to say that there isn't, but It's not clear to me.

Hi.. that's a good question. As you point out, our [queries](http://hledger.org/manual#queries) implementation means

hledger print tag:tag1 tag:tag2

only matches transactions with both tags. Most query terms take a regular expression, so I tried

hledger print tag:'(tag1|tag2')

but no, the argument of tag: is a literal tag name, not a regexp. So the surprising answer is no, it's currently not possible. Someone needs to
1. make tag: take a regexp (matchesPosting and matchesTransaction in hledger-lib:Hledger.Query)
2. add support for full boolean query expressions
##

ascjones:
I needed this, but looks like at least 1. already works. This commit seems to be it:

https://github.com/simonmichael/hledger/commit/fdc507bd1365a98a145f94a48ce7c887b1c1dc7c

hoclun-rigsep:
I'd just like to be able to query transactions that post both to account A and account B, which in the usual case would represent a movement of funds from A to B or B to A. This ability would help keep the account hierarchy manageable in many cases. For instance, I get paid a salary, some of which goes to a retirement account and some of which is deposited in a checking account. If I would like to see what my "take home" pay was for a given period I should be able to query like ``hledger balance Salary AND Checking`` or whatever syntax would be most error-proof. Otherwise I would resort to using Salary:take-home, Salary:DeferredComp, Salary:Taxes and so on, even though there's a one-to-one relationship between these categories and another account.

simonmichael:
The usual way is with a pipe, something like: `hledger print salary | hledger -f - balance'

hoclun-rigsep:
Ha! Years on the shell and this did not occur to me, I guess it didn't occur to me that the output of print is valid journal. Thanks.

simonmichael:
Yep this is quite a common pattern with hledger for more powerful filtering. Probably it should be mentioned in the manual (under QUERIES ?)

the-solipsist:
> The usual way is with a pipe, something like: `hledger print salary | hledger -f - balance'

This will not work if the `hledger print salary` outputs any transactions involving balance assertions, since the balance assertions will not, umm, balance.

simonmichael:
In that situation I add -I.

the-solipsist:
[...]

PR #1314, opened 2020-07-31
OR multiple date terms in queries

thielema:
.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -64,6 +64,7 @@ import Data.Monoid ((<>))
 #endif
 import qualified Data.Text as T
 import Data.Time.Calendar
+import qualified Control.Monad.Trans.State as MS
 import Safe (readDef, maximumByMay, maximumMay, minimumMay)
 import Text.Megaparsec
 import Text.Megaparsec.Char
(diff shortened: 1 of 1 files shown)


Xitian9:
I've always been a bit dissatisfied with how date queries are handled in hledger. It's never been possible to evaluate a query along the lines of `Or [And [Acct "a", Date 2018], And [Acct "b", Date 2017]]`. Certainly it's not possible to make a query like that with the cli, but even if you put something like that in the code, the parser will mangle that and give you something along the lines of `And [Or [Date 2017, Date 2018], Or [Acct a, Acct b]]`.

Do you know what the reasoning behind the special handling of date queries is? I always assumed it was some sort of performance thing, or trying to prevent people from making queries that will give difficult-to-interpret balance reports.

simonmichael:
Thanks for this. Recently I wrote, somewhere I can't find right now: 

- dates are AND'ed because I found that to be more useful in my usage, eg when I had a report command alias and I wanted to restrict the date further.
- I think we could consider changing this as part of general boolean queries (https://github.com/simonmichael/hledger/issues/203), but probably not before then as it's too disruptive a change.

simonmichael:
Closing for now till #203 advances.

PR #1319, opened 2020-08-04
tests: Remove commented test behaviour in 373, better distinguish tree and flat mode in 219.

Xitian9:
In my opinion, the commented test behaviour is incorrect, and the current behaviour is correct. The commented behaviour mostly involves eliding accounts that have zero inclusive balance, *even though* those accounts have non-zero exclusive balance. I think eliding those accounts creates confusion.

For example, in the following, why is the total given as 0?
```
$ hledger bal --tree
                   1  1:2:3
                   1    4:5
--------------------
                   0
```
It's because `1:2` has an exclusive balance of -1, which cancels with the exclusive balance of `1:2:3`. Similarly, there's an exclusive balance of -1 in `1:2:3:4`.
The current behaviour shows those amounts
```
$ hledger bal --tree
                   0  1:2
                   1    3
                   0      4
                   1        5
--------------------
                   0
```
This may be more verbose, but is much clearer.

### DIFF
--- a/tests/balance/219.test
+++ b/tests/balance/219.test
@@ -1,6 +1,6 @@
 # issue 219, --tree and --flat flags should override each other cleanly
 # 1. multiple flags ending with --flat, equivalent to --flat
-hledger -f balance-multicol.journal bal -MEH --no-total date:2013/1 --tree --flat
+hledger -f balance-multicol.journal bal -MH --no-elide --no-total date:2013/1 --tree --flat
 >>>
 Ending balances (historical) in 2013-01:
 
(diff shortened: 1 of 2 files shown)


simonmichael:
+1! I agree we should keep our output here, it's clearer than what Ledger shows.

In "tests: Better distinguish between flat and tree mode in 219.test.", I don't quite see how the description relates to the change - could you possibly reword that ?

simonmichael:
I get it now, --no-elide rather than --empty to clarify what's going on. "tests: clarify 219.test, using the more pertinent --no-elide flag" ?

simonmichael:
I reworded it and merged as 05e33965. (Unfortunately I don't know how to do that and still mark the PR accepted.) Thank you!

Issue #1316, opened 2020-08-01
hledger-web sidebar displays accounts differently since 1.18


simonmichael:
As reported by @thielema on the mail list, master's hledger-web's sidebar was showing accounts in flat mode, making it less useful. I've now made it use tree mode. But also it's not eliding boring parents as it used to. I'm not quite sure why we see this difference in hledger-web but not in hledger. I think related code may be:

New multiBalanceReport code:
```haskell
    -- Accounts interesting for their own sake
    isInteresting name amts =
        d <= depth                                                       -- Throw out anything too deep
        && (empty_ ropts || depth == 0 || not (isZeroRow balance amts))  -- Boring because has only zero entries
      where
        d = accountNameLevel name
        balance | ALTree <- accountlistmode_ ropts, d == depth = aibalance
                | otherwise = aebalance
```

Old balanceReport code used by hledger-web 1.18:
```haskell
    -- | In an account tree with zero-balance leaves removed, mark the
    -- elidable parent accounts (those with one subaccount and no balance
    -- of their own).
    markBoringParentAccounts :: Account -> Account
    markBoringParentAccounts = tieAccountParents . mapAccounts mark
      where
        mark a | length (asubs a) == 1 && mixedAmountLooksZero (aebalance a) = a{aboring=True}
               | otherwise = a
```

The new code allows empty_ (-E) to disable boring parent eliding, like --no-elide. I'm not sure why. I guess hledger-web forces empty_ to true, which is the difference from hledger.

simonmichael:
@Xitian9 any thoughts ?

simonmichael:
Here's an easier way to see the change in behaviour, via CLI:
```
$ hledger-1.18 -f tests/balance/balance-multicol.journal bal -MH --no-total date:2013/1 --tree
Ending balances (historical) in 2013-01:

            || 2013-01-31 
============++============
 assets     ||         10 
   checking ||         10 
$ hledger -f tests/balance/balance-multicol.journal bal -MH --no-total date:2013/1 --tree
Ending balances (historical) in 2013-01:

                 || 2013-01-31 
=================++============
 assets:checking ||         10 
```

Xitian9:
Is 1.18 correct? The posting in assets is on 2013-02-02, so would seem to be excluded by the date constraint.

Xitian9:
[...]

PR #1321, opened 2020-08-05
lib: Elide boring parents in tree mode even when using --empty.

Xitian9:
I'm putting this up to give an idea what a proposed fix for #1316 would look like and what it would do. Note that it currently fails tests, but that will be fixed when either #1319 is merged or the relevant commit is cherry-picked into here.

There's a bit of minor refactoring packaged in here. The relevant change is to line 449 of MultiBalanceReport.hs.

This change has a big effect on the budget report. Budget reports are always called with `--empty`, and so never elided boring parents before. With this change, they now do. This makes things more compact, but I'm not sure if this is desired behaviour. Take a look at the changes in `budget.test` and `budget-sorting.test` to see what it does. Another solution is to make budget reports always use `--no-elide`.


simonmichael:
I did some quick testing and found several cases where this gave better results than 1.18. Merged as 32e184f9, thank you!

simonmichael:
PS I haven't paid too much attention to the budget report, but consistency with a normal balance command sounds good so hopefully it's fine.

PR #1323, opened 2020-08-07
lib: Make ill-formed SmartDates unrepresentable.

Xitian9:
This changes the representation of SmartDate to make ill-formed SmartDates unrepresentable. The represtation has an extra two cases, but makes the different behaviour of SmartDates more clear. This makes `spanFromSmartDate` total, and simplifies a few of the parsers.


Xitian9:
(I'm getting those doctest failures in master as well, so I don't think they're mine)

simonmichael:
Sounds good. Sorry about the doctest failures, I'll fix.

Issue #1325, opened 2020-08-08
fromIntegral considered harmful


simonmichael:
https://github.com/haskell-crypto/cryptonite/issues/330 gives an example of how fromIntegral can easily bite you, causing silent data corruption. We seem to use it in about 40 places, and I would say we should audit those and consider banning fromIntegral in favour of a safer wrapper (or the one from foundation ?).

Xitian9:
We've addressed a number of these. Perhaps this should be closed?

simonmichael:
I see fromIntegral still used in ~36 places. I'd say we should keep this open until we have a documented policy on how to avoid unsafe fromIntegral usage, by replacing it with safer code or at least annotating it as safe.

PR #1326, opened 2020-08-09
Replace some uses of fromIntegral with safer calls

Xitian9:
Partly addressing #1325, a lot of uses of `fromIntegral` come from the fact that `Decimal` uses a `Word8` to store the precision, while `Amount` uses an `Int`. If we use `Word8` for `aprecision`, we save 7 uses of `fromIntegral`.

Issues this raises:
- There are two magic precision values `maxprecision` and `maxprecisionwithpoint`. With a domain of `Word8` rather than `Int`, this is a much larger target for accidentally hitting.
- People might try to input large (i.e. >255) precision values and get wraparound effects. This was also true before, before it only happened when you actually called `roundTo`. Now it appears a bit earlier.

 Another 19 instances are eliminated by replacing `fromIntegral . length` with `genericLength`, replacing `fromIntegral` with `toInteger` when appropriate, or just omitting it entirely (it was occasionally converting from `Int` to `Int`, a glorified identity).

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -136,6 +136,7 @@ import qualified Data.Map as M
 import Data.Map (findWithDefault)
 import Data.Maybe
 import qualified Data.Text as T
+import Data.Word (Word8)
 import Safe (lastDef, maximumMay)
 import Text.Printf
 
@@ -263,7 +264,7 @@ isNegativeAmount Amount{aquantity=q} = q < 0
 -- display precision ?
 amountLooksZero :: Amount -> Bool
 amountLooksZero Amount{aquantity=q, astyle=AmountStyle{asprecision=p}} =
-    roundTo (fromIntegral p) q == 0
+    roundTo p q == 0
 
 -- | Is this amount exactly zero, ignoring its display precision ?
 amountIsZero :: Amount -> Bool
@@ -271,15 +272,15 @@ amountIsZero Amount{aquantity=q} = q == 0
 
 -- | Get the string representation of an amount, based on its commodity's
 -- display settings except using the specified precision.
-showAmountWithPrecision :: Int -> Amount -> String
+showAmountWithPrecision :: Word8 -> Amount -> String
 showAmountWithPrecision p = showAmount . setAmountPrecision p
 
 -- | Set an amount's display precision, flipped.
-withPrecision :: Amount -> Int -> Amount
+withPrecision :: Amount -> Word8 -> Amount
 withPrecision = flip setAmountPrecision
 
 -- | Set an amount's display precision.
-setAmountPrecision :: Int -> Amount -> Amount
+setAmountPrecision :: Word8 -> Amount -> Amount
 setAmountPrecision p a@Amount{astyle=s} = a{astyle=s{asprecision=p}}
 
 -- | Increase an amount's display precision, if needed, to enough
@@ -290,7 +291,7 @@ setFullPrecision a = setAmountPrecision p a
   where
     p                = max displayprecision naturalprecision
     displayprecision = asprecision $ astyle a
-    naturalprecision = fromIntegral $ decimalPlaces $ normalizeDecimal $ aquantity a
+    naturalprecision = decimalPlaces . normalizeDecimal $ aquantity a
 
 -- | Set an amount's display precision to just enough decimal places
 -- to show it exactly (possibly less than the number specified by
@@ -298,16 +299,16 @@ setFullPrecision a = setAmountPrecision p a
 setNaturalPrecision :: Amount -> Amount
 setNaturalPrecision a = setAmountPrecision normalprecision a
   where
-    normalprecision  = fromIntegral $ decimalPlaces $ normalizeDecimal $ aquantity a
+    normalprecision  = decimalPlaces . normalizeDecimal $ aquantity a
 
 -- | Set an amount's display precision to just enough decimal places
 -- to show it exactly (possibly less than the number specified by the
 -- amount's display style), but not more than the given maximum number
 -- of decimal digits.
-setNaturalPrecisionUpTo :: Int -> Amount -> Amount
+setNaturalPrecisionUpTo :: Word8 -> Amount -> Amount
 setNaturalPrecisionUpTo n a = setAmountPrecision (min n normalprecision) a
   where
-    normalprecision  = fromIntegral $ decimalPlaces $ normalizeDecimal $ aquantity a
+    normalprecision  = decimalPlaces . normalizeDecimal $ aquantity a
 
 -- | Get a string representation of an amount for debugging,
 -- appropriate to the current debug level. 9 shows maximum detail.
@@ -328,15 +329,15 @@ showAmountWithoutPrice c a = showamt a{aprice=Nothing}
 -- "If the value ends in 5 then it is rounded to the nearest even value (Banker's Rounding)".
 -- Does not change the amount's display precision.
 -- Intended only for internal use, eg when comparing amounts in tests.
-setAmountInternalPrecision :: Int -> Amount -> Amount
+setAmountInternalPrecision :: Word8 -> Amount -> Amount
 setAmountInternalPrecision p a@Amount{ aquantity=q, astyle=s } = a{
    astyle=s{asprecision=p}
-  ,aquantity=roundTo (fromIntegral p) q
+  ,aquantity=roundTo p q
   }
 
 -- | Set an amount's internal precision, flipped.
 -- Intended only for internal use, eg when comparing amounts in tests.
-withInternalPrecision :: Amount -> Int -> Amount
+withInternalPrecision :: Amount -> Word8 -> Amount
 withInternalPrecision = flip setAmountInternalPrecision
 
 -- | Set (or clear) an amount's display decimal point.
@@ -414,7 +415,7 @@ showamountquantity Amount{aquantity=q, astyle=AmountStyle{asprecision=p, asdecim
       qstr -- p == maxprecision && isint q = printf "%d" (round q::Integer)
         | p == maxprecisionwithpoint = show q
         | p == maxprecision          = chopdotzero $ show q
-        | otherwise                  = show $ roundTo (fromIntegral p) q
+        | otherwise                  = show $ roundTo p q
 
 -- | Replace a number string's decimal mark with the specified
 -- character, and add the specified digit group marks. The last digit
@@ -434,9 +435,9 @@ applyDigitGroupStyle (Just (DigitGroups c gs)) s = addseps (repeatLast gs) s
   where
     addseps [] s = s
     addseps (g:gs) s
-      | length s <= g = s
-      | otherwise     = let (part,rest) = splitAt g s
-                        in part ++ [c] ++ addseps gs rest
+      | genericLength s <= toInteger g = s
+      | otherwise     = let (part,rest) = genericSplitAt g s
+                        in part ++ c : addseps gs rest
     repeatLast [] = []
     repeatLast gs = init gs ++ repeat (last gs)
 
@@ -445,12 +446,12 @@ chopdotzero str = reverse $ case reverse str of
                               s         -> s
 
 -- | For rendering: a special precision value which means show all available digits.
-maxprecision :: Int
-maxprecision = 999998
+maxprecision :: Word8
+maxprecision = 254
 
 -- | For rendering: a special precision value which forces display of a decimal point.
-maxprecisionwithpoint :: Int
-maxprecisionwithpoint = 999999
+maxprecisionwithpoint :: Word8
+maxprecisionwithpoint = 255
 
 -- like journalCanonicaliseAmounts
 -- | Canonicalise an amount's display style using the provided commodity style map.
@@ -601,7 +602,7 @@ multiplyMixedAmountAndPrice n = mapMixedAmount (multiplyAmountAndPrice n)
 -- | Calculate the average of some mixed amounts.
 averageMixedAmounts :: [MixedAmount] -> MixedAmount
 averageMixedAmounts [] = 0
-averageMixedAmounts as = fromIntegral (length as) `divideMixedAmount` sum as
+averageMixedAmounts as = genericLength as `divideMixedAmount` sum as
 
 -- | Is this mixed amount negative, if we can tell that unambiguously?
 -- Ie when normalised, are all individual commodity amounts negative ?
@@ -665,13 +666,13 @@ ltraceamount :: String -> MixedAmount -> MixedAmount
 ltraceamount s = traceWith (((s ++ ": ") ++).showMixedAmount)
 
 -- | Set the display precision in the amount's commodities.
-setMixedAmountPrecision :: Int -> MixedAmount -> MixedAmount
+setMixedAmountPrecision :: Word8 -> MixedAmount -> MixedAmount
 setMixedAmountPrecision p (Mixed as) = Mixed $ map (setAmountPrecision p) as
 
 -- | Get the string representation of a mixed amount, showing each of its
 -- component amounts with the specified precision, ignoring their
 -- commoditys' display precision settings.
-showMixedAmountWithPrecision :: Int -> MixedAmount -> String
+showMixedAmountWithPrecision :: Word8 -> MixedAmount -> String
 showMixedAmountWithPrecision p m =
     vConcatRightAligned $ map (showAmountWithPrecision p) $ amounts $ normaliseMixedAmountSquashPricesForDisplay m
 
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -74,13 +74,9 @@ module Hledger.Data.Amount (
   showAmountWithZeroCommodity,
   showAmountDebug,
   showAmountWithoutPrice,
-  maxprecision,
-  maxprecisionwithpoint,
   setAmountPrecision,
   withPrecision,
   setFullPrecision,
-  setNaturalPrecision,
-  setNaturalPrecisionUpTo,
   setAmountInternalPrecision,
   withInternalPrecision,
   setAmountDecimalPoint,
@@ -129,7 +125,7 @@ module Hledger.Data.Amount (
 
 import Control.Monad (foldM)
 import Data.Char (isDigit)
-import Data.Decimal (roundTo, decimalPlaces, normalizeDecimal)
+import Data.Decimal (decimalPlaces, normalizeDecimal, roundTo)
 import Data.Function (on)
 import Data.List
 import qualified Data.Map as M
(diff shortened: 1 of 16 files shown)


simonmichael:
Thanks very much. I'll just make some general comments here, and ask questions since I don't trust my intuition with this stuff. 

First yes Decimal stores no more than Word8 decimal places, so it seems only fair to use the same for aprecision (at least until we change the number representation again, if ever). I do feel quite unclear about how reducing our bits from 64(?) to 8 could make us more robust, even if it does remove some  fromIntegral calls. And as you say the the magic values seem like a problem; perhaps we'd need to get rid of those.

The new code looks prettier. Are we sure [toInteger](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:toInteger) and [genericLength](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html#v:genericLength) are actually safe ? Their haddock doesn't mention any wrapping behaviour, but neither did  [fromIntegral](https://hackage.haskell.org/package/base-4.14.0.0/docs/Prelude.html#v:fromIntegral)'s.

Xitian9:
`toInteger` is safe, as it's injective, that is different values are always mapped to different values. Replacing `fromIntegral` with `toInteger` is completely safe in cases where the target is an `Integer` (the compiler should do this anyway in a simplifying run). This just makes it explicit.

The problem in `fromIntegral = fromInteger . toInteger` is the `fromInteger` part, which is not injective (in general). I'm not sure if there's a formal specification about what it's supposed to do when fed a value that doesn't fit, but as far as I know all implementations do wrapping (for GHC's behaviour, see section 14.1.2 [here](https://downloads.haskell.org/~ghc/7.8.4/docs/html/users_guide/bugs-and-infelicities.html)). Thus wrapping happens whenever you use `fromIntegral` into a smaller value.

If you run `fromInteger 300 :: Word8`, you can see the result is 44, so it's doing wrapping when it calls `narrow8Word#`. As a result, even if using the larger precision `Int` class, as soon as you call something like `roundTo (fromIntegral 256) a`, the actual call is `roundTo 0 a`, which is surely undesired behaviour.

The fact that this hasn't come up as an issue yet is a good sign that there should be no ill-effects from this change.

`genericLength` will still wrap, so this is only safer if your type is large enough to accommodate the size of your lists. Here it's called into types `Quantity` and `Integer`, which are infinite precision (at least with respect to integral values).

simonmichael:
Great info to study, thanks. I don't follow it in full detail yet but you sound convincing. Whatever we do, we'll also do some quick performance testing just in case. The Precision type sounds nice. We'll have sufficient validation/tests to guard against overflowing it I guess.

simonmichael:
PS I think I saw more than one thing changing to Word8 in these commits, could you list which user-visible things have a new size limit, and we'll figure out where that might show up in docs ?

simonmichael:
PPS if you happen to see any way to reproduce a bug with the existing fromIntegral code, do save it; if such exists it would be neat to show it being fixed.

Xitian9:
I've filed #1329. It's pretty contrived, but is at least an edge case we can test for.

Xitian9:
I also don't seem to be able to see that behaviour at all, either in master or 1.18.1. Trying `hledger -f test.journal add`, the docs suggest that you shoud see the following:
```
Date [2015/05/22]:
Description: supermarket
Account 1: expenses:food
Amount  1: $10
Account 2: assets:checking
Amount  2 [$-10.0]:
```
However I'm seeing:
```
Date [2015/05/22]:
Description: supermarket
Account 1: expenses:food
Amount  1: $10
Account 2: assets:checking
Amount  2 [$-10]:
```
Is this where the trailing `.0` is supposed to be seen? What can I do to see the desired behaviour?

Xitian9:
I've changed the parser to fail on more than 255 decimal points, and changed the type of `asprecision` to be a sum type with two constructors: `NaturalPrecision` and `Precision Word8`. I haven't added an extra case equivalent to the old `maxprecision`, and it hasn't broken anything that I've noticed.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -192,7 +193,7 @@ instance NFData AmountPrice
 data AmountStyle = AmountStyle {
       ascommodityside   :: Side,                 -- ^ does the symbol appear on the left or the right ?
       ascommodityspaced :: Bool,                 -- ^ space between symbol and quantity ?
-      asprecision       :: !Int,                 -- ^ number of digits displayed after the decimal point
+      asprecision       :: !Word8,               -- ^ number of digits displayed after the decimal point
```
simonmichael: Since this appears in a bunch of type signatures, what do you think about a NumDigits type synonym ? 
[Testing commenting from IDEA.]

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -214,7 +215,7 @@ instance Show AmountStyle where
 -- point), and the size of each group, starting with the one nearest
 -- the decimal point. The last group size is assumed to repeat. Eg,
 -- comma between thousands is DigitGroups ',' [3].
-data DigitGroupStyle = DigitGroups Char [Int]
+data DigitGroupStyle = DigitGroups Char [Word8]
```
simonmichael, line 231: Let's mention DigitGroupStyle too in the commit summary

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -434,9 +435,9 @@ applyDigitGroupStyle (Just (DigitGroups c gs)) s = addseps (repeatLast gs) s
   where
     addseps [] s = s
     addseps (g:gs) s
-      | length s <= g = s
-      | otherwise     = let (part,rest) = splitAt g s
-                        in part ++ [c] ++ addseps gs rest
+      | genericLength s <= toInteger g = s
```
simonmichael, line 420: What do you think about `genericLength (length s) < genericLength g` here, as a general pattern to avoid any risk of slowness from genericLength ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -830,7 +831,7 @@ numberp suggestedStyle = label "number" $ do
       Left errMsg -> Fail.fail errMsg
       Right (q, p, d, g) -> pure (sign q, p, d, g)
 
-exponentp :: TextParser m Int
+exponentp :: TextParser m Integer
```
simonmichael, line 833: let's mention this in the commit message too. If we get around to it we might even come up with an example showing the improvement, for release notes

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -870,21 +871,25 @@ fromRawNumber raw mExp = case raw of
...
                            (digitGroupNumber $ preDecimalGrp <> postDecimalGrp)
         precision = digitGroupLength postDecimalGrp
 
-    applyExp :: Int -> (Decimal, Int) -> (Decimal, Int)
-    applyExp exponent (quantity, precision) =
-      (quantity * 10^^exponent, max 0 (precision - exponent))
+    applyExp :: Integer -> (Decimal, Word8) -> (Decimal, Word8)
+    applyExp exponent (quantity, precision) = (quantity * 10^^exponent, newPrecision)
+      where
+        newPrecision | precisionDiff >= 255 = maxBound
```
simonmichael: what's happening here ? this case should be impossible, is that right ? If maxBound is used, that is 255, ie maxprecisionwithpoint, right ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -870,21 +871,25 @@ fromRawNumber raw mExp = case raw of
...
+                     | precisionDiff <= 0   = 0
+                     | otherwise            = fromInteger precisionDiff
```
simonmichael: and if precisionDiff is 254, that's maxprecision ? shouldn't we avoid assigning those special values as a result of parsing numbers ? So the max number of decimal places we should allow
 is 253 ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -842,50 +842,40 @@ exponentp = char' 'e' *> signp <*> decimal <?> "exponent"
...
+    return (quantity, precision, mDecPt raw, digitGroupStyle raw)
   where
+    toQuantity :: Integer -> DigitGrp -> DigitGrp -> Either String (Quantity, Word8)
+    toQuantity e preDecimalGrp postDecimalGrp
+      | precision < 0   = Right (Decimal 0 (digitGrpNum * 10^(-precision)), 0)
+      | precision < 256 = Right (Decimal precision8 digitGrpNum, precision8)
+      | otherwise = Left "invalid number: numbers with more than 255 decimal digits are now allowed at this time"
+      where
+        digitGrpNum = digitGroupNumber $ preDecimalGrp <> postDecimalGrp
+        precision   = toInteger (digitGroupLength postDecimalGrp) - e
```
simonmichael, line 861: can e be negative ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -842,50 +842,40 @@ exponentp = char' 'e' *> signp <*> decimal <?> "exponent"
...
+fromRawNumber (WithSeparators _ _ _) (Just _) =
+    Left "invalid number: mixing digit separators with exponents is not allowed"
+fromRawNumber raw mExp = do
+    (quantity, precision) <- toQuantity (fromMaybe 0 mExp) (digitGroup raw) (decimalGroup raw)
+    return (quantity, precision, mDecPt raw, digitGroupStyle raw)
   where
+    toQuantity :: Integer -> DigitGrp -> DigitGrp -> Either String (Quantity, Word8)
+    toQuantity e preDecimalGrp postDecimalGrp
+      | precision < 0   = Right (Decimal 0 (digitGrpNum * 10^(-precision)), 0)
+      | precision < 256 = Right (Decimal precision8 digitGrpNum, precision8)
```
simonmichael, line 857: as noted before, shouldn't this avoid the special 254, 255 values ? Guess I'll see what happens in your next commit

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -842,50 +842,40 @@ exponentp = char' 'e' *> signp <*> decimal <?> "exponent"
...
+      | otherwise = Left "invalid number: numbers with more than 255 decimal digits are now allowed at this time"
```
simonmichael, line 858: s/now/not/

Won't this message be shown in wrong situations, when an exponent is used ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -1011,17 +1001,15 @@ data AmbiguousNumber = AmbiguousNumber DigitGrp Char DigitGrp
 -- | Description of a single digit group in a number literal.
 -- "Thousands" is one well known digit grouping, but there are others.
 data DigitGrp = DigitGrp {
-  digitGroupLength :: !Int,    -- ^ The number of digits in this group.
-  digitGroupNumber :: !Integer -- ^ The natural number formed by this group's digits.
+  digitGroupLength :: !Word,    -- ^ The number of digits in this group.
```
simonmichael, line 1004: why the change to Word ? to allow groups of more than 255 digits in the integral part ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -1011,17 +1001,15 @@ data AmbiguousNumber = AmbiguousNumber DigitGrp Char DigitGrp
...
+  digitGroupNumber :: !Integer  -- ^ The natural number formed by this group's digits.
```
simonmichael, line 1005: since we're changing things.. is there a better type for this ? we don't need/want this to be negative

hledger-lib/Hledger/Data/Types.hs
simonmichael: Changed later I see.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -208,13 +213,22 @@ instance Show AmountStyle where
     (show asdecimalpoint)
     (show asdigitgroups)
 
+data AmountPrecision = Precision !Word8 | NaturalPrecision deriving (Eq,Ord,Read,Show,Typeable,Data,Generic)
+
+instance NFData AmountPrecision
+
```
simonmichael, line 219: Semigroup instance for AmountPrecision: sums the two precisions, capping the result at the maximum of 255. I see

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -208,13 +213,22 @@ instance Show AmountStyle where
     (show asdecimalpoint)
     (show asdigitgroups)
```
simonmichael, line 215: -- | The number of decimal places to display for an amount: 0 to 255, or NaturalPrecision meaning all available decimal digits, up to the 255 limit.

simonmichael: PS do you think "precision" can be ambiguous here ? It's not the number of significant digits in the amount, but the number of decimal places (digits after the decimal mark). It might be a tedious as my use of "precision" may be pervasive, but in principle we could rename this DecimalPlaces = Places ...  or such ?

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -259,55 +260,43 @@ multiplyAmountAndPrice n a@Amount{aquantity=q,aprice=p} = a{aquantity=q*n, apric
 isNegativeAmount :: Amount -> Bool
 isNegativeAmount Amount{aquantity=q} = q < 0
 
+-- | Round an Amount to its specified precision
```
simonmichael, line 263: Round an Amount's Quantity to its specified display precision. If the specified precision is NaturalPrecision, this is a no-op.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: I presume you mean `toInteger (length s) < toInteger g`. I concur.

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -434,24 +417,12 @@ applyDigitGroupStyle (Just (DigitGroups c gs)) s = addseps (repeatLast gs) s
...
-
--- | For rendering: a special precision value which means show all available digits.
-maxprecision :: Int
-maxprecision = 999998
-
--- | For rendering: a special precision value which forces display of a decimal point.
-maxprecisionwithpoint :: Int
-maxprecisionwithpoint = 999999
-
 -- like journalCanonicaliseAmounts
```
simonmichael, line 426: Old stuff to the chopping block. I can't remember why not and we'll see if any hledger add users notice.

hledger-lib/Hledger/Data/Amount.hs
simonmichael: Doh. Yes thank you for reading my mind.

hledger-lib/Hledger/Data/Types.hs
simonmichael: Isn't there a typeclass that would let us still write `1` instead of `Precision 1` in tests, and when creating Amounts in ghci, etc. ?

And if we went that route, would displaying just `1` in the Show output work as well ? If not, it's probably not worth the inconsistency.

hledger-lib/Hledger/Read/Common.hs
Xitian9: Yes, e can be negative.

hledger-lib/Hledger/Data/Types.hs
Xitian9: The typeclass for doing that is `Num`. I'm hesitant to introduce an instance for which we don't want most of the functions used.

hledger-lib/Hledger/Read/Common.hs
Xitian9: Mentioning this you mean the use of `Word8` in the type of `numberp`? (I feel like the comments are offset from what you're referring to)

hledger-lib/Hledger/Read/Common.hs
Xitian9: Much of this got superceded when I introduced the `Precision` type.

hledger-lib/Hledger/Read/Common.hs
Xitian9: Hmm. Since it can have up to 255 digits, even `Word` wouldn't be big enough, so it would have to be `Natural`. I'm not sure it's worth the change, but we could.

hledger-lib/Hledger/Read/Common.hs
Xitian9: Again, superceded with later code.

hledger-lib/Hledger/Read/Common.hs
Xitian9: I don't think it will be displayed in the wrong circumstances. `5e-7` has precision 7, while `2.555e3` has precision 0.

hledger-lib/Hledger/Read/Common.hs
Xitian9: It was just to get around the need to do overflow checks in the Semigroup instance of `DigitGrp`.

hledger-lib/Hledger/Data/Types.hs
Xitian9: I don't object to changing the name to DecimalPlaces.

simonmichael:
Hi @Xitian9. I've been a bottleneck for PRs lately, unfortunately. But maybe I can get a few more merged in time for the 1.19 release on tuesday. This one only needs some minor doc fixups I believe, I can merge and do them if you prefer ?

hledger-lib/Hledger/Read/Common.hs
Xitian9: Yes, large exponents will do bad things. I think we'll actually get issues here if you translate total prices to unit prices, which adds 1 to the precision.

Xitian9:
I've rebased and I think I've added the rest of the requested code comments. I've also removed the Semigroup instance for `AmountPrecision`. It is only used in one place, introduces two CPP-guarded imports of Semigroup, and is not obviously best general definition. (I think taking the max of two precisions is more commonly the right thing to do than summing them).

hledger-lib/Hledger/Read/Common.hs
Xitian9: I've added an overflow check when calculating unit prices to address the above point.

simonmichael:
Great. I plan to do a merge commit with this description, does it sound accurate ?
```
avoid some uses of fromIntegral, parse numbers more robustly

This is PR #1326, addressing #1325 (fromIntegral considered harmful).
User-visible changes include:

- parsing numbers with more than 255 decimal places now gives an error
  instead of silently misparsing.

- digit groups are now limited to at most 255 digits each.

- exponents greater than 9223372036854775807 or less than
  -9223372036854775808 are now parsed correctly, in theory. (In
  practice, very large exponents will cause hledger to eat all your
  memory, so avoid them for now.)

API/internal changes:

- some fromIntegral calls have been replaced with safer code
  avoiding potential bugs due to numeric wrapping.

- asprecision is now a sum type with Word8 instead of an Int with
  magic values.

- DigitGroupStyle uses Word8 instead of Int.
  
- exponents are parsed as Integer rather than Int.
```

Xitian9:
Looks good!

simonmichael:
Merged, after some tussles with git. Thank you!

PR #1327, opened 2020-08-10
cli: Replace use of fromIntegral in anon with unsafeCoerce. Much safer!

Xitian9:
But seriously, I think this is fine, since we're restricting to `unsafeCoerce :: Int -> Word` and the output is a random anonymised hex string. Also fixes a theoretical anonymisation bug, where the anonymisation function on a 64-bit system is periodic of order `2^32`. I don't know, I guess a determined actor could de-anonymise your account names?

If you're uncomfortable with `unsafeCoerce`, but still want to fix the security bug, changing the type of `fromIntegral` to `Int -> Word` should give identical results to this (in slightly more steps).

BTW: Is there a reason this module is in hledger instead of hledger-lib? It seems like it more naturally lives there.

Edit: Actually, perhaps the reason you chose `Word32` is to keep the size of the output to 8 hex characters, and avoid swamping the screen, in which case this may be counterproductive.

Edit 2: In fact, it looks like there's a rewrite rule which changes `fromIntegral :: Int -> Word` into something more or less equivalent to this, so I think there's no value to this PR in either case. I'll leave it open to preserve this one-sided discussion about this particular use of `fromIntegral`, feel free to close when you've seen it.

### DIFF
--- a/hledger/Hledger/Cli/Anon.hs
+++ b/hledger/Hledger/Cli/Anon.hs
@@ -5,6 +5,7 @@ Instances for anonymizing sensitive data in various types.
 Note that there is no clear way to anonymize numbers.
 
 -}
+{-# LANGUAGE OverloadedStrings #-}
 
 module Hledger.Cli.Anon
     ( Anon(..)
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks. I'm not sure I was thinking about 8 hex characters, but I'm happy to close this if it's not yet important.

PR #1328, opened 2020-08-11
`-split-sections` everywhere

fosskers:
This PR enables `-split-sections` for the stack build of `hledger` and its dependencies. This reduces the size of the `hledger` binary on my machine from 27mb to 13mb.

How are the official releases made? Should this setting be activated elsewhere?

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -4,7 +4,11 @@ nix:
   pure: false
   packages: [perl gmp ncurses zlib]
 
-resolver: lts-16.3
+resolver: lts-16.9
+
+# Provides a very nice reduction in final binary size.
+ghc-options:
+  $everything: -split-sections
 
 packages:
 - hledger-lib


simonmichael:
@fosskers: thanks! This sounds like a good thing. If you have time, could you compare how long a full `stack build` (of the four hledger packages) takes with and without this ?

simonmichael:
Also, which platform(s) have you tried this on ?

fosskers:
Sure, I can do a quick comparison. I'm on Arch Linux.

fosskers:
Enabling `-split-sections` has a one-time CI cost, since all dependencies need to be rebuilt. After they're cached, that cost disappears. As for building *just* the `hledger-*` binaries after all "split" deps are built, let's see.

With `-split-sections`:
```
 Executed in  172.84 secs   fish           external 
   usr time  177.51 secs  575.00 micros  177.51 secs 
   sys time   10.19 secs  257.00 micros   10.19 secs
```
Without:
```
Executed in  172.05 secs   fish           external 
   usr time  181.13 secs  347.00 micros  181.13 secs 
   sys time    5.93 secs  166.00 micros    5.93 secs
```
So effectively no difference! I had thought it would be a bit slower with `-split-sections`, honestly. Seems like a free binary size reduction?

simonmichael:
CI (and developers) rebuild all dependencies more often that one would think, for various reasons (CI cache snafus, new ghc version, new machine, profiling). But "no difference" sounds promising! If it's truly a free optimization, I guess we'd just want to enable it everywhere: all stack yaml files, all .github/workflow scripts, the homebrew formula, etc. I'll do some testing as well, and start enabling it where I can.

simonmichael:
Step 1, merge PR. Thank you very much!

fosskers:
The trick here is that doing it in the `stack.yaml` with the target of `$everything` ensures that all dependencies were built with it too. Here's a related thread from the `stack` repo: https://github.com/commercialhaskell/stack/issues/3467#issuecomment-508474014

simonmichael:
Good tip, thanks. 

On mac, it says

    ghc: on the commandline: -split-sections is not useful on this platform since it always uses subsections via symbols.

and the GHC 8.8 build fails. GHC 8.10 seems to carry on though, and might succeed.

simonmichael:
Ah interesting, we already use it in .github/workflows/linux.yml, which builds the `hledger-ubuntu.zip` CI binaries:
```yaml
      run: |
        $stack install --test --bench --force-dirty --ghc-options=-fforce-recomp --ghc-options=-Werror --ghc-options=-split-sections --no-terminal
        # -split-sections shrinks binaries by 30% on average here
```
I'm guessing this is going to be the only place it'll work right now.

fosskers:
As per what I mentioned above, can we be sure that specifying it on the command line like that actually has the intended effect?

simonmichael:
Not entirely.. I assume I did measure the resulting binary size when I wrote that "30%" comment. This CI does often rebuild all deps but it's possible I haven't seen the full effect of split-sections.

fosskers:
The effect of it in the case of this PR was a 50%+ reduction.

Issue #1312, opened 2020-07-31
parseQuery is partial


amigalemming:
parseQuery is partial:
~~~~
hledger-lib> parseQuery nulldate $ T.pack "date::2020"
(*** Exception: "date::2020" gave a date parse error ()
~~~~

In hledger-web this means that entering an ill-formed query causes an Internal Server Error.
Generally, a parser should be total and return a human-readable error message in case of a parser error.
The type should be:
~~~~
parseQueryM :: Day -> Text -> Either Text (Query,[QueryOpt])
~~~~
We might maintain `parseQuery` as a partial wrapper around `parseQueryM` for compatibility,
but may add a warning.

See #1245, also #889.

simonmichael:
WIP

simonmichael:
So I started pulling on this piece of string, and I'm not yet sure where it will end!

There are a lot of calls to error in hledger. One big reason is to bail out when presented with a malformed regular expression, which can easily happen at runtime. For the command line tool, where we want to exit (with a decent message) at the first error, this has been fine, but it's not a good thing to have in library code and for long-running apps. Eg you only have to search for `?` or `(` to kill hledger-web's current request, showing an ugly internal server error (this issue), or to kill hledger-ui entirely. I wasn't totally aware of this since it doesn't happen to me in practice.

Solutions ? We could try to catch all those error exceptions, converting partial code into pure code. I think doing this everywhere would be asking for trouble, though it might be useful as a workaround for a partial library we depend on.

Or we can actually convert all the partial code to pure code. This is already under way, touching eg Hledger.Utils.Regex and Hledger.Query. This of course ripples outward, affecting Hledger.Reports and ultimately perhaps every part of hledger. It will mean more complicated code and numerous small api changes (lots of functions now dealing with Eithers instead of simple values). This sounds a bit unpleasant but I think that's basically unavoidable if we want to fix this. Let me know if you agree or if I'm missing some alternative.

simonmichael:
PS so far I'm minimising api changes, instead just adding alternative total functions, generally with a `_` suffix. But to avoid doubling the code I expect at some point these should be renamed, replacing the old partial functions. If there are cases where the old partial functions are worth keeping, they'd be given a `'` suffix. Heads up for addon authors like @hpdeifel , @peti , @adept , @apauley and active hackers like @Xitian9 .

simonmichael:
[...]

Issue #1245, opened 2020-05-25
Web: The query "(date:2020 )" causes Internal Server Error


thielema:
Try this with hledger-web-1.17.99 from github-HEAD:
~~~~
http://localhost:5000/journal?q=(date%3A2020+)
~~~~
The space before the closing parenthesis causes the problem.
It seems that `parseQuery` is partial. I may try to fix that.

simonmichael:
Thanks for the bug report. I don't think parentheses are supported in queries, but you're right, it shouldn't crash.

simonmichael:
Oh, I suppose the parenthesised text gets parsed as a regular expression (with acct: prefix assumed). Yes:
```
$ hledger print "(date:2020 )" --debug=2
...
query from opts & args: Acct "(date:2020 )"
...
$ (no error)
```

amigalemming:
What would be the proposed way to report the error?
Show a read parser error message below the query field?
Show the unfiltered journal? Show an empty query result?

simonmichael:
@amigalemming good question, I am guessing the usual way hledger-web shows errors - try adding an unbalanced transaction to see it.

Reading this again though I'm not sure why it should be an error, hledger-web should interpret this as a valid regular expression, matching an account name, as hledger does above.

simonmichael:
PS we're not sure if the UI allowed the user to create this situation unintentionally, that would be a separate issue. I'm assuming a url typed manually.

amigalemming:
I see we are discussing two issues: The particular parse error and the parse error causing an Internal Server Error. I like to open two new issues and close this one.

amigalemming:
Hm, I hesitate to phrase an issue for the parser error. On the one hand I find the parser result surprising, on the other hand it seems to be intended behaviour.

amigalemming:
I hope this can be resolved by #1315.

simonmichael:
I'll keep this open until the actual issue is fixed (repro: paste `(date:2020 )` into hledger-web search field and hit enter -> internal server error).

Xitian9:
I think this issue can be closed. We currently get a ‘regex cannot be compiled’ message rather than an internal server error.

PR #1330, opened 2020-08-17
Make a wrapper around compiled Regex, for more total functions.

Xitian9:
This is a work in progress, dealing with option 2 of #1312. I'd be interested in people's opinions.

It compiles and passes all tests, but some things are perhaps not quite right, in particular, the `Data` instance of `Regexp` calls error a lot. This was taken from similar instances in `Data.Generics.Instances`, but who knows if it's the right thing to do.

A lot of instances of `Data`, `NFData` may not serve any purpose. If we can remove them without compliation errors, should that happen?

Sorry it's a bit of a monolithic PR.

### DIFF
--- a/hledger-lib/Hledger/Utils/String.hs
+++ b/hledger-lib/Hledger/Utils/String.hs
@@ -134,10 +134,10 @@ whitespacechars = " \t\n\r"
 redirectchars   = "<>"
 
 escapeDoubleQuotes :: String -> String
-escapeDoubleQuotes = regexReplace "\"" "\""
+escapeDoubleQuotes = id  -- regexReplace "\"" "\""
 
 escapeQuotes :: String -> String
-escapeQuotes = regexReplace "([\"'])" "\\1"
+escapeQuotes = id  -- regexReplace "([\"'])" "\\1"
 
 -- | Quote-aware version of words - don't split on spaces which are inside quotes.
 -- NB correctly handles "a'b" but not "''a''". Can raise an error if parsing fails.
@@ -346,7 +346,7 @@ strWidth s = maximum $ map (foldr (\a b -> charWidth a + b) 0) $ lines s'
   where s' = stripAnsi s
 
 stripAnsi :: String -> String
-stripAnsi = regexReplace "\ESC\\[([0-9]+;)*([0-9]+)?[ABCDHJKfmsu]" ""
+stripAnsi = regexReplace (toRegex' "\ESC\\[([0-9]+;)*([0-9]+)?[ABCDHJKfmsu]") "" -- PARTIAL: should never happen, no backreferences
 
 -- | Get the designated render width of a character: 0 for a combining
 -- character, 1 for a regular character, 2 for a wide character.
--- a/hledger-lib/Hledger/Utils/String.hs
+++ b/hledger-lib/Hledger/Utils/String.hs
@@ -49,14 +49,14 @@ module Hledger.Utils.String (
  ) where
 
 
-import Data.Char
-import Data.List
-import Text.Megaparsec
-import Text.Megaparsec.Char
+import Data.Char (isDigit, isSpace, toLower, toUpper)
+import Data.List (intercalate, transpose)
+import Text.Megaparsec (Parsec, (<|>), (<?>), between, many, noneOf, oneOf,
+                        parseMaybe, sepBy, takeWhile1P)
+import Text.Megaparsec.Char (char, string)
 import Text.Printf (printf)
 
 import Hledger.Utils.Parse
-import Hledger.Utils.Regex
 
 
 -- | Take elements from the end of a list.
@@ -341,12 +341,15 @@ takeWidth w (c:cs) | cw <= w   = c:takeWidth (w-cw) cs
 -- (not counted), and line breaks (in a multi-line string, the longest
 -- line determines the width).
 strWidth :: String -> Int
-strWidth "" = 0
-strWidth s = maximum $ map (foldr (\a b -> charWidth a + b) 0) $ lines s'
-  where s' = stripAnsi s
+strWidth = maximum . (0:) . map (foldr (\a b -> charWidth a + b) 0) . lines . stripAnsi
 
 stripAnsi :: String -> String
-stripAnsi = regexReplace (toRegex' "\ESC\\[([0-9]+;)*([0-9]+)?[ABCDHJKfmsu]") "" -- PARTIAL: should never happen, no backreferences
+stripAnsi s = maybe s concat $ parseMaybe (many $ takeWhile1P Nothing (/='\ESC') <|> "" <$ ansi) s
+  where
+    -- This parses lots of invalid ANSI escape codes, but that should be fine
+    ansi = string "\ESC[" *> digitSemicolons *> suffix <?> "ansi" :: Parsec CustomErr String Char
+    digitSemicolons = takeWhile1P Nothing (\c -> isDigit c || c == ';')
+    suffix = oneOf ['A', 'B', 'C', 'D', 'H', 'J', 'K', 'f', 'm', 's', 'u']
 
 -- | Get the designated render width of a character: 0 for a combining
 -- character, 1 for a regular character, 2 for a wide character.
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -14,7 +14,7 @@ import Data.List.Extra (groupSort, groupOn)
 import Data.Maybe (fromMaybe)
 import Data.Ord (Down(..))
 import qualified Data.Map as M
-import Data.Text (pack,unpack)
+import qualified Data.Text as T
 import Safe (headMay, lookupJustDef)
 import Text.Printf
 
@@ -28,11 +28,12 @@ import Hledger.Utils
 -- deriving instance Show Account
 instance Show Account where
     show Account{..} = printf "Account %s (boring:%s, postings:%d, ebalance:%s, ibalance:%s)"
-                       (pack $ regexReplace ":" "_" $ unpack aname)  -- hide : so pretty-show doesn't break line
+                       (T.map colonToUnderscore aname)  -- hide : so pretty-show doesn't break line
                        (if aboring then "y" else "n" :: String)
                        anumpostings
                        (showMixedAmount aebalance)
                        (showMixedAmount aibalance)
+      where colonToUnderscore x = if x == ':' then '_' else x
 
 instance Eq Account where
   (==) a b = aname a == aname b -- quick equality test for speed
--- a/hledger-lib/Hledger/Data/AccountName.hs
+++ b/hledger-lib/Hledger/Data/AccountName.hs
@@ -18,7 +18,6 @@ module Hledger.Data.AccountName (
   ,accountNameToAccountOnlyRegex
   ,accountNameToAccountRegex
   ,accountNameTreeFrom
-  ,accountRegexToAccountName
   ,accountSummarisedName
   ,acctsep
   ,acctsepchar
@@ -48,7 +47,6 @@ import Data.Monoid
 import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Tree
-import Text.Printf
 
 import Hledger.Data.Types
 import Hledger.Utils
@@ -210,23 +208,17 @@ clipOrEllipsifyAccountName n        = clipAccountName n
 -- | Escape an AccountName for use within a regular expression.
 -- >>> putStr $ escapeName "First?!#$*?$(*) !@^#*? %)*!@#"
 -- First\?!#\$\*\?\$\(\*\) !@\^#\*\? %\)\*!@#
-escapeName :: AccountName -> Regexp
-escapeName = regexReplaceBy "[[?+|()*\\\\^$]" ("\\" <>)
+escapeName :: AccountName -> String
+escapeName = replaceAllBy (toRegex' "[[?+|()*\\\\^$]") ("\\" <>)  -- PARTIAL: should not happen
            . T.unpack
 
 -- | Convert an account name to a regular expression matching it and its subaccounts.
 accountNameToAccountRegex :: AccountName -> Regexp
-accountNameToAccountRegex "" = ""
-accountNameToAccountRegex a = printf "^%s(:|$)" (escapeName a)
+accountNameToAccountRegex a = toRegex' $ '^' : escapeName a ++ "(:|$)"  -- PARTIAL: Is this safe after escapeName?
 
 -- | Convert an account name to a regular expression matching it but not its subaccounts.
 accountNameToAccountOnlyRegex :: AccountName -> Regexp
-accountNameToAccountOnlyRegex "" = ""
-accountNameToAccountOnlyRegex a = printf "^%s$"  $ escapeName a -- XXX pack
-
--- | Convert an exact account-matching regular expression to a plain account name.
-accountRegexToAccountName :: Regexp -> AccountName
-accountRegexToAccountName = T.pack . regexReplace "^\\^(.*?)\\(:\\|\\$\\)$" "\\1" -- XXX pack
+accountNameToAccountOnlyRegex a = toRegex' $ '^' : escapeName a ++ "$" -- PARTIAL: Is this safe after escapeName?
 
 -- -- | Does this string look like an exact account-matching regular expression ?
 --isAccountRegex  :: String -> Bool
--- a/hledger-lib/Hledger/Data/AccountName.hs
+++ b/hledger-lib/Hledger/Data/AccountName.hs
@@ -39,14 +39,14 @@ module Hledger.Data.AccountName (
 )
 where
 
-import Data.List
 import Data.List.Extra (nubSort)
+import qualified Data.List.NonEmpty as NE
 #if !(MIN_VERSION_base(4,11,0))
-import Data.Monoid
+import Data.Semigroup ((<>))
 #endif
 import Data.Text (Text)
 import qualified Data.Text as T
-import Data.Tree
+import Data.Tree (Tree(..))
 
 import Hledger.Data.Types
 import Hledger.Utils
@@ -115,7 +115,7 @@ expandAccountNames as = nubSort $ concatMap expandAccountName as
 
 -- | "a:b:c" -> ["a","a:b","a:b:c"]
 expandAccountName :: AccountName -> [AccountName]
-expandAccountName = map accountNameFromComponents . tail . inits . accountNameComponents
+expandAccountName = map accountNameFromComponents . NE.tail . NE.inits . accountNameComponents
 
 -- | ["a:b:c","d:e"] -> ["a","d"]
 topAccountNames :: [AccountName] -> [AccountName]
@@ -209,8 +209,10 @@ clipOrEllipsifyAccountName n        = clipAccountName n
 -- >>> putStr $ escapeName "First?!#$*?$(*) !@^#*? %)*!@#"
 -- First\?!#\$\*\?\$\(\*\) !@\^#\*\? %\)\*!@#
 escapeName :: AccountName -> String
-escapeName = replaceAllBy (toRegex' "[[?+|()*\\\\^$]") ("\\" <>)  -- PARTIAL: should not happen
-           . T.unpack
+escapeName = T.unpack . T.concatMap escapeChar
+  where
+    escapeChar c = if c `elem` escapedChars then T.snoc "\\" c else T.singleton c
+    escapedChars = ['[', '?', '+', '|', '(', ')', '*', '$', '^', '\\']
 
 -- | Convert an account name to a regular expression matching it and its subaccounts.
 accountNameToAccountRegex :: AccountName -> Regexp
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -74,7 +74,6 @@ module Hledger.Data.Journal (
   journalCashAccountQuery,
   -- * Misc
   canonicalStyleFrom,
-  matchpats,
   nulljournal,
   journalCheckBalanceAssertions,
   journalNumberAndTieTransactions,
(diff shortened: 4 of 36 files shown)


simonmichael:
I had another look at this today. It's looking pretty good @Xitian9 !

Yes it's unfortunate the Data implementation is partial. Yes, if we don't need the Data classes we can certainly drop them. It's interesting that dropping Journal's Typeable, Data, NFData instances seems to have no ill effects. Maybe they were required by servant or swagger in the old hledger-api ? If so, that'll be great news.

The separate constructor for case-insensitive Regexps is interesting, does that simplify things overall ?

escape(Double)?Quotes are now a no-op - is that just WIP ?

I did a few quick benchmark runs, and they seem to suggest a slight improvement in regex matching performance (note the `ff` commands):
```
$ BENCHEXES=hledger-1.18,hledger make bench
quickbench -w hledger-1.18,hledger
Running 12 tests 1 times with 2 executables at 2020-08-29 13:46:05 PDT:

Best times:
+-----------------------------------------------++--------------+---------+
|                                               || hledger-1.18 | hledger |
+===============================================++==============+=========+
| -f examples/100x100x10.journal print          ||         0.30 |    0.28 |
| -f examples/1000x1000x10.journal print        ||         0.30 |    0.31 |
| -f examples/10000x1000x10.journal print       ||         2.69 |    2.36 |
| -f examples/10000x1000x10.journal print ff    ||         1.16 |    1.04 |
| -f examples/100x100x10.journal register       ||         0.12 |    0.08 |
| -f examples/1000x1000x10.journal register     ||         0.38 |    0.38 |
| -f examples/10000x1000x10.journal register    ||         3.11 |    3.17 |
| -f examples/10000x1000x10.journal register ff ||         1.26 |    1.09 |
| -f examples/100x100x10.journal balance        ||         0.08 |    0.08 |
| -f examples/1000x1000x10.journal balance      ||         0.27 |    0.27 |
| -f examples/10000x1000x10.journal balance     ||         1.26 |    1.16 |
| -f examples/10000x1000x10.journal balance ff  ||         1.12 |    0.98 |
+-----------------------------------------------++--------------+---------+
```

Dropping the old Hledger.Utils.Regex API entirely might affect users of hledger-lib that we don't know about. I'm not against it necessarily.

Is there any need for the underscore suffix of toRegex_, toRegexCI_ ?

"match" clashes with a lot of other packages, but I guess it's a kind of standard, maybe that's ok.

simonmichael:
Benchmarking a little more carefully confirms... perhaps we could say it's "10% faster":
```
$ quickbench -w hledger-1.18,hledger -n 3 -N 3
Running 12 tests 3 times with 2 executables at 2020-08-29 13:58:07 PDT:

Best times 1:
+-----------------------------------------------++--------------+---------+
|                                               || hledger-1.18 | hledger |
+===============================================++==============+=========+
| -f examples/100x100x10.journal print          ||         0.07 |    0.07 |
| -f examples/1000x1000x10.journal print        ||         0.28 |    0.28 |
| -f examples/10000x1000x10.journal print       ||         2.56 |    2.30 |
| -f examples/10000x1000x10.journal print ff    ||         1.17 |    0.95 |
| -f examples/100x100x10.journal register       ||         0.08 |    0.07 |
| -f examples/1000x1000x10.journal register     ||         0.37 |    0.33 |
| -f examples/10000x1000x10.journal register    ||         3.47 |    3.17 |
| -f examples/10000x1000x10.journal register ff ||         1.25 |    1.07 |
| -f examples/100x100x10.journal balance        ||         0.07 |    0.06 |
| -f examples/1000x1000x10.journal balance      ||         0.27 |    0.25 |
| -f examples/10000x1000x10.journal balance     ||         1.38 |    1.17 |
| -f examples/10000x1000x10.journal balance ff  ||         1.19 |    0.92 |
+-----------------------------------------------++--------------+---------+

Best times 2:
+-----------------------------------------------++--------------+---------+
|                                               || hledger-1.18 | hledger |
+===============================================++==============+=========+
| -f examples/100x100x10.journal print          ||         0.07 |    0.06 |
| -f examples/1000x1000x10.journal print        ||         0.29 |    0.26 |
| -f examples/10000x1000x10.journal print       ||         2.67 |    2.61 |
| -f examples/10000x1000x10.journal print ff    ||         1.15 |    0.98 |
| -f examples/100x100x10.journal register       ||         0.08 |    0.07 |
| -f examples/1000x1000x10.journal register     ||         0.34 |    0.32 |
| -f examples/10000x1000x10.journal register    ||         3.49 |    3.21 |
| -f examples/10000x1000x10.journal register ff ||         1.24 |    1.01 |
| -f examples/100x100x10.journal balance        ||         0.06 |    0.06 |
| -f examples/1000x1000x10.journal balance      ||         0.26 |    0.28 |
| -f examples/10000x1000x10.journal balance     ||         1.28 |    1.14 |
| -f examples/10000x1000x10.journal balance ff  ||         1.07 |    0.96 |
+-----------------------------------------------++--------------+---------+

Best times 3:
+-----------------------------------------------++--------------+---------+
|                                               || hledger-1.18 | hledger |
+===============================================++==============+=========+
| -f examples/100x100x10.journal print          ||         0.08 |    0.07 |
| -f examples/1000x1000x10.journal print        ||         0.29 |    0.27 |
| -f examples/10000x1000x10.journal print       ||         2.50 |    2.29 |
| -f examples/10000x1000x10.journal print ff    ||         1.13 |    1.02 |
| -f examples/100x100x10.journal register       ||         0.07 |    0.07 |
| -f examples/1000x1000x10.journal register     ||         0.39 |    0.32 |
| -f examples/10000x1000x10.journal register    ||         3.38 |    3.02 |
| -f examples/10000x1000x10.journal register ff ||         1.18 |    1.00 |
| -f examples/100x100x10.journal balance        ||         0.06 |    0.06 |
| -f examples/1000x1000x10.journal balance      ||         0.25 |    0.27 |
| -f examples/10000x1000x10.journal balance     ||         1.31 |    1.17 |
| -f examples/10000x1000x10.journal balance ff  ||         1.22 |    1.03 |
+-----------------------------------------------++--------------+---------+
```

simonmichael:
And, confirming this really does fix things like the hledger-web internal server error reported in #1245:
Screen Shot 2020-08-29 at 2 12 25 PM

Xitian9:
> Yes it's unfortunate the Data implementation is partial. Yes, if we don't need the Data classes we can certainly drop them. It's interesting that dropping Journal's Typeable, Data, NFData instances seems to have no ill effects. Maybe they were required just for servant or swagger in the old hledger-api ? If so, that'll be great news.

I'll try removing `Data` and `NFData` and see if anything happens. Since GHC 7.10, `Typeable` is automatically derived even without a `deriving` clause, so removing them has no effect.

> The separate constructor for case-insensitive Regexps is interesting, does that simplify things overall ?

It doesn't simplify things, but it's there to give some instances more reasonable/lawful definitions. This way `Eq`, `Ord`, `Show`, and `Read` can work with the string used to generate them, but not equate the case-insensitve and non-case-insensitive versions.

> escape(Double)?Quotes are now a no-op - is that just WIP ?

WIP. I was trying to see what would break if I took those out. Apparently nothing that's tested for? What are those there to do?

> Dropping the old Hledger.Utils.Regex API entirely might affect users of hledger-lib that we don't know about. I'm not against it necessarily.

Yes, this definitely breaks things. I'm happy to take direction on what should happen for backwards compatibility.

> Is there any need for the underscore suffix of toRegex_, toRegexCI_ ?

No, they're just there for now while things are finalised.

Lastly, the biggest (and ideally only) use of `toRegex'` and `toRegexCI'` is for hard-coded regular expressions that we know compile. We could get rid of these by using a quasiquoter. I had [this one](https://hackage.haskell.org/package/regex-tdfa-quasiquoter) in mind, but I don't know if that's the best one. What do you think?

Xitian9:
Glad to see there's a speedup too! I didn't specifically aim for that, but I think this comes from regular expressions being compiled everywhere now.

simonmichael:
> escape(Double)?Quotes are now a no-op - is that just WIP ?
> 
> WIP. I was trying to see what would break if I took those out. Apparently nothing that's tested for? What are those there to do?
> 
Apparently, indeed. They are just general utilities I've needed from time to time to avoid bugs when handling quoted/spaced text. I think at least one of them is used indirectly eg by accountQuery.

> Dropping the old Hledger.Utils.Regex API entirely might affect users of hledger-lib that we don't know about. I'm not against it necessarily.
> 
> Yes, this definitely breaks things. I'm happy to take direction on what should happen for backwards compatibility.
> 
I suppose the kind thing would be to keep old Hledger.Utils.Regex for at least one release cycle and add new Hledger.Utils.Regexp. We could also poll the mail list for input. @apauley, @adept, @hpdeifel, @peti, will a greatly altered Hledger.Utils.Regex api affect you much ?

> Lastly, the biggest (and ideally only) use of toRegex' and toRegexCI' is for hard-coded regular expressions that we know compile. We could get rid of these by using a quasiquoter. I had this one <https://hackage.haskell.org/package/regex-tdfa-quasiquoter> in mind, but I don't know if that's the best one. What do you think?
> 
It looks unmaintained, so I wouldn't be too keen. Secondly it's another tie to a specific regex library which we might want to change in future, but that's minor.

In general, adding a quasiquoter for hard-coded regexps might be a nice improvement if it's not too much effort or maintenance cost or impactful on compile time. I probably would keep the partial helpers around for a while though.

hledger-lib/Hledger/Utils/String.hs
```diff
@@ -134,10 +134,10 @@ whitespacechars = " \t\n\r"
 redirectchars   = "<>"
 
 escapeDoubleQuotes :: String -> String
-escapeDoubleQuotes = regexReplace "\"" "\""
+escapeDoubleQuotes = id  -- regexReplace "\"" "\""
```
Xitian9, line 137: Actually, I think this is a no-op in any case. The original code isn't actually escaping anything, just replacing every quotation mark with another quotation mark. This should either be fixed if it is supposed to do something, or be removed.

hledger-lib/Hledger/Utils/String.hs
```diff
@@ -134,10 +134,10 @@ whitespacechars = " \t\n\r"
...
 
 escapeQuotes :: String -> String
-escapeQuotes = regexReplace "([\"'])" "\\1"
+escapeQuotes = id  -- regexReplace "([\"'])" "\\1"
```
Xitian9, line 140: This is not used anywhere, but is exported. Should it be removed or kept?

Xitian9:
It looks like the `Data` and `NFData` instances are all unnecessary, and the `Typeable` are all redundant. I've removed them all.

hpdeifel:
> will a greatly altered Hledger.Utils.Regex api affect you much ?

I've never used this module :+1:

apauley:
> will a greatly altered Hledger.Utils.Regex api affect you much ?

Not at all, [hledger-flow](https://github.com/apauley/hledger-flow) does not use any of the hledger libraries directly, it runs hledger in a new process, similar to what a shell script would do.

simonmichael:
On Github PRs like this one, I struggle to understand which commits are old, rebased, amended, new.. Could you help me out: after my last comment at https://github.com/simonmichael/hledger/pull/1330#issuecomment-683468629 I see :

Xitian9 added 4 commits 16 days ago
@Xitian9 lib: Remove unused Tree functions. ccd6fdd
@Xitian9 lib,cli,ui,web: Make Regexp a wrapper for Regex. e5371d5
@Xitian9 lib: Remove unneeded total Query code. e3b2c94
@Xitian9 lib: Improve Read and Show instances for Regexp, get rid of custom sh… 01f5a92

These are all commits I've already seen, just rebased, am I right ? 

Further down I see 5 more commits, which are clear enough (all new).

hledger-lib/Hledger/Utils/String.hs
simonmichael: You're right! String.escapeDoubleQuotes never worked. Committed in 2014 with message: "More crazy quote engineering.. seems more robust than before."

Text.escapeDoubleQuotes was fixed in 2018:
```haskell
escapeDoubleQuotes = T.replace "\"" "\\\""
```
but not the original String version.

It's not exported, so might as well be moved into quoteIfNeeded, its only user. 

Feel free to fix it, or ignore it and I'll do after this PR.

hledger-lib/Hledger/Utils/String.hs
simonmichael: I imagine these things could be handy for hledger script authors working with command line arguments, but quoteIfNeeded probably is enough. Sure let's remove it.

simonmichael:
> It looks like the Data and NFData instances are all unnecessary, and the Typeable are all redundant.

Fantastic, good cleanup. 

Slight doubt about removing NFData, which gives us the ability to force hledger data values, which seems a useful thing as we get more into performance optimisation. Here was the comment when added by 790d42bfa4119541950cbf8e74b1c538eab267ea:
```
derive NFData (and Generic) for all types  so we can benchmark things more easily with criterion.  As well as NFData, the Generic instance and a bunch more GHC extensions seemed necessary. This is a little scary, impact unknown.
```
So actually it was for Criterion, which I used occasionally in dev.hs or tools/progressionbench.hs.

simonmichael:
@hpdeifel, @apauley cool, thanks. I'm ok with changing the Hledger.Utils.Regex api in tomorrow's 1.19, on the basis that no-one has ever mentioned it.

simonmichael:
I compared build time, twice, just to see if dropping those instances had an impact. I'm going to say it made builds "3% faster". :)

I'm ready to merge this, but probably without the commit removing NFData. Ok with you ?

simonmichael:
```
Time to build the hledger packages (`gtime make buildtest`):

master (5b647902):
120.21user 41.17system 2:49.20elapsed  95%CPU (0avgtext+0avgdata 785948maxresident)k 0inputs+1outputs (272829major+5038291minor)pagefaults 0swaps
117.73user 39.58system 2:38.10elapsed  99%CPU (0avgtext+0avgdata 784672maxresident)k 0inputs+0outputs (239116major+4956271minor)pagefaults 0swaps

1330 (b91b391d):
116.29user 38.28system 2:35.59elapsed  99%CPU (0avgtext+0avgdata 800392maxresident)k 0inputs+0outputs (278876major+4928780minor)pagefaults 0swaps
113.78user 38.31system 2:30.35elapsed 101%CPU (0avgtext+0avgdata 796108maxresident)k 0inputs+0outputs (230555major+4971542minor)pagefaults 0swaps
```

simonmichael:
I forgot that `make buildtest` does a non-optimised build. I'll say the normal optimised build is "7% faster":

Time to optimised-build the hledger packages (`gtime stack build --force-dirty --ghc-options=-fforce-recomp`):

```
master (5b647902):
306.66user 42.28system 6:09.03elapsed  94%CPU (0avgtext+0avgdata 1189180maxresident)k 0inputs+0outputs (167198major+5350911minor)pagefaults 0swaps
287.85user 37.09system 5:04.17elapsed 106%CPU (0avgtext+0avgdata 1160420maxresident)k 0inputs+0outputs (163114major+5068612minor)pagefaults 0swaps

1330 (b91b391d):
275.41user 37.90system 4:58.18elapsed 105%CPU (0avgtext+0avgdata 1191832maxresident)k 0inputs+0outputs (155186major+5094106minor)pagefaults 0swaps
268.65user 36.46system 4:43.16elapsed 107%CPU (0avgtext+0avgdata 1218752maxresident)k 0inputs+1outputs (151460major+5016105minor)pagefaults 0swaps
```

simonmichael:
And... one more pair of measurements with the NFData instances restored:

Time to optimised-build the hledger packages (`gtime stack build --force-dirty --ghc-options=-fforce-recomp`):
```
1330 minus NFData removal (198ed0bd):
321.78user 43.01system 6:00.61elapsed 101%CPU (0avgtext+0avgdata 1163576maxresident)k 0inputs+0outputs (191687major+5096229minor)pagefaults 0swaps
299.74user 41.14system 5:33.15elapsed 102%CPU (0avgtext+0avgdata 1108892maxresident)k 0inputs+0outputs (192408major+5077066minor)pagefaults 0swaps
```

simonmichael:
Some more notes, for future work. Based on the above numbers and some discussion on #haskell, I think: 

- it's the NFData instances that make our builds ~7% slower
- particularly because they are auto-derived using Generic (*Are you using generics derived NFData instances? Generics tend to be slow, especially if you have lots of ADTs and/or large ADTs.*)
- hand-written NFData instances might be faster
- NFData instances could also be hidden behind a build flag, since at the moment we only use them for (rare) timing experiments
- we also rely on Generic to auto-derive aeson ToJSON/FromJSON instances
- hand-written instances for those are rumoured to be quite a bit faster (*Remove those generic aeson instances and watch your compile speed jump up. :-)*)
- Data.Aeson.TH is a third option that sounds worth trying (*Instead of manually writing your FromJSON instance, there are two options to do it automatically: 1. Data.Aeson.TH provides Template Haskell functions which will derive an instance at compile time. The generated instance is optimized for your type so it will probably be more efficient than: 2. The compiler can provide a default generic implementation for parseJSON.*
   *Perhaps Data.Aeson.TH might offer better performance than Generics based instances for aeson. (It especially does if you have some mechanism of manually splicing the generated instances, of course)*
   *you can manually splice the instances (perhaps using -ddump-splices), then you're paying the price only once - at the cost of having to do this job every time an ADT gets changed/added/removed*
   *well I dont mean to write out the TH yourself. Just use https://hackage.haskell.org/package/aeson-1.5.4.0/docs/Data-Aeson-TH.html* )

So back to this PR: I'm good with removing the auto-derived NFData instances. Ready to merge.

Xitian9:
That got merged faster than I expected. There are still a couple of cleanups to do around removing trailing underscores from `toRegex_` and removing/fixing some of the ‘escape string’ functions. I'll put those in a separate PR.

simonmichael:
Ah, sorry @Xitian9. With release tomorrow and other things scheduled I'm trying to close some open loops. I also made a mistake, rebase-merging when I wanted to add a merge description, so I rolled back master and will be merging this again. So feel free to push more here, I won't merge till you are ready.

simonmichael:
Today or not too late tomorrow would be fine.

Here's a draft of the summary message:
```
replace our stringly Regex with safer compiled Regexp type

This PR #1330, addressing #1312 (parseQuery is partial) and #1245
(internal server error).

User-visible changes:

- hledger-web now handles malformed regular expressions
  (eg, a query consisting of the single character `?`) gracefully,
  showing a tidy error message instead "internal server error".

API/internal changes:

- The stringly Regex type has been replaced by the Regexp ADT, which
  contains the compiled regular expression (so is guaranteed to be
  usable at runtime) and the original string (so can be serialised,
  printed, compared, etc.)

- Typeable and Data instances are no longer derived for hledger's
  data types; they were redundant/no longer needed

- NFData instances are no longer derived for hledger's data types.
  This speeds up a full build by roughly 7%. But, it means we can't
  deep-evaluate hledger values or time hledger code with Criterion, for
  now. https://github.com/simonmichael/hledger/pull/1330#issuecomment-684075129
  has some ideas on this.

- Some internal use of regular expressions was replaced by string
  replacement or parsers.

- Query no longer has a custom Show instance

- some old utilities in Hledger.Utils.Tree and some obsolete code for
  --display was removed.
```

simonmichael:
I noticed hledger-ui still exits if you search for a bad regexp, if you have time feel free to look at that.

simonmichael:
Re the Hledger.Utils.Regex api: what would you think of renaming match/replaceAllBy to regexMatch/regexReplaceAllBy ? Though longer, it'd be consistent and might make a little clearer what code is doing.

simonmichael:
Either that, or go the other way and rename regexReplace/regexReplaceMemo_ to replace/replaceMemo..

Xitian9:
> On Github PRs like this one, I struggle to understand which commits are old, rebased, amended, new.. Could you help me out: after my last comment at [#1330 (comment)](https://github.com/simonmichael/hledger/pull/1330#issuecomment-683468629) I see :
> 
> Xitian9 added 4 commits 16 days ago
> @Xitian9 lib: Remove unused Tree functions. [ccd6fdd](https://github.com/simonmichael/hledger/commit/ccd6fdd7b903a2eaed71f4fc03c994db91c7936f)
> @Xitian9 lib,cli,ui,web: Make Regexp a wrapper for Regex. [e5371d5](https://github.com/simonmichael/hledger/commit/e5371d5a6a2cbe6661111f57e231a668aafe5355)
> @Xitian9 lib: Remove unneeded total Query code. [e3b2c94](https://github.com/simonmichael/hledger/commit/e3b2c94353ded9c5fa9fc1fda3ba9f7c599104b0)
> @Xitian9 lib: Improve Read and Show instances for Regexp, get rid of custom sh… [01f5a92](https://github.com/simonmichael/hledger/commit/01f5a92761f7f95a104a2ada1d1cd36885a6ab5c)
> 
> These are all commits I've already seen, just rebased, am I right ?
> 
> Further down I see 5 more commits, which are clear enough (all new).

Yes, those were just rebases.

I'm actually a bit unclear on the etiquette of when to rebase, when to rewrite history (e.g. fixing a typo in an unmerged commit), and when to just add a new commit. Sorry if I get it wrong sometimes.

Xitian9:
I've made the requested changes to the Hledger.Utils.Regex API, and fixed quoteIfNeeded. I haven't yet found the problem in hledger-ui, but I'll keep looking.

Did I accidentally close this pull request? Should I resubmit? I wouldn't want to lose this conversation.

simonmichael:
> I've made the requested changes to the Hledger.Utils.Regex API, and fixed quoteIfNeeded. I haven't yet found the problem in hledger-ui, but I'll keep looking.
> 
> Did I accidentally close this pull request? Should I resubmit? I wouldn't want to lose this conversation.
> 

Thanks! No you're doing fine, it was me who closed it with my hasty first merge attempt. Will look in the morning.

simonmichael:
@Xitian9: looks good. I'm going to merge this shortly if you have no objections, with the latest changes from your repo, except for the hledger-ui change. (I think it's fine for it to exit with an error at startup, where I think it should not exit is when you enter a query while the app is running, setFilter function in UIState).

simonmichael:
This has landed in master. Thanks a lot, @Xitian9.

PR #1334, opened 2020-08-26
lib,cli,ui: Remove old CPP directives made redundant by version bounds.

Xitian9:
Also gets rid of the partial functions `parsedate` and `mkdatespan`, used entirely with hard-coded arguments (with one exception). Use the total functions `fromGregorian` or `parsedateM` instead.


simonmichael:
Great CPP cleanup, thanks!

The mac build failure is a recurring unrelated CI issue.

The parsedate/mkdatespan could have been a separate PR perhaps. These were little helpers to save typing/brain load in test code/GHCI exploration, but the new code is not much longer and keeps things total. Looks good to me.

Issue #1338, opened 2020-09-01
hledger-ui crashes on malformed regular expression


Xitian9:
The problem comes from `queryFromOpts` and `queryOptsFromOpts`.

https://github.com/simonmichael/hledger/blob/5b647902c9f89b84ce814dec5bdc50778c82a153/hledger-lib/Hledger/Reports/ReportOptions.hs#L465-L469

Basically, we need to reckon with the fact that generating a `Query` from a `ReportOpts` might yield a `RegexError`. I can see three options:
1. Regex errors are handled as `None` (or some other Query). The disadvantage is that users won't get a warning on a malformed Regular expression.
2. Functions which call `query(Opts)?FromOpts` return `Either RegexError a` instead of `a`. The disadvantage is that this means functions like `multiBalanceReport` and `compoundBalanceReport` meet these criteria, and we went to a lot of effort to avoid this. (On the plus side, `(multi|compound)BalanceReportWith` are fine).
3. We store the parsed `Query` and `[QueryOpt]` inside `ReportOpts` itself, rather than the `String` that it will be parsed from. The disadvantage is that this is the biggest interface change. The advantage that we can stop passing around both `ReportOpts` and `Query` in lots of places (a quick grep reveals at least 15 functions with `ReportOpts -> Query ->`). This would also remove any ambiguity about whether you should use the `Query` or `ReportOpts`.

My inclination is to go with 3, and move the `queryFromOpts` logic into `rawOptsToReportOpts`. We should probably store all three of `queryFromOpts`, `queryFromOptsOnly`, and `queryOptsFromOpts`. Maybe we can use some friendlier names. Thoughts?

Xitian9:
Interestingly, `queryFromOptsOnly` is used in only two places (Hledger.Cli.Commands.Add and Hledger.Cli.Commands.Registermatch), and in both places replacing it with `queryFromOpts` doesn't cause any tests to fail. Is it necessary to have this at all?

simonmichael:
Argh, more deep-rooted partiality (sorry!). 

This will be something to explore for next release I think, unless we want to implement 1 in hledger-ui as an interim fix.

Xitian9:
I don't use `hledger-ui`, so I have no strong opinions about whether to implement 1 in the interim. It's a simple change, the only disadvantage I can think of is the lack of error message for people using `hledger-cli`.

simonmichael:
[...]

PR #1340, opened 2020-09-03
Make sure ReportOpts only stores valid values.

Xitian9:
This addresses #1338. It's a big api change, so let me know your thoughts.

I'd particularly appreciate checking out how hledger-ui and hledger-web fare. I've checked myself and they seem to work fine, but since I don't normally use these tools I might be missing something.

### DIFF
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -82,6 +82,8 @@ type CompoundBalanceReport = CompoundPeriodicReport DisplayName MixedAmount
 -- type alias just to remind us which AccountNames might be depth-clipped, below.
 type ClippedAccountName = AccountName
 
+-- Type alias for a valuation function
+type Valuation = Day -> MixedAmount -> MixedAmount
 
 
 -- | Generate a multicolumn balance report for the matched accounts,
@@ -106,6 +108,8 @@ multiBalanceReportWith ropts' j priceoracle = report
     -- Queries, report/column dates.
     reportspan = dbg "reportspan" $ calculateReportSpan ropts' j
     ropts      = dbg "reportopts" $ makeReportQuery ropts' reportspan
+    valuation  = makeValuation ropts' j priceoracle  -- Must use ropts' instead of ropts,
+                                                     -- so the reportspan isn't used for valuation
 
     -- Group postings into their columns.
     colps    = dbg'' "colps"  $ getPostingsByColumn ropts j reportspan
@@ -117,7 +121,7 @@ multiBalanceReportWith ropts' j priceoracle = report
 
     -- Generate and postprocess the report, negating balances and taking percentages if needed
     report = dbg' "report" $
-      generateMultiBalanceReport ropts j priceoracle colspans colps startbals
+      generateMultiBalanceReport ropts j valuation colspans colps startbals
 
 -- | Generate a compound balance report from a list of CBCSubreportSpec. This
 -- shares postings between the subreports.
@@ -134,6 +138,8 @@ compoundBalanceReportWith ropts' j priceoracle subreportspecs = cbr
     -- Queries, report/column dates.
     reportspan = dbg "reportspan" $ calculateReportSpan ropts' j
     ropts      = dbg "reportopts" $ makeReportQuery ropts' reportspan
+    valuation  = makeValuation ropts' j priceoracle  -- Must use ropts' instead of ropts,
+                                                     -- so the reportspan isn't used for valuation
 
     -- Group postings into their columns.
     colps    = dbg'' "colps"  $ getPostingsByColumn ropts{empty_=True} j reportspan
@@ -149,7 +155,7 @@ compoundBalanceReportWith ropts' j priceoracle subreportspecs = cbr
             ( cbcsubreporttitle
             -- Postprocess the report, negating balances and taking percentages if needed
             , prNormaliseSign cbcsubreportnormalsign $
-                generateMultiBalanceReport ropts' j priceoracle colspans colps' startbals'
+                generateMultiBalanceReport ropts' j valuation colspans colps' startbals'
             , cbcsubreportincreasestotal
             )
           where
@@ -234,6 +240,18 @@ makeReportQuery ropts reportspan
     dateless         = dbg "dateless" . filterQuery (not . queryIsDateOrDate2)
     dateqcons        = if date2_ ropts then Date2 else Date
 
+-- | Make a valuation function for valuating MixedAmounts and a given Day
+makeValuation :: ReportOpts -> Journal -> PriceOracle -> Valuation
+makeValuation ropts j priceoracle day = case value_ ropts of
+    Nothing -> id
+    Just v  -> mixedAmountApplyValuation priceoracle styles day mreportlast today multiperiod v
+  where
+    -- Some things needed if doing valuation.
+    styles = journalCommodityStyles j
+    mreportlast = reportPeriodLastDay ropts
+    today = fromMaybe (error' "multiBalanceReport: could not pick a valuation date, ReportOpts today_ is unset") $ today_ ropts  -- XXX shouldn't happen
+    multiperiod = interval_ ropts /= NoInterval
+
 -- | Group postings, grouped by their column
 getPostingsByColumn :: ReportOpts -> Journal -> DateSpan -> Map DateSpan [Posting]
 getPostingsByColumn ropts j reportspan = columns
@@ -314,11 +332,11 @@ acctChangesFromPostings ropts ps = HM.fromList [(aname a, a) | a <- as]
 -- | Accumulate and value amounts, as specified by the report options.
 --
 -- Makes sure all report columns have an entry.
-accumValueAmounts :: ReportOpts -> Journal -> PriceOracle -> [DateSpan]
+accumValueAmounts :: ReportOpts -> Valuation -> [DateSpan]
                   -> HashMap ClippedAccountName Account
                   -> HashMap ClippedAccountName (Map DateSpan Account)
                   -> HashMap ClippedAccountName (Map DateSpan Account)
-accumValueAmounts ropts j priceoracle colspans startbals acctchanges =  -- PARTIAL:
+accumValueAmounts ropts valuation colspans startbals acctchanges =  -- PARTIAL:
     HM.mapWithKey processRow $ acctchanges <> (mempty <$ startbals)
   where
     -- Must accumulate before valuing, since valuation can change without any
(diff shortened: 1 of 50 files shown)


simonmichael (commented):
Nice work! A few comments:

Would you mind documenting the new overline width argument somewhere, eg in StringFormat's haddock (and for my own understanding, could you show an example of the user-visible effect ?)

ReportOpts now includes the notion of time (current day), it's more of a "report spec" than a pure description of report options. If you want to reuse the report options for a different day, you have to regenerate and overwrite its query_. The change seems to simpify code in most cases so I'm ok with this.

"Valuation" isn't obviously a function (to convert amounts to other amounts on a given day) to me, but I'm ok with that name for now.

Rather than adding the query string to UIState, shall we keep it in ReportOpts alongside query_ ? Analogous to Regexp holding both raw and parsed forms.

simonmichael:
(PS, and of course overall this PR reduces partiality / possibility of error calls after startup, which is great.)

Xitian9:
> Would you mind documenting the new overline width argument somewhere, eg in StringFormat's haddock (and for my own understanding, could you show an example of the user-visible effect ?)

Sure thing,
 
> "Valuation" isn't obviously a function (to convert amounts to other amounts on a given day) to me, but I'm ok with that name for now.

I'm happy to remove the type and just leave the full `Day -> MixedAmount -> MixedAmount` in there.

> Rather than adding the query string to UIState, shall we keep it in ReportOpts alongside query_ ? Analogous to Regexp holding both raw and parsed forms.

I'm a bit hesitant to introduce redundancy which could get out of sync, but perhaps this could be ameliorateb by including a `regenerateReportOpts :: Day -> String -> ReportOpts -> Either String ReportOpts` function, which sets a new date, query string, and parsed `Query`.

Xitian9:
Actually, I'm not sure about moving the query string into ReportOpts.

The lack of proper boolean algebra support for query parsing means there are plenty of Queries that cannot be obtained from a query string. What should be stored in the query string in these cases? This comes up for me fairly frequently, as I define a number of Queries in code, rather than parsing them from strings.

simonmichael:
Good point. Maybe String is one option. 

I felt the redundancy exists anyway in hledger-ui, and probably other places sooner or later, so might as well manage it in one place. But I'm not sure about this either. The naming and purpose of regenerateReportOpts feels non-obvious which is a sign that maybe this data model isn't optimal yet. Maybe it's not a big deal, maybe we'll find clearer names. Let's think a bit more.

hledger-lib/Hledger/Reports/MultiBalanceReport.hs
```diff
@@ -241,7 +238,7 @@ makeReportQuery ropts reportspan
     dateqcons        = if date2_ ropts then Date2 else Date
 
 -- | Make a valuation function for valuating MixedAmounts and a given Day
-makeValuation :: ReportOpts -> Journal -> PriceOracle -> Valuation
+makeValuation :: ReportOpts -> Journal -> PriceOracle -> Day -> MixedAmount -> MixedAmount
```
simonmichael, line 241: Maybe parenthesise (Day -> MixedAmount -> MixedAmount) here too, like the others below.

simonmichael:
Brainstorming alternatives: we could keep ReportOpts as a pure description of options, like the other Opts types, and add something new that includes the current date and derived Query, like:

```haskell
data ReportSpec = ReportSpec {
   rsopts :: ReportOpts
  ,rstoday :: Day
  ,rsquery :: Query
}
```

Or perhaps it's just not worth the code noise and attention cost.

simonmichael:
The easier alternative: allow ReportOpts' query string to be Nothing (signifying a Query not currently parseable from command line arguments), and maybe rename regenerateReportOpts ?
```haskell
-- | Set a new today-date, query string, and Query derived from these in this ReportOpts.
-- Or return a query parse error.
reportOptsSetQuery :: Day -> T.Text -> ReportOpts -> Either String ReportOpts
```

Xitian9:
> Brainstorming alternatives: we could keep ReportOpts as a pure description of options, like the other Opts types, and add something new that includes the current date and derived Query, like:

This is more or less the current approach used in `multiBalanceReportWith`, except also with a `Day` extracted. Would there remain a query string in `ReportOpts`?

The hesitation I have with any option that keeps a query string in the same place as the derived `Query` is the need to keep them synced. When the query string is floated out into `UIOpts`, it's at least clear that this is something only for handling the UI, and not for report processing. As far as I know, the only use in keeping the query string around is so it can be modified in hledger-ui. Correct me if I'm wrong. Or do you have in mind a use case where there is a pure description of options, and you need to apply the same description on different days?

simonmichael:
Just trying to guess which formulation will be least hassle and error-prone to work with, yes including when you want to reuse some report options with different "today" dates. Maybe that's rare.

I'm realising that ReportOpts isn't a "pure description of options" even now, the today_ field is used for multiple things: parsing the query string, interpreting partial dates in -b/-e/-p, selecting prices for market valuation, etc.

Xitian9:
I'm actually not 100% clear on what constitutes a pure description of options. Here are some questions I have.

1. Does `today_` belong in `ReportOpts`? If so, what is it and should it be used for?
    a. Valuation date?
    b. Resolving `SmartDate`s parsed from a query string?
    c. Resolving `SmartDate`s parsed from -b/-e/-p
2. There are dates involved in `today_`, `period_`, and `query_`. Which takes priority when all are present? Can some of these be discarded when we have the others?
3. There's some amount of redundancy between `period_` and `interval_`. Should we clarify the relationship?
4. A bunch of flags in `ReportOpts` are later put into a `Query` in some way, including `period_`, `statuses_`, `depth_`, `date2_`, `empty_`, `real_`, and a query string. Should we separate the query from these so it's clear which to use?
5. There are some options which are related to display on a cli, including `pretty_tables_`, `color_`, and `transpose_`. Do these belong in `CliOpts`?

simonmichael:
Answering re the current state of things:
> Does today_ belong in ReportOpts? If so, what is it and should it be used for?
> a. Valuation date?
> b. Resolving SmartDates parsed from a query string?
> c. Resolving SmartDates parsed from -b/-e/-p
It has been there for a long time and I think is used for all of these
> There are dates involved in today_, period_, and query_. Which takes priority when all are present? Can some of these be discarded when we have the others?
https://hledger.org/hledger.html#report-start-end-date <https://hledger.org/hledger.html#report-start-end-date>  talks about how dates from -b, -e, -p, date:, date2: are combined:

"among start/end dates specified with options, the last (i.e. right-most) option takes precedence... The effective report start and end dates are the intersection of the start/end dates from options and that from date: queries."
> There's some amount of redundancy between period_ and interval_. Should we clarify the relationship?
Hmm.. is there ?  In ReportOpts, I think Interval defines the frequency/length/transitions of report subperiods, while Period defines the span of time that is the overall report period.
> A bunch of flags in ReportOpts are later put into a Query in some way, including period_, statuses_, depth_, date2_, empty_, real_, and a query string. Should we separate the query from these so it's clear which to use?
We could. Sometimes you need to check a flag, sometimes you need to apply a query, so the two forms can be handy. I think usually ReportOpts are built once and we don't need to alter them so perhaps this redundancy isn't a pressing problem ?
> There are some options which are related to display on a cli, including pretty_tables_, color_, and transpose_. Do these belong in CliOpts?
Possibly. I think some things are kept in ReportOpts because report code needs to know about them (and it can't access CliOpts). Isn't transpose one of those..

simonmichael:
Part of the reason ReportOpts holds all these things is its job as a kind of universal "app state/reader environment" argument that we pass everywhere (in higher code, wrapped in a CliOpts/UIOpts/WebOpts). This helps keep function arguments manageable and reasonably consistent. We could instead put everything into a custom Hledger reader monad (or perhaps mess with implicit parameters - let's not go there). I usually feel that would add code complexity/difficulty that isn't really worth the benefit.

Xitian9:
Alright, ReportSpec introduced. Making that change (especially in hledger-ui) involved changing a lot of boilerplate. If I'm going to make any other changes in there I'm sorely tempted to introduce some lenses.

Xitian9:
Given the deep nesting of `ReportOpts` in `ReportSpec`, `CliOpts`, `UIOpts`, and `UIState`, and the fact that it's needed so often, it's probably worth it to make some lenses. We could minimise dependencies by depending only on `microlens` and `microlens-th`. We could use `makeClassy ReportOpts`, and make an instance of `HasReportOpts` for each of the above types. 

Is this something you'd be open to? We'd need to put some thougth into what the lenses would be called, but we could replace monstrosities like this:
```
-- | Toggle between showing all and showing only nonempty (more precisely, nonzero) items.
toggleEmpty :: UIState -> UIState
toggleEmpty ui@UIState{aopts=uopts@UIOpts{cliopts_=copts@CliOpts{reportspec_=rspec@ReportSpec{rsOpts=ropts}}}} =
  ui{aopts=uopts{cliopts_=copts{reportspec_=rspec{rsOpts=toggleEmpty ropts}}}}
  where
    toggleEmpty ropts = ropts{empty_=not $ empty_ ropts}
```
with the more understandable (naming of `empty_` TBD)
```
-- | Toggle between showing all and showing only nonempty (more precisely, nonzero) items.
toggleEmpty :: UIState -> UIState
toggleEmpty = over empty_ not
```

Xitian9:
It might also be worth rethinking some of the naming here. `ReportOpts` has a trailing underscore, `ReportSpec` prefixed rs and uses camelCase, and others do some combination of these. I admit I haven't contributed to consistency with my choices here, so I'm happy to change them once we decide what to do.

simonmichael:
Hi @Xitian9.. I just had another look at your latest, where you have implemented the new ReportSpec type. I don't have a super strong opinion on it. ReportOpts still has some dependency on the current date, but less so now. Overall this seems to clean up code a bit. Do you feel it's a good improvement overall, justifying the new type ? If so I'm ok with merging the work so far. We would just need a haddock describing the purpose of ReportSpec. Let's discuss lens later just to limit the scope.

simonmichael:
PS I'm going to install this now and try to exercise it a bit.

Xitian9:
Sure thing. Let me know your thoughts.

simonmichael:
I haven't noticed any problems, I'm ready to merge this but could we have a haddock describing the purpose of ReportSpec ?

Xitian9:
Documentation added and rebased.

simonmichael:
Ok, thank you very much!

simonmichael:
Found a regression in hledger-ui due to fcbe511d (ui,web: Adopt new ReportOptions interface, store literal query string in UIOpts): pressing number keys changes the depth: indicator but the displayed accounts don't change.

simonmichael:
PS: `make ghci-ui` & :reload will be handy, if you didn't already know it.

Xitian9:
This seems to be part of a larger suite of regressions. The same should happen when toggling Unmarked/Pending/Cleared, when filtering by real or not, or anything else should affect both ReportOpts and the query of ReportSpec. Currently it seems we're only updating the `Query` when `querystring_` is updated. I'll look to fix that.

simonmichael:
Fixed by #1383.

Issue #1067, opened 2019-07-15
Cleanup makefile for shell completions


alerque:
#1065

schoettl:
On 3.:
If we use templates for the targets, then we have to define a variable in the Makefile that lists the hledger sub-commands, right? That is, hard-coding hledger sub-commands in the Makefile. Currently, they are generated.

schoettl:
On 1.:
You mean `make -C shell-completion/ <target>` in the root Makefile? Or do you plan to integrate the content of shell-completion/Makefile into the root Makefile?

On 4.:
What do you mean with "output option"? Why not just the (target) file `hledger-completion.zsh`?

alerque:
On 1.: Probably the latter, recursive make calls are notoriously problematic and you loose a lot of the build dependency tree it would otherwise already know about, introducing more delays.

Re 3.: No I would never hard code the list of sub-commands into a Makefile. That can be determined at runtime based on the output of `hledger` as it does now, and new make targets generated for each one.

Re 4.: That's what I  meant. Sorry about the bad wording there.

schoettl:
Re 1.: Hm, I still would prefer keeping the separate Makefile for maintenance reasons. Because what dependency would you lose? The final completion target only depend on the hledger binary and that is usually built anyway, on a package or install build. And even if not – a few seconds more but much better maintainability (and no thousand-line Makefile).

Re 2.: How can you generate make targets dynamically (using templates)? With a placeholder file for each sub-command?

alerque:
Re 1: Doing it that way you would loose the ability to understand whether it needed rebuilding in without invoking a bunch of extra processes and loosing things like path variables that might be useful, especially for installation (not currently being done at all). If maintainability is what you are after I suggest an `include` is more appropriate.

Re 2: Here is [an example of templates targets](https://www.gnu.org/software/make/manual/html_node/Eval-Function.html). The use case for this would be even simpler than that example.

If you are unfamiliar with these issues in `make` I would kindly request to hold the objections until you can make in light of a PR where I (or somebody) actually demonstrate how it works. This is just an issue report and to-do list for some adventurous coder, not a PR, and if/when I get a chance to actually write the code I feel pretty confident I can demonstrate that it works, and even works well with fewer external dependencies and process overhead slowing the whole thing down. If not of course it will be a moot point.

schoettl:
Thanks for pointing me to `include` and `$(shell ...)` for variable assignment.

simonmichael:
[...]

PR #1343, opened 2020-09-05
Update shell completion for current master

schoettl:
I updated the shell completions for new options and new commands in version 1.19.

I also contacted the Arch packager if he can add the completion in the installation script.

### DIFF
--- a/shell-completion/hledger-completion.bash
+++ b/shell-completion/hledger-completion.bash
@@ -119,6 +119,7 @@ check-dupes
 close
 diff
 rewrite
+aregister
 balancesheet
 balancesheetequity
 cashflow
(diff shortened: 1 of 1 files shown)


alerque:
I've had [a bug report open](https://bugs.archlinux.org/task/63197?project=5&string=hledger) about that on the Arch package for a year. I've tried contacting the updaters in IRC as well. For some reason it keeps getting ignored even as version bumps eventually go through.

schoettl:
Oh, I also had an bug report. The reason for closing/deferring was:

> The completion script is not in released source code which we package.

- https://bugs.archlinux.org/task/61546

As Simon commented [here](https://github.com/simonmichael/hledger/issues/1067#issuecomment-687632830),
we probably have to update the package.yaml to include the completion script.

simonmichael:
Thanks @schoettl !

PR #1344, opened 2020-09-05
web: Put session file in $XDG_DATA_DIR.

legrostdg:
https://github.com/simonmichael/hledger/pull/1228
#1081 

It would be more correct to put the file in XDG_RUNTIME_DIR, but it is not supported by `directory`. But XDG_DATA_DIR seems OK, too. `directory` handles platforms other than Linux.

### DIFF
--- a/hledger-web/Hledger/Web/Foundation.hs
+++ b/hledger-web/Hledger/Web/Foundation.hs
@@ -31,7 +31,9 @@ import Data.Time.Calendar (Day)
 import Network.HTTP.Conduit (Manager)
 import Network.HTTP.Types (status403)
 import Network.Wai (requestHeaders)
-import System.FilePath (takeFileName)
+import System.Directory (XdgDirectory (..), createDirectoryIfMissing,
+                         getXdgDirectory)
+import System.FilePath (takeFileName, (</>))
 import Text.Blaze (Markup)
 import Text.Hamlet (hamletFile)
 import Yesod
@@ -100,9 +102,11 @@ type Form x = Html -> MForm (HandlerT App IO) (FormResult x, Widget)
 instance Yesod App where
   approot = ApprootMaster $ appRoot . settings
 
-  makeSessionBackend _ =
+  makeSessionBackend _ = do
+    hledgerdata <- getXdgDirectory XdgData "hledger"
+    createDirectoryIfMissing True hledgerdata
     let sessionexpirysecs = 120
-    in  Just <$> defaultClientSessionBackend sessionexpirysecs ".hledger-web_client_session_key.aes"
+    Just <$> defaultClientSessionBackend sessionexpirysecs (hledgerdata </> "hledger-web_client_session_key.aes")
 
   -- defaultLayout :: WidgetFor site () -> HandlerFor site Html
   defaultLayout widget = do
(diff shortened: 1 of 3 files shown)


hledger-web/Hledger/Web/Foundation.hs
```diff
@@ -31,7 +31,9 @@ import Data.Time.Calendar (Day)
 import Network.HTTP.Conduit (Manager)
 import Network.HTTP.Types (status403)
 import Network.Wai (requestHeaders)
-import System.FilePath (takeFileName)
+import System.Directory (XdgDirectory (..), createDirectoryIfMissing,
+                         getXdgDirectory)
```
simonmichael, line 35: Can you add a `>=1.2.3.0` bound to directory in hledger-web/package.yaml.

hledger-web/Hledger/Web/Foundation.hs
```diff
@@ -100,9 +102,11 @@ type Form x = Html -> MForm (HandlerT App IO) (FormResult x, Widget)
 instance Yesod App where
   approot = ApprootMaster $ appRoot . settings
 
-  makeSessionBackend _ =
+  makeSessionBackend _ = do
+    hledgerdata <- getXdgDirectory XdgData "hledger"
```
simonmichael, line 106: For this runtime-generated file, is XdgData (`~/.local/share`, `C:/Users/<user>/AppData/Roaming`) the best place, or is XdgCache (`~/.cache`, `C:/Users/<user>/AppData/Local`) ? (I see XDG_RUNTIME_DIR is not yet supported, a pity.)

hledger-web/Hledger/Web/Foundation.hs
```diff
@@ -100,9 +102,11 @@ type Form x = Html -> MForm (HandlerT App IO) (FormResult x, Widget)
...
+    createDirectoryIfMissing True hledgerdata
     let sessionexpirysecs = 120
-    in  Just <$> defaultClientSessionBackend sessionexpirysecs ".hledger-web_client_session_key.aes"
+    Just <$> defaultClientSessionBackend sessionexpirysecs (hledgerdata </> "hledger-web_client_session_key.aes")
```
simonmichael, line 109: I was wondering if we should add a process ID to that filename but it sounds unnecessary (#1228: *"It won't make it so the sessions will get messed up or anything like that, it's just the service (single or multiple instances) will use the same key to decrypt session for all instances, and that's ok... unless you have some unusual scenario going on"*)

simonmichael (changes requested):
Thanks @legrostdg.

hledger-web/Hledger/Web/Foundation.hs
legrostdg: done. I also updated the .cabal file.

hledger-web/Hledger/Web/Foundation.hs
legrostdg: You're right. Updated.

hledger-web/Hledger/Web/Foundation.hs
legrostdg: OK.

simonmichael:
I tend to not update the .cabal file in the first commit, as it always causes annoying conflicts, maybe just omit that file. (I commit them later/periodically with ./Shake.hs cabalfiles.)

legrostdg:
Done.

simonmichael:
Thanks!

PR #1347, opened 2020-09-06
;ci: mac: update gha mac CI

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/60579

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -144,7 +144,7 @@ jobs:
         stack: ${{ matrix.plan.stack }}
       run: |
         brew install shelltestrunner
-        shelltestrunner --version
+        $shelltestrunner --version
 
     - name: Test functional tests (excluding addons)
       env:
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -26,6 +26,10 @@ on:
       - '!**.info'
       - '!**.txt'
 
+  pull_request:
+    paths:
+      - '.github/workflows/mac.yml'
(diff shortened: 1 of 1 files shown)


simonmichael:
@chenrui333, thanks for the PR!

What's the goal of these changes ? It seems to be to make hledger's mac CI more like homebrew's CI, and to run it more often. But hledger's CI has some different goals, such as test each version of GHC somewhere (8.6 happens to be the one tested on mac right now), avoid unnecessary building, maintain similarity of workflows across the platforms.

chenrui333:
I see, yeah, my goal was to keep the build expectation consistent with the homebrew one (so that I can always cross reference), but it sounds like the different workflows are testing against different ghc versions?

chenrui333:
happy to close the PR. I am curious why dont use use matrix config to setup the workflow?

simonmichael:
Yeah, instead of building with all ghc versions on one or more workflows I have ensured each of them uses a different ghc which gives roughly the same assurance with less carbon footprint.

Matrix is used a bit; do you mean why not use one matrix/workflow covering all three platforms ? I separated them to focus on one platform at a time and allow whatever custom scripting was needed for each. Perhaps a single platform-adaptive workflow could make sense now.

chenrui333:
> Perhaps a single platform-adaptive workflow could make sense now.

yeah, that is actually what I was suggesting, maybe one workflow with multiple ghc versions.

simonmichael:
Closing this for now, thanks.

Issue #1345, opened 2020-09-05
Negative amounts not aligned


markokocic:
Looks like negative amounts in reports are not anymore aligned with positive amounts. I am at the latest git version at the moment.

Steps to reproduce:
```
~/tmp $ cat test 
2020/09/01 I
  assets:a  1
  assets:a:d  1
  assets:c  1
  income

2020/09/01 A
  expenses:a  1
  assets:b

~/tmp $ hledger bs -f test 
Balance Sheet 2020-09-01

             || 2020-09-01 
=============++============
 Assets      ||            
-------------++------------
 assets:a    ||          1 
 assets:a:d  ||          1 
 assets:b    || -1 
 assets:c    ||          1 
-------------++------------
             ||          2 
=============++============
 Liabilities ||            
-------------++------------
-------------++------------
             ||            
=============++============
 Net:        ||          2 
~/tmp $ hledger bal -f test 
                   1  assets:a
                   1  assets:a:d
  -1  assets:b
                   1  assets:c
                   1  expenses:a
  -3  income
--------------------
                   0
~/tmp $ 

```

Previous and expected behaviour is that both negative and positive amounts are right aligned.

simonmichael:
Yikes! Thank you, good catch.

simonmichael:
@Xitian9, the new stripAnsi doesn't remove this ANSI:
```
ghci> stripAnsi $ showAmountWithoutPrice True (-1)
-- "\ESC[31m-1\ESC[m"
```
I don't quite see why not, do you ?

simonmichael:
Fixed by 1.19.1.

PR #1348, opened 2020-09-07
lib: Correctly strip ansi sequences with no numbers/semicolons.

Xitian9:
Replaced an incorrect use of `takeWhile1P` with `takeWhileP`. Addresses #1345.


simonmichael:
Sweet, thank you.

Issue #1352, opened 2020-09-11
balance --format in color mode can over-truncate or print corrupted ANSI codes


Xitian9:
Call `hledger balance --format "%.7(total)"` on the following journal file.
```
2020-09-01 Test
    a  -500 AAA
    b
```
The result is:
```
..[m
500 AAA
-------
0
```

The problem is that `takeWidth` is not aware of ansi escape sequences, and so will mangle any which are there.

I think perhaps maximum width should be ignored when formatting amounts. I can't think of an example where you'd want to clip an amount, and this is what the fancier mixedAmount ellision functions are for.

simonmichael:
Ouch, another one. Good catch. 1.18 does this which was more correct I assume:
```
$ hledger-1.15 -f a.j bal -N --format "%.7(total)"
..AA
500 AAA
```

simonmichael:
I suppose dropping the --format feature is out of the question...

simonmichael:
1.18's output doesn't look very correct either. Here's what it's supposed to be when there are no ANSI codes messing things up (the amount is elided to a maximum of 7 characters):
```
$ hledger bal -N a --format "%.7(total)" --color=never
..0 AAA
```

Xitian9:
Yes, I don't think this was ever handled correctly, as it would take some fancy logic to handle it correctly, and it would be strange for that to be removed.

Handling this correctling would involve:
1. Stripping ansi escape codes
2. Clipping the resulting ouptut
3. Re-inserting ansi escape codes in the appropriate places (which is tricky to even define)

Alternately, you could teach `takeWidth` to handle ansi escape codes, but it would have to know how to recognise them forwards and backwards.

Both of these options seem like a lot of work for something of questionable value in the first place. I propose that maximum widths on fields containing `MixedAmount` are not honoured.

Xitian9:
Version 1.18.1 will also give you ansi escape garbage.

```
$ hledger balance -N a --format "%.4(total)"
..[m
```

Xitian9:
Options I can think of:
1. Do not honour maximum width on `%(total)`
2. Calling maximum width on `%(total)` turns off colour
3. Document a warning that using a maximum width for `%(total)` may do bad things, and is not recommended unless you turn off colour

simonmichael:
2 sounds good to me (--format always disables colour), with a note in the docs.

Xitian9:
[...]

Issue #1350, opened 2020-09-10
performance regression in 1.19.1 ?


simonmichael:
I see these kind of numbers with 1.19.1 - rough measurements, but suggesting a 3x slowdown:
```
$ BENCHEXES=hledger-1.18,hledger-1.19,hledger-1.19.1 make bench

run quick performance benchmarks in bench.sh (approximate, can be skewed):
quickbench -w hledger-1.18,hledger-1.19,hledger-1.19.1  # -v
Running 12 tests 1 times with 3 executables at 2020-09-09 17:26:16 PDT:

Best times:
+-----------------------------------------------++--------------+--------------+----------------+
|                                               || hledger-1.18 | hledger-1.19 | hledger-1.19.1 |
+===============================================++==============+==============+================+
| -f examples/100x100x10.journal print          ||         0.13 |         0.11 |           0.17 |
| -f examples/1000x1000x10.journal print        ||         0.39 |         0.27 |           1.08 |
| -f examples/10000x1000x10.journal print       ||         3.43 |         2.69 |           9.72 |
| -f examples/10000x1000x10.journal print ff    ||         1.42 |         1.48 |           4.73 |
| -f examples/100x100x10.journal register       ||         0.10 |         0.09 |           0.16 |
| -f examples/1000x1000x10.journal register     ||         0.42 |         0.42 |           1.10 |
| -f examples/10000x1000x10.journal register    ||         4.97 |         3.57 |          11.45 |
| -f examples/10000x1000x10.journal register ff ||         1.75 |         1.39 |           5.48 |
| -f examples/100x100x10.journal balance        ||         0.08 |         0.07 |           0.13 |
| -f examples/1000x1000x10.journal balance      ||         0.31 |         0.30 |           0.89 |
| -f examples/10000x1000x10.journal balance     ||         1.99 |         1.59 |           5.95 |
| -f examples/10000x1000x10.journal balance ff  ||         1.48 |         1.24 |           5.03 |
+-----------------------------------------------++--------------+--------------+----------------+

```

simonmichael:
It's harder to see the difference in the profiles, but notice the increased memory allocation and the appearance of stripAnsi as top cost center. It seems likely that https://github.com/simonmichael/hledger/commit/600dab3976cabfcaa4dfaedc67ebc6fd99d2a340 was a pessimisation.

[...]

PR #1354, opened 2020-09-14
Write some smarter showMixedAmount functions.

Xitian9:
showMixedAmount* functions are now more aware of their actual width, and safer to colourise. This allows us to stop stripping ansi escape sequences when calculating `strWidth`, addressing #1350. It also sets the stage to address #1352, though there are a lot of conflicts with PR #1340, so I want to get that merged before addressing this.


Xitian9:
Now addresses #1352 as well.

Xitian9:
I'm a bit unsatisfied with the final API for `showMixed*`. We have 12 different functions which display `MixedAmount` with various different options. Prior to this PR there are 9 such functions, so the situation was better but still not great.

Perhaps a better option is to create a data type `ShowMixedOpts` which contains the different display options.
- Uses colour?
- Single line or multi-line?
- Minimum width?
- Maximum width?
- Normalised or unnormalised?
- Show zero commodity?
- With or without price?
- With Precision? (This option is only used in one place, so might be dropped)

We could create a `ShowMixedOpts` for each of the common use cases, and replace e.g. `showMixedAmountOneLineWithoutPrice` with `showMixed oneLineNoPrice`, or something like that. Old functions could deprecated or removed entirely.

Thoughts?

hledger-lib/Hledger/Utils/String.hs
```diff
@@ -338,17 +336,7 @@ takeWidth w (c:cs) | cw <= w   = c:takeWidth (w-cw) cs
...
--- | Strip ANSI escape sequences from a string.
---
--- >>> stripAnsi "\ESC[31m-1\ESC[m"
--- "-1"
-stripAnsi :: String -> String
-stripAnsi s = either err id $ regexReplace ansire "" s
- where
-   err    = error "stripAnsi: invalid replacement pattern"      -- PARTIAL, shouldn't happen
-   ansire = toRegex' "\ESC\\[([0-9]+;)*([0-9]+)?[ABCDHJKfmsu]"  -- PARTIAL, should succeed
+strWidth = maximum . (0:) . map (foldr (\a b -> charWidth a + b) 0) . lines
```
simonmichael, line 339: Utils is sort of a grab bag of helpers that may come in handy - I don't mind having a few things in there that aren't currently being used. Maybe it's worth keeping stripAnsi and a strWidthAnsi in there for now ?

simonmichael (commented):
Belatedly tackling this, @Xitian9. There's a lot here so I figure I'll start with the first three commits. I think these three cause no user-visible changes yet, except for possibly a rendering speedup, right ?

simonmichael:
Ps and we can remove ansi from strWidth's haddock.

hledger-lib/Hledger/Utils/String.hs
Xitian9: Fine by me. Should I amend the commit to add it back in, or re-add it in a new commit on the end?

Xitian9:
> There's a lot here so I figure I'll start with the first three commits. I think these three cause no user-visible changes yet, except for possibly a rendering speedup, right ?

Correct. At this point there is not even a rendering speedup, because the new interface is just introduced but not yet used in any of the report code.

simonmichael:
Yes please, alter the commit and repush all.

simonmichael:
Looking ahead.. what are the user visible changes from the next three commits ? ("lib: renderTable can now receive alignment and width specifications on all cells, and has an option to display the outer border." .. "cli: Ensure register reports are correctly aligned for negative numbers.")

Xitian9:
Oh, there is one user-visible change from the first commit. Previously `showMixedAmountElided` would show two amounts and then the elision string if necessary. Now it will display as many Amounts as it can subject to the condition that the amounts plus the elision string fit within 22 characters.

Xitian9:
- _lib: renderTable can now receive alignment and width specifications on all cells, and has an option to display the outer border._
This gives `renderTable` a little more customisation. Before any of the commits of this PR, `render` would just receive a string to display in each cell. After the second commit of this PR it would also receive a width of the string (in place of stripping ANSI sequences and then calculating the width). After this commit, it now also takes an alignment, so you can make cells left or right aligned. The function `render` calls `renderTable` with appropriate options to give the same behaviour as before.
Also, previously `render` would always put a border around the table. We would take this output, and would sometimes strip the border by dropping the first and last rows, and first and last characters of every row. I've just added an option to control whether to put the border in, so we can just not add it in the first place, rather than stripping it later. Note that this is again just defining helper functions; this extra power is not yet used anywhere.

- _lib: BudgetReport uses new renderTable inteface, now has more compact output._
This changes `showMixedAmountElided` so that the width to elide to is given as an argument, rather than fixed at 22 characters.
This actually uses the new `renderTable` interface. Mostly this is just an internal change, but since we have more information about the widths of things, we can actually get rid of some superfluous spaces in the budget report output, previously there to make sure it stayed aligned with the largest reasonable contents.

- _cli: Ensure register reports are correctly aligned for negative numbers._
This uses the new `showMixed*` function for the register report. This fixes some misaligned negative numbers which appeared in one of the earlier commits, and adds a test for it.

- _lib,cli,ui: Introduce showMixed*Unnormalised_
This introduces some new helper functions which are exactly the same as what we had before, but do not call `normaliseMixedAmountSquashPricesForDisplay`, so that we can use the new functions for displaying `Transaction` and `Posting`.  It also goes through and gets rid of most uses of the old `showMixed*` functions which would benefit from using the new interface.

simonmichael:
Thanks. PS while you're there I noticed some inconsistent spelling of elide/elision (ellision) in later commits - I think it should be just one L.

Xitian9:
Haha, at some point I noticed my mis-spelling and went back to fix it. I guess I missed some.

Xitian9:
Changes made. Let me know if you have other questions or requests.

simonmichael:
> (change notes)

Great - I'll save these in the commit message(s) when merging, unless you do it first.

Re your [question](https://github.com/simonmichael/hledger/pull/1354#issuecomment-718307600): you might be right - I don't have a strong opinion yet. Once we have the desired behaviour, I don't mind leaving internals a bit over-complex until a better design becomes clear. Here, I would probably first just add & validate the new helpers, then discuss removal of old API (exported functions) as a follow-on task.

simonmichael:
Merged, thank you very much!

I'll be pushing updated changelogs shortly.

simonmichael:
Sorry, after finding some issues/questions I am quickly unmerging this so we can test more, without pressure. Here's one:
```
$ hledger-pre1354 -f examples/sample.journal  bal -Y
Balance changes in 2008:

                    || 2008 
====================++======
 assets:bank:saving ||   $1 
 assets:cash        ||  $-2 
 expenses:food      ||   $1 
 expenses:supplies  ||   $1 
 income:gifts       ||  $-1 
 income:salary      ||  $-1 
 liabilities:debts  ||   $1 
--------------------++------
                    ||    0 
$ hledger -f examples/sample.journal  bal -Y
Balance changes in 2008:

                    || 2008 
====================++======
 assets:bank:saving ||   $1 
 assets:cash        || $-2 
 expenses:food      ||   $1 
 expenses:supplies  ||   $1 
 income:gifts       || $-1 
 income:salary      || $-1 
 liabilities:debts  ||   $1 
--------------------++------
                    ||    0 
```

simonmichael:
While we're tweaking this more, perhaps you can also save those extra change notes in the relevant commit messages - it'll be nicer than adding them to the merge commit.

simonmichael:
Here's an example of the new multicommodity amount eliding behaviour:
```
$ hledger-pre1354 -f examples/bcexample.hledger bs -1
Balance Sheet 2014-10-11

             ||                      2014-10-11 
=============++=================================
 Assets      ||                                 
-------------++---------------------------------
 Assets      || 70.00 GLD, 17.00 ITOT, 6 more.. 
-------------++---------------------------------
             || 70.00 GLD, 17.00 ITOT, 6 more.. 
=============++=================================
 Liabilities ||                                 
-------------++---------------------------------
 Liabilities ||                     2891.85 USD 
-------------++---------------------------------
             ||                     2891.85 USD 
=============++=================================
 Net:        || 70.00 GLD, 17.00 ITOT, 6 more.. 
$ hledger -f examples/bcexample.hledger bs -1
Balance Sheet 2014-10-11

             ||          2014-10-11 
=============++=====================
 Assets      ||                     
-------------++---------------------
 Assets      || 70.00 GLD, 7 more.. 
-------------++---------------------
             || 70.00 GLD, 7 more.. 
=============++=====================
 Liabilities ||                     
-------------++---------------------
 Liabilities ||         2891.85 USD 
-------------++---------------------
             ||         2891.85 USD 
=============++=====================
 Net:        || 70.00 GLD, 7 more.. 
```
The old eliding was already bothering some people.. now we are eliding harder. On the upside multicolumn reports will be more compact and more likely to fit the window by default. Maybe we'll think of better UX later.

Xitian9:
The choice of 22 characters for the default elision width was pretty arbitrary. I recall doing some back-of-the-envelope calculation to figure out what size two amounts were likely to fit in. Your example there would fit in 31 characters; I could set the width to be 32? Thoughts?

Also, we can save a character on all elision by replacing `..` with `…`, at the cost of using a unicode character.

simonmichael:
> Your example there would fit in 31 characters; I could set the width to be 32? Thoughts?

Sure, why don't we try that for the moment. It'll reduce UI churn, and showing at least two commodities helps communicate that it's a multicommodity amount.

> Also, we can save a character on all elision by replacing .. with …, at the cost of using a unicode character.

Perhaps a nice option later, but I'm inclined to stick with our default ascii output to reduce headaches.

Xitian9:
I've fixed the bug and changed the default width, as well as altering the commit messages. It looks like this PR remains closed. Should I start a new one?

simonmichael:
No need for a new PR. Thanks, I'll test soon.

simonmichael:
What is the new rule for when multicommodity amounts will get elided ? 

Here are some more cases to look at:

Totals are aligned differently in `hledger.1354 -f examples/bcexample.hledger bal -t`.

Paypal balances are aligned differently here (sorry, no shareable repro yet):
```
$ hledger.1354 bal -Y banking:paypal banking:stripe
Balance changes in 2020:

                            ||                             2020 
============================++==================================
 JS:expenses:banking:paypal || $ 14.51, C$ 1.44, € 0.79 
 JS:expenses:banking:stripe ||                           $ 8.18 
----------------------------++----------------------------------
                            || $ 22.69, C$ 1.44, € 0.79 
```

Xitian9:
The new rule (not how it's implemented, but essentially):
Render the multicommodity amount as usual:
`1.00 USD, 1.00 GBP, 1.00 AUD, 1.00 CAD, 1.00 MEP` 
Check whether it fits within 32 characters. In this case, it is 48 characters, so it does not.
Progressively elide from the right until it fits within 32 characters.
`1.00 USD, 1.00 GBP, 1.00 AUD, 1.00 CAD, 1 more..` is also 48 characters, so it still doesn't fit.
`1.00 USD, 1.00 GBP, 1.00 AUD, 2 more..`  is 38 characters, so it still doesn't fit.
`1.00 USD, 1.00 GBP, 3 more..` is 28 characters, so it now fits.
Result is `1.00 USD, 1.00 GBP, 3 more..`.

simonmichael:
A repro for the paypal/stripe case above:
```
2020-02-22
    26018c6e:ced6cffd:c3c182f1:7b433efb          $ 9.41
    ea50865f:3bfb86b7:bf72f75a:a7cad1ac        $ -10.00
    ea50865f:325566ed:216fec7e:7b433efb          $ 0.59

2020-02-27
    26018c6e:ced6cffd:c3c182f1:7b433efb        £ -19.16
    ea50865f:325566ed:216fec7e:7b433efb          £ 0.91
    ea50865f:325566ed:47134948                  £ 18.25

2020-04-12
    26018c6e:ced6cffd:c3c182f1:7b433efb          € 9.21
    ea50865f:3bfb86b7:bf72f75a:a7cad1ac        € -10.00
    ea50865f:325566ed:216fec7e:7b433efb          € 0.79

2020-05-16
    26018c6e:ced6cffd:c3c182f1:7b433efb        C$ 24.56
    ea50865f:3bfb86b7:bf72f75a:a7cad1ac       C$ -26.00
    ea50865f:325566ed:216fec7e:7b433efb         C$ 1.44

```
```
$ hledger.1354 -f a.j bal -Y 
Balance changes in 2020:

                                     ||                            2020 
=====================================++=================================
 26018c6e:ced6cffd:c3c182f1:7b433efb ||      $ 9.41, C$ 24.56, 2 more.. 
 ea50865f:325566ed:216fec7e:7b433efb || $ 0.59, C$ 1.44, € 0.79 
 ea50865f:325566ed:47134948          ||                         £ 18.25 
 ea50865f:3bfb86b7:bf72f75a:a7cad1ac ||   $ -10.00, C$ -26.00, € -10.00 
-------------------------------------++---------------------------------
                                     ||                               0 
```

Xitian9:
I'll look into both these cases. I think one or both of them should be implemented as functests; there's a bit of a dearth of examples with multiple commodities and negative amounts and colour, etc.

simonmichael:
Indeed, we have a lot of tests but they seem not to have kept up with our layout functionality.

simonmichael:
PS, related: organising the tests so that

- there is an obvious place to put/find each one
- their coverage is clear
- excessive duplication/slowness is avoided

has become a bit challenging. Perhaps with some thought and cleanup we could improve this.

Xitian9:
The process of tackling these bugs has opened up an interesting opportunity.

(One of) the problems is that code calling `concat(Top|Bottom)Aligned` is not aware of ansi sequences anymore, so these functions need to be rewritten to support them. This has significant code duplication with the `Text.Tabular.AsciiWide` stuff, so I'm looking at implementing them as just some sort of wrapper around `renderTable`. To do this, I've adapted the `Text.Tabular.AsciiWide` code to allow cells that have multiple lines, and are top or bottom-aligned amongst themselves.

It's possible to just implement this and use it only in the `concat(Top|Bottom)Aligned` code, but it also opens the possibility of allowing our balance reports to display multi-commodity amounts on separate lines. This could make elision less necessary. I'll continue as is for now, but I'd be interested in hearing your thoughts on this possibility.

simonmichael:
No strong opinion, just some general thoughts: 

- Amounts on separate lines doesn't seem urgent to me, in Ledger & our basic balance report they have always confused newcomers.
- We like refactorings giving us cleaner code, more power and new possibilities at reasonable cost. We should be careful about over-extending and getting mired in costly long-running reworks without a real need.
- Personally I like a balance between forward-looking internal work, and externally-visible high-immediate-user-value work. 
- All else being equal, simple is good and less is more...

simonmichael:
To be clear I'm not disagreeing with your ideas here. My bandwidth for internal work is a bit limited between now and our next major release on dec 1, so if this gets too involved I may not be much help. 

Might some of this make sense to move to the tabular library ? That could also help.

Xitian9:
I've pushed some changes which resolve the two issues you raised earlier. Let me know what you think.

Xitian9:
> Might some of this make sense to move to the tabular library ? That could also help.

I think it might, but that would require them deciding to handle wide characters, where they don't currently. I'm not sure how amenable they are to that, but it's worth asking I guess.

simonmichael:
The changes are looking good! Quick benchmarks:
```
$ BENCHEXES=hledger-1.19,hledger.pre1354,hledger.1354 make bench

quickbench -w hledger-1.19,hledger.pre1354,hledger.1354  # -v
Running 5 tests 1 times with 3 executables at 2020-11-03 17:18:42 HST:

Best times:
+----------------------------------------------------++--------------+-----------------+--------------+
|                                                    || hledger-1.19 | hledger.pre1354 | hledger.1354 |
+====================================================++==============+=================+==============+
| -f examples/10000x1000x10.journal print            ||         2.07 |            2.46 |         1.43 |
| -f examples/10000x1000x10.journal register         ||         2.91 |            3.21 |         1.75 |
| -f examples/10000x1000x10.journal balance          ||         1.09 |            1.05 |         1.31 |
| -f examples/1000x1000x10.journal balance --weekly  ||         1.92 |            2.10 |         1.14 |
| -f examples/10000x1000x10.journal balance --weekly ||        18.73 |           20.40 |        15.70 |
+----------------------------------------------------++--------------+-----------------+--------------+
```
I'll merge today or tomorrow if all seems well.

Xitian9:
I've fixed the typo and improved the commit message for allowing multiline cells in tables.

simonmichael:
Merged to master. Thanks!

simonmichael:
Here's what I have for the hledger package's changelog (user-visible changes):
```
- Eliding of multicommodity amounts now makes better use of available space. (Stephen Morgan)

- Amount eliding no longer displays corrupted ANSI codes (Stephen Morgan)

- Console reports are rendered more efficiently, making many of them
  about 30% faster (Stephen Morgan)

- bal: --budget reports no longer insert an extra space inside the brackets (Stephen Morgan)
```

simonmichael:
And for the hledger-lib changelog:
```
- Implement concat(Top|Bottom)Padded in terms of renderRow, allowing them to be width aware. (Stephen Morgan)

- Expand Tabular.AsciiWide to allow multiline, custom-width,
  vertically/horizontally-aligned cells, and optional table borders.
  (Stephen Morgan)

- Introduce showMixed*Unnormalised, eliminate most direct calls of strWidth. (Stephen Morgan)

- showMixedAmountElided now makes better use of space, showing as many
  Amounts possible as long as they and the elision string fit within
  32 characters.
```

simonmichael:
I called your nice speedup "about 30%" but maybe that was over-optimistic, here's another benchmark:
```
$ LEDGER_FILE=examples/1000x1000x10.journal  quickbench -n2 -N2 -w hledger-1.19,hledger 'hledger bal -M'  'hledger bal -W'  'hledger bal -D' 
Running 3 tests 2 times with 2 executables at 2020-11-04 08:50:15 HST:

Best times 1:
+--------++--------------+---------+
|        || hledger-1.19 | hledger |
+========++==============+=========+
| bal -M ||         0.45 |    0.39 |
| bal -W ||         1.82 |    1.19 |
| bal -D ||        11.21 |   10.26 |
+--------++--------------+---------+

Best times 2:
+--------++--------------+---------+
|        || hledger-1.19 | hledger |
+========++==============+=========+
| bal -M ||         0.43 |    0.49 |
| bal -W ||         1.88 |    1.15 |
| bal -D ||        11.86 |   11.57 |
+--------++--------------+---------+
```

Anyway, inspired by this and a String-vs-text discussion on #haskell yesterday, just a thought: do you have a sense of how String-based (vs text-based) is all this rendering/table code you've been working on ? When I converted hledger from String to text, memory usage went down by about 30% (not sure about speed). As usual this is less than the drastic improvement always claimed by haskellers for text-vs-string.. but I know I must have missed some parts.

Xitian9:
Currently it is entirely String-based, but that was an intentional choice. I wanted to get the functionality out of the way with minimal other changes before moving on to performance aspects. I've actually got another branch I'm working on (https://github.com/Xitian9/hledger/tree/reporttext) with the goal of switching the data structures to strict Text and the report code to Lazy Text Builder, to see what kind of speedups and improvements we can get. Now that this PR has been merged, it needs a rebase, but is otherwise mostly current.

I'm always a bit hesitant about benchmarks, because I'm never sure if stack has given me optimised builds (it seems to give unoptimised builds after `make functest`, and doesn't necessarily optimise them if you try to rebuild later), but it looks like it's giving me significant speedups on long register reports (1s versus 3s).

simonmichael:
No hurry on this, but good news that you're looking into it! For the record, that #haskell String-vs-text discussion starts [here](https://ircbrowse.tomsmeding.com/browse/haskell?id=104553&timestamp=1604439506#t1604439506). 

You're right that make functest builds in --fast mode. I have the idea that it will use an optimised build if it's up to date (make `stack build hledger` first).

Issue #1364, opened 2020-10-02
Query fails when account name has a special character preceded by a space


Fornost461:
#### Journal
```
2020-09-19 bus tickets
    assets:bank           -5
    assets:a et            5

2020-09-20 bus tickets
    assets:bank           -10
    assets:a ét            10
```
(The accounts `assets:a et` and `assets:a ét` do not have the same name: the second one has the special character `é`.)
#### Terminal
```
$ hledger bal 'assets:a'       # fine
                   5  assets:a et
                  10  assets:a ét
--------------------
                  15
```
```
$ hledger bal 'assets:a et'     # fine
                   5  assets:a et
--------------------
                   5
$ hledger bal 'assets:a ét'    # wrong: missing output
--------------------
                   0
```
```
$ hledger bal ' et'    # fine
                   5  assets:a et
--------------------
                   5
$ hledger bal ' ét'   # wrong: missing output (notice the space before the special character)
--------------------
                   0
```
```
$ hledger bal 'ét'    # fine    (no space before the special character)
                  10  assets:a ét
--------------------
                  10
```
I also tried with the following special character: `ê`. It seems to have the same issue.
I also tried running `register` instead of `balance`, but output is missing too as above.
```
$ echo $LANG
en_US.UTF-8
```
```
$ hledger --version
hledger 1.19.1
```
```
$ echo $SHELL
/bin/bash
```
```
$ bash --version
GNU bash, version 5.0.18(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
Platform: _Manjaro Linux_
### Related docs I found:
https://hledger.org/hledger.html#unicode-characters
https://hledger.org/hledger.html#troubleshooting

simonmichael:
Thanks for the report - fixed by #1367.

PR #1367, opened 2020-10-07
lib: quoteIfNeeded should not escape the backslashes in unicode points

Xitian9:
Addresses #1364.

The commit which broke this was 7d1e6d7d12f3c3a042661c1d6963523b5d5f676b, which changed `quoteIfNeeded` so that it actually quotes things (albeit apparently incorrectly). It previously spent several years as the identity function with nobody noticing. This raises the question: is this function actually needed at all?

Edit: Correction to the above. quoteIfNeeded was not the identity function, but it did spend several years not escaping quotes. Should it attempt to do so? Can we find an example of something which doesn't work if it doesn't escape quotes?


Xitian9:
Related to #1368.

simonmichael:
Thanks @Xitian9.. as this seems to work better, merging.

Issue #1368, opened 2020-10-07
Cannot match account names with quotes


Xitian9:
Trying to match the second transaction without matching the first seems impossible.

```
2020-09-19 Quoting
    assets:bank           -5
    assets:unquoted        5

2020-09-20 Quoting
    assets:bank           -5
    assets:"quoted"        5
```

master:
```
$ hledger -f - balance -N --flat 'quoted'
                   5  assets:"quoted"
                   5  assets:unquoted
$ hledger -f - balance -N --flat '"quoted'
hledger: this regular expression could not be compiled: \
$ hledger -f - balance -N --flat '\"quoted'
hledger: this regular expression could not be compiled: \\\
```

Following #1367 the situation changes, but is not solved:
```
$ hledger -f - balance -N 'quoted'
                   5  assets:"quoted"
                   5  assets:unquoted
$ hledger -f - balance -N '"quoted'
hledger: this regular expression could not be compiled: \
$ hledger -f - balance -N '\"quoted'
(succeeds with empty output)
```

1.18.1 (and into the past) again had different output, none of it correct.
```
$ hledger -f - balance -N 'quoted'
                   5  assets:"quoted"
                   5  assets:unquoted
$ hledger -f - balance -N '"quoted'
                   5  assets:"quoted"
                 -10  assets:bank
                   5  assets:unquoted
$ hledger -f - balance -N '\"quoted'
hledger: this regular expression could not be compiled: \\
```

There's some complicated interaction going on regarding escaping quotes and regular expressions.

PR #1376, opened 2020-10-20
Use [Text] for querystring_ instead of Text, to get around quoting problems.

Xitian9:
This is proof of concept for addessing #1368 by sidestepping quoting altogether. Queries are no longer wrapped in quotes and escaped, because they are stored term-by-term in a list, so no quoting is necessary.

I do not propose adopting this as is (at the very least there are conflicts with #1340), but to solicit feedback on the approach. The important changes are in `Query.hs` and `ReportOptions.hs`.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -20,6 +20,7 @@ module Hledger.Query (
   generatedTransactionTag,
   -- * parsing
   parseQuery,
+  parseQueryList,
   simplifyQuery,
   filterQuery,
   -- * accessors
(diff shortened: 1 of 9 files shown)


simonmichael:
To conserve brain space, I'll look at this after we have resolved some of the older PRs.

hledger-lib/Hledger/Query.hs
```diff
@@ -177,9 +178,13 @@ data QueryOpt = QueryOptInAcctOnly AccountName  -- ^ show an account register fo
 -- >>> parseQuery nulldate "\"expenses:dining out\""
 -- Right (Acct (RegexpCI "expenses:dining out"),[])
```
simonmichael, line 179: doctest needs update I think

simonmichael: or removal

simonmichael (changes requested):
Looks good to me - I'll double check when rebased.

hledger-lib/Hledger/Query.hs
Xitian9: I agree that these doctests don't make a lot of sense for `parseQueryList`, but I think as long as we're leaving `parseQuery` in we should be sure it's dealing with quoting issues correctly. On the other hand, if you're in favour of just removing `parseQuery` altogether (and renaming `parseQueryList`), then we could get rid of these.

Xitian9:
Rebased

simonmichael:
Sorry, I misread the diff. Looks good, thanks!

Issue #1353, opened 2020-09-12
bs -V --change doesn't report changes in investment values


chvp:
hledger version: 1.19.1 (on NixOS)

minimal example:
```
P 2020-01-01 AAPL € 5

2020-01-01 Investment
    assets:cash  € -5
    assets:investment   1 AAPL  @@ € 5

P 2020-01-02 AAPL € 10
P 2020-01-03 AAPL € 7.5
```
`hledger bs -f example.journal --daily -V -e 2020-01-04` gives the following:
```
Balance Sheet 2020-01-01..2020-01-03, valued at period ends

                   || 2020-01-01  2020-01-02  2020-01-03 
===================++====================================
 Assets            ||                                    
-------------------++------------------------------------
 assets:cash       ||     € -5.0      € -5.0      € -5.0 
 assets:investment ||      € 5.0      € 10.0       € 7.5 
-------------------++------------------------------------
                   ||          0       € 5.0       € 2.5 
===================++====================================
 Liabilities       ||                                    
-------------------++------------------------------------
-------------------++------------------------------------
                   ||                                    
===================++====================================
 Net:              ||          0       € 5.0       € 2.5 
```
`hledger bs -f example.journal --daily -V --change -e 2020-01-04 --debug` gives
```
Balance Sheet 2020-01-01..2020-01-03 (Balance Changes), valued at period ends

valuation date: 2020-01-01
default valuation commodity for €: Nothing
valuation date: 2020-01-01
default valuation commodity for AAPL: Just "\8364"
market price for AAPL -> €: 5.00000000
                   || 2020-01-01  2020-01-02  2020-01-03 
===================++====================================
 Assets            ||                                    
-------------------++------------------------------------
 assets:cash       ||     € -5.0           0           0 
 assets:investment ||      € 5.0           0           0 
-------------------++------------------------------------
                   ||          0           0           0 
===================++====================================
 Liabilities       ||                                    
-------------------++------------------------------------
-------------------++------------------------------------
                   ||                                    
===================++====================================
 Net:              ||          0           0           0
```
As far as I can tell from the debug output, the valuation of the investment is only calculated for the first date.

simonmichael:
Thanks @charvp. When showing change reports with -V, it seems we have always shown the value of the changes, and it would seem more useful to show the change of the values instead. (I think it's unlikely anyone will need the former.. ?)

This will require changes in accumValueAmounts/generateMultiBalanceReport, postingsReport and accountTransactionsReport I think.

[...]

PR #1380, opened 2020-10-26
lib: For MultiBalanceReport, report change in valuation rather than valuation of change.

Xitian9:
Addresses the MultiBalanceReport part of #1353


Xitian9:
The history of this PR seems to have been contaminated a bit. I'll rebase this when we've sorted out #1354.

Xitian9:
Rebased.

simonmichael:
Four tests changed, but their descriptions didn't change - are those still correct ?

I guess we'll need changes somewhere under https://hledger.org/hledger.html#valuation - at https://hledger.org/hledger.html#effect-of-valuation-on-reports and perhaps somewhere above.

simonmichael:
Looking good, I confirm that it fixes @charvp's bs example too.

Perhaps the test descriptions, and report headings, should change something like this ?
```
# 35. multicolumn balance report valued at period end
$ hledger -f- bal -M --value=end
Balance changes in 2000Q1, valued at period ends:
```
->
```
# 35. multicolumn balance report showing changes in period-end values
$ hledger -f- bal -M --value=end
Period-end value changes in 2000Q1:
```

Xitian9:
That change in doc/description sounds good to me.

Xitian9:
Hmm. Thinking about the doc/heading a bit more. Should we distinguish between cases where it's checking the change in valued  balance, and cases where it's valuing the change? The latter happens whenever there's no valuation, or when valuation happens at a fixed date for all columns.

simonmichael:
Yes it sounds like we should, probably. For each report we support, the title should be something that helps users keep straight on what the numbers mean.

simonmichael:
For the docs: it looks like only the "Effect of valuation on reports" table needs updating. Maybe we should split the "balances (with report interval)" row into "end balances (with report interval)" and "balance changes (with report interval)" rows.

Xitian9:
Alright, I did my best. If you can think of a more succinct way of describing the Effect on valuation reports, feel free to change it.

tests/journal/valuation.test
```diff
@@ -547,7 +547,7 @@ Budget performance in 2000Q1, valued at cost:
 ---++------------------------------------------------------------------------------------------------
    || 6 B [300% of 2 B]  7 B [350% of 2 B]  8 B [400% of 2 B]  21 B [350% of 6 B]  7 B [350% of 2 B] 
 
-# 48. budget report, valued at period ends.
+# 48. budget report, showing changes in period-end values.
 $ hledger -f- bal -MTA --budget --value=e
 Budget performance in 2000Q1, valued at period ends:
```
simonmichael, line 552: I can add this as a fixup, but need a sanity check: the budget report will also show value changes, and we can title it "Budget performance in 2000Q1, period-end value changes:"  ?

simonmichael: Oh I see, BudgetReport is separate and has not changed. 

Looks good to me, thank you!

simonmichael:
https://hledger.org/hledger.html#effect-of-valuation-on-reports tweaked

Issue #1158, opened 2020-01-01
Convert decimal separator for CSV imports


xnyhps:
I use a journal file which uses `.` for thousands and `,` for decimals:

```
commodity €1.000,00
```

One of the banks that I use uses `.` for the decimal separator for the CSV exports. This means that when I import these entries, the amounts are incorrect. Currently, I manually fix this by searching for the regex `\d\.\d\d^` and editing the `.`, but that is messy.

In the documentation for rules files, I did not see an option to convert the decimal separator (only the CSV field separator). It would be nice if that could also be supported.

This would probably mean defining how numbers are formatted (like `date-format` but for amounts) and how they should be formatted for the output (like a full `commodity` declaration, not the `currency` as it is in a rules file).

simonmichael:
Thanks for the report. I would welcome reproducible examples, a specification or a PR for this.

xnyhps:
I might be able to create a PR later, but for now a minimal example:

**hledger.journal**:
```
commodity €1.000,00
```

**import.rules**:
```
account1 assets:bank:checking

comment: %3, %14 %15 %16 %18

fields date,,,description,,,,,,,amount,,,,,,,,

date-format %d-%m-%Y

currency €
```

**import.csv**:
```
01-12-2019,NL12341234,,,,,,EUR,50.44,EUR,-42.79,01-12-2019,01-12-2019,1234,BEA,1234,'...','...',
```

Importing results in:
```
$ hledger import import.csv --rules-file import.rules -f hledger.journal
imported 1 new transactions
$ cat hledger.journal
commodity €1.000,00

2019/12/01    ; , 1234 BEA 1234 '...'
    assets:bank:checking         €-42.79
    expenses:unknown              €42.79
$ hledger -f hledger.journal bal
          €-4.279,00  assets:bank:checking
           €4.279,00  expenses:unknown
--------------------
                   0

simonmichael:
Thanks! I see. When the import command is generating journal entries, it should format them according to any applicable commodity directives in the journal file it is importing to,

simonmichael:
@xnyhps you can now use the https://hledger.org/csv.html#decimal-mark rule, in master.

Flimm:
Here is the new link to the `decimal-mark` documentation: https://hledger.org/hledger.html#decimal-mark

Here is what it says now:

>     decimal-mark .
> 
> or:
> 
>     decimal-mark ,
> 
[...]

PR #1382, opened 2020-11-07
csv decimal mark

simonmichael:
- cleanup
- ;csv: refactor amount parsing
- csv: decimal-mark rule to help with number parsing

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -175,6 +175,7 @@ instance Semigroup Journal where
   j1 <> j2 = Journal {
      jparsedefaultyear          = jparsedefaultyear          j2
     ,jparsedefaultcommodity     = jparsedefaultcommodity     j2
+    ,jparsedecimalmark          = jparsedecimalmark          j2
     ,jparseparentaccounts       = jparseparentaccounts       j2
     ,jparsealiases              = jparsealiases              j2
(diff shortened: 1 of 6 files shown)


simonmichael:
A user on #hledger had thousands separators and no decimal points in their CSV (exported from a spreadsheet), and there was no way to tell hledger that `1,000` in the CSV was one thousand not one point zero. Also requested earlier by #1158. I've added a `decimal-mark` CSV rule to make that possible.

Possibly todo: When reading journal files we can configure the input decimal mark for each commodity using commodity directives. But that must be done for each commodity and forces the output decimal mark to be the same. It might be worth adding a similar `decimal-markup` directive to journal format. I think this is touched on in #793.

simonmichael:
This helps parse numbers correctly, eg here it correctly treats the period as a digit group separator and parses one thousand two hundred and thirty four. But also the number format propagates to the output, which is according to spec but can be confusing, and there's no way to override the display style for amounts parsed from CSV I think.

```
2020-01-01,12.34
```
```
fields date,amount
decimal-mark ,
```
```
$ hledger print amt:'>1000'
2020-01-01
    expenses:unknown           12.34
    income:unknown            -12.34

```

simonmichael:
But this is as before: output from a CSV reflects the (canonicalised) style of the original CSV numbers. In practice I think this is usually fine; the exception is the import command, where we'd like to apply the journal's commodity styles to the new entries, which I'll do as a separate task.

simonmichael:
[Doc](https://hledger.org/csv.html#decimal-mark)

PR #1383, opened 2020-11-09
lib,ui: Make sure ReportSpec is updated when updating ReportOpts.

Xitian9:
Addresses some regressions. Changing the depth in hledger-ui now actually filters the postings, rather than just update the counter.

[Fix for #1340]

### DIFF
--- a/hledger-ui/Hledger/UI/UIState.hs
+++ b/hledger-ui/Hledger/UI/UIState.hs
@@ -53,14 +53,16 @@ uiShowStatus copts ss =
     showstatus Unmarked = "unmarked"
 
 reportSpecToggleStatusSomehow :: Status -> CliOpts -> ReportSpec -> ReportSpec
-reportSpecToggleStatusSomehow s copts rspec = rspec{rsOpts=ropts}
+reportSpecToggleStatusSomehow s copts =
+    either (error "reportSpecToggleStatusSomehow: updating Status should not result in an error") id  -- PARTIAL:
+    . updateReportSpecFromOpts update
   where
-    ropts = case maybeposintopt "status-toggles" $ rawopts_ copts of
-      Just 2 -> reportOptsToggleStatus2 s ropts
-      Just 3 -> reportOptsToggleStatus3 s ropts
---      Just 4 -> reportOptsToggleStatus4 s ropts
---      Just 5 -> reportOptsToggleStatus5 s ropts
-      _      -> reportOptsToggleStatus1 s ropts
+    update = case maybeposintopt "status-toggles" $ rawopts_ copts of
+      Just 2 -> reportOptsToggleStatus2 s
+      Just 3 -> reportOptsToggleStatus3 s
+--      Just 4 -> reportOptsToggleStatus4 s
+--      Just 5 -> reportOptsToggleStatus5 s
+      _      -> reportOptsToggleStatus1 s
 
 -- 1 UPC toggles only X/all
 reportOptsToggleStatus1 s ropts@ReportOpts{statuses_=ss}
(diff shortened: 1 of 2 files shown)


hledger-ui/Hledger/UI/UIState.hs
```diff
@@ -186,10 +188,11 @@ toggleForecast d ui@UIState{aopts=uopts@UIOpts{cliopts_=copts@CliOpts{reportspec
 
 -- | Toggle between showing all and showing only real (non-virtual) items.
 toggleReal :: UIState -> UIState
-toggleReal ui@UIState{aopts=uopts@UIOpts{cliopts_=copts@CliOpts{reportspec_=rspec@ReportSpec{rsOpts=ropts}}}} =
-  ui{aopts=uopts{cliopts_=copts{reportspec_=rspec{rsOpts=toggleReal ropts}}}}
+toggleReal ui@UIState{aopts=uopts@UIOpts{cliopts_=copts@CliOpts{reportspec_=rspec}}} =
+    ui{aopts=uopts{cliopts_=copts{reportspec_=update rspec}}}
   where
-    toggleReal ropts = ropts{real_=not $ real_ ropts}
+    update = either (error "toggleReal: updating Real should not result in an error") id  -- PARTIAL:
```
simonmichael, line 194: It's unfortunate to have to add more partials when we're trying to minimise those, even though we can convince ourselves they'll never fire. I wonder if we could have a total version of updateReportSpecFromOpts that we could use for safe updates like toggling a flag ?

simonmichael: Eg I suppose toggleReal could just perform a custom change to the ReportSpec & ReportOpts without actually regenerating the former ?

Xitian9: It could, but this would require climbing through the Query to alter any Real flags. That's tractable, but it wouldn't get around the problem for changing the report periods, which would require a much more complicated algorithm. I just don't see this being a reasonable path.

Xitian9: The main problem I see here is that `Query` is a very expressive data type. The CLI generates `Query`s of a very simple form (Dates and flags all at the top, etc.), but this might in the future not be the case. How would we toggle `Real` on the query `Or [And [Acct "a", Real True], Acct "b"]`? It's an un-natural operation. Toggling Real only really makes sense as an option flag, not on the underlying Query.

This gets worse if we start allowing deeply nested depth queries (as I dream might someday be permitted).

Xitian9: If the objective is just to get rid of the `error` calls, then we could replace `either (error "…") with `either (const origrspec)`. This is what `setFilter` does. Since the condition will never be fulfilled according to our understanding, it has the same effect as the above code without the partiality.

If the situation changes in the future and this code becomes accessible, it will silently ignore the updates. I'm not sure if this is desirable, or if in a runtime error telling us to fix the code is preferable.

simonmichael:
First, I'm ok with merging this as it gets master back in a good state, especially if you'd rather spend the time on other things.

I do feel like there must be another solution here. First, yes flipping bits inside a complex Query sounds tedious, though maybe it could be generalised, along the lines of filterQuery: queryTransform :: (Query -> Maybe Query) -> Query -> Query ? instance Traversable Query ?

More to the point, the risky/partial part of parsing a query is parsing the user data (query string), right ? It seems like we should only have to do that part once, and then be able to apply safe transformations with no error handling required. I am imagining something in the direction of 

parseQuery :: Text -> Either String (RelativeQuery, [QueryOpt])
fixQuery :: Day -> RelativeQuery -> Query

But I guess ReportOpts should be involved somewhere and I don't remember how. As I say I'm ok with the fix but I suspect we'll find a better way sooner or later.

simonmichael:
PS and I don't love the idea of silently ignoring problems, I agree that a runtime error is a slightly better option.

Xitian9:
Between introducing a bunch of `RelativeQuery` bookkeeping in hledger-lib and adding three inaccessible `error` calls indicating a programming error in hledger-ui, I'd take the second option as the lesser of two evils. I think a better option will become clearer when it's sorted out making `Query` a true boolean algebra.

simonmichael:
Sounds good. Thanks!

PR #1384, opened 2020-11-09
lib: Remove Empty Query constructor, which does nothing and has done so for a very long time.

Xitian9:
The Empty query constructor has done nothing for at least 9 years. I think this is a reasonable candidate for removal.

The only question is whether we should continue to accept the `empty:` prefix when parsing query terms. I've removed it here, but if you think it should continue to parse successfully (presumably as `Any`), let me know. I somewhat doubt anybody uses this ever.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -35,7 +35,6 @@ module Hledger.Query (
   queryIsSym,
   queryIsReal,
   queryIsStatus,
-  queryIsEmpty,
   queryStartDate,
   queryEndDate,
   queryDateSpan,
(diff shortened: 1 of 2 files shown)


simonmichael:
I believe you. Thanks for the cleanup!

simonmichael:
(0.26 release note in 2015: 
`empty: is no longer supported, as it overlaps a bit confusingly with amt:0. The --empty flag is still available.`)

Xitian9:
In fact, it doesn't really overlap with `amt:0`, because that would actually filter results. `Empty True` and `Empty False` would both match any posting or transaction.

PR #1386, opened 2020-11-10
import rewrites amount styles

simonmichael:
- ;lib: reader cleanups
- lib: global commodity display styles can be set in InputOpts or Journal, overriding all others
- ;doc: journal: amount display style -> commodity display style
- import: apply journal's commodity styles to imported amounts
- ;lib: drop a todo, seems ok
- ;update CLI usage texts
- ;update manuals


simonmichael:
Follow-on from #1382: import now applies the main journal's styles to imported amounts, so eg if you use a period decimal mark you can safely import from a csv file using comma decimal mark.

Issue #1398, opened 2020-11-17
roi gives unclear "No solution -- not bracketed" error


simonmichael:
A user on #hledger is getting this error and we don't know why, even though we traced it to "The result of searching for a root of a mathematical function.: The function does not have opposite signs when evaluated at the lower and upper bounds of the search." in the math-functions package. `roi` should make this error (and the others) impossible, or explain them well enough for new hledger users and non-math/investment-experts to understand (or at least, to fix).

adept:
It would be nice to have a reproducible example.

For  IRR part of roi, it tries to solve an equation to find a constant rate of return `r` such that investment (taking into account all cash flows) grows with rate `r` and reaches the same final balance that it actually has. The error likely means that this equation is not solvable. Perhaps final balance is negative, cashflow does not make sense, or some other corner case is hit.

That being said, error message could definitely be better. The problem is that it is hard to guess the exact root cause here

simonmichael:
Keeping this open for a little more exploration.

simonmichael:
We have [documented](https://hledger.org/hledger.html#roi) these errors. We don't yet prevent them (if it's possible) or provide any advice on how to resolve them. I'll keep this issue open until that improves.

psamim:
@simonmichael Thanks for opening this issue. There was an error in my ledger entries. This is an example of a ledger file which cause the same problem. The mistake was mine but maybe there could be better errors or maybe I was not knowledgeable enough to understand the error.

The mistake is in the entry for the date 2020-12-31, the last line is `Assets:Cash     -10 A @ 120` but it should be `Assets:Cash     10 A @ 120`.

```ledger
2019-11-30
    Assets:V       5 A @ 100
    Equities

2020-12-02
    Assets:V       5 A @ 120
    Equities

2020-12-31
    Income
    Assets:V         -5 A @ 100
    Assets:V         -5 A @ 120
    Assets:Cash     -10 A @ 120

2021-01-01
    Income
    Assets:V         10 A @ 100
    Assets:Cash           -1200

```

simonmichael:
@psamim, what command reproduces this error with your example above ? I'd like to check it with latest master and 1.20-branch.

psamim:
@simonmichael I tested it with 1.20 branch, It gives a very nice error message with explanation about possible causes. Thanks! I think we can close this.

PR #1401, opened 2020-11-18
roi: add adept's IRC discussion to doc

simonmichael:
For #1398.


simonmichael:
@adept: I've cleaned up and added your notes from yesterday - comments/changes welcome, or I'll merge in due course if you have no objections.

simonmichael:
PS it still needs one or two simpler examples.. are there any around that I've missed ?

adept:
I'll take a lock on this for a couple of days if you don't mind

PR #1415, opened 2020-12-14
lib: Sorting accounts by amount should also consider account name after amounts.

### LINKED ISSUE #1414: occasional strange sort order of accounts in balance reports
One of the balance command's functional tests is failing, but only on the hledger.org ubuntu machine (not on my mac, and not on github's ubuntu or mac machines apparently).
```
2018/1/1
  (b:j)  2

2018/1/1
  (c)    1

2018/1/1
  (b:i)  1

2018/1/1
    (a:k)   1

```
A balance report sorted by amount (-S) should be sorted first by amount, then account name, like this:
```
hledger bal -NS
                   2  b:j
                   1  a:k
                   1  b:i
                   1  c
```
But on that one machine it gives:
```
hledger bal -NS
                   2  b:j
                   1  c
                   1  a:k
                   1  b:i
```
It seems that in a few places we convert maps with toList and assume the result will be sorted, and in practice it usually is, but apparently not always. 

(All haskell package versions I checked seem to be the same, but perhaps there is some difference in the hashing. int-e said: *So it sorts by hash, but in case of a hash collision it produces the elements in something like insertion order, probably messed up further by deletions.* Perhaps different libgmp versions ?)

I think it means all our uses of `toList` should be replaced by `sort . toList` or similar.

Xitian9:
Should fix #1414 .

### DIFF
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -203,8 +203,9 @@ sortAccountTreeByAmount :: NormalSign -> Account -> Account
 sortAccountTreeByAmount normalsign = mapAccounts $ \a -> a{asubs=sortSubs $ asubs a}
   where
     sortSubs = case normalsign of
-        NormallyPositive -> sortOn (Down . normaliseMixedAmountSquashPricesForDisplay . aibalance)
-        NormallyNegative -> sortOn (       normaliseMixedAmountSquashPricesForDisplay . aibalance)
+        NormallyPositive -> sortOn (\a -> (Down $ amt a, aname a))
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks! This is better than what I was planning, as it makes the sort order explicit.

Perhaps mention in the haddock of buildReportRows that the rows are unsorted. Also, don't we need to sort in the last case of `rowtot`, to ensure "Total for a cumulative/historical report is always the last column." ?

All uses of "toList" elsewhere in hledger are fine.

simonmichael:
Actually it's fine, that's using a different toList which does sort. Merged as 92645cca, with a few extra comments.

Issue #1405, opened 2020-11-28
--value='end' for non-periodic queries ignores P directives after last posting


aragaer:
Sample journal:
```
P 2020-11-20 commodity 1

2020-11-20
 (assets)  1 commodity

P 2020-11-22 commodity 2
```
Now the following two commands show different result:
`hledger bal -f /tmp/test.journal --value='end,' --cumulative -e 2021` shows 1, while `hledger bal -f /tmp/test.journal --value='end,' --cumulative -e 2021 -Y` shows 2. Adding an unrelated posting fixes that:
```
2020-11-22
 unrelated  0
```
I have found this accidentally by comparing yearly results to "current" value. This difference in behavior seems a bit non-intuitive.

simonmichael:
Thanks for the report. I agree there might be an issue here. --debug can give more insight. I thought -V was equivalent to --value=end, but apparently not always:
```journal
P 2020-11-20 A 1B

2020-11-20
 (assets)  1A

P 2020-11-22 A 2B
```
```shell
$ hledger bal -e 2021 --value=end --debug=1 2>&1 | grep 'valuation date'
valuation date: 2020-11-20
$ hledger bal -e 2021 -V --debug=1 2>&1 | grep 'valuation date'
valuation date: 2020-12-31
```

https://hledger.org/hledger.html#valuation-date says:
> For single period reports, if an explicit report end date is specified, that will be used as the valuation date; otherwise the valuation date is "today".

-V used the report end date, but --value=end used the journal end date (date of last transaction).
https://hledger.org/hledger.html#value-flexible-valuation says:
> --value=end : Convert amounts to their value in the default valuation commodity, using market prices on the last day of the report period (or if unspecified, the journal's end date); or in multiperiod reports, market prices on the last day of each subperiod.

Last day of the report period was specified, so I think it should have used that, like -V. I'm only slight concerned that it was set this way for a reason, so we should check the commit history. Another part of the doc seems to acknowledge the difference in behaviour:

https://hledger.org/hledger.html#effect-of-valuation-on-reports > balance > balance changes says for -V:

> value at report end or today of sums of postings

and for --value=end:

> value at report or journal end of sums of postings

aragaer:
The only thing I might propose is that P directive is considered a journal entry when we're looking for "the end of journal/period".

Xitian9:
[...]

Issue #1396, opened 2020-11-17
multiperiod balance report should be more respectful of a specified report period


simonmichael:
`bs` will fill the report period you have specified, even when there are no transactions:
```
2020-01-01
  assets  $1
```
```
$ hledger -f a.j bs -M -b 202001 -e 202004
Balance Sheet 2020-01-31..2020-03-31

             || 2020-01-31  2020-02-29  2020-03-31 
=============++====================================
 Assets      ||                                    
-------------++------------------------------------
 assets      ||         $1          $1          $1 
-------------++------------------------------------
             ||         $1          $1          $1 
=============++====================================
 Liabilities ||                                    
-------------++------------------------------------
-------------++------------------------------------
             ||                                    
=============++====================================
 Net:        ||         $1          $1          $1 
```
`balance` should do the same, but does not (and has not at least since 1.10, surprisingly):
```
$ hledger -f a.j bal -M -b 202001 -e 202004
Balance changes in 2020-01:

        || Jan 
========++=====
 assets ||  $1 
--------++-----
        ||  $1 
```

simonmichael:
Ah I guess it's because `balance` also hides leading/trailing all-zero columns by default, so you must add -E:
```
$ hledger -f a.j bal -M -b 202001 -e 202004 -E
Balance changes in 2020Q1:

        || Jan  Feb  Mar 
========++===============
 assets ||  $1    0    0 
--------++---------------
        ||  $1    0    0 
```

`bs` doesn't do this:
```
2020-02-01
  (assets)    $1
```
$ hledger -f a.j bs -M -b 202001 -e 202004 
Balance Sheet 2020-01-31..2020-03-31

             || 2020-01-31  2020-02-29  2020-03-31 
=============++====================================
 Assets      ||                                    
-------------++------------------------------------
 assets      ||          0          $1          $1 
-------------++------------------------------------
             ||          0          $1          $1 
=============++====================================
 Liabilities ||                                    
-------------++------------------------------------
-------------++------------------------------------
             ||                                    
=============++====================================
 Net:        ||          0          $1          $1 
```

Should they be more consistent ? Which behaviour is more robust and useful overall in practice ?

Xitian9:
I definitely would agree with this change; it has annoyed me mildly in the past. It's a simple change; I'll look into it when I get some time.

Xitian9:
Just linking to a previous discussion on the topic of empty columns in #1269

Xitian9:
[...]

PR #1416, opened 2020-12-15
lib: Include price directives when calculating journal end date.

Xitian9:
Addresses #1405.

Let me know what you think of these changes.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -1184,16 +1184,22 @@ journalStyleInfluencingAmounts j =
 -- | The fully specified date span enclosing the dates (primary or secondary)
 -- of all this journal's transactions and postings, or DateSpan Nothing Nothing
 -- if there are none.
+--
+-- This will include the dates of any price directives that come after the last
+-- posting/transaction, but not those that come before the first.
 journalDateSpan :: Bool -> Journal -> DateSpan
 journalDateSpan secondary j
     | null ts   = DateSpan Nothing Nothing
     | otherwise = DateSpan (Just earliest) (Just $ addDays 1 latest)
     where
       earliest = minimumStrict dates
-      latest   = maximumStrict dates
+      latest   = case ddates of
+                      [] -> maximumStrict dates
+                      _  -> max (maximumStrict ddates) (maximumStrict dates)  -- Include commodity price directives in journal end
       dates    = pdates ++ tdates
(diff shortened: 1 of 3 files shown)


simonmichael:
This seems like a reasonably useful change of default behaviour, and there's probably no need to make it a new function or dependent on a new argument. 

https://hledger.org/hledger.html#report-start-end-date says:
> Most hledger reports show the full span of time represented by the journal data, by default. So, the effective report start and end dates will be the earliest and latest transaction or posting dates found in the journal.

I guess the new doc will be something like:
> By default, most hledger reports will show the full span of time represented by the journal data, as follows: the default report start date will be the earliest transaction or posting date, and the default report end date will be the latest transaction, posting, or market price date.

Xitian9:
New docs added.

simonmichael:
Thanks!

Xitian9:
Actually, I've just noticed a few bugs with this implementation, and I'm not 100% clear where they're coming from. I have some fixes in mind (which also address #1396), but I need to make sure there are not more bugs being swept under the carpet.

adept:
If you have multiple yearly files, and they all include common prices.journal, then this change would mean that the end date for all reports generated from any of these yearly files becomes the date of the ast price. This looks rather weird if your last price is "yesterday" and you are generating a report for 3 years ago.

I guess I should just reorganize my files, and move prices into the yearly journals. 

Just raising the visibility of this as you probably want to mention this in the changelog.

Xitian9:
I don't have much to say on this subject. My journals are not separated by year, so whenever I need to limit to a specific timeframe I use `-b/-e` options. I'm happy to defer to others about what works best in their workflow.

If we were wanting to make the changes @simonmichael suggests, we might even be able to limit it further. transactions + prices could be limited to just when using `--value=end`, as the other forms are independent of report end date.

An internal question: currently the journal end date is the latest of the transaction and price dates. Making this change would suggest that the journal end date should always be the latest of the transaction dates, which is then modified in certain circumstances. This would restore the output of `journalDateSpan` to its previous behaviour, perhaps good for compatibility.

simonmichael:
>  we might even be able to limit it further. transactions + prices could be limited to just when using --value=end, as the other forms are independent of report end date.

True. I guess the question is, what will be least surprising and most intuitive and easy to explain.

simonmichael:
Resolved as @Xitian9 suggested by #1495. Now only end-valued (`--value-end`) reports pick last transaction or market price date as default report end date. Other reports pick just last transaction date, as before.

Issue #30, opened 2013-04-08
feature; broaden "filter by account/description" field input


simonmichael:
_Original author: davamu...@gmail.com (November 21, 2010 03:51:49)_

to allow for simple filtering on tags, metadata and comments, using established syntax matching.

just as now 
[desc:cat     ] works great, to easily locate description text, I would like to be able to enter

[:nick:fitz:  ] to find matching tags anywhere, (case insensitive, please) and 

[; fence      ] to filter on matching text in the comments, which would work great locating

[; qtty:      ] metadata, even though hledger does not currently otherwise manipulate metadata or tags, it would be great to have the power to filter on them.

_Original issue: http://code.google.com/p/hledger/issues/detail?id=30_

simonmichael:
Needs discussion/developers, closing.

PR #1418, opened 2020-12-17
lib: Include empty columns in a MultiBalanceReports.

Xitian9:
This was previously done for CompoundBalanceReport and when --empty was
called; this makes the behaviour consistent.

Addresses #1396, and should also address part of #1405 (see test #30 in valuation.test).


simonmichael:
Ok let's give this a try. Thanks!

PR #1422, opened 2020-12-19
lib: Make -V behave as --value=end

Xitian9:


simonmichael:
This seems like a nice simplification. Great work!

PR #1423, opened 2020-12-23
[RFC] cli: drop deprecated command aliases a, b, p, r, txns

simonmichael:
These one-letter a, b, p, r command aliases for accounts, balance, print, register are less good than the more common three-letter ones: less mnemonic, less clear when reading, and impractical with shell completions. (And "txns" for print will never catch on I think.) So in master I deprecated these aliases, hiding them from the commands list, though they still work. 

Too many alternatives creates confusion, and I'd like to remove them entirely, but I have seen a few people using them.  I have posted this PR for discussion, and will cc the mail list. Any thoughts ?

A related change that might help with backward compabitility, at the cost of less careful command selection: always run the first command whose name matches the command argument, rather than requiring an unambiguous prefix. [ First alphabetically or first in the command list ? On second thought, might complicate life too much.]

[PS I should also have said: 

The goal of command abbreviations is to save typing. But "hledger b" is not much easier to type than "hledger bal". In practice, to really save typing you'll use eg a shell alias like "alias bal='hledger bal'".

And, the reason for defining "standard" abbreviations was to facilitate discussion and sharing examples, so fewer is better. (The other reason was to disambiguate collisions with other commands, so "bal" would always mean "balance" even as other "bal..." commands are added.)]

### DIFF
--- a/hledger/Hledger/Cli/Commands/Accounts.md
+++ b/hledger/Hledger/Cli/Commands/Accounts.md
@@ -1,4 +1,4 @@
-accounts, a\
+accounts\
 Show account names.
 
 _FLAGS
--- a/hledger/Hledger/Cli/Commands/Accounts.txt
+++ b/hledger/Hledger/Cli/Commands/Accounts.txt
@@ -1,4 +1,4 @@
-accounts, a
+accounts
 Show account names.
 
 _FLAGS
--- a/hledger/Hledger/Cli/Commands/Balance.md
+++ b/hledger/Hledger/Cli/Commands/Balance.md
@@ -1,4 +1,4 @@
-balance, bal, b\
+balance, bal\
 Show accounts and their balances.
 
 _FLAGS
--- a/hledger/Hledger/Cli/Commands/Balance.txt
+++ b/hledger/Hledger/Cli/Commands/Balance.txt
@@ -1,4 +1,4 @@
-balance, bal, b
+balance, bal
 Show accounts and their balances.
 
 _FLAGS
--- a/hledger/Hledger/Cli/Commands/Print.md
+++ b/hledger/Hledger/Cli/Commands/Print.md
@@ -1,4 +1,4 @@
-print, txns, p\
+print\
 Show transaction journal entries, sorted by date. 
 
 _FLAGS
--- a/hledger/Hledger/Cli/Commands/Print.txt
+++ b/hledger/Hledger/Cli/Commands/Print.txt
@@ -1,4 +1,4 @@
-print, txns, p
+print
 Show transaction journal entries, sorted by date.
 
 _FLAGS
--- a/hledger/Hledger/Cli/Commands/Register.md
+++ b/hledger/Hledger/Cli/Commands/Register.md
@@ -1,4 +1,4 @@
-register, reg, r\
+register, reg\
 Show postings and their running total.
 
 _FLAGS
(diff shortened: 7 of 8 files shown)


alerque:
Yes. :+1:

I use a lot of 1 and 2 letter abbreviations in my shell alias, but I'm a firm believer that anything that short should be reserved for individual customization and that projects should not be defining them upstream for everybody. As you say they make document ion harder to follow rather than easier, are near impossible to search for, completely ruin autocompletion, and introduce other problems. At the very most more self explanatory mnemonic abbreviations should be supported out of the box

simonmichael:
Thanks for the input here and elsewhere - done.

markokocic:
I am a heavy user of one letter mnemonics in hledger. Found this issue only after opening #1486 when I realised they are missing.

While I agree with @simonmichael that typing `hledger bal` is not much shorter that `hledger b`, I type only `h b` because I have an alias to call hledger as `h`. I also use `p` as in `h p` frequently, as well as `bs`, `bse`, `is`.

In principle I'm not against removing the shortcuts from the documentation, as long as there's a way for users that needs them to enable them again.

simonmichael:
Once you're defining aliases, wouldn't you rather go all the way and define `b` ? (I define `bal`)

markokocic:
I'm not aware of an easy way to define `h b` to be an alias to `hledger bal` or `h p` to `hledger print`.

I guess it's something that a shell script could do, create a script called h that parses its arguments and transforms them in a format hledger can understand. But then again, it won't work on Windows, at least not unless one writes another batch file.

simonmichael:
I assume your `h` is a shell alias. So in your shell config, I imagine you have something like
```shell
alias h='hledger'
```
I have 
```shell
# short aliases for hledger commands
alias h='hledger'
alias acc='hledger accounts'
alias act='hledger activity'
alias add='hledger add'
alias areg='hledger aregister'
alias bal='hledger balance'
alias check='hledger check'
alias checks='hledger check -s ordereddates'
alias close='hledger close'
alias iadd='hledger-iadd --date-format %Y/%m/%d'
alias print='hledger print'
alias reg='hledger register'
alias stats='hledger stats'
alias tags='hledger tags'
alias bs='hledger bs'
alias bse='hledger bse'
alias is='hledger is'
alias cf='hledger cashflow'
alias budget='hledger bal --budget'
```
Yes it's a little tedious to maintain that list as hledger acquires new commands, and possibly we could make that easier somehow.

The above may not work as easily on non-WSL windows, but you're not on windows and I think windows command line users don't mind typing `bal` instead of `b`.

markokocic:
I just created a short `fish shell` function wrapper around `hledger` to bring back the only functionality the way I like it.
Posting here as a reference if someone else needs a similar functionality.

```fish
function h --wraps=hledger --description 'customized hledger command'
    set -l args $argv
    set -l argc (count $args)
    if test $argc -eq 0
        command hledger
        return
    end

    # substitute command shortcuts for full command names
    set -l ind (contains -i -- b $args) && set args[$ind] bal
    set -l ind (contains -i -- p $args) && set args[$ind] print
    set -l ind (contains -i -- r $args) && set args[$ind] reg

    # add --tree where needed
    if contains bal $args; or contains bs $args; or contains is $args;
        if not contains -- --tree $args; and not contains -- --flat $args;
            set -a args --tree
        end
    end

    # add --change where needed
    if contains bs $args
        set -a args --change
    end

    if contains -- --test $args
        #vset -e args --test
        echo args: $args
    else
        command hledger $args
    end
end
```

This pretty much solves #1486 and #1487 for me.

I suppose something similar could also be created using `bash` if needed.

PR #1432, opened 2020-12-28
lib,cli: Add --gain flag

### LINKED ISSUE #1426: Allow viewing gain by showing the difference between valuation and cost basis
As discussed on IRC.

A small example journal:
```
2020-01-01 Order 1
    assets:cash  € -100
    assets:stocks:20200101  1 AAPL @@ € 100

2020-01-02 Order 2
    assets:cash  € -125
    assets:stocks:20200102  1 AAPL @@ € 125

P 2020-01-03 AAPL € 150
```

Using `hledger bs -f test.journal --infer-value -V -D -e 2020-01-04` gives the following output:
```
Balance Sheet 2020-01-01..2020-01-03

                        || 2020-01-01  2020-01-02  2020-01-03 
========================++====================================
 Assets                 ||                                    
------------------------++------------------------------------
 assets:cash            ||     € -100      € -225      € -225 
 assets:stocks:20200101 ||      € 100       € 125       € 150 
 assets:stocks:20200102 ||          0       € 125       € 150 
------------------------++------------------------------------
                        ||          0        € 25        € 75 
========================++====================================
 Liabilities            ||                                    
------------------------++------------------------------------
------------------------++------------------------------------
                        ||                                    
========================++====================================
 Net:                   ||          0        € 25        € 75 
```

Using `hledger bs -f test.journal --infer-value -B -D -e 2020-01-04` gives the following output:
```
Balance Sheet 2020-01-01..2020-01-03, valued at cost

                        || 2020-01-01  2020-01-02  2020-01-03 
========================++====================================
 Assets                 ||                                    
------------------------++------------------------------------
 assets:cash            ||     € -100      € -225      € -225 
 assets:stocks:20200101 ||      € 100       € 100       € 100 
 assets:stocks:20200102 ||          0       € 125       € 125 
------------------------++------------------------------------
                        ||          0           0           0 
========================++====================================
 Liabilities            ||                                    
------------------------++------------------------------------
------------------------++------------------------------------
                        ||                                    
========================++====================================
 Net:                   ||          0           0           0 
```

I think the difference of these two would be an interesting report (essentially reporting the unrealized gains):
```
Balance Sheet 2020-01-01..2020-01-03

                        || 2020-01-01  2020-01-02  2020-01-03 
========================++====================================
 Assets                 ||                                    
------------------------++------------------------------------
 assets:stocks:20200101 ||          0        € 25        € 50 
 assets:stocks:20200102 ||          0           0        € 25 
------------------------++------------------------------------
                        ||          0        € 25        € 75 
========================++====================================
 Liabilities            ||                                    
------------------------++------------------------------------
------------------------++------------------------------------
                        ||                                    
========================++====================================
 Net:                   ||          0        € 25        € 75 
 ```

The UX for this could be a new flag similar to `-V` and `-B` (`-G` or `--gain` in full, for example). The `--value` flag should probably also be extended, perhaps as the following: `--value=gain,COMM`.

### PR DESCRIPTION
**Pull request by chvp**
Fixes #1426.

Still needs some more tests.

### DIFF
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -336,9 +336,9 @@ aliasReplace (RegexAlias re repl) a =
 -- using the provided price oracle, commodity styles, and reference dates.
 -- Costing is done first if requested, and after that any valuation.
 -- See amountApplyValuation and amountCost.
-postingApplyCostValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> Costing -> Maybe ValuationType -> Posting -> Posting
-postingApplyCostValuation priceoracle styles periodlast today cost v p =
-    postingTransformAmount (mixedAmountApplyCostValuation priceoracle styles periodlast today (postingDate p) cost v) p
+postingApplyCostValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> Costing -> Maybe ValuationType -> Gaining -> Posting -> Posting
+postingApplyCostValuation priceoracle styles periodlast today cost v gain p  =
+    postingTransformAmount (mixedAmountApplyCostValuation priceoracle styles periodlast today (postingDate p) cost v gain) p
 
 -- | Apply a specified valuation to this posting's amount, using the
 -- provided price oracle, commodity styles, and reference dates.
--- a/hledger-lib/Hledger/Data/Transaction.hs
+++ b/hledger-lib/Hledger/Data/Transaction.hs
@@ -618,9 +618,9 @@ transactionTransformPostings f t@Transaction{tpostings=ps} = t{tpostings=map f p
 -- | Apply a specified costing and valuation to this transaction's amounts,
 -- using the provided price oracle, commodity styles, and reference dates.
 -- See amountApplyValuation and amountCost.
-transactionApplyCostValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> Costing -> Maybe ValuationType -> Transaction -> Transaction
-transactionApplyCostValuation priceoracle styles periodlast today cost v =
-  transactionTransformPostings (postingApplyCostValuation priceoracle styles periodlast today cost v)
+transactionApplyCostValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> Costing -> Maybe ValuationType -> Gaining -> Transaction -> Transaction
+transactionApplyCostValuation priceoracle styles periodlast today cost v gain =
+  transactionTransformPostings (postingApplyCostValuation priceoracle styles periodlast today cost v gain)
 
 -- | Apply a specified valuation to this transaction's amounts, using
 -- the provided price oracle, commodity styles, and reference dates.
--- a/hledger-lib/Hledger/Data/Valuation.hs
+++ b/hledger-lib/Hledger/Data/Valuation.hs
@@ -14,6 +14,7 @@ looking up historical market prices (exchange rates) between commodities.
 
 module Hledger.Data.Valuation (
    Costing(..)
+  ,Gaining(..)
   ,ValuationType(..)
   ,PriceOracle
   ,journalPriceOracle
@@ -56,6 +57,10 @@ import Text.Printf (printf)
 
 -- | Whether to convert amounts to cost.
 data Costing = Cost | NoCost
+  deriving (Show,Eq)
+
+-- | Whether to convert amounts to gain.
+data Gaining = Gain | NoGain
   deriving (Show,Eq)
 
 -- | What kind of value conversion should be done on amounts ?
@@ -104,9 +109,9 @@ priceDirectiveToMarketPrice PriceDirective{..} =
 -- using the provided price oracle, commodity styles, and reference dates.
 -- Costing is done first if requested, and after that any valuation.
 -- See amountApplyValuation and amountCost.
-mixedAmountApplyCostValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> Day -> Costing -> Maybe ValuationType -> MixedAmount -> MixedAmount
-mixedAmountApplyCostValuation priceoracle styles periodlast today postingdate cost v =
-    mapMixedAmount (amountApplyCostValuation priceoracle styles periodlast today postingdate cost v)
+mixedAmountApplyCostValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> Day -> Costing -> Maybe ValuationType -> Gaining -> MixedAmount -> MixedAmount
+mixedAmountApplyCostValuation priceoracle styles periodlast today postingdate cost v gain =
+    mapMixedAmount (amountApplyCostValuation priceoracle styles periodlast today postingdate cost v gain)
 
 -- | Apply a specified valuation to this mixed amount, using the
 -- provided price oracle, commodity styles, and reference dates.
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

**Comment by simonmichael** [comment]
Thanks; the test failure is mine, disregard.

**Comment by Xitian9** [comment]
Minor point of opinion, happy to be overridden. I would just have a `--value=gain` option, and not a corresponding `-G` flag. I feel that valuation is nuanced enough that making people type out the full option is a reasonable roadblack to make sure they understand what it's doing. See our recent conversations figuring out what exactly the `-V` flag should do.

**Comment by the-solipsist** [comment]
Hi. I'm not sure `--value=gain` is intuitive.  I think of it as one operation with four different parameter.
1. Determine gain/loss (i.e., calculate the difference between the increase/decrease in value from the value on the date of purchase + sale):
  a. Up to a particular date  (`--begin` + `--end` or `--period`)
  b. Using the FIFO/LIFO/HIFO/ACB/WACB method of calculating gain/loss. (a future option `--costbasis`?)  (Default should either be FIFO or WACB, imho, since the first is what is common under most tax systems for capital gains and the latter is the sensible option for currency conversions where capital gains doesn't need to be calculated.)
  c. Report the amount in a particular commodity/currency (Default should be what is declared using the the `D` declaration.)
  d. For unrealized gains/losses, use the rate (`--value`) as of `now` / a specific date (`then` doesn't really make sense in this regard, unless I'm mistaken).  (Default should be `--value=now`.)  (For realized gains/losses, of course, the rate as of the date of sale are used, rather than any other value: it is only for unrealized gains/losses that a particular date for valuation needs to be mentioned.)

As I see it, `value=` is thus a parameter, rather than the function/operation that is being carried out to determine gain.  Do I understand it incorrectly?

**Comment by simonmichael** [comment]
@charvp this relies on the original cost being stored in the Amount, and Amounts of different costs not being collapsed together. I don't remember how that all works exactly, have you found --value=gain to be robust in various report modes ? (If so I'm quite impressed.)

The PR currently hard-codes --value=gain to always use period-end value (like --value=end). Does this make sense ?

**Comment by Xitian9** [comment]
@charvp: Thanks for picking this up again! I've been meaning to get around to it.

We had a discussion around UI (buried somewhere in a long thread), and I think we decided that a gain report should be a ReportType, rather than a separate option. That is, another constructor called GainReport should be added to the ReportType, and the call to calculate the gain should happen somewhere around here, rather than in the `amountApplyCostValuation` function:
https://github.com/simonmichael/hledger/blob/71b45567375b83fa7440498d9d4dd2471b9138d0/hledger-lib/Hledger/Reports/MultiBalanceReport.hs#L303-L306

I'm happy to either leave this to you, or to do it myself. Let me know if you have strong feelings either way.

**Comment by chvp** [comment]
Ah, I missed that (or maybe I just didn't remember). Unfortunately I don't have any time left today, but I can look into doing it that way somewhere this week.

**Comment by chvp** [comment]
Hmm, I'm not sure what you mean actually. A gain report can be combined with the valuechange flag in the way I've written it now and that report seems useful to me. This is the current output for a historical report (taken from my personal ledger):
```
Ending balances (historical) in 2020-07-01..2021-04-30, valued at period ends with cost substracted:

                       || 2020-07-31  2020-08-31  2020-09-30  2020-10-31  2020-11-30  2020-12-31  2021-01-31  2021-02-28  2021-03-31  2021-04-30 
=======================++========================================================================================================================
 pension fund:20200724 ||   € -14.06     € 13.22      € 6.63    € -19.60     € 67.10     € 88.09     € 82.02     € 95.66    € 135.00    € 156.73 
 pension fund:20210104 ||          0           0           0           0           0           0     € -7.10      € 5.40     € 41.47     € 61.40 
-----------------------++------------------------------------------------------------------------------------------------------------------------
                       ||   € -14.06     € 13.22      € 6.63    € -19.60     € 67.10     € 88.09     € 74.91    € 101.06    € 176.47    € 218.13
```
This is a value change report:
```
Period-end value changes in 2020-07-01..2021-04-30 with cost substracted:

                       ||  2020-07  2020-08  2020-09   2020-10  2020-11  2020-12   2021-01  2021-02  2021-03  2021-04     Total 
=======================++=======================================================================================================
 pension fund:20200724 || € -14.06  € 27.28  € -6.59  € -26.23  € 86.71  € 20.98   € -6.07  € 13.64  € 39.34  € 21.73  € 156.73 
 pension fund:20210104 ||        0        0        0         0        0        0   € -7.10  € 12.51  € 36.07  € 19.93   € 61.40 
-----------------------++-------------------------------------------------------------------------------------------------------
                       || € -14.06  € 27.28  € -6.59  € -26.23  € 86.71  € 20.98  € -13.17  € 26.14  € 75.42  € 41.66  € 218.13 
```

I would like to have access to both these reports, and I don't really see how that would work if this became a report type.

**Comment by Xitian9** [comment]
I think the idea is that the --gain flag would output what you currently output for the have for the combined --gain and --value-change report. That is, a gain report would basically be a value change report, but would subtract the cost from the period in which it first appears.

The problem with not having gain be a separate report type, is that the output of a non-value-change multi-period gain report doesn't seem to have much meaning.

**Comment by chvp** [comment]
The historical multi-period gain report has a lot of meaning, I would think.

**Comment by Xitian9** [comment]
Maybe it's best to hash this out with examples. My understanding is that the intended UI is as follows.

```
P 2019-12-01 stock $1.00
P 2020-01-01 stock $1.50
P 2020-02-01 stock $2.00
P 2020-03-01 stock $2.50

2019-12-01
    (assets:old)   1 stock @@ $2.00

2020-01-01
    (assets:new)   1 stock @@ $2.00
```

```
$ hledger balance --gain --no-total -b 2020 -M
            ||    Jan    Feb    Mar
=============++=====================
 assets:new || -$0.50  $0.50  $0.50
 assets:old ||  $0.50  $0.50  $0.50

$ hledger balance --gain --no-total -b 2020 -M --historical
            ||    Jan    Feb    Mar
=============++=====================
 assets:new || -$0.50  $0.00  $0.50
 assets:old || -$0.50  $0.00  $0.50

$ hledger balance --gain --no-total -b 2020 -M --cumulative  # Weird one, not sure how meaningful this is
            ||    Jan    Feb    Mar
=============++=====================
 assets:new || -$0.50  $0.00  $0.50
 assets:old ||  $0.50  $1.00  $1.50
```

Is there another set of numbers you think would be useful here?

chvp:
I don't think there's other numbers that would be that useful (at least I can't think of any). I also agree that cumulative is kinda weird, but I feel that way about cumulative in other contexts as well, so I'm probably not a good judge of that.

Xitian9:
Good point. Since tabular reports would give weird results for every valuation type except `--value=end`, I wonder what we should do. Should a multi-period report force valuation type to `--value=end`?

Xitian9:
Actually, I don't think this would need special treatment: just the same as `--valuechange`. If no valuation is specified, default to `--value=end`, and otherwise just honour the valuation mode specified.

chvp:
Given this info, do you still think this should be a report type? Because I'm not entirely sure how the interactions with (value)change/historical/cumulative should work in that case.

Xitian9:
I think it's still reasonable. I'll give it a go if you'd like.

Xitian9:
I think the only thing that needs to happen is in this block of code:

https://github.com/simonmichael/hledger/blob/b6044aa6c89a9b71a6af46f499cdbb47ab274877/hledger-lib/Hledger/Reports/MultiBalanceReport.hs#L303-L309

We just need to make sure `historical` is the historical gain at that point when we've specified `GainReport`, i.e. it should be calculated with the gain function instead of `avalue` in `cumulativeSum`; and `changeamt` is `periodChanges startingGain historical`.

chvp:
Okay, I'll take another look this weekend.

Xitian9:
The main hesitation I'm having is that I feel like the valuation code is in the middle of a lot of conflicting changes: those here and those in #1554. I feel like we have too many places where we can be converting to cost, performing valuation, checking gain, and it's done a bit inconsistently at the moment.

I'm going to try to rationalise this, so that costs are dealt with in `journalSelectingAmountFromOpts` whenever possible. Then we can just subtract the cost from the value at that point whenever running a Gain report.

chvp:
Ok, maybe you can ping me when you've done the changes you think are required to consolidate the valuation and cost conversion and I'll rebase this PR at that point? You can work on it as well if you want, but I wouldn't mind putting in the work required so you can work on other things.

Xitian9:
@chvp. I've got a commit which implements some of the desired functionality, but I don't seem able to submit a PR to your repository. What's the best way to get it to you?

Xitian9:
I've also submitted #1560, which should hopefully make the valuation/costing/gain interface clearer. We should now be able to implement the GainReport only by modifying `mixedAmountApplyValuationFromOptsWith` and `valuationAfterSum` in ReportOptions.hs, along with a new case in `changeamts` in MultiBalanceReport.hs.

PR #1433, opened 2020-12-28
ci: also test compilation of bin/ add-on scripts

simonmichael:
The [add-on scripts in bin/](https://hledger.org/scripting.html#haskell-scripts) always fall behind hledger-lib/hledger API changes because we don't test them. We describe them as just examples and possibly broken, but that is unsatisfying and leads to intermittent manual catchup work. I think we need to include them in tests so as to keep them building, at least. This PR adds a test for that. If this is a good idea, next we need to update the scripts to make the tests pass.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -214,7 +214,7 @@ jobs:
         stack: ${{ matrix.plan.stack }}
       run: |
         export PATH=~/.local/bin:$PATH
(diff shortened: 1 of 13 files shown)


simonmichael:
PS help welcome, just push to the test-addons branch.

Xitian9:
I don't have permission to push to your branch, but I've pushed here: https://github.com/Xitian9/hledger/tree/test-addons

simonmichael:
Thanks!

PR #1434, opened 2020-12-29
make csvtest.sh script POSIX and shellchecked

ninewise:
As discussed in IRC. Should probably be tested by someone on BSD by running `make functest`.


simonmichael:
Sorry but it doesn't pass tests on a mac (and I don't want to spend time on it myself). We can give up on this for now, or I have triggered a mac CI build so you can see the failures, or perhaps another mac/BSD user wants to help.

simonmichael:
Thanks!

PR #1435, opened 2020-12-30
;lib: add 'payee' directive

DamienCassou:
Add (ignored, for now) "payee" directive. The idea would be to support ledger's `--check-payees` [CLI argument](https://www.ledger-cli.org/3.0/doc/ledger3.html#Error-Checking-and-Calculation-Options).


simonmichael:
Thanks!

PR #1438, opened 2020-12-31
journal: Ignore AUTO commodity when strict checking

### LINKED ISSUE #1419: strict mode: undeclared commodity "AUTO"
When in strict mode, hledger does not allow you to use automatically inferred posting prices. For example, take this file:

```
commodity USD 1 000.00
account assets
account revenues

2020/01/01 test
    assets  USD 100
    revenues
```

You would expect that this would parse correctly in strict mode, since all the accounts and the commodities have been declared. However, it doesn't:

```
$ hledger -f bla.journal balance -s
hledger: 
strict mode: undeclared commodity "AUTO"
in transaction at: "example/bla.journal" (lines 5-7)
```

aragaer:
AUTO commodity is a placeholder for postings with missing amounts. It
should be ignored when doing a strict commodity check.

Fixes #1419


simonmichael:
Thanks @aragaer! I was trying to get to this one.

Issue #1428, opened 2020-12-27
-T and -X. Any way to get "immediate" exchange rate?


aragaer:
```
P 2020-01-01 A 2.0

2020-01-02
 assets  2
 income

P 2020-01-03 A 1.0

2020-01-04
 assets  2
 income
```
I want some report that would show that I've got 3A. However either A is valued at 1 (end of journal) or income/assets get revaluated at P-record on 2020-01-03. Either way I'm getting 4A.

```
$ hledger bal -X A --transpose -N -T --invert income
                 A4  income
$ hledger bal -X A --transpose -N -T --invert income -D
Period-end value changes in 2020-01-02..2020-01-04:

            || income 
============++========
 2020-01-02 ||     A1 
 2020-01-03 ||     A1 
 2020-01-04 ||     A2 
   Total    ||     A4 
```

Is there a way to get number 3 using this journal?

aragaer:
It seems that I found an actual problem with this -- when currency gets revaluated, it recalculates the total value of account even outside the requested period:
```
$ cat a.journal 
2019-01-02
 assets  50
 income

P 2020-01-01 A 2.0

2020-01-02
 assets  2
 income

P 2020-01-03 A 1.0

2020-01-04
 assets  2
 income
$ hledger bal  -f a.journal -X A --transpose -N -T --invert income -D -b 2020
Period-end value changes in 2020-01-02..2020-01-04:

            || income 
============++========
 2020-01-02 ||     A1 
 2020-01-03 ||    A26 
 2020-01-04 ||     A2 
   Total    ||    A29 
```
Which looks like I've magically earned 26 units of A just because the money I've earned last year suddenly got more valuable.

simonmichael:
Thanks for the report. Re your first example, I can't figure out how you could have 3A, can you explain ?

Re the second, I don't get that output, please check it.

aragaer:
Just tried it again. Also added the --version output.
```
$ cat a.journal 
2019-01-02
 assets  50
 income

P 2020-01-01 A 2.0

2020-01-02
 assets  2
 income

P 2020-01-03 A 1.0

2020-01-04
 assets  2
 income
$ hledger bal -f a.journal -X A --transpose -N -T --invert income -D -b 2020
Period-end value changes in 2020-01-02..2020-01-04:

            || income 
============++========
 2020-01-02 ||     A1 
 2020-01-03 ||    A26 
 2020-01-04 ||     A2 
   Total    ||    A29 
[...]

Issue #1424, opened 2020-12-24
Different `--value=end` for `-p 2020` vs `-p 2020 -Y` (in 1.20)


julian-klode:
The file below produces inconsistent `--value=end` results for `-p 2020` and `-p 2020 -Y`; I have not calculated which one is correct. It's a bit hard to say - given that the last price is before the January purchase, I'd probably say there were no gains on that purchase, only on the one in December:

```
D 1,000.00€
commodity 1,000.00€

2019-12-23 buy 200 IE00BF4RFH31
    assets:investments    200 "IE00BF4RFH31" @ 4.9785€
    assets:investments                        -995.70€

2020-01-22 buy 103 IE00BF4RFH31
    assets:investments    103 "IE00BF4RFH31" @ 5.033€
    assets:investments                       -518.40€

P 2019/12/30 "IE00BF4RFH31" 4.957€
P 2020/01/02 "IE00BF4RFH31" 4.938€

```

The outcomes are:

```
; $ hledger -f f bal -p 2020   --value=end  assets -Y
; Period-end value changes in 2020:
; 
;                    ||    2020 
; ====================++=========
; assets:investments || -13.59€ 
; --------------------++---------
;                    || -13.59€ 
; $ hledger -f f bal -p 2020   --value=end  assets 
;              -9.79€  assets:investments
; --------------------
;              -9.79€
```

julian-klode:
So it seems that `-p 2020` only takes account the investment made in 2020, whereas `-p 2020 -Y` also takes into account the investment made in 2019 or something like it.

simonmichael:
Thanks for the report. Here's a simpler reproducer:
```
2019-12-15
    (a)    1 A

2020-01-15
    (a)    1 A

P 2019-12-30 A 10 B
P 2020-01-02 A 20 B
```

hledger 1.19 picks valuation dates 2020/1/15 and then 2020/12/31, per docs, giving the same results:

```
$ hledger-1.19 bal -p 2020  --value=end --debug
valuation date: 2020-01-15
default valuation commodity for A: Just "B"
market price for A -> B: 20.00000000
                20 B  a
--------------------
                20 B
```
```
$ hledger-1.19 bal -p 2020 -Y --value=end --debug
Balance changes in 2020, valued at period ends:

valuation date: 2020-12-31
default valuation commodity for A: Just "B"
market price for A -> B: 20.00000000
   || 2020 
===++======
 a || 20 B 
---++------
   || 20 B 
```

[...]

PR #1440, opened 2021-01-01
lib,cli: Extend --value=then valuation to all report types.

Xitian9:
Addresses #1428. @aragaer, could you let me know if this gives you the results you expect?

Also adds a postingDate argument to amountApplyValuation, and re-orders the ValuationType and (Transaction/Posting) arguments to (transaction/posting)ApplyValuation, to be consistent with amountApplyValuation.

Also addresses #1424.

The test failures are present in master. I've put a fix in commit a023fc880b00b85be3cde19f16af014c81edfde5 in PR #1427. I'll rebase this once that is merged or cherry-picked.

### DIFF
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -334,27 +334,13 @@ aliasReplace (RegexAlias re repl) a =
 -- provided price oracle, commodity styles, reference dates, and
 -- whether this is for a multiperiod report or not. See
 -- amountApplyValuation.
-postingApplyValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> Posting -> ValuationType -> Posting
-postingApplyValuation priceoracle styles periodlast today p v =
-  case v of
-    AtCost    Nothing -> postingToCost styles p
-    AtCost    mc      -> postingValueAtDate priceoracle styles mc periodlast $ postingToCost styles p
-    AtThen    mc      -> postingValueAtDate priceoracle styles mc (postingDate p) p
-    AtEnd     mc      -> postingValueAtDate priceoracle styles mc periodlast p
-    AtNow     mc      -> postingValueAtDate priceoracle styles mc today p
-    AtDate d  mc      -> postingValueAtDate priceoracle styles mc d p
+postingApplyValuation :: PriceOracle -> M.Map CommoditySymbol AmountStyle -> Day -> Day -> ValuationType -> Posting -> Posting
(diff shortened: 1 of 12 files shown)


aragaer:
Yep, `hledger bal --value="then,A"` gives me the result I wanted to see.

Xitian9:
Rebased.

Xitian9:
It might be worth reconsidering what the best valuation behaviour for -V is. We've already had two bug reports which expressed surprise at the new behaviour of `AtEnd with `PeriodChange` (see #1424 for an example). I myself have been caught by surprise with this, and I wrote the code! Though I think the new behaviour is indeed the correct behaviour, it wonder if we'll get lots of bug reports complaining about this.

The new `AtThen` behaviour might actually be a reasonable default behaviour for `PeriodChange` reports: it's very similar to the old behaviour of `AtEnd`, and is in fact identical when using a daily report (it differs more when you have longer report periods and more frequent market price changes). However, I think `AtEnd` is clearly better default behaviour for `Historical` reports.

I can think of a few possible courses of action:
1. Leave `AtEnd` as the behaviour of `-V`
2. Change to `AtThen` for the behaviour of `-V`
3. Use `AtThen` when using `-V` with `PeriodChange` and `AtEnd` when using `Historical` (I'm not sure what's best for `Cumulative`)
4. Remove `-V` as a flag, and make people use the `--value=X` interface; it's more explicit and less likely to be surpising
5. Add a warning to `-V`, encouraging people to use the `--value=X` interface, perhaps as a first step to later implementing (4)

Thoughts?

simonmichael:
I like the look of these changes, as they fill out a missing case that we previously didn't support (--value=then support for the balance commands, not just print and register).

I hesitate to merge it because the valuation situation is so fluid. Trying not to add confusion or cause behaviour to flip-flop from one release to the next. 

Or, can we be fairly sure that this change is standalone, not complicating other issues like #1353 and #1428 ? I'm not fully following your comment above @Xitian9 - I think it's discussing larger issues, not immediate effects of this PR, am I right ? Can you remind me what is the "new behaviour of AtEnd with PeriodChange" ?

Xitian9:
Yes, the above discussion is about larger issues that this PR (there was not an established place to discuss that at the the time I posted it). The ‘new behaviour of AtEnd with PeriodChange’ is the ‘change of valuation’ behaviour introduced in hledger-1.20.

I think this issue is fairly standalone; I don't think it complicates the above issues, as there behaviour with respect to valuation is clear (valuation is only calculated once for each posting, at a well-defined date).

simonmichael:
Great. Merging, thank you !

Issue #1431, opened 2020-12-28
Space leak in Amount arithmetic


Xitian9:
There seems to be a space leak somewhere in `Amount` arithmetic. Running `hledger register +RTS -h` and viewing the resulting profiling report gives the characteristic spike. [hledger-register-master.pdf](https://github.com/simonmichael/hledger/files/5746750/hledger-register-master.pdf)

I've tried making `Amount` strict in its contents and adding some strictness annotations to `similarAmountsOp`, but this has only changed the precise function the spike is attributed to. I'm not very experienced in squashing space leaks, so any pointers on how to get rid of this would be helpful.

Xitian9:
Okay, I think I've tracked this down to these lines. https://github.com/simonmichael/hledger/blob/7c9303a15c64859f11aec8fa75546793827e3086/hledger-lib/Hledger/Data/Amount.hs#L545-L549

I've gotten good results by replacing these lists with strict `Map` (total allocated memory drops in my example from ~250MB to ~35MB), but I'm having trouble getting the correct behaviour with respect to ensuring transactions with `AmountPrice` balance correctly (in particular `TotalPrice`). I've looked through the documentation, but I can't find where the correct behaviour is specified. Could you perhaps point to where this is specified?

Maybe related: #1177

simonmichael:
Great! I'd love to read any notes about your tracking-down process. 
Does this help:
https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Data/Transaction.hs#L360

simonmichael:
Incidentally, this will be interesting! http://www.well-typed.com/blog/2021/01/first-look-at-hi-profiling-mode

Xitian9:
Thanks!

A quick question about `MixedAmount` normalisation. The tests specify that `TotalPrice`s should never be combined under any circumstances. Is there a reason for such a strict treatment? Why can we not safely combine the following
```
1A @@ 2B
2A @@ 3B
```
to `3A @@ 5B`?

simonmichael:
Hmm.. it looks easy enough in this example, indeed. I don't remember the thinking exactly, unless it was to do with preserving per-lot information. I'm not against trying it.

Xitian9:
Though I'm sure there are other space leaks around, this one has been fixed by #1445.

PR #1445, opened 2021-01-06
lib: Fix some space leaks

Xitian9:
Addresses #1431.

I'm sure there's more to find here, but I'm getting decent results so I thought I'd put this up at least for discussion. Here's some profiling when running on my journal files.

```
$ hledger balance -Y +RTS -s    # master
   5,382,427,616 bytes allocated in the heap
   2,018,625,736 bytes copied during GC
      57,702,808 bytes maximum residency (67 sample(s))
       1,214,056 bytes maximum slop
              55 MB total memory in use (0 MB lost due to fragmentation)

  Gen  0      5140 colls,     0 par    0.318s   0.322s     0.0001s    0.0008s
  Gen  1        67 colls,     0 par    2.873s   2.950s     0.0440s    0.0924s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time    6.228s  (  6.309s elapsed)
  GC      time    1.404s  (  1.460s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    1.787s  (  1.812s elapsed)
  EXIT    time    0.000s  (  0.001s elapsed)
  Total   time    9.422s  (  9.583s elapsed)

  Alloc rate    864,181,667 bytes per MUT second

  Productivity  85.1% of total user, 84.7% of total elapsed

$ hledger balance -Y +RTS -s    # this PR
   4,999,963,024 bytes allocated in the heap
   1,718,454,752 bytes copied during GC
      49,167,928 bytes maximum residency (63 sample(s))
       1,301,792 bytes maximum slop
              46 MB total memory in use (0 MB lost due to fragmentation)

  Gen  0      4775 colls,     0 par    0.290s   0.289s     0.0001s    0.0010s
  Gen  1        63 colls,     0 par    2.517s   2.557s     0.0406s    0.0869s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time    5.840s  (  5.847s elapsed)
  GC      time    1.182s  (  1.218s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    1.626s  (  1.629s elapsed)
  EXIT    time    0.000s  (  0.001s elapsed)
  Total   time    8.649s  (  8.695s elapsed)

  Alloc rate    856,198,369 bytes per MUT second

  Productivity  86.3% of total user, 86.0% of total elapsed

$ hledger register +RTS -s    # master
  21,222,212,224 bytes allocated in the heap
  17,586,635,464 bytes copied during GC
     356,007,792 bytes maximum residency (140 sample(s))
       3,850,384 bytes maximum slop
             339 MB total memory in use (0 MB lost due to fragmentation)

  Gen  0     20618 colls,     0 par    0.765s   0.765s     0.0000s    0.0008s
  Gen  1       140 colls,     0 par   19.332s  19.498s     0.1393s    0.4020s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time   13.573s  ( 13.629s elapsed)
  GC      time   10.619s  ( 10.788s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    9.478s  (  9.476s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   33.672s  ( 33.894s elapsed)

  Alloc rate    1,563,590,976 bytes per MUT second

  Productivity  68.5% of total user, 68.2% of total elapsed

$ hledger register +RTS -s    # this PR
  13,106,052,448 bytes allocated in the heap
   3,136,387,864 bytes copied during GC
      56,339,968 bytes maximum residency (104 sample(s))
       1,184,256 bytes maximum slop
              53 MB total memory in use (0 MB lost due to fragmentation)

  Gen  0     12614 colls,     0 par    0.397s   0.399s     0.0000s    0.0008s
  Gen  1       104 colls,     0 par    4.425s   4.507s     0.0433s    0.0948s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time    9.914s  ( 10.035s elapsed)
  GC      time    2.126s  (  2.178s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    2.697s  (  2.728s elapsed)
  EXIT    time    0.000s  (  0.001s elapsed)
  Total   time   14.738s  ( 14.943s elapsed)

  Alloc rate    1,321,958,888 bytes per MUT second

  Productivity  85.6% of total user, 85.4% of total elapsed
```

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -40,6 +40,7 @@ exchange rates.
 
 -}
 
+{-# LANGUAGE BangPatterns       #-}
 {-# LANGUAGE CPP                #-}
 {-# LANGUAGE OverloadedStrings  #-}
 {-# LANGUAGE RecordWildCards    #-}
@@ -144,12 +145,10 @@ module Hledger.Data.Amount (
 import Control.Monad (foldM)
 import Data.Decimal (DecimalRaw(..), decimalPlaces, normalizeDecimal, roundTo)
 import Data.Default (Default(..))
-import Data.Function (on)
-import Data.List (groupBy, intercalate, intersperse, mapAccumL, partition,
-                  sortBy)
+import Data.Foldable (toList)
+import Data.List (intercalate, intersperse, mapAccumL, partition)
 import Data.List.NonEmpty (NonEmpty(..), nonEmpty)
-import qualified Data.Map as M
-import Data.Map (findWithDefault)
+import qualified Data.Map.Strict as M
 import Data.Maybe (fromMaybe)
 #if !(MIN_VERSION_base(4,11,0))
 import Data.Semigroup ((<>))
@@ -246,8 +245,8 @@ amt @@ priceamt = amt{aprice=Just $ TotalPrice priceamt}
 -- Prices are ignored and discarded.
 -- Remember: the caller is responsible for ensuring both amounts have the same commodity.
 similarAmountsOp :: (Quantity -> Quantity -> Quantity) -> Amount -> Amount -> Amount
-similarAmountsOp op Amount{acommodity=_,  aquantity=q1, astyle=AmountStyle{asprecision=p1}}
-                    Amount{acommodity=c2, aquantity=q2, astyle=s2@AmountStyle{asprecision=p2}} =
+similarAmountsOp op !Amount{acommodity=_,  aquantity=q1, astyle=AmountStyle{asprecision=p1}}
+                    !Amount{acommodity=c2, aquantity=q2, astyle=s2@AmountStyle{asprecision=p2}} =
    -- trace ("a1:"++showAmountDebug a1) $ trace ("a2:"++showAmountDebug a2) $ traceWith (("= :"++).showAmountDebug)
    amount{acommodity=c2, aquantity=q1 `op` q2, astyle=s2{asprecision=max p1 p2}}
   --  c1==c2 || q1==0 || q2==0 =
@@ -559,24 +558,18 @@ normaliseMixedAmount = normaliseHelper False
 
 normaliseHelper :: Bool -> MixedAmount -> MixedAmount
 normaliseHelper squashprices (Mixed as)
-  | missingamt `elem` as = missingmixedamt -- missingamt should always be alone, but detect it even if not
-  | null nonzeros        = Mixed [newzero]
-  | otherwise            = Mixed nonzeros
+  | missingkey `M.member` amtMap = missingmixedamt -- missingamt should always be alone, but detect it even if not
+  | M.null nonzeros= Mixed [newzero]
+  | otherwise      = Mixed $ toList nonzeros
   where
-    newzero = lastDef nullamt $ filter (not . T.null . acommodity) zeros
-    (zeros, nonzeros) = partition amountIsZero $
-                        map sumSimilarAmountsUsingFirstPrice $
-                        groupBy groupfn $
-                        sortBy sortfn
-                        as
-    sortfn  | squashprices = compare `on` acommodity
-            | otherwise    = compare `on` \a -> (acommodity a, aprice a)
-    groupfn | squashprices = (==) `on` acommodity
-            | otherwise    = \a1 a2 -> acommodity a1 == acommodity a2 && combinableprices a1 a2
-
-    combinableprices Amount{aprice=Nothing} Amount{aprice=Nothing} = True
-    combinableprices Amount{aprice=Just (UnitPrice p1)} Amount{aprice=Just (UnitPrice p2)} = p1 == p2
-    combinableprices _ _ = False
+    newzero = maybe nullamt snd . M.lookupMin $ M.filter (not . T.null . acommodity) zeros
+    (zeros, nonzeros) = M.partition amountAndPriceIsZero amtMap
+    amtMap = foldr (\a -> M.insertWith sumSimilarAmountsUsingFirstPrice (key a) a) mempty as
+    key Amount{acommodity=c,aprice=p} = (c, if squashprices then Nothing else priceKey <$> p)
+      where
+        priceKey (UnitPrice  x) = (acommodity x, Just $ aquantity x)
+        priceKey (TotalPrice x) = (acommodity x, Nothing)
+    missingkey = key missingamt
 
 -- | Like normaliseMixedAmount, but combine each commodity's amounts
 -- into just one by throwing away all prices except the first. This is
@@ -600,9 +593,13 @@ unifyMixedAmount = foldM combine 0 . amounts
 -- | Sum same-commodity amounts in a lossy way, applying the first
 -- price to the result and discarding any other prices. Only used as a
 -- rendering helper.
-sumSimilarAmountsUsingFirstPrice :: [Amount] -> Amount
-sumSimilarAmountsUsingFirstPrice [] = nullamt
-sumSimilarAmountsUsingFirstPrice as = (sumStrict as){aprice=aprice $ head as}
+sumSimilarAmountsUsingFirstPrice :: Amount -> Amount -> Amount
+sumSimilarAmountsUsingFirstPrice a b = (a + b){aprice=p}
+  where
+    p = case (aprice a, aprice b) of
+        (Just (TotalPrice ap), Just (TotalPrice bp))
+          -> Just . TotalPrice $ ap{aquantity = aquantity ap + aquantity bp }
+        _ -> aprice a
 
 -- -- | Sum same-commodity amounts. If there were different prices, set
 -- -- the price to a special marker indicating "various". Only used as a
@@ -945,9 +942,7 @@ tests_Amount = tests "Amount" [
        [usd 1 @@ eur 1
        ,usd (-2) @@ eur 1
        ])
-        @?= Mixed [usd 1 @@ eur 1
-                   ,usd (-2) @@ eur 1
-                   ]
+        @?= Mixed [usd (-1) @@ eur 2 ]
 
     ,test "showMixedAmount" $ do
        showMixedAmount (Mixed [usd 1]) @?= "$1.00"
@@ -970,8 +965,8 @@ tests_Amount = tests "Amount" [
         normaliseMixedAmount (Mixed [usd 1 `at` eur 1, usd 1 `at` eur 1]) @?= Mixed [usd 2 `at` eur 1]
       ,test "amounts with different unit prices are not combined" $
         normaliseMixedAmount (Mixed [usd 1 `at` eur 1, usd 1 `at` eur 2]) @?= Mixed [usd 1 `at` eur 1, usd 1 `at` eur 2]
-      ,test "amounts with total prices are not combined" $
-        normaliseMixedAmount (Mixed  [usd 1 @@ eur 1, usd 1 @@ eur 1]) @?= Mixed [usd 1 @@ eur 1, usd 1 @@ eur 1]
+      ,test "amounts with total prices are combined" $
+        normaliseMixedAmount (Mixed  [usd 1 @@ eur 1, usd 1 @@ eur 1]) @?= Mixed [usd 2 @@ eur 2]
     ]
 
     ,test "normaliseMixedAmountSquashPricesForDisplay" $ do
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -209,7 +209,7 @@ instance Num Amount where
     abs a@Amount{aquantity=q}    = a{aquantity=abs q}
     signum a@Amount{aquantity=q} = a{aquantity=signum q}
     fromInteger i                = nullamt{aquantity=fromInteger i}
-    negate a@Amount{aquantity=q} = a{aquantity= -q}
+    negate a                     = transformAmountAndPrice negate a
     (+)                          = similarAmountsOp (+)
     (-)                          = similarAmountsOp (-)
     (*)                          = similarAmountsOp (*)
@@ -260,14 +260,14 @@ amountWithCommodity c a = a{acommodity=c, aprice=Nothing}
 -- - price amounts must be MixedAmounts with exactly one component Amount
 --   (or there will be a runtime error XXX)
 --
--- - price amounts should be positive
+-- - price amounts should be positive in the Journal
 --   (though this is currently not enforced)
 amountCost :: Amount -> Amount
 amountCost a@Amount{aquantity=q, aprice=mp} =
     case mp of
       Nothing                                  -> a
       Just (UnitPrice  p@Amount{aquantity=pq}) -> p{aquantity=pq * q}
-      Just (TotalPrice p@Amount{aquantity=pq}) -> p{aquantity=pq * signum q}
+      Just (TotalPrice p@Amount{aquantity=pq}) -> p{aquantity=pq}
 
 -- | Replace an amount's TotalPrice, if it has one, with an equivalent UnitPrice.
 -- Has no effect on amounts without one.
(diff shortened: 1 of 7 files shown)


Xitian9:
Point for discussion:

Currently, we have `newtype MixedAmount = Mixed [Amount]`. I've always thought that there should really be two different data representations: `[Amount]` and `newtype NormalisedMixedAmount = Map AmountKey Amount`.

The bulk of fixing this space leak comes from using this latter representation within `normaliseHelper`, and then switching back to the former when returning. This results in a lot of converting back-and-forth between these representations. Perhaps it might be worth keeping this `NormalisedMixedAmount` representation around longer, to keep things more efficient.

Xitian9:
Some notes on methodology:

To find the space leak fixed in the first commit I used this technique: http://neilmitchell.blogspot.com/2015/09/detecting-space-leaks.html.

The second space leak is much bigger, but was not detected by the above method: presumably it uses a lot of threads, but no individual thread uses up much memory. I found that by looking at the heap profile generated by running `hledger command +RTS -h`, seeing the functions which were using big memory spikes, and poking around inside by trial and error to see what fixed it.

simonmichael:
I didn't totally understand the second commit yet. "This makes journal entries like the following impossible to represent:  a  0 @@ 10 A" - I guess this refers to the new code after this PR ? (Such an entry seems to work right now).

Could you update the third commit's message to mention the other changes (or split it): I think it also makes some fields strict, and it changes mixed amount normalisation to merge @@-priced amounts ?

Also, I'm having trouble merging this locally for testing even though github thinks it has no conflicts - maybe you could rebase against latest master.

Xitian9:
> I didn't totally understand the second commit yet. "This makes journal entries like the following impossible to represent: a 0 @@ 10 A" - I guess this refers to the new code after this PR ? (Such an entry seems to work right now).

Actually, I meant the previous code. If you look inside `amountToCost`, for `TotalPrice` case it previously had the cost as `pq * signum q`. Since `signum 0 == 0`, this meant that anything with quantity 0 would be converted to 0 cost, even if it had a non-zero total cost. So you could put `0 @@ 10 A` in a journal file, but `amountToCost` would not convert it to 10 A.

I'll rebase the PR and write some more detailsed commit messages.

Xitian9:
Rebased and split, with more detailed commit messages.

Xitian9:
I'm actually coming around to a slightly more aggressive normalisation, where all prices are converted to total prices, then summed in the normalised amount. I don't really see what advantage there is to keeping amounts with different unit prices (in the same commodity) separate.

Xitian9:
> I'm actually coming around to a slightly more aggressive normalisation, where all prices are converted to total prices, then summed in the normalised amount. I don't really see what advantage there is to keeping amounts with different unit prices (in the same commodity) separate.

It looks like doing this would change how some things are displayed (e.g. in opening/closing reports). It's debatable whether these changes are good, bad, or neutral, but I think there's not a huge argument for making this change.

That said, I would still be interested in opinions on https://github.com/simonmichael/hledger/pull/1445#issuecomment-755785519. My impression is that the current definition `newtype MixedAmount = Mixed [Amount]` is basically just there so we can define a `Num` instance for `MixedAmount`. I would propose having `Posting` contain `[Amount]`, and most other uses changed to the internal representation `newtype MixedAmount = Mixed (M.Map AmountKey Amount)`. Anything where we've performed arithmetic or normalisation loses its individual amount and lot information, so I don't see the advantage to keeping that representation around.

Xitian9:
I've been running this (and some subsequent work built upon it) for a while now, and I'm not seeing any issues so far. The only potentially controversial changes here are:
- Combining amounts with total prices
- Storing the sign inside a total price (rather than inferring it from the quantity of the amount)

I think given the huge memory leak plugged, this is a pretty clear win. Let me know if you have any issues or concerns.

simonmichael:
You seem to have a massive improvement in runtime (half) for register. I'm seeing far more modest improvements for either balance or register with examples/10000x1000x10.journal, or my own 2020 journal. I wonder what could explain this.

Xitian9:
Interesting. Are you also seeing the big drop in maximum residency for the register report? Most of the saved time is in garbage collection.

I can't say exactly why the savings are big for register and smaller for balance, but my wild guess was:
- Register sums the mixed amounts for each posting, but then doesn't do much else with them (aside from keeping a running total). So it probably keeps all the intermediate lists around for a long time before consuming them when it displays the results.
- Balance doesn't track individual postings, so it doesn't keep them around after it sums them in the running balance. This allows it to garbage collect earlier.

simonmichael:
With the 10k-txn journal it goes from

      33,865,088 bytes maximum residency (10 sample(s))

to

      29,092,952 bytes maximum residency (10 sample(s))

That's the biggest of the changes I've seen in my tests.

Can you reproduce yours ? I wonder if both builds were fully optimised. Eg `make functest` builds an unoptimised hledger by default. Also please try with examples/10000x1000x10.journal when you get a chance (`make samplejournals`).

simonmichael:
**lib: Consume list immediately in commodityStylesFromAmounts. **

The new code is less obvious, but more efficient. I'm ok with this one.

**lib: Include sign in TotalPrice in Amount, rather than relying on the sign of aquantity.**

"you could put 0 @@ 10 A in a journal file, but amountToCost would not convert it to 10 A, rather to 0 A". 
Is that better ? I always thought of the quantity (`0`) as primary and the price (`@@ 10 A`) as an extra property. Is there any reason to let 0 @@ 10 A mean 10 A ? Here's how it would look:
```
2021-01-01
    a      0 @@ 10 A
    b
```
Old behaviour:
```
$ hledger print -x --cost
2021-01-01
    a               0
    b               0
```
New behaviour:
```
$ hledger.1445 print -x --cost
2021-01-01
    a            10 A
    b           -10 A
```

simonmichael:
(Also, allowing the price be signed seems to complicate things. What would be the meaning of a negative price ? Now there'd be two signs to think about.)

Xitian9:
I agree that the use case for the actually writing `0 @@ 10A` is limited, but being able to represent such a thing is necessary if you're going to be combining total prices. Otherwise what is the sum of the following?
```
10 B @@ 20 A
-10 B @@ 10 A
```
Negative price is necessary to deal with sums like:
```
10 B @@ 10 A
-8 B @@ 12 A
```

Xitian9:
Also, if somebody did write `0 @@ 10 A`, I don't see how they would expect anything other than the second behaviour. If you were expecting the first, wouldn't you write `0 @@ 0 A`?

simonmichael:
The sums there should be.. 
10B @@ 10A and 2B @@ 4A, am I right ?

simonmichael:
Up to this point at least my mental model has been that if the quantity is zero, the amount is zero and a price doesn't mean anything. I wouldn't expect that something written with zero quantity could become non-zero if converted to cost. I assume and grant that it's more mathematically consistent the way you are proposing..

Xitian9:
I don't think so. These are total prices, so if you buy 10 B of something at a cost of 20 A, and then sell 10 B of something at a cost of 10 A, your overall change is of 0 B, but you've lost 10 A in cost. `(10 B @@ 20 A) + (-10 B @@ 10 A) = (0 B @@ 10 A)`.

Similarly, if you buy 10 B at a cost of 10 A, then sell 8 B at a cost of 12 A, you have 2 B left to you, and have gained 2 A.
`(10 B @@ 10 A) + (-8 B @@ 12 A) = (2 B @@ -2 A)`

Xitian9:
I expect this change to make no difference with how people are actually writing things. I don't see a reasonable case for writing any of the things you don't expect. But if we're going to be combining total prices, instead of keeping them entirely separate as was done previously, then being able to represent these kinds of things is absolutely necessary.

Basically, I view this an an internal change, not an API change.

Xitian9:
Here's a real-life use case for negative prices. There was a brief period last year when the price of oil was negative: people would pay you to take their (contracts to purchase) oil (because they didn't have a place to store it). This was of course a highly unusual situation, not something you would generally expect, but we cannot represent such transactions within the current cost interface.

Xitian9:
I've rebuilt the two branches with the following commands: `stack clean; stack --profile build hledger`. I then proceeded to profile with the following command: `hledger register +RTS -poregister -s >/dev/null 2> register.txt`. The results were as follows:

master:
```
  22,580,919,416 bytes allocated in the heap
   1,951,921,616 bytes copied during GC
     384,091,328 bytes maximum residency (15 sample(s))
       4,111,168 bytes maximum slop
             814 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0     22064 colls,     0 par    0.999s   1.026s     0.0000s    0.0008s
  Gen  1        15 colls,     0 par    0.847s   0.857s     0.0571s    0.3345s

  INIT    time    0.003s  (  0.002s elapsed)
  MUT     time   14.579s  ( 14.644s elapsed)
  GC      time    1.847s  (  1.883s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.001s elapsed)
  Total   time   16.429s  ( 16.530s elapsed)

  Alloc rate    1,548,837,475 bytes per MUT second

  Productivity  88.7% of total user, 88.6% of total elapsed
```

spaceleak
```
  13,401,875,184 bytes allocated in the heap
     573,245,336 bytes copied during GC
      50,188,384 bytes maximum residency (13 sample(s))
       1,183,648 bytes maximum slop
             170 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0     12991 colls,     0 par    0.460s   0.471s     0.0000s    0.0008s
  Gen  1        13 colls,     0 par    0.223s   0.225s     0.0173s    0.0419s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time    9.903s  (  9.923s elapsed)
  GC      time    0.683s  (  0.696s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.001s elapsed)
  Total   time   10.588s  ( 10.622s elapsed)

  Alloc rate    1,353,271,615 bytes per MUT second

  Productivity  93.5% of total user, 93.4% of total elapsed
```

I'm not sure why my numbers have gone up significantly for both branches, but the ratio of the two remains roughly the same order of magnitude.

Running the same binaries on `example/10000x1000x10.journal` gives more modest results. Is there really something so weird about my journal file?

master:
```
   7,890,991,336 bytes allocated in the heap
     563,512,592 bytes copied during GC
      62,905,232 bytes maximum residency (13 sample(s))
         577,992 bytes maximum slop
             174 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0      7630 colls,     0 par    0.380s   0.392s     0.0001s    0.0029s
  Gen  1        13 colls,     0 par    0.225s   0.228s     0.0176s    0.0660s

  INIT    time    0.009s  (  0.009s elapsed)
  MUT     time    7.632s  (  7.693s elapsed)
  GC      time    0.605s  (  0.620s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time    8.246s  (  8.322s elapsed)

  Alloc rate    1,033,943,375 bytes per MUT second

  Productivity  92.6% of total user, 92.4% of total elapsed
```

spaceleak:
```
   7,711,814,400 bytes allocated in the heap
     457,122,296 bytes copied during GC
      55,695,664 bytes maximum residency (12 sample(s))
         558,800 bytes maximum slop
             154 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0      7450 colls,     0 par    0.333s   0.343s     0.0000s    0.0011s
  Gen  1        12 colls,     0 par    0.179s   0.182s     0.0151s    0.0522s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time    8.296s  (  8.376s elapsed)
  GC      time    0.512s  (  0.524s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.001s elapsed)
  Total   time    8.811s  (  8.903s elapsed)

  Alloc rate    929,558,366 bytes per MUT second

  Productivity  94.2% of total user, 94.1% of total elapsed
```

Xitian9:
Aha. I think I've found the problem. I noticed that the sample journals have no costs in them, so I altered `generatejournal` to include a cost of `@ {amt} C` in each transaction. This again produced minor differences.

However, if I altered `generatejournal` to instead include a cots of `@@ {amt} C` in each transaction, then I saw a big difference. In fact, I couldn't even run the master branch on `examples/10000x1000x10.journal`: it exhausted the RAM on my computer and then crashed.

Here are the results on the modified `examples/1000x1000x10.journal` with both the master and spaceleak branches.

master:
```
  11,615,918,552 bytes allocated in the heap
   2,202,465,024 bytes copied during GC
     353,124,496 bytes maximum residency (16 sample(s))
       3,407,728 bytes maximum slop
             796 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0     11253 colls,     0 par    0.913s   0.933s     0.0001s    0.0053s
  Gen  1        16 colls,     0 par    0.898s   1.234s     0.0771s    0.3474s

  INIT    time    0.003s  (  0.016s elapsed)
  MUT     time    6.331s  (  6.373s elapsed)
  GC      time    1.811s  (  2.167s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time    8.145s  (  8.557s elapsed)

  Alloc rate    1,834,693,780 bytes per MUT second

  Productivity  77.7% of total user, 74.5% of total elapsed
```

spaceleak:
```
     951,920,624 bytes allocated in the heap
      63,541,616 bytes copied during GC
       7,599,600 bytes maximum residency (9 sample(s))
          64,016 bytes maximum slop
              22 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0       913 colls,     0 par    0.046s   0.046s     0.0001s    0.0005s
  Gen  1         9 colls,     0 par    0.040s   0.040s     0.0045s    0.0072s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time    0.959s  (  0.961s elapsed)
  GC      time    0.086s  (  0.087s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time    1.046s  (  1.050s elapsed)

  Alloc rate    992,830,827 bytes per MUT second

  Productivity  91.6% of total user, 91.5% of total elapsed
```

So it looks like it's the presence of total costs which is running up the memory. This is completely consistent with the fact that `master` never combines total prices, so they're all kept around forever, making a bunch of giant list whenever you try to sum MixedAmounts, while `spaceleak` allows them to be combined.

Xitian9:
Some information on my journals, to give context to the results above: there are a total of 8709 transactions, and 176 postings which contain total prices.

adept:
Oh man. I have 8336 total prices :( This explains a thing or two I have been seeing. Thanks!

simonmichael:
That's great, exactly the sort of clue I was hoping for. I'll look into adding more such prices to the test journals. More on this soon.

Xitian9:
If you're updating the test journals, it may be a good idea to include more commodities as well. The old code slows down with the number of commodities (I think quadratically?), so this will give a more robust of amount arithmetic performance.

simonmichael:
> On Feb 10, 2021, at 7:32 PM, Stephen Morgan <notifications@github.com> wrote:
> Here's a real-life use case for negative prices. There was a brief period last year when the price of oil was negative:
> 
Good example. So indeed negative prices should be supported, to model situations like that.

What about 0 A @@ $10 being converted to $10 by --cost ? I'm still not keen on that. Is there any reason it must work that way ?

simonmichael:
> What about 0 A @@ $10 being converted to $10 by --cost ? I'm still not keen on that. Is there any reason it must work that way ?

Well I guess it's not a big deal, I won't sweat it.

simonmichael:
This second commit does change user-visible behaviour, though hopefully not in a way that affects existing journals. But I think we should have at least one test demonstrating it, maybe https://github.com/simonmichael/hledger/pull/1445#issuecomment-777156666.

simonmichael:
**lib: Distinguish between an Amount having quantity (or rounded quanti…**

Now we can check that both posting amount and price appear to be zero. I guess this is required because a total price can now override a zero posting amount.

The "XAndPriceLooksZero" naming seems slightly off though, since it doesn't check that a unit price looks zero. 

If you just used the old names ("amountLooksZero"), would a lot of tests break I wonder ?

simonmichael:
And maybe we could sharpen the haddocks a little.

Current: "Does mixed amount and its price appear to be zero when rendered with its display precision ?"

Perhaps: "Does this mixed amount (and its price, if it's a total price) appear to be zero when rendered with its display precision ?"

Or should it be: "Does this mixed amount (or its total price, if it has one) appear to be zero when rendered with its display precision ?" 

In fact could we enumerate the ways an amount can be zero ? This could get confusing.

0A - yes
0A @ $1 - yes
0A @@ $1 - no
2A @ $0 - yes ?
2A @@ $0 - yes

simonmichael:
I'm getting confused even about current hledger, let me show that as well:
```
2021-01-01
  (a)  0A      
  (b)  0A @  $1
  (c)  0A @@ $1
  (d)  2A @  $0
  (e)  2A @@ $0
```
```
$ hledger-1.20.4 bal -N -E
                   0  a
                   0  b
                   0  c
                  2A  d
                  2A  e
$ hledger-1.20.4 bal -N -E -B
                   0  a
                   0  b
                   0  c
                   0  d
                   0  e
```

simonmichael:
And with this PR currently:
```
$ stack exec -- hledger bal -N -E
                   0  a
                   0  b
                   0  c
                  2A  d
                  2A  e
$ stack exec -- hledger bal -N -E -B
                   0  a
                   0  b
                  $1  c
                   0  d
                   0  e
```

Xitian9:
> If you just used the old names ("amountLooksZero"), would a lot of tests break I wonder ?

Yes, I think the names are a bit clunky as well. Only one test breaks if I just replace the old functions with the new ones (`journal/transaction-prices.test:17`), but this can be solved by more carefully thinking about when we check zeros and when we strip prices. In fact, making this change on a different branch building upon this results in no tests failing at all.

Of course, this would still leave a small API change which might affect others. If you're happy with that, I can go ahead.

Xitian9:
Yes, making those changes to the haddocks seems reasonable.

Your enumerations look good to me. Should those be added to the haddocks as well?

Xitian9:
We could just change the description to ‘Do this mixed amount and its cost appear to be zero…’.

Xitian9:
As an aside, parallel to the `(mixedA|a)mount(Looks|Is)Zero` versus `(mixedA|a)mountAndPrice(Looks|Is)Zero` ugliness, there's a similar ugliness with `(multiply|divide)(Mixed)?Amount` versus `(multiply|divide)(Mixed)?AmountAndPrice`. I think we almost always want to do the latter.

If we're making minor breaking API changes on the first set of functions, we may also want to change the second.

Xitian9:
I've added a commit which makes the `*AndPrice` variants of these functions the default, and removed the old versions which only acted on the quantity. I think this is probably the better way to do things, but if you'd prefer to leave the old ones in let me know and I can edit this accordingly.

simonmichael:
Merged as 7bfbcde. Thank you!

PR #1447, opened 2021-01-07
lib,cli: Revert --value=end to hledger-1.19 ‘value of change’ behaviour.

Xitian9:
Include a new flag ‘--change-of-value’ to instead report the change in the valuations.

Documentation hasn't been updated yet (that valuation table is going to be a nightmare).

[Addressing #1428]


simonmichael:
Thanks!

I assume this is not strictly a revert of the original commit, but includes some new cleanups (or is it ?)

I feel there's some unclarity in the original commit message and eg in this commit's proposed help text: "with -V/-X/--value=end, report the change in the...". I believe this change-of-value vs value-of-change question is not specific to --value=end, in fact it's orthogonal to the valuation type.

simonmichael:
What would you think about this instead:
```
     --change                show balance change in each period (or value of change) (default)
     --change-of-value       show balance change in each period (or change of value)
```

simonmichael:
Or
```
     --change                show balance change or value of change in each period (default)
     --change-of-value       show balance change or change of value in each period
```
possible with a "(with -V)" in there..

Xitian9:
Yes, it's not a strict revert: it's just restoring the original behaviour with the default options.

I don't think this is orthogonal, in fact this is not an issue at all with anything other than the `AtEnd` (and fallback behaviour of `AtCost`).

The issue is that for every other valuation type, the valuation of a posting is evaluated a fixed date. This means that the two different behaviours (valuation-of-change versus change-of-valuation) are identical. For `AtEnd`, we valuate the same amount with different valuations in different periods, which gives the difference.

In formulae:
```
Initial amount = x1
Final amount = x2
Valuation at initial = r1
Valuation at final = r2
Valuation of change = r2 * (x2 - x1) = r2 * x2 - r2 * x1
Change of valuation = r2 * x2 - r1 * x1 = (valuation of change) + (r2 - r1) * x1
```
Note that if `r1 = r2`, there is no difference. For each given posting, if the valuation used is the same at every point in time (as it is for all other valuation types), then there's no difference between the two.

I have no strong opinions on help messages.

Xitian9:
Note: This is why we're able to do the valuation directly on the posting for all other valuation types. You can see that none of the tests started failing when that change was made.

simonmichael:
Really. Aha. Thanks for explaining, I hadn't understood that.

Xitian9:
Would you still like the help text changed? Any of the descriptions offered are correct.

simonmichael:
I guess your wording is fine, thanks.

Help me out once more - when using --value=then, will --change and --change-of-value be the same ?

Xitian9:
Yes, they will. The reasoning is slightly more complicated than that given above (since different postings are valued at different dates), but since each posting is always valued at the same date regardless of the period you're considering, it all cancels out.

simonmichael:
If so, I'm not sure it's worth introducing --change-of-value at all. The alternative is to just remove that change from 1.20 ("bal, is, bs --change: Valued multiperiod balance change reports now show changes of value, rather than the value of changes. (#1353, Stephen Morgan)") and keep the new stuff for 1.21.

simonmichael:
And I don't mean to create more work for you (sorry), I can also do it when I get a chance. But opinions welcome.

Xitian9:
Also reasonable. It would leave @charvp without their desired functionality for a bit longer, but it would avoid creating a command line flag which will be removed if (when) we update the behaviour of AtEnd in two months. All the changes are relatively simple, the only hard part is deciding on the wording of the documentation and renumbering all the tests!

Xitian9:
Flag removed. Let me know what you think.

charvp:
I'll just not update to 1.20.3, so that's fine for me. I'll give my thoughts on #1353 on the weekend (after thinking it through first).

Xitian9:
Rebased

simonmichael:
Thanks, but could you rebase against 1.20-branch, not master ? Whenever you can, no hurry.

Xitian9:
Oops, I accidentally rebased against hledger-1.20. Give me a bit to fix this.

Xitian9:
Ah, here's an issue. The multi balance report code went through a significant refactoring in #1440; rebasing would be a lot more effort than just starting from scratch. Is that fine?

Xitian9:
The new branch is here: https://github.com/Xitian9/hledger/tree/newrevert. I'm not sure how to change the target of this PR to 1.20-branch. Should I just rebase and that's something you can do, or should I make a new PR?

simonmichael:
> The new branch is here: https://github.com/Xitian9/hledger/tree/newrevert <https://github.com/Xitian9/hledger/tree/newrevert>. I'm not sure how to change the target of this PR to 1.20-branch. Should I just rebase and that's something you can do, or should I make a new PR?
> 
Thanks! Got it merged, more tomorrow.

simonmichael:
This is in https://github.com/simonmichael/hledger/tree/1.20-branch. Thanks!

PR #1448, opened 2021-01-09
;doc: adjust traditional top level accounts

cardoe:
Across the rest of the documentation and other accounting documentation
it appears to imply that the more traditional top level name would be
revenue. In fact the income statement says that is a subtraction between
revenue and expenses to derive the income. While the account types that
hledger supports are named revenue so it seems more natural to name this
revenue.

### DIFF
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -1751,7 +1751,7 @@ flag or `real:1` query.
 Account names typically have several parts separated by a full colon, from
 which hledger derives a hierarchical chart of accounts. They can be
 anything you like, but in finance there are traditionally five top-level
-accounts: `assets`, `liabilities`, `income`, `expenses`, and `equity`.
+accounts: `assets`, `liabilities`, `revenue`, `expenses`, and `equity`.
 
 Account names may contain single spaces, eg: `assets:accounts receivable`.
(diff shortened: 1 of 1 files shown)


cardoe:
A bigger commit would be to rename the places in the example that use "income" to "revenue". My goal here isn't to say the docs should dictate things but that they be internally consistent. When a person is employed as a W-2 employee I've seen it commonly referred to as "income" while a person that gets paid primarily through 1099s will have things referred to as "revenue". I'm guessing that's because a W-2 person has their taxes taken out while the 1099 person will not. Anyway, the goal here was to make the docs more consistent and I'm willing to do other adjustments as suggested or drop this commit entirely.

simonmichael:
Thanks for the consistency fix. Would you mind re-pushing this as an edit to the master file, which is hledger.m4.md. No need to change hledger.info, hledger.1, hledger.txt (I update those periodically as a separate commit to reduce merge conflicts).

cardoe:
Done! Thanks for the quick response. Didn't realize that was the master doc. Thanks for the pointer.

simonmichael:
Thanks!

Re further income->revenue edits - I probably have some inconsistent language because "income" is so heavily used and more familiar to most users. Perhaps in some places I say "revenue (AKA income)" or similar. But if you see more cases like this we can certainly consider fixing them.

Issue #1456, opened 2021-01-12
--drop N doesn't affect other output formats


simonmichael:
--drop affects text (terminal) output, but not CSV or HTML output (probably all formats except text). Eg
```
hledger bal -O csv --drop 1
```

simonmichael:
This was fixed.

PR #1459, opened 2021-01-12
cli,csv: Fix --drop option in csv output format

aragaer:
When printing balance in csv format the `--drop` option is ignored. Found this when playing with my reports and decided to just fix it.


simonmichael:
Nice timing, thank you! (Cf #1456). Merged as 7bde3345.

PR #1460, opened 2021-01-12
Fix insane precision in roi (re-fixes #1417)

### LINKED ISSUE #1417: -X/--value doesn't work for 'roi' command
My default currency is quite volatile and I prefer reports in some other, somewhat more stable currency. Let's say I want to track some investment in commodity A while getting report in currency B (here 1/B is the volatile default currency).

```
P 2020-12-01 A 50
P 2020-12-01 B 100

2020-12-01 * buy A
 assets  10 A @ 50
 cash

P 2020-12-10 A 55
P 2020-12-10 B 125

2020-12-10 * reevaluate
 assets  -10 A @ 50
 assets  10 A @ 55
 equity:unrealized gains
```

I want to see that my investment actually went down -- It was 5B and now it's 4.4B. But I can't do that since `hledger roi` ignores the `-X` or `--value` flag (but it is in the documentation):

```
$ hledger -f /tmp/hl.journal roi --inv "assets cur:A" --pnl "unrealized gains" -X B
+---++------------+------------++---------------+----------+-------------+-----++----------+----------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) | PnL ||      IRR |      TWR |
+===++============+============++===============+==========+=============+=====++==========+==========+
| 1 || 2020-12-01 | 2020-12-10 ||             0 |      500 |         550 |  50 || 3142.15% | 3142.15% |
+---++------------+------------++---------------+----------+-------------+-----++----------+----------+
```

adept:
#1417 is already fix, but fix introduced a potential for high-precision Decimals to be computed inside roi.

Let's ensure that we never print them to full precision. While we are at it, let's use MixedAmount more in Roi, in particular let's report all commodity amounts in the styles used in the journal


simonmichael:
Merged as 07a9f1196. Thanks!

Issue #538, opened 2017-04-14
hledger-web pie chart


NorfairKing:
This is a request:

It would be nice to see a breakdown per category in the form of a pie chart in the web UI.

NorfairKing:
I've tried something here: https://github.com/NorfairKing/hledger/tree/pie-chart, but I'm rather terrible at JS and not familiar where the appropriate data is in hledger.

simonmichael:
Nice! Can you show a demo or screenshot ?

NorfairKing:
@simonmichael, sure but there's nothing to see really. I haven't hooked up the data and I'm terrible at making things look good. I used the jquery plotting library that you're already using.

![pie](https://cloud.githubusercontent.com/assets/3521180/25045776/33d79504-212e-11e7-8059-6d26a01ea6f6.png)

NorfairKing:
I had an idea for 2 pie charts:

- One that shows where the money is divided among the subcategories
- One that shows where the different commodities are divided among this category.

I think the second idea isn't always feasible because you can't always compare comoddities.

NorfairKing:
Also: I'm really happy with what you've made here. It integrates nicely with my own systems :D 👍

simonmichael:
Thanks, that's good to know! Do feel free to blog/tweet about your setup (and if the latter, don't forget #hledger #plaintextaccounting tags so it gets seen).

I installed your branch on the demo1.hledger.org test instance to see how it works and follow the WIP, eg: http://demo1.hledger.org/register?q=inacct%3AExpenses . You'll get more efficient feedback if you run your own public instance, otherwise I'll update this one when I notice changes.

simonmichael:
As with hledger-chart, there's an unresolved question about pie charts which you're probably already aware of: how will you handle the case when some amounts have different sign ?

NorfairKing:
> I installed your branch on the demo1.hledger.org test instance to see how it works and follow the WIP, eg: http://demo1.hledger.org/register?q=inacct%3AExpenses . You'll get more efficient feedback if you run your own public instance, otherwise I'll update this one when I notice changes.

Cool!
Any chance you could help me with this feature?

NorfairKing:
>  How will you handle the case when some amounts have different sign ?

No idea! 
I guess you could just _not_ draw a pie chart in that case?

simonmichael:
Sure, what do you need ? Feel free to join #hledger also.

NorfairKing:
[...]

PR #1467, opened 2021-01-19
Pie chart

raboof:
See issue #538

Took https://github.com/NorfairKing/hledger/commits/pie-chart as a starting point

### DIFF
--- a/hledger-web/Hledger/Web/Foundation.hs
+++ b/hledger-web/Hledger/Web/Foundation.hs
@@ -161,6 +161,7 @@ instance Yesod App where
       addScript $ StaticR js_jquery_hotkeys_js
       addScript $ StaticR js_jquery_flot_min_js
       addScript $ StaticR js_jquery_flot_selection_min_js
+      addScript $ StaticR js_jquery_flot_pie_min_js
       addScript $ StaticR js_jquery_flot_time_min_js
       addScript $ StaticR js_jquery_flot_tooltip_min_js
       toWidget [hamlet| \<!--[if lte IE 8]> <script type="text/javascript" src="@{StaticR js_excanvas_min_js}"></script> <![endif]--> |]
--- a/hledger-web/Hledger/Web/Handler/RegisterR.hs
+++ b/hledger-web/Hledger/Web/Handler/RegisterR.hs
@@ -119,19 +119,28 @@ registerChartHtml q title percommoditytxnreports = $(hamletFile "templates/chart
 
 data PieHalf = Positive | Negative
 
+moreThanOne :: Eq a => [a] -> Bool
+moreThanOne [] = False
+moreThanOne (x : xs) = rec xs
+  where
+    rec [] = False
+    rec (y : ys) = x /= y || (rec ys)
+
 -- | Generate javascript/html for a mockup pie chart
 registerPieChartHtml :: PieHalf -> Text -> BalanceReport -> HtmlUrl AppRoute
 registerPieChartHtml half q (items, _) = $(hamletFile "templates/piechart.hamlet")
   where
     charttitle = "Pie Chart" :: String
-    labelDataTuples =
+    labelData =
       reverse $
-      sortOn snd $
-      filter (\(_, quant) -> case half of Positive -> quant >= 0
-                                          Negative -> quant < 0) $
+      sortOn (\(_, quant, _) -> quant) $
+      filter (\(_, quant, _) -> case half of Positive -> quant >= 0
+                                             Negative -> quant < 0) $
       flip concatMap items $ \(accname, _, _, Mixed as) ->
-        flip map as $ \a -> (accname, aquantity a)
-    showChart = if ((length labelDataTuples) > 1) then "true" else "false" :: String
+        flip map as $ \a -> (accname, aquantity a, acommodity a)
+    moreThanOneAcct = moreThanOne $ map (\(acct, _, _) -> acct) labelData
+    moreThanOneCommodity = moreThanOne $ map (\(_, _, com) -> com) labelData
+    showChart = if (moreThanOneAcct && not moreThanOneCommodity) then "true" else "false" :: String
     noacctlink = (RegisterR, [("q", T.unwords $ removeInacct q)])
     chartId = case half of Positive -> "postive" :: String
                            Negative -> "negative" :: String
--- a/hledger-web/Hledger/Web/Handler/RegisterR.hs
+++ b/hledger-web/Hledger/Web/Handler/RegisterR.hs
@@ -9,7 +9,7 @@
 
 module Hledger.Web.Handler.RegisterR where
 
-import Data.List (intersperse, nub, partition)
+import Data.List (intersperse, nub, partition, sortOn)
 import qualified Data.Text as T
 import Text.Hamlet (hamletFile)
 
@@ -51,6 +51,12 @@ getRegisterR = do
         | isJust (inAccount qopts) = "Period Total"
         | otherwise                = "Total"
       transactionFrag = transactionFragment j
+      (chartQuery, depth) =
+          case (,) <$> inAccount qopts <*> inAccountQuery qopts of
+              Nothing -> (Any, 0)
+              Just ((an, _), aq) -> (aq, accountNameLevel an)
+      rspecWithQuery = rspec { rsQuery = (And [chartQuery, Depth $ depth + 1, m]) }
+      balancereport = balanceReport rspecWithQuery j
   defaultLayout $ do
     setTitle "register - hledger-web"
     $(widgetFile "register")
@@ -112,10 +118,15 @@ registerChartHtml q title percommoditytxnreports = $(hamletFile "templates/chart
    nodatelink = (RegisterR, [("q", T.unwords $ removeDates q)])
 
 -- | Generate javascript/html for a mockup pie chart
-registerPieChartHtml :: HtmlUrl AppRoute
-registerPieChartHtml title = $(hamletFile "templates/piechart.hamlet")
- where
-   charttitle = "Pie Chart" :: String
+registerPieChartHtml :: BalanceReport -> HtmlUrl AppRoute
+registerPieChartHtml (items, _) = $(hamletFile "templates/piechart.hamlet")
+  where
+    charttitle = "Pie Chart" :: String
+    labelDataTuples =
+      reverse $
+      sortOn snd $
+      flip concatMap items $ \(accname, _, _, Mixed as) ->
+        flip map as $ \a -> (accname, aquantity a)
 
 dayToJsTimestamp :: Day -> Integer
 dayToJsTimestamp d =
--- a/hledger-web/Hledger/Web/Handler/RegisterR.hs
+++ b/hledger-web/Hledger/Web/Handler/RegisterR.hs
@@ -111,6 +111,12 @@ registerChartHtml q title percommoditytxnreports = $(hamletFile "templates/chart
    shownull c = if null c then " " else c
    nodatelink = (RegisterR, [("q", T.unwords $ removeDates q)])
 
+-- | Generate javascript/html for a mockup pie chart
+registerPieChartHtml :: HtmlUrl AppRoute
+registerPieChartHtml title = $(hamletFile "templates/piechart.hamlet")
+ where
+   charttitle = "Pie Chart" :: String
+
 dayToJsTimestamp :: Day -> Integer
 dayToJsTimestamp d =
   read (formatTime defaultTimeLocale "%s" t) * 1000 -- XXX read
--- a/hledger-web/Hledger/Web/Handler/RegisterR.hs
+++ b/hledger-web/Hledger/Web/Handler/RegisterR.hs
@@ -118,8 +118,8 @@ registerChartHtml q title percommoditytxnreports = $(hamletFile "templates/chart
    nodatelink = (RegisterR, [("q", T.unwords $ removeDates q)])
 
 -- | Generate javascript/html for a mockup pie chart
-registerPieChartHtml :: BalanceReport -> HtmlUrl AppRoute
-registerPieChartHtml (items, _) = $(hamletFile "templates/piechart.hamlet")
+registerPieChartHtml :: Text -> BalanceReport -> HtmlUrl AppRoute
+registerPieChartHtml q (items, _) = $(hamletFile "templates/piechart.hamlet")
   where
     charttitle = "Pie Chart" :: String
     labelDataTuples =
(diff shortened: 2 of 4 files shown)


simonmichael:
Thanks for reviving this!

`make help-buildtest` shows some make rules that may help show errors like the CI does.

simonmichael:
Also, screenshots welcome when you get to that point!

raboof:
Yes, planning to get back to fixing the warnings and adding some more things later ;). Pretty basic for now, but already on par with the existing work I think:

hledger

(snipped out most data for privacy - is there a '3DBenchy for hledger' canonical test dataset somewhere bychance?)

simonmichael:
Nifty. Is it in addition to/instead of the line chart ?

examples/bcexample.hledger is a pretty good example journal.

raboof:
That's a lot of taxes! (it's looks like it is showing the USD and the IRAUSD taxes as separate slices, also haven't checked those are converted correctly so the % values make sense)

pie

raboof:
showing 2 pie charts when there are both positive and negative transactions might still be a good improvement, but I think this PR is ready to be reviewed even without that.

simonmichael:
> showing 2 pie charts when there are both positive and negative transactions might still be a good improvement, but I think this PR is ready to be reviewed even without that.

(Thought I replied to this earlier, but I don't see it.) @raboof, won't it potentially show misleading results without that capability ? I think we need it.

raboof:
> won't it potentially show misleading results without that capability ? I think we need it.

Updated to split the pie chart into positive and negative accounts (only showing them if there is more than one account to show).

The chart would still be confusing when accounts use multiple currencies. I don't think splitting those into multiple pie charts as well would make too much sense - I'm just hiding the pie charts in that case for now.

simonmichael:
That sounds good. Will review as soon as I clear some PR backlog, thank you! Other reviewers welcome.

hledger-web/hledger-web.cabal
```diff
@@ -103,6 +103,7 @@ extra-source-files:
     templates/edit-form.hamlet
     templates/journal.hamlet
     templates/manage.hamlet
+    templates/piechart.hamlet
```
simonmichael, line 106: Would you mind making this same change in the package.yaml file, from which this is generated. (Changing the .cabal file as well is optional, stack does it and I sometimes commit that change separately to reduce conflicts.)

simonmichael (changes requested):
@raboof: I've got this running locally. Great work so far. I'd like to improve usability a bit more if you have time. Do you think some of these could be done:

- Always show both charts, in the same stable positions. When one can't be shown for whatever reason, show it greyed out, or at least some placeholder. 
- Show the charts side by side, to use less vertical space. Negative amounts always on the left, positives on the right I guess.
- In the legend, show the amounts. I know you'll be short of space, maybe the percentage is not necessary there ?

simonmichael:
Also, I have seen it show two pies for one account, but most of the time it shows only one, or none. Perhaps there's a bug ? Eg with examples/sample.journal, a single pie is shown for expenses and for income, and no pies for any other account.

raboof:
> I'd like to improve usability a bit more (...) Always show both charts, in the same stable positions. When one can't be shown for whatever reason, show it greyed out, or at least some placeholder

Funny, I introduced hiding charts when there is no meaningful data a usability feature, not a shortcoming :D. Happy to adapt though.

> I have seen it show two pies for one particular account, but most of the time it shows only one, or none. Perhaps there's a bug ? Eg with examples/sample.journal, a single pie is shown for expenses and for income, and no pies for any other account.

It shows a pie chart when there is more than one (sub)account being reported on, unless those (sub)accounts contain multiple commodities. That was intentional, but if it creates more confusion than it reduces noise then we should indeed implement a more straightforward behaviour ;).

simonmichael:
I limited my testing to a single commodity, I'm pretty sure. See if you can reproduce the issue with examples/sample.journal. As a user I'm confused when the pies come and go, especially when I can't see the reason.

raboof:
> Eg with examples/sample.journal, a single pie is shown for expenses and for income, and no pies for any other account.

This matches what I would expect:
* expenses has 2 subaccounts with a positive balance, so those are shown in a pie chart
* income has 2 subaccounts with a negative balance, so those are shown in a pie chart
* 'assets' has 2 subaccoutns: 1 with negative balance and 1 with positive balance. Since showing a pie chart with just one '100%' category didn't feel very useful to me, I wasn't showing those.
* all other accounts don't have subaccounts, so don't have pie charts

So this was indeed my intended behavior - but evidently that wasn't very intuitive. I'll look into always showing both charts.

simonmichael:
My idea might not be right either. I just found it confusing as is.

simonmichael:
Hi @raboof, I just tested your latest push. 

The solid pie charts are a bit clunky of course, but for now, they do reassure that the feature is working..

Any idea why no charts are shown for the sample journal's assets:bank:checking ? http://127.0.0.1:5000/register?q=inacct%3Aassets%3Abank%3Achecking%20

Wouldn't it be more familiar to have the chart for positives on the right, and negatives on the left ?

I wonder how it feels if you show amounts instead of percentages ?

Do you think it's feasible to have some useful info (eg amount) always showing on or near the pie slices themselves ?

Xitian9:
I hate to be “that person”, especially when so much great work has been put in here, but I feel compelled to raise the point that it's notoriously difficult for people to be able to [extract accurate information from pie charts](https://scc.ms.unimelb.edu.au/resources-list/data-visualisation-and-exploration/no_pie-charts).

When I generate my financial charts, I use a stacked bar chart with absolute amounts on the left y-axis and percentage of total on the right y-axis for this purpose. It's easier to read, easier to display, and allows you to include both positive and negative amounts in a sensible way.

raboof:
> Any idea why no charts are shown for the sample journal's `assets:bank:checking`?

(will check when I get back to this)

> Wouldn't it be more familiar to have the chart for positives on the right, and negatives on the left?

Sounds reasonable, will update.

> I wonder how it feels if you show amounts instead of percentages ? Do you think it's feasible to have some useful info (eg amount) always showing on or near the pie slices themselves ?

I think that would be a great idea (though there's something to be said about percentages as well, as they are less likely to invoke 'fear of big numbers' :D ). Perhaps not absolutely necessary for this first iteration though - haven't looked into how hard it would be yet.

> I hate to be “that person”, especially when so much great work has been put in here

Don't worry ;)

> I feel compelled to raise the point that it's notoriously difficult for people to be able to [extract accurate information from pie charts](https://scc.ms.unimelb.edu.au/resources-list/data-visualisation-and-exploration/no_pie-charts).

To be honest I feel like many of those points don't apply so strongly here. Obviously the '3d' pie charts are pure evil. We don't really seem to suffer from any of the '4 signs that your pie is failing', mainly because it's interactive rather than a static image. The 'small percentages' aren't particularly important in our case.
 
> When I generate my financial charts, I use a stacked bar chart with absolute amounts on the left y-axis and percentage of total on the right y-axis for this purpose. It's easier to read, easier to display, and allows you to include both positive and negative amounts in a sensible way.

I would love to have those, and they might indeed replace pie charts. Until we have those, though, I think at least having the pie charts would be a significant improvement over not having them :).

simonmichael:
Good discussion. @raboof stop me if the below seems not the latest code, I think it is though. I'm testing with `make ghci`, `:main -f examples/sample.journal`.

- I want to see amounts, it's hard to understand what these pies are showing otherwise
- shows 0, 1 or 2 pies for each account, it's hard to understand why
- still showing positives on left, negatives on right
- click assets, move mouse pointer sideways across both pies; one of them is left in highlighted mouseover state
- should this show the subaccounts within the displayed account, or the other accounts being transacted with ?
  Currently it shows the former, but I think people are going to struggle with this. Maybe it should show the other instead,
  or maybe there's some way to make it more obvious what's being shown.

Sorry this is hard to get merged. The UI polish is a big part of the task, and in the present state I'm not sure it adds enough value to release it.

raboof:
> stop me if the below seems not the latest code, I think it is though

Yeah it is - I should've converted the PR to 'draft' though since I didn't yet find time to pick up some of the comments that I do plan to.

> Sorry this is hard to get merged.

No worries - I only find time to work on this occasionally so that rather drags things out wall-clock-wise, but so far I'm happy with the process.

raboof:
> should this show the subaccounts within the displayed account, or the other accounts being transacted with ?
Currently it shows the former, but I think people are going to struggle with this. Maybe it should show the other instead,
or maybe there's some way to make it more obvious what's being shown.

My use case is definitely 'subaccounts': I import and auto-classify my banking statements, and then use hledger-web to get a feel for 'where my money is going' (and later I'd like to add budgeting). The pies are nice for that while browsing the 'expenses' tree, like 'x % of my expenses in this period where utilities', with the option to from there 'zoom in' to the individual accounts under 'utilities', etc.

Showing charts of the 'accounts being transacted with' would also *kinda* work: I could open `assets:bank` and see what it transacts with. However, that would lose the hierarchy in the expenses. Also for more complicated journals it would seem browsing the 'expenses' tree hierarchically would make more sense than 'accounts being transacted with'.

simonmichael:
My guess is people most often are in the account register of their real world bank/credit card accounts. That's where I was, at least, and was getting confused. 

The same information would be available either way, it's just a question of which is shown where. I'd be curious to compare the other way.

raboof:
> My guess is people most often are in the account register of their real world bank/credit card accounts. That's where I was, at least, and was getting confused.

Interesting, I'm rarely there ;). Perhaps when we show (100%, but labeled) pie charts in more cases this will become less confusing.

> The same information would be available either way, it's just a question of which is shown where

The hierarchy would be different though, right?

simonmichael:
> The same information would be available either way, it's just a question of which is shown where
> 
> The hierarchy would be different though, right?
> 

You'd see pie charts of assets/liabilities in the revenue/expense account registers, and you'd see pie charts of revenues/expenses in the asset/liability account registers, I assume.

raboof:
> You'd see pie charts of assets/liabilities in the revenue/expense account registers, and you'd see pie charts of revenues/expenses in the asset/liability account registers, I assume.

Well, yes, but consider the example where you have an 'expense' account 'utilities' broken down in subaccounts 'gas', 'water' and 'electricity' that are paid out of the 'bank' assets account.

If you choose to show pie charts breaking down the 'subaccounts' of the selected account, then when browsing 'expenses' you see how many % of your expenses is 'utilities', and then can 'zoom in' to the 'utilities' account to see this broken down in 'gas', 'water' and 'electricity'.

If you choose to show pie charts breaking down the 'accounts being transacted with', then viewing 'bank' you'll only see *all* the expense accounts in one pie chart. Looking at the expense accounts you'll just see "100% bank". That seems to provide less insight to me. If you have multiple asset accounts then you can now see which asset accounts are used to cover which kinds of expenses, but TBQH that doesn't seem particularly interesting information to me.

simonmichael:
Does anyone have more thoughts on this ? Here are mine: there are a lot of different needs for charts and different aspects to making them usable and robust. I think we should do the R & D and prototyping outside of hledger-web where possible, eg using separate tools like ploterific (cf my recent discussion with semaj in #hledger and the new https://hledger.org/charting.html), as it's much cheaper to iterate and doesn't commit us prematurely. When we see some charts that seem to have wide usefulness, we can tackle moving them into hledger-web as a separate task.

simonmichael:
PS and so we can test and compare somewhat realistic examples, perhaps examples/bcexample.hledger will work as a standard data set ?

Xitian9:
I can supply a copy of the charts I generate from my hledger files (applied to the bcexample.journal file, modified slightly so that it works and produces nice-looking results), if that's helpful. What is the best way to share images?

simonmichael:
Yes please.  👍🏻 I guess just drag into the GitHub issue when commenting, probably dragging into an email reply also works. If we're chatting, dragging into Element works similarly, and non-Element can users can try the paste site shortcut in topic - I think it's https://pasteimg.hledger.org

simonmichael:
> the paste site shortcut in topic - I think it's https://pasteimg.hledger.org <https://pasteimg.hledger.org/>
(No longer in the topic, it's http not https, and now has ads so deprecating this one.

Xitian9:
Here is an overall breakdown of expenses and savings; the graph I use to take the place of a pie chart in showing proportion.
0-breakdown-0

A graph showing total cashflow in and out over the last year. Amounts are a simple moving average over the previous 14 days.
1-cashflow-0

Graph of net position, i.e. total assets minus total liabilities.
2-netposition-0

Graph of net liquid position.
3-liquid-0

Total assets and liabilities.  Since I haven't bothered separating liquid from non-liquid assets and liabilities, this is similar to the previous graph.
4-wealth-0

The next three are graphs of expense categories over the past year, grouped together so the axes aren't too different.
5-expenses-0
5-expenses-1
5-expenses-2

raboof:
> there are a lot of different needs for charts and different aspects to making them usable and robust.

Sure!

> I think we should do the R & D and prototyping outside of hledger-web where possible, eg using separate tools like ploterific

Can't hurt ;)

> the new https://hledger.org/charting.html

(https://hledger.org/charts.html?)

> When we see some charts that seem to have wide usefulness, we can tackle moving them into hledger-web as a separate task.

I'm personally mainly interested in 'interactively' browsing the data (selecting date ranges and subaccounts). Some graphs might not really be that useful when you have to switch between several tools to see them, but could be useful when integrated into hledger-web - so if we wait for them to "seem to have wide usefullness" outside of hledger-web they might never.

> we can test and compare somewhat realistic examples, perhaps examples/bcexample.hledger will work as a standard data set

I agree a standard example might be helpful. A complication in this particular dataset is that it includes expenses in different currencies (USD and IRAUSD), making them trickier to graph of course - which might be an advantage or a disadvantage depending how you look at it.

Do you have ideas on how to handle multiple currencies in one graph?

> Here is an overall breakdown of expenses and savings; the graph I use to take the place of a pie chart in showing proportion.

This is not bcexample.hledger, right? It's not clear to me why "Food" is split into ''Irregular expenses" and "Regular expenses", how does that work?

Xitian9:
> This is not bcexample.hledger, right? It's not clear to me why "Food" is split into ''Irregular expenses" and "Regular expenses", how does that work?

I classify groceries as regular expenses and restaurants and irregular expenses. I grant that this breakdown is specific to me, and it's probably not desirable to apply this widely.

simonmichael:
What should we do with this PR ?

raboof:
> What should we do with this PR ?

It's up to you: I still like the feature and am happily using this branch of hledger. Of course I'd hoped it'd be of use to more people than just myself, but I respect that you don't consider it acceptable for inclusion in this state. I don't really see how to make it more attractive, though :). I'm OK with either closing the PR or keeping it open for future suggestions.

simonmichael:
Hi @raboof. Are you still using this 2021 hledger pr: Pie chart ?

IIRC I felt it needs more refinement to be baked in to hledger-web. If you or anyone would like to rebase it, I'm happy to re-review it.

PR #1470, opened 2021-01-26
Fix ‘change of value’ reports to not see historical postings. Implement independent costing and valuation.

Xitian9:
This implements Tier 1 of https://github.com/simonmichael/hledger/issues/1353#issuecomment-762623077, minus the mockups of the documentation for Tier 1.1.


simonmichael:
Merged as 3429601. Thanks @Xitian9!

PR #1471, opened 2021-01-27
Allow selecting the date range from the chart

raboof:
By dragging a region with the mouse

select

### DIFF
--- a/hledger-lib/hledger-lib.cabal
+++ b/hledger-lib/hledger-lib.cabal
@@ -1,10 +1,10 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.33.0.
+-- This file has been generated from package.yaml by hpack version 0.34.3.
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 02ce04c7a484333ff2c8fbcaccbc26164fc238840f86cb69976f4d73280c9e7f
+-- hash: aa9217e7e1959660c6302d2761ceae1c141739e5e6524850c2b631ad00b41061
 
 name:           hledger-lib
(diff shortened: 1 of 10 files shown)


simonmichael:
Great!

When zoomed in, it doesn't seem to always pick the dates you'd want. Eg to view the transactions on a single day, you'd expect that selecting a region that contains the dot will show the transactions on that day, but sometimes that day is excluded. See if you can reproduce, and maybe expand the selection in a smart way ?

raboof:
> to view the transactions on a single day, you'd expect that selecting a region that contains the dot will show the transactions on that day, but sometimes that day is excluded. See if you can reproduce, and maybe expand the selection in a smart way ?

Aah, indeed it was rounding down for the 'to' date, where it should been rounding up. Updated.

simonmichael:
That's much better. One more thing that may not be your fault, but would improve the feature I think. Testing with `make ghci-web`, `:main -f examples/sample.journal`:

- Click assets, select the two months enclosing the three central points.
  In the query field I see the span I selected, eg `date:2008/4/30-2008/6/29`.
  But the chart is shrunk to the points' span (6/1 to 6/3).
  It would be more intuitive if it showed the span that was selected.

raboof:
> One more thing that may not be your fault, but would improve the feature I think

I agree, but indeed pretty sure that's not directly related to this PR. Maybe let's create a separate issue for that?

simonmichael:
Ok, would you mind creating it ? Merging this one, thanks a lot !

raboof:
> Ok, would you mind creating it ?

Sure, #1485

> Merging this one, thanks a lot !

Thanks for the merge!

PR #1477, opened 2021-02-04
print, ui transaction screen: always show all decimal places  (#931)

### LINKED ISSUE #931: print can show unbalanced entries due to display rounding
This frequently annoys me when I need to re-parse print's output, eg when doing a query that requires a pipeline. Time to do something about it.

```
; limit display precision to 2 places
commodity $0.00

1/1
   a  -$0.01
   b   $0.005
   c   $0.005
```
print shows an unbalanced transaction:
```
$ hledger print
2018/01/01
    a          $-0.01
    b               0
    c               0

```
which can not be re-parsed:
```
$ hledger print | hledger -f- print
hledger: "-" (lines 1-4)
could not balance this transaction (real postings are off by $-0.01)
2018/01/01
    a          $-0.01
    b               0
    c               0


```

simonmichael:
Ensures parseable and more sensible-looking output in more cases, and behaves more like Ledger's print.

There is still an issue with adding trailing zeroes, which would be nice to prevent.

Fixes #931 (and makes it easier/more reliable to chain hledger commands in a pipeline).


simonmichael:
I'm just going to optimistically merge this, let me know if you discover any problems.

simonmichael:
timeclock amounts often have recurring decimals, which looked ugly when printed (#1527). Resolved by always rounding these to two decimal places internally.

PR #1480, opened 2021-02-08
Implement new --valuechange interface, restore old --value=end behaviour.

Xitian9:
Addresses some of the discussions of #1353.

I've done my best to make sure this matches up with the agreed-upon interface, and that the tests are correct and comprehensive, but given how much change there's been here in the last few months it's very possible I've made a mistake somewhere. Extra eyes would be appreciated.

### DIFF
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -50,7 +50,7 @@ import Data.Semigroup ((<>))
 #endif
 import Data.Semigroup (sconcat)
 import Data.Time.Calendar (Day, addDays, fromGregorian)
-import Safe (headMay, lastDef, lastMay)
+import Safe (headMay, lastDef, lastMay, minimumMay)
 
 import Hledger.Data
 import Hledger.Query
@@ -197,7 +197,9 @@ startingBalances rspec@ReportSpec{rsQuery=query,rsOpts=ropts} j priceoracle repo
     rspec' = rspec{rsQuery=startbalq,rsOpts=ropts'}
     -- If we're re-valuing every period, we need to have the unvalued start
     -- balance, so we can do it ourselves later.
-    ropts' = if changingValuation ropts then ropts''{value_=Nothing} else ropts''
+    ropts' = case value_ ropts of
+        Just (AtEnd _) -> ropts''{value_=Nothing}
+        _              -> ropts''
       where ropts'' = ropts{period_=precedingperiod, no_elide_=accountlistmode_ ropts == ALTree}
 
     -- q projected back before the report start date.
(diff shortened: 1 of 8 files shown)


Xitian9:
Are we looking at getting this merged in time for 1.21? Let me know what I can do to make this easier to merge/send back for changes/reject.

simonmichael:
Thanks for the ping. I have left it pretty late, but yes I'll try to merge #1445 and this asap.

simonmichael:
This PR seems to *rename* --change to --periodic. Isn't that different from what we were planning on #1353 ? Copying here for easy reference:

> Report types - what is calculated/shown in each cell:
> --change : show sum of posting amounts (default)
> --budget : show sum and goal
> --valuechange : show change of value of period-end historical balances
> --gain : show change of value of period-end historical balances caused by market price fluctuations
> 
> Accumulation mode - which earlier values are included in calculations:
> --periodic : from column start to column end (default)
> --cumulative : from report start (specified by -b/--begin, eg) to column end
> --historical/-H : from journal start to column end
> 
> Valuation type - which valuation date(s) to use (and optionally which target commodity):
> --value=then : transaction dates
> --value=end : period end date(s)
> --value=now : today's date
> --value=DATE : other date
> --value=cost : legacy/convenience alias for --cost/-B: don't value, just convert to cost (I'm finding this still useful for comparing different reports)

simonmichael:
Oh maybe disregard, I see --change added again in the second commit.

simonmichael:
It looks good to me. 

Not a blocker but I think it's going to be a little hard for folks to understand how and when to use/combine these 6+ report type/accumulation type flags - perhaps we'll tweak the help at some point, maybe something like this ?

```
balance [OPTIONS] [REPORTTYPE] [ACCUMULATIONTYPE] [VALUATIONTYPE] [QUERY]
  Show accounts and their balances.
  REPORTTYPE is --change|--budget|--valuechange (default: --change).
  ACCUMULATIONTYPE is --periodic|--cumulative|--historical|-H (default: --periodic for most reports, --historical for balancesheet/balancesheetequity/cashflow)
  VALUATIONTYPE is --cost|--market|--exchange|--value|-B|-V|-X (default: no valuation)
```

(I've put --budget after --change this time since those are more like each other than --valuechange.)

I'll update my balance docs mockup and merge it, separately from this PR I think since it has a bunch of other rewrites.

Xitian9:
That sounds good to me. It's probably not worth mentioning at that point in the documentation that the default valuation type is `--value=end` for `--valuechange`.

simonmichael:
I haven't quite finished working on docs, but they'll follow soon. Merging this. Thank you very much!

simonmichael:
Docs updated: https://hledger.org/hledger.html#balance

Issue #1473, opened 2021-01-30
[ROI] Simplify the `roi` command 


artkpv:
As per this discussion at https://groups.google.com/g/hledger/c/nDlfa-j2GRw

Suggestions are

1. Remove --investment parameter but just leave the query so it is similar to `bal` and `reg`. Thus `roi` acts "the same", ie. filtering only those transactions. And --pnl would be used in the same way, ie. it is required, it specifies profit/loss account, etc. 
2. Into docs something like: "Only transactions included in --investment are taken for roi, eg. if you wish to track dividends their transactions should include --investment account. All outcoming amounts from the --investment account not in --pnl account are counted as withdrawals or incomes, eg. to track your end of year roi your fake transaction should be only wihtin the --investment account.

Example
```

; hledger roi -f scz_example.ledger --inv assets:equity:A --pnl income:equity:A -e 2017-01-01 --cashflow -Y -X D

2016-02-02 Buying A
    assets:equity:A                                     100.0 A @ 100.0 D
    assets:bank

2016-06-28 Dividends
    assets:bank                                         200 D
    assets:equity:A                                       0             ;  So that --inv gets this.
    income:equity:A

2016-06-28 Dividends
    assets:bank                                        -200 D
    assets:equity:A                                       0 
    assets:bank

2016-12-31 Fake - to track PnL
    assets:equity:A                                    -100.0 A @ 100.0 D
    assets:equity:A                                   11000.0 D            ; Should be in --inv!
    income:equity:A                                   -1000 D

2017-01-01 Fake - to track PnL
    assets:equity:A                                     100.0 A @ 100.0 D
    assets:equity:A                                  -11000.0 D
    income:equity:A                                    1000 D

```

artkpv:
Btw, here is a script for creating fake end transactions. https://gist.github.com/artkpv/5f682827841763c827db6edef78a608b

adept:
#1483 should remove the need for this script

PR #1483, opened 2021-02-09
roi: use Valuations throughout the code

### LINKED ISSUE #1267: Improve UX of ROI command
Hello all,

First things first, thank you all for this very nice software ! 👍 

Few days ago, I suggested via IRC (ping @adept) that the ROI command could be a little bit more ux-friendly by removing the need of the --pnl flag.

Here's an example of how I understand the ROI commands works now

1. Buy assets and record offset of pnl gain/loss
```
2019-06-20 Gettin' Rich
  Assets:cash   - $300
  Investment:APPL

2020-06-23 pnl valuation
  investment:APPL   = $25
  Unrealized:APPL
```
2. Enjoy roi
```
➜ hledger roi --investment='Investment:APPL' --pnl='Unrealized:APPL'
+---++------------+------------++---------------+----------+-------------+-----++-------+-------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) | PnL ||   IRR |   TWR |
+===++============+============++===============+==========+=============+=====++=======+=======+
| 1 || 2019-06-20 | 2020-06-23 ||             0 |      300 |         325 |  25 || 8.22% | 8.21% |
+---++------------+------------++---------------+----------+-------------+-----++-------+-------+
```

This approach requires to offset pnl of each investments into a dedicated account (eg. Unrealized:APPL in this case), which I found somehow not very practical since it spams my journal by 'not real' transaction.

Moreover, I guess this needs of offseting latest pnl is redundant with the P directive (cf. https://hledger.org/journal.html#declaring-market-prices) since the pnl is only a subtraction of the latest price minus the buying price and then could be easily calculated by the command.

In my opinion, the ROI commands should work this way:
1. Buy assets and record new price with the P directive
```
2019-06-20 Gettin' Rich ; buy some APPL
  Assets:cash   - $300
  Investment:APPL

P 2020-06-15 APPL  $ 280.80 ; record new price
P 2020-06-15 APPL  $ 290.43 ; same
```
And then the ROI command should be able to calculate the latest pnl for the selected commodity, maybe with a new flag like that for example
```
➜ hledger roi --investment='Investment:APPL' --commodity='AAPL'
+---++------------+------------++---------------+----------+-------------+-----++-------+-------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) | PnL ||   IRR |   TWR |
+===++============+============++===============+==========+=============+=====++=======+=======+
| 1 || 2019-06-20 | 2020-06-23 ||             0 |      300 |         325 |  25 || 8.22% | 8.21% |
+---++------------+------------++---------------+----------+-------------+-----++-------+-------+
```

Curious to have opinon around this wish, let me know if it seems relevant 😳

adept:
with this change, roi supports full variety of --value switches AND takes into account value changes due to price directives.

Instead of valuing all transactions upfront and working with the result, roi now applies valuations if/when needed, with proper date/periodend values, and additionally visits all dates on which we have price directives to make sure it "samples" value of the investment on that date.

This renders unnecessary "fake sell-rebuy" transactions that were shown in examples/roi-unrealised.journal (and mentioned as part of the workaround script in #1473) and closes #1267

### DIFF
--- a/examples/roi-unrealised.ledger
+++ b/examples/roi-unrealised.ledger
@@ -3,14 +3,14 @@
 ; By adding some fake transactions you can make it show unrealised gains.
 
 comment
-$ hledger -f examples/roi-unrealised.ledger roi --inv Assets -b 2015-01-01 -e 2019-01-01 --pnl  Income --yearly
+$ hledger -f examples/roi-unrealised.ledger roi --inv Assets -b 2015-01-01 -e 2019-01-01 --pnl  Income --yearly --value=then
 +---++------------+------------++---------------+----------+-------------+--------++--------+--------+
 |   ||      Begin |        End || Value (begin) | Cashflow | Value (end) |    PnL ||    IRR |    TWR |
 +===++============+============++===============+==========+=============+========++========+========+
-| 1 || 2015/01/01 | 2015/12/31 ||         717.4 |        0 |      756.30 |  38.90 ||  5.42% |  5.42% |
-| 2 || 2016/01/01 | 2016/12/31 ||        756.30 |        0 |      998.80 | 242.50 || 31.96% | 31.96% |
-| 3 || 2017/01/01 | 2017/12/31 ||        998.80 |        0 |     1151.20 | 152.40 || 15.26% | 15.26% |
-| 4 || 2018/01/01 | 2018/12/31 ||       1151.20 |        0 |     1145.00 |  -6.20 || -0.54% | -0.54% |
+| 1 || 2015-01-01 | 2015-12-31 ||        725.50 |        0 |      756.30 |  30.80 ||  4.24% |  4.25% |
+| 2 || 2016-01-01 | 2016-12-31 ||        756.30 |        0 |      998.80 | 242.50 || 31.96% | 31.96% |
+| 3 || 2017-01-01 | 2017-12-31 ||        998.80 |        0 |     1151.20 | 152.40 || 15.26% | 15.26% |
+| 4 || 2018-01-01 | 2018-12-31 ||       1151.20 |        0 |     1145.00 |  -6.20 || -0.54% | -0.54% |
 +---++------------+------------++---------------+----------+-------------+--------++--------+--------+
 end comment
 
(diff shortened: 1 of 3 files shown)


Xitian9:
This looks great, but I think there's a slight problem in terms of `--value=then` valuation. Looking at the code, it looks like instead of using the posting date it's using the beginning of the period span in which it lies.

The fundamental problem: `--value=then` requires keeping the postings around for valuation, `--value=end` requires knowing the Datespan in which you're valuing. The code needs to be refactored so that either both these pieces of information are available at the same time, or the valuation needs to be done in two different places depending on the valuation strategy.

Functions which may be of use:
https://github.com/simonmichael/hledger/blob/94c0832b85f1100e7cb04a9d5189e064261744ae/hledger-lib/Hledger/Reports/MultiBalanceReport.hs#L250-L252

https://github.com/simonmichael/hledger/blob/94c0832b85f1100e7cb04a9d5189e064261744ae/hledger-lib/Hledger/Reports/MultiBalanceReport.hs#L573-L577

adept:
> This looks great, but I think there's a slight problem in terms of `--value=then` valuation. Looking at the code, it looks like instead of using the posting date it's using the beginning of the period span in which it lies.

Hmm. Perhaps I misunderstood the meaning of arguments for `mixedAmountApplyCostValuation`? There is `periodlast`, `today` and `date`, and I took them to mean:

1. last date of (reporting) period
2. literally today 
3. date of the posting

I do not remember consciously writing a line of code that would value the posting using the start of the period span in which it lies, but I could of course have made all kinds of bugs there. Can you give me a pointer to the place in code where you saw that?

adept:
Sorry, I am looking at the functions you have referenced and I must confess that I just can't see what problem they could solve for me (perhaps because I am not seeing the problem yet :).  In my (surely, somewhat limited) testing my uses of `mixedAmountApplyCostValuation` behaved exactly how I wanted them to. Perhaps I am missing a compelling example that would show where my computations go wrong.

Xitian9:
> Hmm. Perhaps I misunderstood the meaning of arguments for `mixedAmountApplyCostValuation`? There are `periodlast`, `today` and `date`, and I took them to mean:
> 
> 1. last date of (reporting) period
> 2. literally today
> 3. date of the posting

That is correct.

Ah, this is probably my mistake. I was seeing the beginning of the span being used in valueBefore (and the end of the span in valueAfter), but completely missed its use in cashFlowApplyCostValue. I apologise for the noise.

adept:
Upd: fixes #1484 as well

PR #1491, opened 2021-02-24
MixedAmount and other cleanups

Xitian9:
This is a collection of refactorings, cleanups, and performance improvements. I think that all of this should be relatively uncontroversial, though there are a couple of follow-ups this suggests which may be controversial (I have therefore broken them off into a separate PR which will follow this).

Points to consider:
- The `Num` instance for `MixedAmount` is no longer used anywhere in this codebase. We could just remove it (morally the right thing to do), but this would have a big effect on the API.
- I have gone through and replaced every usage of `sum` for `MixedAmount` with `mconcat` (or `foldMap` applied to an appropriate function), and every usage of `sumStrict` with `foldMap'` applied to an appropriate function. Since all these locations are now marked, we may want to consider if any of the non-strict versions should be made strict (or vice versa). (Edit: Apparently I've already strictified some uses of `sum`, but there may be more.)

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -530,7 +530,7 @@ nullmixedamt = Mixed []
 
 -- | A temporary value for parsed transactions which had no amount specified.
 missingmixedamt :: MixedAmount
-missingmixedamt = Mixed [missingamt]
+missingmixedamt = mixedAmount missingamt
 
 -- | Convert amounts in various commodities into a normalised MixedAmount.
 mixed :: [Amount] -> MixedAmount
@@ -964,37 +964,37 @@ tests_Amount = tests "Amount" [
         @?= Mixed [usd (-1) @@ eur 2 ]
 
     ,test "showMixedAmount" $ do
-       showMixedAmount (Mixed [usd 1]) @?= "$1.00"
-       showMixedAmount (Mixed [usd 1 `at` eur 2]) @?= "$1.00 @ €2.00"
-       showMixedAmount (Mixed [usd 0]) @?= "0"
-       showMixedAmount (Mixed []) @?= "0"
+       showMixedAmount (mixedAmount (usd 1)) @?= "$1.00"
+       showMixedAmount (mixedAmount (usd 1 `at` eur 2)) @?= "$1.00 @ €2.00"
+       showMixedAmount (mixedAmount (usd 0)) @?= "0"
+       showMixedAmount nullmixedamt @?= "0"
        showMixedAmount missingmixedamt @?= ""
 
     ,test "showMixedAmountWithoutPrice" $ do
       let a = usd 1 `at` eur 2
-      showMixedAmountWithoutPrice False (Mixed [a]) @?= "$1.00"
-      showMixedAmountWithoutPrice False (Mixed [a, -a]) @?= "0"
+      showMixedAmountWithoutPrice False (mixedAmount (a)) @?= "$1.00"
+      showMixedAmountWithoutPrice False (mixed [a, -a]) @?= "0"
 
     ,tests "normaliseMixedAmount" [
        test "a missing amount overrides any other amounts" $
-        normaliseMixedAmount (Mixed [usd 1, missingamt]) @?= missingmixedamt
+        amounts (normaliseMixedAmount $ mixed [usd 1, missingamt]) @?= [missingamt]
       ,test "unpriced same-commodity amounts are combined" $
-        normaliseMixedAmount (Mixed [usd 0, usd 2]) @?= Mixed [usd 2]
+        amounts (normaliseMixedAmount $ mixed [usd 0, usd 2]) @?= [usd 2]
       ,test "amounts with same unit price are combined" $
-        normaliseMixedAmount (Mixed [usd 1 `at` eur 1, usd 1 `at` eur 1]) @?= Mixed [usd 2 `at` eur 1]
+        amounts (normaliseMixedAmount $ mixed [usd 1 `at` eur 1, usd 1 `at` eur 1]) @?= [usd 2 `at` eur 1]
       ,test "amounts with different unit prices are not combined" $
-        normaliseMixedAmount (Mixed [usd 1 `at` eur 1, usd 1 `at` eur 2]) @?= Mixed [usd 1 `at` eur 1, usd 1 `at` eur 2]
+        amounts (normaliseMixedAmount $ mixed [usd 1 `at` eur 1, usd 1 `at` eur 2]) @?= [usd 1 `at` eur 1, usd 1 `at` eur 2]
       ,test "amounts with total prices are combined" $
-        normaliseMixedAmount (Mixed  [usd 1 @@ eur 1, usd 1 @@ eur 1]) @?= Mixed [usd 2 @@ eur 2]
+        amounts (normaliseMixedAmount $ mixed [usd 1 @@ eur 1, usd 1 @@ eur 1]) @?= [usd 2 @@ eur 2]
     ]
 
     ,test "normaliseMixedAmountSquashPricesForDisplay" $ do
-       normaliseMixedAmountSquashPricesForDisplay (Mixed []) @?= Mixed [nullamt]
+       amounts (normaliseMixedAmountSquashPricesForDisplay nullmixedamt) @?= [nullamt]
        assertBool "" $ mixedAmountLooksZero $ normaliseMixedAmountSquashPricesForDisplay
-        (Mixed [usd 10
+        (mixed [usd 10
                ,usd 10 @@ eur 7
                ,usd (-10)
-               ,usd (-10) @@ eur 7
+               ,usd (-10) @@ eur (-7)
                ])
 
   ]
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -94,6 +94,9 @@ module Hledger.Data.Amount (
   nullmixedamt,
   missingmixedamt,
   mixed,
+  mixedAmount,
+  maAddAmount,
+  maAddAmounts,
   amounts,
   filterMixedAmount,
   filterMixedAmountByCommodity,
@@ -104,12 +107,18 @@ module Hledger.Data.Amount (
   mixedAmountStripPrices,
   -- ** arithmetic
   mixedAmountCost,
+  maNegate,
+  maPlus,
+  maMinus,
+  maSum,
   divideMixedAmount,
   multiplyMixedAmount,
   averageMixedAmounts,
   isNegativeAmount,
   isNegativeMixedAmount,
   mixedAmountIsZero,
+  maIsZero,
+  maIsNonZero,
   mixedAmountLooksZero,
   mixedAmountTotalPriceToUnitPrice,
   -- ** rendering
@@ -138,12 +147,12 @@ import Control.Monad (foldM)
 import Data.Decimal (DecimalRaw(..), decimalPlaces, normalizeDecimal, roundTo)
 import Data.Default (Default(..))
 import Data.Foldable (toList)
-import Data.List (intercalate, intersperse, mapAccumL, partition)
+import Data.List (foldl', intercalate, intersperse, mapAccumL, partition)
 import Data.List.NonEmpty (NonEmpty(..), nonEmpty)
 import qualified Data.Map.Strict as M
 import Data.Maybe (fromMaybe)
 #if !(MIN_VERSION_base(4,11,0))
-import Data.Semigroup ((<>))
+import Data.Semigroup (Semigroup(..))
 #endif
 import qualified Data.Text as T
 import qualified Data.Text.Lazy.Builder as TB
@@ -494,13 +503,26 @@ canonicaliseAmount styles a@Amount{acommodity=c, astyle=s} = a{astyle=s'}
 -------------------------------------------------------------------------------
 -- MixedAmount
 
+instance Semigroup MixedAmount where
+  (<>) = maPlus
+
+instance Monoid MixedAmount where
+  mempty = nullmixedamt
+#if !(MIN_VERSION_base(4,11,0))
+  mappend = (<>)
+#endif
+
 instance Num MixedAmount where
-    fromInteger i = Mixed [fromInteger i]
-    negate (Mixed as) = Mixed $ map negate as
-    (+) (Mixed as) (Mixed bs) = normaliseMixedAmount $ Mixed $ as ++ bs
-    (*)    = error' "error, mixed amounts do not support multiplication" -- PARTIAL:
-    abs    = error' "error, mixed amounts do not support abs"
-    signum = error' "error, mixed amounts do not support signum"
+  fromInteger i = Mixed [fromInteger i]
+  negate = maNegate
+  (+)    = maPlus
+  (*)    = error' "error, mixed amounts do not support multiplication" -- PARTIAL:
+  abs    = error' "error, mixed amounts do not support abs"
+  signum = error' "error, mixed amounts do not support signum"
+
+-- | Get a mixed amount's component amounts.
+amounts :: MixedAmount -> [Amount]
+amounts (Mixed as) = as
 
 -- | The empty mixed amount.
 nullmixedamt :: MixedAmount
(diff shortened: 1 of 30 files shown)


Xitian9:
Some quick benchmarks:

```
Running 9 tests 1 times with 2 executables at 2021-03-03 10:43:46 AEDT:
 (error: )  (error: ) 
Best times:
+-----------------------------------------------------++--------------------+-------------------------------+
|                                                     || bin/hledger-master | bin/hledger-mixedamt-cleanups |
+=====================================================++====================+===============================+
| -f examples/10000x1000x10.journal print             ||               8.27 |                          8.28 |
| -f examples/10000x10000x10.journal print            ||               8.28 |                          8.43 |
| -f examples/10000x1000x10.journal register          ||               8.67 |                          8.93 |
| -f examples/10000x10000x10.journal register         ||               8.81 |                          8.60 |
| -f examples/10000x1000x10.journal balance           ||               7.52 |                          7.59 |
| -f examples/10000x10000x10.journal balance          ||              14.54 |                          8.30 |
| -f examples/1000x1000x10.journal balance --weekly   ||               2.16 |                          2.13 |
| -f examples/10000x1000x10.journal balance --weekly  ||              23.34 |                         23.64 |
| -f examples/10000x10000x10.journal balance --weekly ||             105.70 |                        105.67 |
+-----------------------------------------------------++--------------------+-------------------------------+
```

There are mostly no changes, but the performance improvements in `-f examples/10000x10000x10.journal balance` are due to f1f36c7, because that's the case in which sorting the accounts becomes the dominant performance sink.

Xitian9:
Benchmarks updated to include ad4da85

```
+-----------------------------------------------------++---------------------------+----------------------------+
|                                                     || hledger-mixedamt-cleanups | hledger-mixedamt-cleanups' |
+=====================================================++===========================+============================+
| -f examples/10000x1000x10.journal print             ||                      8.39 |                       7.83 |
| -f examples/10000x10000x10.journal print            ||                      7.86 |                       7.61 |
| -f examples/10000x1000x10.journal register          ||                      8.31 |                       8.11 |
| -f examples/10000x10000x10.journal register         ||                      8.94 |                       9.92 |
| -f examples/10000x1000x10.journal balance           ||                      7.96 |                       7.04 |
| -f examples/10000x10000x10.journal balance          ||                      7.89 |                       7.78 |
| -f examples/1000x1000x10.journal balance --weekly   ||                      2.07 |                       1.89 |
| -f examples/10000x1000x10.journal balance --weekly  ||                     22.13 |                      21.21 |
| -f examples/10000x10000x10.journal balance --weekly ||                    103.22 |                      95.66 |
+-----------------------------------------------------++---------------------------+----------------------------+

```

Xitian9:
By the way, I'm assuming nobody's looked at this yet, and so am adding more commits of a similar nature, rather than making separate PRs. If you'd like me to stop so this PR isn't a moving target, just say so and I can break them off into a separate PR.

simonmichael:
Correct, thanks. I'm not sure which way will be easiest to manage when the time comes, so as you think best. Sorry for the time lag.

simonmichael:
The slowdown was with this PR, right ? I'm still assuming it's not vital to include this in 1.21.

Xitian9:
Yes, it's not vital to include this in 1.21.

Yes the slowdown was within this PR. It was due to two minor bugs in master which cancelled each other out; when I fixed one of them, the other took effect and slowed everything down.

hledger-lib/Text/Megaparsec/Custom.hs
```diff
@@ -244,7 +244,7 @@ customErrorBundlePretty errBundle =
       _ -> Nothing
 
     finds :: (Foldable t) => (a -> Maybe b) -> t a -> Maybe b
-    finds f = asum . map f . toList
+    finds f = getAlt . foldMap (Alt . f)
```
simonmichael, line 247: I think Alt will be new to many readers. Is it an improvement ?

simonmichael: (It was new to me, and I had to go have a discussion on #haskell about it. But I just realised I don't care, disregard this comment.)

Xitian9: For a simple one-liner, I actually went back an forth a lot on this one. Ideally I would have used `First` instead of `Alt`, but there are actually two different `First`s: one in `Data.Monoid`, which would be `finds f = getFirst . foldMap (First . f)`, and one in `Data.Semigroup`, which would be `finds f = fmap getFirst . foldMap (fmap First . f)`. The former is in all supported versions of base, but was scheduled to disappear a couple of versions ago, so I don't think it will last much longer. The latter will be around for the forseeable future, but requires a Semigroup instance for Maybe, which isn't in all supported versions. I decided `Alt` was the best choice since it doesn't require messing around with any CPP.

simonmichael:
**lib,cli,ui,bin: Add Semigroup, Monoid, don't use Num instances of MixedAmount.**: I'm a bit sad about this one. The Num instance, even if imperfect, was quite convenient at times, I suspect pleasing to API users/contributors, and concise. Now instead of 0, -, +, sum, we have zeromixedamt, negateMixedAmount, <>, foldMap', and a fair bit of additional import and CPP noise. On the other hand, it may be worth it for a more rigorous API with fewer ways to misuse it. I'd like a more informative alias for foldMap' perhaps, sumMixedAmounts.

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -712,6 +734,9 @@ showMixedAmountDebug m | m == missingmixedamt = "(missing)"
                        | otherwise       = printf "Mixed [%s]" as
     where as = intercalate "\n       " $ map showAmountDebug $ amounts m
 
+showMixedAmountB :: AmountDisplayOpts -> MixedAmount -> WideBuilder
```
simonmichael: Let's give this a haddock, and update the other two haddocks. They should help explain when to use each of these, ie mention the different normalisation behaviour.

Xitian9:
> **lib,cli,ui,bin: Add Semigroup, Monoid, don't use Num instances of MixedAmount.**: I'm a bit sad about this one. The Num instance, even if imperfect, was quite convenient at times, I suspect pleasing to API users/contributors, and concise. Now instead of 0, -, +, sum, we have zeromixedamt, negateMixedAmount, <>, foldMap', and a fair bit of additional import and CPP noise. On the other hand, it may be worth it for a more rigorous API with fewer ways to misuse it.

I agree that we would be sacrificing ergonomics for rigour. The most annoying change is that `a - b` becomes `a <> negateMixedAmount b`. To me, the big advantage of the semigroup/monoid interface is that the typeclasses give more control over behaviour. For example, `sum` is a `Foldable` method, but `mconcat` is a `Monoid` method. So if we ever decide we can improve efficiency by making sum strict, we can do that at one point.

> I'd like a more informative alias for foldMap' perhaps, sumMixedAmounts.
Perhaps `sumOver`, so the dominant use case becomes `sumOver mixed`.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: Haddocks are added in the next commit. Sorry, I structured this and the following three commits to minimise the diffs. I was trying to track down a subtle bug here, and couldn't figure out where it was.

I could add a note on the different normalisation behaviour, but my hope is that such a note would become redundant after this and the next PR. The eventual goal is that `MixedAmount` is always normalised by construction, and in the instances where you don't want it normalised (i.e. just the contents of `pamount`) you just use `[Amount]`. Let me know if you want me to update those haddocks now.

Xitian9:
As a side note: we could actually go for a more powerful abstraction than Monoid for MixedAmount. Mathematically it should be a vector space over the ring `Quantity`, but such a typeclass does not exist outside of Numeric Haskell and the like. If we really felt like it, it would be a natural `Quantity`-reducer ([see here](https://hackage.haskell.org/package/reducers-3.12.3/docs/Data-Semigroup-Reducer.html)), but I highly doubt we want to go down that road.

Xitian9:
It should be noted that this PR does not actually remove the `Num` instance: it just adds the `Semigroup` and `Monoid` instances and gets rid of all uses of the `Num` instance internally. Actually removing the `Num` instance would break a lot of people's code, so if we were to do it we would need a migration plan.

The “bad Num usage” that I view as most likely is people trying to do something like `2 * a`, instead of `multiplyMixedAmount 2 a`. It looks like it should work, makes sense intuitively, and compiles, but it will throw a runtime error. `abs` also looks like it should work, and is something that has tripped me up in the past: the lack of a reasonable corresponding definition of `signum` is the only thing that makes it a bad idea.

Xitian9:
I've added a commit which eliminates `zeromixedamt`. Using `nullmixedamt`/`mempty` now produced identical results in all our tests. This should simplify the ergonomics.

Xitian9:
Looking into it, it seems that almost every use of `foldMap'` could be replaced with a use of `sumPostings`, and the two left over could be refactored away without much fuss. I've added a commit which does this.

hledger-lib/Hledger/Data/Amount.hs
simonmichael: Yes since it's a small thing let's add and update the haddocks now, making this commit/PR more self-contained and not dependent on future work. Haddocks don't have to be fancy, one clear sentence is usually enough.

simonmichael:
Good discussion. My feeling is that in the long run, highly readable code and intuitive APIs are the most important thing. Performance/efficiency is next most important. 

For greatest abstraction of implementation details, wouldn't it be best to provide a full custom API/DSL for all amount operations, and avoid Num/Monoid/Semigroup operators entirely ?

simonmichael:
That said, I'm not yet arguing that we shouldn't use (internally) or offer (externally) convenient typeclass-based operators, perhaps in addition. As a matter of practicality, not to mention backward compatibility as you rightly point out. I don't even necessarily mind partial/unlawful typeclass implementation, if we provide clear guidance on it. Speaking of, I wonder if there are creative ways to make a more fully "lawful" Num instance for MixedAmount ? What's the problem with `*` and `abs` ?

Xitian9:
I think the main problem with `(*)` is that it doesn't make sense semantically. We may be able to make sense of `Num 2 * $3`, what what about `2€ * $3`? We would need to create a new currency, the Dollar-Euro, and what use is it?

We could create an implementation for `abs`, taking the absolute value of each commodity in turn, but what would `signum` be? We need to pick a single number, which doesn't work if we're multiplying each commodity by a different number. It wouldn't be the biggest problem in the world to define `abs` without `signum`, but I'm not sure there's a big use for it.

simonmichael:
Ah yes, I was thinking of `* N` (multiply by a scalar), which is not a full Num `*`. Agreed that there's no sensible full implementation of these and they'd have to be left as unimplemented partials (error).

Xitian9:
My biggest problem with the `Num` instance comes down to the fact that it makes it convenient to write subtly buggy code. What is the zero of `MixedAmount`? Is it `nullmixedamt`/`mempty`, or is it `Num 0`/`zeromixedamt`/`0`. It's not obvious, and the easiest thing to write is `0`, so people will go with that.

But `zeromixedamt` doesn't behave like `0` usually does. It's close, but `0 + a /= a` in general. This means that adding `zeromixedamt` will change the value, and you can see lots of places in the code where people add zero to things. This is actually necessary: if you remove those (without a little thought), the code gives a runtime error. This is quite common in `Roi.hs` – look in af1fbc7c03e29bf82db6aeebb536ec426f56fea5 and you can see a couple – but there are also some instances in `MultiBalanceReport.hs` and other places. `Roi.hs` even filters on whether a MixedAmount `/=0`. You might think that you can get rid of these by just replacing those instances of `0` with `nullmixedamt`, but that doesn't fully work, because `sum` returns `zeromixedamt` on an empty list.

Basically, so long as we're using the Num instance, we have to live with the fact that there are two special values – `zeromixedamt` and `nullmixedamt` – which need to be checked for and handled appropriately. If we don't use the Num instance, then `zeromixedamt` never shows up, and the code just has to handle `nullmixedamt`, which as the Monoid unit is generally gracefully handled. (Another result of this PR is that all API functions now handle `nullmixedamt` and `zeromixedamt` the same, to the greatest extend possible, so unless you're doing something low-level, like checking equality, they should now be interchangeable).

This is not a huge issue, but I've actually lost a fair amount of time squashing these sorts of bugs.

Xitian9:
I'll squash ad8ed79 into 3a8c3e4 (and update the haddocks to mention normalisation), 0935c7e into e40223b, and af1fbc7 into 573cfe4.

simonmichael:
I don't have recent familiarity with this code, but I trust your remarks. Would I be right in saying this PR 

- includes both performance fixes and typeclass cleanups, which are more or less independent
- and the intent of typeclass cleanups is that we stop treating MixedAmounts like Num-bers and start treating them like Monoids
- we're not removing the Num instance yet, but we're thinking about it for later, and this might affect API users
-

Xitian9:
Yes, correct. The only sense in which they are not independent is that rebasing to disentangle them is a bit of a nightmare.

simonmichael:
Ok, don't worry about that then.

simonmichael:
I expect we can merge this when you are ready, but would you mind also summarising the user-visible API changes in the style of http://hackage.haskell.org/package/hledger-lib-1.21/changelog - here is fine - so we can take another look at the amounts API.

Xitian9:
We are now de-emphasising the Num instance of MixedAmount in favour of the Monoid instance. Though the old interface will continue to work for the time being, we recommend the following API changes.
Hledger.Data.MixedAmount:
 (+) -> (<>)
 a - b -> a <> negateMixedAmount b
 negate -> negateMixedAmount
 sum -> mconcat
 0 -> nullmixedamt / mempty
 (==0) -> mixedAmountIsZero
 (/=0) -> not . mixedAmountIsZero

Some functions on Journals now return Sets of elements instead of lists. You can get the lists back by using either Data.Foldable (toList) or Data.Set (toList).
Hledger.Data.Journal:
 journalCommoditiesDeclared
 journalDescriptions
 journalPayeesDeclared
 journalPayeesUsed
 journalPayeesDeclaredOrUsed
 journalAccountNamesUsed
 journalAccountNamesImplied
 journalAccountNamesDeclared
 journalAccountNamesDeclaredOrImplied
 journalAccountNames
Hledger.Data.Posting:
 accountNamesFromPostings

simonmichael:
Thanks, that is very helpful!

Your MixedAmount changes again, in slightly different form and favouring the Monoid API:
```
0              ->  mempty
a + b          ->  a <> b
a - b          ->  a <> negateMixedAmount b
sum as         ->  mconcat as
(-a), negate a ->  negateMixedAmount a
a == 0         ->  mixedAmountIsZero a
a /= 0         ->  not . mixedAmountIsZero a
```
Here's a more abstracted version hiding the Monoid. Note I tend to
migrate function names towards "typeOperation", as it scales well and
is similar to OO "Class.method", but we haven't yet gone all the way
with this:
```
0              ->  nullmixedamt
a + b          ->  a `mixedAmountPlus` b
a - b          ->  a `mixedAmountMinus` b
sum as         ->  sumMixedAmounts as
(-a), negate a ->  negateMixedAmount a
a == 0         ->  mixedAmountIsZero a
a /= 0         ->  mixedAmountIsNonZero a
```
A less verbose version could be:
```
0              ->  nullmixedamt
a + b          ->  a `maPlus` b
a - b          ->  a `maMinus` b
sum as         ->  maSum as
(-a), negate a ->  maNegate a
a == 0         ->  maIsZero a
a /= 0         ->  maIsNonZero a
```

Operators (":+:") are also an option but probably not worth it right now.

Perhaps this is too much for this PR but since we are starting to change the API it's good to think about where we want to take it.

Also I'll note some extra complexity in MixedAmount's zero:

- is it pure zero or a commoditised zero (used eg for hledger-web's chart)
- is it a normalised pure zero (Mixed []) or unnormalised

simonmichael:
> Some functions on Journals now return Sets of elements instead of lists.

Is this necessary ? I believe these generally return results in the order parsed, which can be useful information (though it's not yet mentioned in their haddocks).

Xitian9:
Either of those interfaces looks good to me; let me know which you'd prefer.

Each of those functions that I changed to Set calls `nubSort`, so the Set has exactly the same order. In fact, `journalAccountNames` called `nubSort` several times in total, concatenating all the results together. In one of the benchmarks (I can't remember which), `nubSort` was the biggest cost centre and came from these functions, which was why I did this.

simonmichael:
> Either of those interfaces looks good to me; let me know which you'd prefer.

Great, let's aim for the short one. No great urgency on a total switch though, I would keep the old names working too for now.
> Each of those functions that I changed to Set calls nubSort, so the Set has exactly the same order.
> 
I don't quite follow.. isn't a Set by definition unordered ? Even if its implementation does currently spits out things in the same order, that's not guaranteed, as it was with List. Can we fix the nubSort inefficiencies while still keeping the ordered type ? And by the way are we using https://hackage.haskell.org/package/data-ordlist-0.4.7.0/docs/Data-List-Ordered.html#v:nubSort <https://hackage.haskell.org/package/data-ordlist-0.4.7.0/docs/Data-List-Ordered.html#v:nubSort> , which seems to be the fast one ?

Xitian9:
A Haskell `Set` is by definition ordered, and it is maintained as such as an invariant. While a mathematical set is unordered, the implementation here requires an `Ord` constraint, and stores the keys in that order.

In fact, `Journal` actually stores these things internally as the keys of a Map, which is in fact a Set, so this actually just maintains the internal structure longer.

simonmichael:
> A Haskell Set is by definition ordered

Woah, TIL. Thanks. 

Next objection: lists are more idiomatic and better supported in Haskell syntax, and Haskellers of all experience levels are familiar with them. Is it really necessary to give them up, and to create this API churn ?

Xitian9:
We might be able to have our cake and eat it too. Each of those functions has lots of variants: Declared, Used, Implied, DeclaredOrImplied, etc. In each case, there looks to be one end-point function which is really the one people want to use: `journalAccountNames`, `journalPayeesDeclaredOrUsed`, `journalDescriptions`. We could have these return lists, but the intermediate functions deal with Sets for efficiency reasons.

Xitian9:
I could also implement a Set version of each and have the existing functions spit out calling toList on the Set version.

simonmichael:
> I could also implement a Set version of each and have the existing functions spit out calling toList on the Set version.
> 
For now, maybe that's a good approach ? Use Set-based variants as needed internally, but keep the exported functions List-based.

Xitian9:
Set versions created.

simonmichael:
There's no hurry on this, but in case you are waiting for my input, these were my quick notes on the latest:

**lib,cli,ui,bin: Add Semigroup, Monoid, don't use Num instances of MixedAmount.**
Not an improvement as it stands, the old code was clearer.

**lib: Split showMixedAmountB into showMixedAmountB and showAmountsB, the former being a simple wrapper around the latter.**
Do we still have simple MixedAmount -> String and/or MixedAmount -> Text show functions ?
Should both of showAmountsB and showAmountsLinesB be exported ? They don't seem especially user friendly.

**lib,cli: Remove special handling of nullmixedamt. Remove all uses of 0 as a MixedAmount.**
It seems like this and some previous commits could be squashed.

**lib,cli,ui,bin: Eliminate all uses of Mixed outside of Hledger.Data.Amount.**
Perhaps mixedFromAmount -> mixedAmount ?

Do you plan to add the short `ma` names in this PR ? Should we postpone that ?

Generally I'd like to simplify this history a bit before merging, removing side tracks. It may be a tedious, I can have a go if you run out of time.

Xitian9:
Okay, here's what I'll do. This requires a fair amount of history rewriting, so let me know if it agrees with your ideas.

I'll move the small independent commits to the beginning of this PR to the greatest extent possible.
I will rewrite the commit which adds the Monoid typeclass: it will create the short form interface you suggest, with a couple of additions (`maSum'`, 'mixedAmount`, `maAddAmount`, `maAddAmounts`, `maAddAmounts'`, prime versions are strict). The old long forms I added will be removed. The `Num` and `Monoid` typeclasses will be implemented using these functions, but will not be used in code. I'll squish that “remove special handling of zeromixedamt” commit where possible, but it actually depends on a couple of previous commits, so it may not disappear entirely.

The `showAmountsLinesB` function replaces the previous `showMixedAmountLinesB` function. Users probably don't need it, but we use it in some reports for making tables. `showAmountsB` and `showMixedAmountB` are probably the only functions users need in their lives. The simple `MixedAmount -> String` functions remain unchanged.

simonmichael:
It sounds good as far as I can tell.

Xitian9:
This PR now just addresses the new interface for MixedAmounts, and implements the discussed interface. It conflicts very slightly with #1506, but either one can be rebased easily if the other is merged.

simonmichael:
This looks good. maZero looks a little superfluous now, but maybe it'll be handy later as we clarify the variations of zero. Perhaps mixedAmount does too (we have mixed [..]) but that's ok.

Do we actually need the lazy variants, could we just have the strict ones ?

Xitian9:
I'd be quite happy to leave out `maZero` and the lazy variants.

`mixedAmount` is called quite frequently, and looks a lot neater than `mixed . (:[])` or `mixed . pure`. In a later commit I generalise `mixed` to work on Foldables, in which case `mixedAmount` avoids the ambiguous type that `mixed . pure` gives, so I think it is still useful.

simonmichael:
Cool let's do as you say.

Xitian9:
Regarding clarifying the variations of zero:
I think the correct thing to do is make the variations of zero behave identically, so that people can use ‘the wrong one’ and it just works. Then we provide only a single zero, `nullmixedamt`. I actually think that has been more or less accomplished here: unless people are actually checking equality with `==` and `/=`, the zeros should behave identically.

simonmichael:
Great. For release notes I've summarised it like so:
```
new API for MixedAmount arithmetic (#1491)

Previously we relied on MixedAmount being a Num, which allows +, -,
sum, negate, fromInteger etc. to be used with MixedAmounts. While
convenient, this Num instance is not (and can't be) fully implemented
or law abiding, so it's possible to misuse it, potentially leading to
bugs.

Now, MixedAmount is a lawful Monoid (and a Semigroup), so you can
combine (add) MixedAmounts with <> or mconcat and represent zero with
mempty.

However, we recommend using the following more abstract API, which
will insulate you from future implementation changes:

maPlus       (instead of +)
maMinus      (instead of -)
maNegate     (instead of negate)
maSum        (instead of sum/sumStrict)
nullmixedamt (instead of 0)

And when constructing MixedAmounts, avoid the Mixed constructor,
instead use:

mixed        :: [Amount] -> MixedAmount
mixedAmount  :: Amount -> MixedAmount
maAddAmount  :: MixedAmount -> Amount -> MixedAmount
maAddAmounts :: MixedAmount -> [Amount] -> MixedAmount

For now the Num instance remains, as a convenience for scripters and
for backward compatibility, but for production code you should
probably consider it deprecated.
```

Having to warn people not to use Mixed is a slight wart, but allows us to continue avoiding the busywork of declaring all of Hledger.Data.Types exports. I'm ready to merge if you are.

Xitian9:
I'm happy to merge.

One possibility for the Mixed constructor is to export it from Hledger.Data.Types, but hide it from the export list for Hledger.Data. I actually do this in a later commit.

simonmichael:
Merged. Hurrah! Thank you @Xitian9 !

PR #1495, opened 2021-03-02
Do not include price directives in report span unless valuing at period ends.

Xitian9:
Addresses points raised in #1416. If it is decided that price directives should also be included for any valuation, rather than just period end, let me know and that can be adjusted.

Also gets rid of some code duplication for calculating report spans.

As a side effect, this enlarges the reportspan to encompass full intervals for budget goals. This was documented in the tests, but it's not clear whether that was desired behaviour or just documenting quirks. If we want budget goals to have a different range than the actual report, let me know and I can reinstate the old behaviour.


adept:
FWIW, i think that "budget period==report period" is the right semantics

simonmichael:
Lovely. Thank you!

PR #1496, opened 2021-03-03
Fix --valuechange for CompoundBalanceReport.

Xitian9:
Addresses #1353.

### DIFF
--- a/hledger/Hledger/Cli/CompoundBalanceCommand.hs
+++ b/hledger/Hledger/Cli/CompoundBalanceCommand.hs
@@ -111,7 +111,7 @@ compoundBalanceCommand CompoundBalanceCommandSpec{..} opts@CliOpts{reportspec_=r
         parse = \case
           "historical" -> Just HistoricalBalance
           "cumulative" -> Just CumulativeChange
-          "change"     -> Just PeriodChange
+          "periodic"   -> Just PeriodChange
           _            -> Nothing
     balancetype = fromMaybe cbctype mBalanceTypeOverride
     -- Set balance type in the report options.
@@ -135,8 +135,7 @@ compoundBalanceCommand CompoundBalanceCommandSpec{..} opts@CliOpts{reportspec_=r
             _                 -> showDateSpan requestedspan
           where
             enddates = map (addDays (-1)) . mapMaybe spanEnd $ cbrDates cbr  -- these spans will always have a definite end date
-            requestedspan = queryDateSpan date2_ (rsQuery rspec)
-                                `spanDefaultsFrom` journalDateSpan date2_ j
+            requestedspan = reportSpan j rspec
 
         -- when user overrides, add an indication to the report title
         -- Do we need to deal with overridden ReportType?
--- a/hledger/test/journal/valuechange.test
+++ b/hledger/test/journal/valuechange.test
@@ -7,16 +7,16 @@ P 2000/03/01 A  3 B
 P 2000/04/01 A  4 B
 
 1999/01/01
-  (a)      2 A @ 4 B
+  (assets)      2 A @ 4 B
 
 2000/01/01
-  (a)      1 A @ 6 B
+  (assets)      1 A @ 6 B
 
 2000/02/01
-  (a)      1 A @ 7 B
+  (assets)      1 A @ 7 B
 
 2000/03/01
-  (a)      1 A @ 8 B
+  (assets)      1 A @ 8 B
 
 # 1. multicolumn balance report showing changes in period-end values
 # Initial balance 2 A, valued at 10 B each, total 20 B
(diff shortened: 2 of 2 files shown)


simonmichael:
Thanks!

simonmichael:
Actually.. I still don't see --valuechange affecting bs, except the title. Eg, replicating valuechange.test:4:
```
$ hledger bs -MN -b 2000 -V
Balance Sheet 2000-01-31..2000-04-30, valued at period ends

             || 2000-01-31  2000-02-29  2000-03-31  2000-04-30 
=============++================================================
 Assets      ||                                                
-------------++------------------------------------------------
 assets      ||       15 B         8 B        15 B        20 B 
=============++================================================
 Liabilities ||                                                
-------------++------------------------------------------------
$ hledger bs -MN -b 2000 -V --valuechange
Balance Sheet 2000-01-31..2000-04-30

             || 2000-01-31  2000-02-29  2000-03-31  2000-04-30 
=============++================================================
 Assets      ||                                                
-------------++------------------------------------------------
 assets      ||       15 B         8 B        15 B        20 B 
=============++================================================
 Liabilities ||                                                
-------------++------------------------------------------------
```

Xitian9:
That's because the default report type for balancesheet is Historical, and `--valuechange --historical` is indistinguishable from `--value=end --historical`. If you add `--periodic` or `--cumulative` you will see the difference (note this is what we do in the balancesheet test added in this commit).

simonmichael:
Ah. Similarly, I was surprised when bs --change didn't work any more. But of course it's bs --periodic now. 

This is going to create confusion, isn't it.

Xitian9:
Probably… We could always restore the old `--change` in place of the new `--periodic`, but then we would need to find a new flag for `ChangeReport`.

simonmichael:
Let's sleep on that. Also whatever we call it, maybe --valuechange should imply it ? So eg bal --valuechange and bs --valuechange just work ?

simonmichael:
Perhaps:

Report type: The general thing that is calculated/shown in each cell. It is one of:

- --sum : show a sum of posting amounts (default)
- --budget : like --change but also show a budget goal amount
- --valuechange : show change of value of period-end historical balances

Accumulation type: Which previous periods' postings should be included in calculations. It is one of:

--change : postings from column start to column end. Ie, show changes in each period. Typically used when reviewing revenues/expenses. (default for balance, incomestatement)
--cumulative : postings from report start (specified by -b/--begin, eg) to column end. Ie, show accumulated changes since start of report. Rarely used.
--historical/-H : postings from journal start to column end. Ie, show historical balance at end of each period. Typically used when reviewing assets/liabilities/equity. (default for balancesheet, balancesheetequity, cashflow)

?

simonmichael:
Or, if we feel --periodic is clearer for the long term, --change could be a legacy alias for it.

simonmichael:
I think it's worth a little more experimentation, I'd rather hold 1.21 a bit and reduce fixup releases.

Xitian9:
>  Also whatever we call it, maybe --valuechange should imply it ? So eg bal -M --valuechange and bs -M --valuechange just work ?

Could do. That's a bit trickier than adjusting the default valuation type, but is definitely doable.

Using `--sum` in place of `--change` is reasonable.

Xitian9:
Maybe a clarification. Do you mean that `--valuechange` implies `--periodic` (or whatever we call it), or just that it changes the default? Would `--valuechange --cumulative/--historical` produce the same results as `--valuechange --periodic`, produce different results, or throw a usage error?

Xitian9:
I think it might be reasonable to change the `reporttype_` and `balancetype_` fields of ReportOpts to be `Maybe` instead of the bare value, so we can distinguish between not being specified and being specified as the default value. This would also allow us to get rid of the extra opt parsing in CompoundBalanceReport for overriding cbctype. It would of course be an API change.

simonmichael:
> Maybe a clarification. Do you mean that --valuechange implies --periodic (or whatever we call it), or just that it changes the default? Would --valuechange --cumulative/--historical produce the same results as --valuechange --periodic, produce different results, or throw a usage error?
> 

Good question. I'm not sure. My concern is that there'll be too many twisty modes, all alike, some of them useful and some of them dead ends.

Xitian9:
Perhaps a summary of useful and non-useful `--valuechange` reports:
- Does not make sense for any valuation other than `--value=end`. We currently change to `--value=end` if no valuation is specified, and if any other valuation method is specified we throw a usage error.
- `--periodic/--change` is definitely the most useful default report
- `--cumulative` produces a new report. It's unclear whether there's a strong use case for this report, but it's easy to implement and pretty clearly what you would want if you called it with these options.
- `--historical` makes sense as a report, and it's clear what it should give you, but what it should give you is identical to the standard report with `--value=end --historical`. Are we okay with having two paths to the same report?

Xitian9:
My feeling is that so long as there are sensible defaults, I'm not opposed to having two different paths to the same report. I'm also not opposed to having reports of questionable use if they are natural extensions of the options we provide and are relatively clear and unambiguous. If there's a report which really doesn't make sense which is directly requested e.g. `--valuechange --value=then`, there should be a usage error.

simonmichael:
Nice summary. 

I thought I had come up with a scheme that avoided breaking old commands, but I hadn't. I feel we don't have the optimal CLI for choosing report variations yet, and it might take a few more releases to get there. That is fine, and I shouldn't sweat it too much. But I do want to minimise CLI churn and especially avoid causing things to stop working after an upgrade. That kind of thing is against hledger goals and can be costly to the project. Making --change = --periodic would help a bit.

I also feel users want a menu that has one main list of choices, with optional extras on the side (diner menu), rather than multiple lists they have to choose from (complicated build-your-own-meal vegan-fast-food menu). 

The separate report type and accumulation type (and valuation type) lists make it feel more like the latter. Report type and accumulation type, as currently presented, are somewhat orthogonal, but not entirely. 

I think --budget could be generalised to work with all report types, making it a side option rather than a main choice. That would get us a bit closer to a single main list.

But, perhaps the higher-level commands are the simple meals and it's ok for `balance` to be the confusing build-your-own-pizza command. It used to be *the* command though, so people rely on it, so backward compatibility needs to be considered.

simonmichael:
PS while thinking aloud, one other option I haven't really thought about is to add these new reports only as new commands, eg `valuechange`. How valuable is it to have all reports available within the `balance` command ?

Xitian9:
It wouldn't be unreasonable to use a different command name. The main point against it is that there is very little difference between a valuechange report and a normal balance report: the only thing that differs is the numbers presented. It would probably produce a fair amount of code duplication to make a new command, though the code in `hledger-lib` would probably be unchanged.

One question I have: how often is this feature going to be used? I think the vast majority of users will never use it, which means they will never even go look at the options in order to presented with the choice. If we document it but not prominently, then people who need it can look it up and use it, and others will not be aware of its existence, living their lives happily in ignorance.

I view this as similar to the `--value=XXX` options. Most people don't need it, or `-V` is perfectly sufficient for them. Only if they actually need more fine-grained control do they look into things and find the other options. I think this type of menu-choice is acceptable, and falls more on the optional extras side than a required choice.

So my view is that the proposed interface (with `--periodic=--change`) is sufficient. Don't advertise `--valuechange` prominently , but hide it down in the options menu along with `--value=XXX`, and people won't encounter it unless they go looking for it.

simonmichael:
Thanks, that is persuasive for now. Though I expect more people to start using `--valuechange` etc. once it exists, especially if it's easy.

I realise also I have expanded scope, from just the `balance` command, to the higher-level commands (`bs`) and how the new `balance` options interact with them. Lingering issues with the latter are not quite so important.

 I'll continue tweaking the docs, here's a new draft: https://gist.github.com/simonmichael/2f55199007f87a15b91dde64eda0f756
Changes needed to match this (PR welcome, or I'll do it later):

- rename `--change` to `--sum` and `--periodic` back to `--change` (I chose to support just the old flag for now)

And we should review how balance options work with the higher-level commands now, but that's a lower priority.

simonmichael:
PR merged, https://hledger.org/hledger.html#balance updated.

PR #1497, opened 2021-03-05
Use --change for PeriodChange, --sum for ChangeReport, and --valuechange defaults to PeriodChange unless overridden

Xitian9:


simonmichael:
Great! Thanks!

PR #1504, opened 2021-03-13
Remove more reference to old man page stuff

felixonmars:
They are not cleaned up in 9824c9683ae80a2ebd5be72b8dad88f00c8588bd


simonmichael:
Thanks!

PR #1505, opened 2021-03-17
Performance cleanups

Xitian9:
I've broken off some of the simpler commits originally in #1491. All these commits are minor refactors, documentation fixups, and efficiency improvements which are mostly minor, but become significant with larger journals.

The only API changes are the addition of a few Set-based functions getting various journal parameters, a list of which is in e120861.

### DIFF
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -1,4 +1,6 @@
-{-# LANGUAGE RecordWildCards, OverloadedStrings #-}
+{-# LANGUAGE CPP               #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards   #-}
 {-|
 
 
@@ -9,12 +11,13 @@ account, and subaccounting-excluding and -including balances.
 
 module Hledger.Data.Account
 where
+import qualified Data.HashSet as HS
+import qualified Data.HashMap.Strict as HM
 import Data.List (find, sortOn)
-import Data.List.Extra (groupSort, groupOn)
-import Data.Maybe (fromMaybe)
-import Data.Ord (Down(..))
+import Data.List.Extra (groupOn)
 import qualified Data.Map as M
-import Safe (headMay, lookupJustDef)
+import Data.Ord (Down(..))
+import Safe (headMay)
 import Text.Printf
 
 import Hledger.Data.AccountName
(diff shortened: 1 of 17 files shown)


simonmichael:
**lib: In sorting account names, perform lookups on HashSets and HashMaps,** - makes benchmarks run faster

Could you give a rough idea of the performance impact from the other two ?
**lib: More efficiently check whether Amounts are or look zero.**
**lib,cli,ui,web: A number of AccountName and Journal functions which are**

simonmichael:
Ie, anything users would notice, and in what kind of situation ?

simonmichael:
(And feel free to add it to the commit messages, it helps explain to future readers why we added a bunch more API functions etc.)

Xitian9:
I've added an idea of the performance impacts to the first two of the performance commits.

For the last one (working with Sets instead of lists for a bunch of functions which are supposed to be sorted), I recall implementing that because of seeing `nubSort` high on the list of cost centres, but it looks like I was misremembering. There is negligible performance difference in most cases.

I still think this commit in some form is worthwhile. The fact that most of these functions take what is a Set (the keys of a Map), convert it to a list, concatenate the two lists, and then call `nubSort` on the result is a very roundabout way of just calling `<>` on the original Sets. That said, the larger API probably isn't justified by the potential gains.

I'm unsure about the best path forward. On the one hand I think most code would either work as-is or require minimal changes to work fine with Sets instead of lists. Anything which relies on just being Foldable will work out of the box, anything testing for membership will work more efficiently with a `Set.member` in place of `elem`, and anything which really requires a list can just call `toList`. On the other hand creating API churn, even if the fixes are simple, is annoying for others.

One solution which I think is reasonable is to have most of these functions return Sets, and the most-likely-to-be-used versions of each continuing to return a list: I'm thinking `expandAccountNames`, `journalAccountNames`, and maybe an alias for `journalPayeesDeclaredOrUsed`. All the other functions really feel internal to me, and I have a hard time coming up with examples where somebody will want to call them.

What do you think?

Xitian9:
Okay, I have rewritten the Set commit. The number of new functions introduced is cut down to just one: the most useful (i.e. most inclusive) of each series (`expandAccountNamesSet`, `journalAccountNamesDeclaredOrImpliedSet`, `journalPayeesDeclaredOrUsedSet`, and `journalCommodities`). The other functions have been rewritten to use Sets internally where more efficient, but this doesn't leak out into the API.

simonmichael:
Sorry for the slow reply. The performance data looks great. Re the Set functions, I see the value of using them internally if they improve performance, but is there any reason to expose them in the public API yet ? I would be inclined to wait until we have a real use case.

Xitian9:
Do we have a good option for exposing them on the internal API without exposing them on the public API? This commit uses each of these functions outside of the module in which they're defined, and also outside of hledger-lib.

simonmichael:
Oh I see.

I feel that the Set functions, even though they improve efficiency and elegance internally, don't add enough value to be worth complicating the API, yet. By the principle of Do The Simplest Thing That Could Possibly Work, and If It Ain't Broke Don't Fix It.

Xitian9:
Okay, I'll remove those functions from the API.

Should I also remove `journalCommodities`? This is a new function which is used in hledger, and does not duplicate an existing function in the API.

simonmichael:
journalCommodities is fine.

Xitian9:
The Set-producing functions have been pruned.

simonmichael:
Lovely, thank you!

PR #1506, opened 2021-03-17
Allow showing lists of Amounts, make showing register reports and lists of postings more efficient.

Xitian9:
These are the commits from #1491 concerning the `showMixedAmountB` function, and splitting it into `showMixedAmountB` and `showAmountsB`.

It continues to be the case that most users should continue using the simple `MixedAmount -> String` variants, and for those who want more control, `showMixedAmountB` is almost always what you want. `showAmountsB` is for when you have a list of amounts which you do not want normalised before being shown, as in when you're showing a transaction or posting (in which case, you should probably be using `postingsAsLines`, which calls this internally). `showAmountsLinesB` is a minor variant which gives you a builder for each line (since we're using builders for efficiency, we cannot call T.lines without strictifying).

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -77,6 +77,8 @@ module Hledger.Data.Amount (
   styleAmountExceptPrecision,
   amountUnstyled,
   showAmountB,
+  showAmountsB,
+  showAmountsLinesB,
   showAmount,
   cshowAmount,
   showAmountWithZeroCommodity,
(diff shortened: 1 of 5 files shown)


simonmichael:
What sort of performance impacts might users notice from this PR ?

Xitian9:
I've added performance numbers in the commit messages. As you can see, the commit which strips prices results in a very large performance improvement. I haven't teased out precisely how much of that is due to stripping prices and how much of that is from the other changes.

simonmichael:
Awesome. Thank you!

Issue #1360, opened 2020-09-29
JSON: account->aibalance->astyle->asprecision is an object in 1.19.1


real-dam:
Hi,

I got a crash report from two MoLe users that seems to be a problem in the JSON API of hledger-web. The `asprecision` element, which usually is a number, is an object, and that breaks the parser. Here's the beginning of the `/accounts` output. The interesting part is at the middle:

```
[
   {
      "aboring" : false,
      "aibalance" : [
         {
            "acommodity" : "INR",
            "aismultiplier" : false,
            "aprice" : null,
            "aquantity" : {
               "decimalMantissa" : 0,
               "decimalPlaces" : 10,
               "floatingPoint" : 0
            },
            "astyle" : {
               "ascommodityside" : "R",
               "ascommodityspaced" : true,
               "asdecimalpoint" : ".",
               "asdigitgroups" : null,
               "asprecision" : {
                  "contents" : 2,
                  "tag" : "Precision"
               }
            }
         }
      ],
      "aname" : "root",
      "aparent_" : "",
      "asubs" : [],
      "asubs_" : [
         "ASSETS",
         "EQUITY",
         "EXPENSES",
         "INCOME",
         "…"
      ]
   }
]
```

hledger-web version is reported to be `1.19.1`. I can't reproduce this with the demo.hledger.org installation that always uses a plain number for `asprecision`. The version I personally use with MoLe is `1.16.2` and that is fine too.

Perhaps there is something in the journal file that triggers the issue, but I don't have access to the raw data.

I hope this rings a bell. I should be able to get more information about the journal content if you tell me what to ask for.

In theory, I could spawn another parser hierarchy that requires that `asprecision` is an object. That would be the way if the change in the API was intentional and is there to stay.

Cheers.

Xitian9:
This is because of a change in the internal representation of `Precision`, implemented in f6fa76bba7530af3be825445a1097ae42498b1cd. The change was used to get rid of magic values for `Precision`, but since it is now a sum type it will be serialised in JSON as an object by default.

An option would be to write a custom JSON serialisation which allows it to be represented by a bare number.

real-dam:
Thanks for the analysis.

Would you say that this was an intentional change in the JSON output and I should adapt the parser?

Xitian9:
It was an intentional change in the internal representation. I wouldn't quite say it was intentional that the JSON output changed, but it was a natural consequence.

I would recommend adapting the parser, unless there's a specific reason to not.

[...]

PR #1509, opened 2021-03-22
Change format of MixedAmount so it is normalised by design

Xitian9:
This PR changes the internal format of MixedAmount to a Strict Map, which is normalised by construction. This format is used internally in `normaliseHelper` already. Anything which should not be normalised should be stored as an `[Amount]`.

I'll add performance numbers here, but I'll wait until #1505 is merged to add those to the commit messages, since the improvements will interfere with one another.


Xitian9:
Here are the preliminary performance numbers (pre-#1505). All other metrics and tests that I've checked produce changes that are small enough that we can reasonably assume no effect.

`hledger reg -f examples/10000x10000x10.journal`
- Total memory in use drops from 189MiB to 167MiB, a 12% drop

`hledger bal -f examples/10000x10000x10.journal`
- Total bytes allocated in the heap drops from 26.9GB to 13.1GB, a 51% drop
- Total (profiled) time drops from 21.8s to 16.0s, a 27% drop

`hledger bal -f examples/10000x10000x10.journal -M`
- Total bytes allocated in the heap drops from 195.5GB to 53.9GB, a 72% drop
- Total (profiled) time drops from 93.4s to 48.6s, a 48% drop

Xitian9:
Running the above tests with the more realistically sized `examples/10000x1000x10.journal` produced smaller but consistent results (except I'm getting a slightly higher “total memory in use” stat for `hledger bal -M`, which is not reproduced for `hledger bal`).

Xitian9:
This also adds some tests for JSON output, at least partially addressing #1360.

Xitian9:
I've added performance comments in the commit which changes the internal representation of MixedAmount (at time of writing, the second commit in this PR). Overall there's a large improvement for journals with lots of accounts, and a smaller but still significant improvement for journals with a more reasonable number of accounts.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -299,7 +299,7 @@ data Posting = Posting {
       pdate2            :: Maybe Day,         -- ^ this posting's secondary date, if different from the transaction's
       pstatus           :: Status,
       paccount          :: AccountName,
-      pamount           :: MixedAmount,
+      pamount           :: [Amount],
```
simonmichael, line 303: I approve of modelling what's needed more precisely with types. But is there not some loss of overall conceptual simplicity ? We tell users "a posting has a (possibly multicommodity) amount". This is still true on the outside, but developers now must deal with two different representations of multicommodity amounts (MixedAmount vs [Amount]). Could you explain/motivate more why it's better for Posting not to use a MixedAmount ? Maybe we could include it in the field's haddock. I do see that it seems to reduce code noise in many places. (Also, "pamount" singular now sounds a bit odd.)

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -230,7 +230,8 @@ data Amount = Amount {
       aprice      :: !(Maybe AmountPrice)  -- ^ the (fixed, transaction-specific) price for this amount, if any
     } deriving (Eq,Ord,Generic,Show)
 
-newtype MixedAmount = Mixed [Amount] deriving (Eq,Ord,Generic,Show)
+newtype MixedAmount = Mixed (M.Map MixedAmountKey Amount) deriving (Eq,Ord,Generic,Show)
```
simonmichael, line 233: We usually seem to end up replacing Map with HashMap - any thoughts ?

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -230,7 +230,8 @@ data Amount = Amount {
...
+type MixedAmountKey = (CommoditySymbol, Maybe (CommoditySymbol, Maybe Quantity))
```
simonmichael, line 234: Interesting, what's going on here ? We start to need some haddocks here.

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -580,56 +579,64 @@ canonicaliseAmount styles a@Amount{acommodity=c, astyle=s} = a{astyle=s'}
 
 instance Semigroup MixedAmount where
   (<>) = maPlus
+  sconcat = maSum
+  stimes n = multiplyMixedAmount (fromIntegral n)
 
 instance Monoid MixedAmount where
   mempty = nullmixedamt
+  mconcat = maSum
```
simonmichael, line 587: Just curious, why do you implement more instance methods now ? They were using default implementations before.. and now you've provided more efficient implementations ?

hledger-lib/Hledger/Data/Types.hs
simonmichael, line 234: Any concerns about using a tuple here rather than an ADT, since they often seem to get blamed for space leaks/excessive laziness/thunk buildup ?

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -580,56 +579,64 @@ canonicaliseAmount styles a@Amount{acommodity=c, astyle=s} = a{astyle=s'}
...
 missingmixedamt :: MixedAmount
 missingmixedamt = mixedAmount missingamt
 
--- | Convert amounts in various commodities into a normalised MixedAmount.
-mixed :: [Amount] -> MixedAmount
-mixed = normaliseMixedAmount . Mixed
+-- | Convert amounts in various commodities into a mixed amount.
+mixed :: Foldable t => t Amount -> MixedAmount
+mixed = maAddAmounts nullmixedamt
+{-# SPECIALISE mixed :: [Amount] -> MixedAmount #-}
```
simonmichael, line 618: Adding these SPECIALISE lines causes GHC to generate an extra version that is optimised for lists. Did this make a noticeable difference ? (And just for the record, with which GHC version ?)

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -759,22 +765,27 @@ sumSimilarAmountsUsingFirstPrice a b = (a + b){aprice=p}
...
+  where ma' = M.filterWithKey (\k _ -> (==c) $ fst k) ma
 
 -- | Apply a transform to a mixed amount's component 'Amount's.
 mapMixedAmount :: (Amount -> Amount) -> MixedAmount -> MixedAmount
-mapMixedAmount f (Mixed as) = Mixed $ map f as
+mapMixedAmount f (Mixed ma) = mixed . map f $ toList ma
+
+-- | Apply a transform to a mixed amount's component 'Amount's, which does not
+-- affect the key of the amount (i.e. doesn't change the commodity, price
+-- commodity, or unit price amount). This condition is not checked.
```
simonmichael, line 786: Why provide/use this.. is it faster ? Easier to use ? I guess it's unexported/for internal use only ?

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -759,22 +765,27 @@ sumSimilarAmountsUsingFirstPrice a b = (a + b){aprice=p}
...
+mapMixedAmountUnsafe :: (Amount -> Amount) -> MixedAmount -> MixedAmount
+mapMixedAmountUnsafe f (Mixed ma) = Mixed $ M.map f ma  -- Use M.map instead of fmap to maintain strictness
```
simonmichael, line 788: Nice strictness comment/improvement.

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -702,34 +710,32 @@ maIsNonZero = not . mixedAmountIsZero
...
-        priceKey (UnitPrice  x) = (acommodity x, Just $ aquantity x)
-        priceKey (TotalPrice x) = (acommodity x, Nothing)
-    missingkey = key missingamt
+    (zeros, nonzeros) = M.partition amountIsZero ma
+    missingkey = amountKey missingamt
+
+-- | Get a mixed amount's component amounts without normalising zero and missing
+-- amounts.
+amountsRaw :: MixedAmount -> [Amount]
+amountsRaw (Mixed ma) = toList ma
```
simonmichael, line 726: I guess we need to export this for JSON and balancing.. is toList's ordering reliable ?

simonmichael (commented):
The speedups sound terrific.

simonmichael:
Some casual time benchmarking: 

Simple reports on a 15 year journal - a slight improvement:
```
$ hledger.1509 -f ~/adm/all.journal stats | grep -iE '(accounts|transactions) +:'
Transactions             : 15119 (2.9 per day)
Accounts                 : 1678 (depth 8)
$ quickbench "hledger.pre1509 -f all.journal stats" "hledger.1509 -f all.journal stats"
Running 2 tests 1 times at 2021-04-05 16:07:53 HST:

Best times:
+--------------------------------------++------+
|                                      ||      |
+======================================++======+
| hledger.pre1509 -f all.journal stats || 2.39 |
| hledger.1509 -f all.journal stats    || 2.21 |
+--------------------------------------++------+
$ quickbench "hledger.pre1509 -f all.journal bal" "hledger.1509 -f all.journal bal"
Running 2 tests 1 times at 2021-04-05 16:08:04 HST:

Best times:
+------------------------------------++------+
|                                    ||      |
+====================================++======+
| hledger.pre1509 -f all.journal bal || 2.41 |
| hledger.1509 -f all.journal bal    || 2.16 |
+------------------------------------++------+
```

Two runs of the standard benchmarks - ~10% faster print, ~15% slower register, ~10-25% faster balance (but not with large files perhaps ?):
```
$ BENCHEXES=hledger.pre1509,hledger.1509 make bench

Running 5 tests 1 times with 2 executables at 2021-04-05 16:08:27 HST:

Best times:
+----------------------------------------------------++-----------------+--------------+
|                                                    || hledger.pre1509 | hledger.1509 |
+====================================================++=================+==============+
| -f examples/10000x1000x10.journal print            ||            0.97 |         0.86 |
| -f examples/10000x1000x10.journal register         ||            8.02 |         8.61 |
| -f examples/10000x1000x10.journal balance          ||            1.24 |         0.86 |
| -f examples/1000x1000x10.journal balance --weekly  ||            0.83 |         0.68 |
| -f examples/10000x1000x10.journal balance --weekly ||           11.04 |        11.40 |
+----------------------------------------------------++-----------------+--------------+
$ BENCHEXES=hledger.pre1509,hledger.1509 make bench

Running 5 tests 1 times with 2 executables at 2021-04-05 16:20:59 HST:

Best times:
+----------------------------------------------------++-----------------+--------------+
|                                                    || hledger.pre1509 | hledger.1509 |
+====================================================++=================+==============+
| -f examples/10000x1000x10.journal print            ||            1.02 |         0.87 |
| -f examples/10000x1000x10.journal register         ||            7.56 |         8.57 |
| -f examples/10000x1000x10.journal balance          ||            1.20 |         0.90 |
| -f examples/1000x1000x10.journal balance --weekly  ||            0.86 |         0.72 |
| -f examples/10000x1000x10.journal balance --weekly ||           11.01 |        10.51 |
+----------------------------------------------------++-----------------+--------------+
```

Two runs of your benchmarks - register is a little faster, balance is >100% faster (2x), balance -M ~30-40% faster:
```
$ quickbench -w hledger.pre1509,hledger.1509 "hledger reg -f examples/10000x10000x10.journal" "hledger bal -f examples/10000x10000x10.journal" "hledger bal -f examples/10000x10000x10.journal -M"
Running 3 tests 1 times with 2 executables at 2021-04-05 16:14:56 HST:

Best times:
+-------------------------------------------++-----------------+--------------+
|                                           || hledger.pre1509 | hledger.1509 |
+===========================================++=================+==============+
| reg -f examples/10000x10000x10.journal    ||            8.03 |         7.98 |
| bal -f examples/10000x10000x10.journal    ||            3.03 |         1.29 |
| bal -f examples/10000x10000x10.journal -M ||           36.63 |        24.79 |
+-------------------------------------------++-----------------+--------------+
$ quickbench -w hledger.pre1509,hledger.1509 "hledger reg -f examples/10000x10000x10.journal" "hledger bal -f examples/10000x10000x10.journal" "hledger bal -f examples/10000x10000x10.journal -M"
Running 3 tests 1 times with 2 executables at 2021-04-05 16:17:03 HST:

Best times:
+-------------------------------------------++-----------------+--------------+
|                                           || hledger.pre1509 | hledger.1509 |
+===========================================++=================+==============+
| reg -f examples/10000x10000x10.journal    ||            9.48 |         8.47 |
| bal -f examples/10000x10000x10.journal    ||            5.22 |         2.07 |
| bal -f examples/10000x10000x10.journal -M ||           44.83 |        24.23 |
+-------------------------------------------++-----------------+--------------+
```

hledger-lib/Hledger/Data/Types.hs
Xitian9: I originally tried using the same `MixedAmount` type for `pamount`, but it ended up breaking some of the tests. I seem to recall that it was in the close command.

That said, I started work on this part a long time ago, and perhaps the problem was not actually with the data representation, but another issue which was resolved elsewhere, or could be. I'll do another check to see whether that can be done.

hledger-lib/Hledger/Data/Types.hs
Xitian9: Could do. We would lose the ordering information, which I believe would make the commodity when you combine multiple zero amounts unpredictable. Perhaps not the biggest issue, especially since this step is done only when calling `amounts` after this PR, so we could do the more expensive operation of sorting at that stage.

hledger-lib/Hledger/Data/Types.hs
Xitian9: Yes, I can add some haddocks. These fields are 
1. The commodity symbol of the main amount,
2. `Nothing` if there is not price, or if there is a price then `Just` the following: 
      a.  The commodity symbol of the price amount
      b.  `Just` the unit price, or `Nothing` if it's a total price

Xitian9: It's probably better to use and ADT, I was just copying over the same data representation as used in the previous normalisation command.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: Internal use only. Any function which does not change the key used can just operate on the values, preventing the need to deconstruct and reconstruct the Map each time. Many commonly used functions meet this criteria, so it will save a lot of heap space.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: It is reliable if we use a `Map` rather than a `HashMap`: it will have the same ordering as the keys.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: Correct. In particular, this now does a strict sum, rather than a lazy sum.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: I did not check if there was any performance improvement. Previously `mixed` worked only on `[Amount]`, and I added the specialise pragma to make sure that this most-common case was always optimised for.

hledger-lib/Hledger/Data/Types.hs
Xitian9: One issue is that HashMap does not have `partition` or `partitionWithKey` defined, both of which we use here. It seems this is just a matter of them not having been implemented yet, rather than any fundamental difficulties.

hledger-lib/Hledger/Data/Amount.hs
simonmichael: I tend to feel we shouldn't bother adding such pragmas unless/until we know they give a significant benefit, since they tend to be unpredictable across GHC versions and they cost extra maintenance and code-reading effort.

Xitian9: Got it.

hledger-lib/Hledger/Data/Types.hs
Xitian9: It turns out I was able to use MixedAmount for pamount.

Xitian9:
In the end I was able to use `MixedAmount` for the `pamount` field of Postings without any problem: whatever blocker there was originally must have gotten resolved. This also improves the performance numbers, which I've updated in the commit message. There is no longer anything that runs slower/uses more memory: it's all-around improvements. Those improvements are actually slightly better than before, but since we're also comparing against a better baseline the percentage improvements are slightly lower.

Let me know if you have any other questions.

Xitian9:
And of course, the requisite build failure due to Semigroups in ghc8.2…

Xitian9:
Alright, all checks passing now.

simonmichael (commented):
I'm going to run this locally for a bit, but I'm not seeing a reason not to merge it. If anyone else has thoughts, let me know @adept @ony @glguy @thielema ...

simonmichael:
With the latest I'm seeing benchmark results like this:
```
$ BENCHEXES=hledger.pre1509,hledger.1509 make bench -n2 -N3
...
+----------------------------------------------------++-----------------+--------------+
|                                                    || hledger.pre1509 | hledger.1509 |
+====================================================++=================+==============+
| -f examples/10000x1000x10.journal print            ||            2.83 |         0.85 |
| -f examples/10000x1000x10.journal register         ||           12.69 |         7.10 |
| -f examples/10000x1000x10.journal balance          ||            2.76 |         0.83 |
| -f examples/1000x1000x10.journal balance --weekly  ||            1.74 |         0.72 |
| -f examples/10000x1000x10.journal balance --weekly ||           23.08 |         9.83 |
+----------------------------------------------------++-----------------+--------------+
```
which seems outstanding.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -230,7 +235,30 @@ data Amount = Amount {
...
+
+-- | Stores the CommoditySymbol of the Amount, along with the CommoditySymbol of
+-- the price, and its unit price if being used.
+data MixedAmountKey
+  = MixedAmountKeyNoPrice    !CommoditySymbol
+  | MixedAmountKeyTotalPrice !CommoditySymbol !CommoditySymbol
+  | MixedAmountKeyUnitPrice  !CommoditySymbol !CommoditySymbol !Quantity
+  deriving (Eq,Generic,Show)
+
+instance Ord MixedAmountKey where
```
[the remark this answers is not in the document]
Xitian9: It's because it wouldn't be the order we want. The auto-derived instance would put all Amounts with no price before all Amounts with a price, regardless of their commodity. I can add this in a comment.

Xitian9:
Those numbers look a bit too good to me. In particular, I wouldn't expect any benefit in the print command. You may be using an unoptimised build for pre1509.

Xitian9:
Hold off merging this. Changing back to using `MixedAmount` for `pamount` has brought up some bugs with balance assertions that the tests didn't catch.

simonmichael:
Thanks. Here are more realistic numbers. (I ran this a bunch of times and saw a slowdown in register surprisingly often.. but let's assume it's just variance from the low-tech benchmarking method).
```
$ quickbench -w hledger.pre1509,hledger.1509 -n3 -N3
Running 5 tests 3 times with 2 executables at 2021-04-12 17:51:59 HST:

Best times 1:
+----------------------------------------------------++-----------------+--------------+
|                                                    || hledger.pre1509 | hledger.1509 |
+====================================================++=================+==============+
| -f examples/10000x1000x10.journal print            ||            0.96 |         0.87 |
| -f examples/10000x1000x10.journal register         ||            7.01 |         6.63 |
| -f examples/10000x1000x10.journal balance          ||            0.96 |         0.82 |
| -f examples/1000x1000x10.journal balance --weekly  ||            0.79 |         0.67 |
| -f examples/10000x1000x10.journal balance --weekly ||           10.48 |        10.17 |
+----------------------------------------------------++-----------------+--------------+

Best times 2:
+----------------------------------------------------++-----------------+--------------+
|                                                    || hledger.pre1509 | hledger.1509 |
+====================================================++=================+==============+
| -f examples/10000x1000x10.journal print            ||            0.93 |         0.86 |
| -f examples/10000x1000x10.journal register         ||            6.46 |         6.18 |
| -f examples/10000x1000x10.journal balance          ||            0.94 |         0.84 |
| -f examples/1000x1000x10.journal balance --weekly  ||            0.79 |         0.69 |
| -f examples/10000x1000x10.journal balance --weekly ||           11.17 |        10.33 |
+----------------------------------------------------++-----------------+--------------+

Best times 3:
+----------------------------------------------------++-----------------+--------------+
|                                                    || hledger.pre1509 | hledger.1509 |
+====================================================++=================+==============+
| -f examples/10000x1000x10.journal print            ||            0.94 |         0.87 |
| -f examples/10000x1000x10.journal register         ||            6.09 |         6.89 |
| -f examples/10000x1000x10.journal balance          ||            0.96 |         0.83 |
| -f examples/1000x1000x10.journal balance --weekly  ||            0.80 |         0.69 |
| -f examples/10000x1000x10.journal balance --weekly ||           11.07 |         9.63 |
+----------------------------------------------------++-----------------+--------------+
```

Xitian9:
Okay, I think I've got it. I added another test and refactored `priceInferrerFor`.

hledger-lib/Hledger/Data/Amount.hs
simonmichael: I imagine someone refactoring in future could easily overlook ordering here and start generating different/variable JSON output as a result. I feel we should say something about it in the haddock, specify some kind of invariant. The order of amounts will be... what ? Determined by the ordering of MixedAmountKey, I guess. Which means.. amounts will be ordered by commodity symbol (alphabetically), then by price type (none, unit, total), then by price amount (from most negative to most positive) ?

simonmichael: PS and I probably got that ordering wrong - I have now checked MixedAmountKey and I think we could still clarify that MixedAmountKey haddock a bit to make the ordering more obvious.

hledger-lib/Hledger/Data/Transaction.hs
```diff
@@ -551,38 +551,37 @@ inferBalancingPrices t@Transaction{tpostings=ps} = t{tpostings=ps'}
...
-
-    inferprice p@Posting{pamount=Mixed [a]}
-      | caninferprices && ptype p == pt && acommodity a == fromcommodity
-        = p{pamount=mixedAmount $ a{aprice=Just conversionprice}, poriginal=Just $ originalPosting p}
+    postings     = filter ((==pt).ptype) $ tpostings t
+    pcommodities = map acommodity $ concatMap (amounts . pamount) postings
+    sumamounts   = amounts $ sumPostings postings  -- sum normalises to one amount per commodity & price
+    noprices     = all (isNothing . aprice) sumamounts
+
+    -- Determine the amounts we are inferring from and to based on the order of commodities in pcommodities
```
simonmichael, line 561: Does this change any user-visible behaviour ? 

More generally, priceInferrerFor looks a bit hard to understand - my fault, but since it's fresh in your mind if you see any additional comments that would help please feel free to add.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: Sure, I'll add some more comments to both of those.

hledger-lib/Hledger/Data/Transaction.hs
Xitian9: It should not change any user-visible behaviour. I made the changes one replacement at a time, checking that all tests passed at each stage. I agree that the function is a bit hard to understand, and I'm happy to add more comments.

Xitian9: I've added some more comments. Let me know if there's anything that you think could be clarified further.

simonmichael:
priceInferrerFor comments look great!

I'm out of time for today, a few things still on my todo list:
- test/understand balance assertions test (I tried once and it wouldn't balance, complaining about same sign)
- compare old/new debug output
- test in production for a bit

Xitian9:
Sounds good to me. The transactions which consist of only balance assignments are easy to mess up, and I encountered complaints about the same sign too. That's the reason for the first set-up transaction in the tests I added, so that the quantity of A in account `c` is going down by 50 A, while the quantity of B is going up by 50.

```
2013/1/5
  (c)    100 A

2013/1/5
  c      = 50 B
  c      = 50 A
```

simonmichael:
Old debug output (from `hledger -f examples/sample.journal bal --debug=7`):
```
     , prrAmounts =
       [ Mixed
         [ Amount
           { acommodity = "$"
           , aquantity = 0
           , aismultiplier = False
           , astyle = AmountStylePP "L False Precision 0 Just '.' Nothing.."
           , aprice = Nothing
           }
         ]
       ]
     , prrTotal = Mixed
       [ Amount
         { acommodity = "$"
         , aquantity = 0
         , aismultiplier = False
         , astyle = AmountStylePP "L False Precision 0 Just '.' Nothing.."
         , aprice = Nothing
         }
       ]
     , prrAverage = Mixed
       [ Amount
         { acommodity = "$"
         , aquantity = 0
         , aismultiplier = False
         , astyle = AmountStylePP "L False Precision 0 Just '.' Nothing.."
         , aprice = Nothing
         }
       ]
```
New:
```
     , prrAmounts =
       [ Mixed
         ( fromList
           [
             ( MixedAmountKeyNoPrice "$"
             , Amount
               { acommodity = "$"
               , aquantity = 0
               , aismultiplier = False
               , astyle = AmountStylePP "L False Precision 0 Just '.' Nothing.."
               , aprice = Nothing
               }
             )
           ]
         )
       ]
     , prrTotal = Mixed
       ( fromList
         [
           ( MixedAmountKeyNoPrice "$"
           , Amount
             { acommodity = "$"
             , aquantity = 0
             , aismultiplier = False
             , astyle = AmountStylePP "L False Precision 0 Just '.' Nothing.."
             , aprice = Nothing
             }
           )
         ]
       )
     , prrAverage = Mixed
       ( fromList
         [
           ( MixedAmountKeyNoPrice "$"
           , Amount
             { acommodity = "$"
             , aquantity = 0
             , aismultiplier = False
             , astyle = AmountStylePP "L False Precision 0 Just '.' Nothing.."
             , aprice = Nothing
             }
           )
         ]
       )
```

simonmichael:
Or with a price (from examples/100x100x10.journal):
Old:
```
     , pamount=Mixed
       [ Amount
         { acommodity = "N"
         , aquantity = 14
         , aismultiplier = False
         , astyle = AmountStylePP "R True Precision 2 Just '.' Nothing.."
         , aprice = Just
           ( TotalPrice
             ( Amount
               { acommodity = "O"
               , aquantity = 14
               , aismultiplier = False
               , astyle = AmountStylePP "R True Precision 0 Nothing Nothing.."
               , aprice = Nothing
               }
             )
           )
         }
       ]
```
New:
```
     , pamount=Mixed
       ( fromList
         [
           ( MixedAmountKeyTotalPrice "N" "O"
           , Amount
             { acommodity = "N"
             , aquantity = 14
             , aismultiplier = False
             , astyle = AmountStylePP "R True Precision 2 Just '.' Nothing.."
             , aprice = Just
               ( TotalPrice
                 ( Amount
                   { acommodity = "O"
                   , aquantity = 14
                   , aismultiplier = False
                   , astyle = AmountStylePP "R True Precision 0 Nothing Nothing.."
                   , aprice = Nothing
                   }
                 )
               )
             }
           )
         ]
       )
```

simonmichael:
I found the behaviour in the new balance assertions test a little surprising:
```
2013/1/1
  a    $1  =$1
  b        =-$1

2013/1/3
  a    $6  = $7
  b   $-1  = $-2
  b   $-1  = $-3
  b   $-7    = $-10
  b   $-1
  b   $-1  = $-12
  b

2013/1/4
  a    $0 = $7
  b   $0   = $-7

2013/1/5
  (c)    100 A

2013/1/5
  c      = 50 B
  c      = 50 A
```
```
$ hledger print -x c
2013-01-05
    (c)           100 A

2013-01-05
    c    50 B @@ 50 A = 50 B
    c           -50 A = 50 A

```
Because it's a single-commodity assertion (`=`), I expected the first assertion on c to just post 50B; the `@@ 50 A` price seems unnecessary. Ie:
```
2013-01-05
    c            50 B = 50 B
    c           -50 A = 50 A

```
I see this is not new behaviour, all releases back to 1.19 do it. Not part of this issue I guess.

simonmichael:
Ohh... they are the same. Got it.

Xitian9:
Actually, it occurs to me there's the option for some generalisation in priceInferrerFor. Now that we allow negative total prices, we can actually produce valid prices in more cases that previously.

In the following line, we can replace `sign (abs toAmount)` with `negate toAmount`. This still passes all tests, except two tests which are supposed to produce parse errors now give output. It also allows us to clean up the code a little, as we no longer need to check `totalpricesign`.

https://github.com/simonmichael/hledger/pull/1509/commits/50d428e384034482101c61c702a6009fcbd47351#diff-051c1411484e9789d47bafacc03e4c238801a585b51eaa70a02239ae712f3480R586

On the other hand, negative prices are not commonly needed, so automatically inferring them may cause more problems than it solves. If somebody really wants a negative price, requiring them to enter it explicitly is reasonable.

However, I strongly suspect that the current implementation is masking some bugs we're not testing for. I'll see if I can find any.

Xitian9:
Looking a bit more closely, it looks like the price inferrer will produce incorrect results only in cases which will produce negative prices, which are then caught by the ‘all amounts have the same posting’ check, so the errors don't end up leaking through.

That said, I think it would be better to check for this in `inferFromAndTo`, and would take the form of checking whether `a` and `b` have opposite signs. What do you think?

Xitian9:
I've added a new commit which implements the above. It should eventually be squashed into the previous commit, but I'll leave it separate for now so the changes are clearer. If you decide that inferring negative prices is reasonable, we can just remove the opposite signs check.

Xitian9:
I have a couple of `inferPriceFor` changes, and there completely independent of the changes to `MixedAmount`, so I've taken that last commit out of here, and it will be in a follow-up PR if and when this one gets merged.

Xitian9:
Have you had a chance to test this out? I've been running this code as my day-to-day hledger for a couple months now, and haven't had any problems with it, so I'm feeling pretty good about the current state of affairs.

simonmichael:
Sorry for the holdup. I hope to get back to PRs/code review fairly soon. I have been running it too. Let's merge this and move forward, thank you very much @Xitian9.

Issue #1507, opened 2021-03-18
Regression: empty accounts in balance and balancesheet reports


markokocic:
I updated today to the latest git commit and observed that some accounts with 0 started to appear in balance and balancesheet report.

I tracked it down, and it seems that it happens only to the accounts that had transactions with more than one currency in the past.

Steps to reproduce:
``` bash
> cat test
2014-07-07 Desc 1
    liabilities:a       1200 inr
    liabilities:p

2014-07-13 Desc 2
    liabilities:t      900 inr
    liabilities:p

2014-07-13 Desc 3
    expenses:m          1,407.00 inr
    expenses:b           550.00 inr
    liabilities:p

2014-07-18 Desc 4
    liabilities:p    4,057.00 inr @@ 50.00 €
    expenses:cash

> hledger bal -f test
          550.00 inr  expenses:b
            -50.00 €  expenses:cash
        1,407.00 inr  expenses:m
        1,200.00 inr  liabilities:a
                   0  liabilities:p
          900.00 inr  liabilities:t
--------------------
        4,057.00 inr
            -50.00 €
```
Here `liablilities:p` account should not be displayed because its value is `0`

Xitian9:
I think the problem here is not the presence of multiple commodities, but in the fact that its amount is balanced only in the primary amount, not the cost amount (with prices the balance is `0.00 inr @@ 50.00 €`). Since we're not displaying prices the fact that the cost is non-zero should obviously be irrelevant.

markokocic:
I'm not sure about this. I just changed the test case as following:

```
> cat test
2014-07-07 Desc 1
    liabilities:a       1200 inr
    liabilities:p

2014-07-13 Desc 2
    liabilities:t      900 inr
    liabilities:p

2014-07-13 Desc 3
    expenses:m          1,407.00 inr
    expenses:b           550.00 inr
    liabilities:p

2014-07-18 Desc 4
    liabilities:p    4,057.00 inr
    expenses:cash  -50 €

> hledger bal -f test
          550.00 inr  expenses:b
               -50 €  expenses:cash
        1,407.00 inr  expenses:m
        1,200.00 inr  liabilities:a
                   0  liabilities:p
          900.00 inr  liabilities:t
--------------------
        4,057.00 inr
               -50 €
```
Now the only postings that appear on `liabilities:p` side are in `inr`, but it still gets displayed as 0.

The minimal test case that demonstrates the issue:
```
> cat test2
2014-07-07 Desc 1
[...]

PR #1511, opened 2021-03-22
Strip prices from balance and register commands if unneeded.

Xitian9:
Addresses #1507.

### DIFF
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -558,14 +558,22 @@ cumulativeSum value start = snd . M.mapAccumWithKey accumValued start
 postingAndAccountValuations :: ReportSpec -> Journal -> PriceOracle
                             -> (DateSpan -> Posting -> Posting, DateSpan -> Account -> Account)
 postingAndAccountValuations ReportSpec{rsToday=today, rsOpts=ropts} j priceoracle = case value_ ropts of
-    Just (AtEnd _) -> (const id, avalue' (cost_ ropts) (value_ ropts))
-    _              -> (pvalue' (cost_ ropts) (value_ ropts), const id)
+    -- If we're doing AtEnd valuation, we may need to value the same posting at different dates
+    -- (for example, when preparing a ValueChange report). So we should only convert to cost and
+    -- maybe strip prices from the Posting, and should do valuation on the Accounts.
+    Just v@(AtEnd _) -> (pvalue Nothing, avalue v)
+    -- Otherwise, all costing and valuation should be done on the Postings.
+    _                -> (pvalue (value_ ropts), const id)
   where
-    avalue' c v span a = a{aibalance = value (aibalance a), aebalance = value (aebalance a)}
-      where value = mixedAmountApplyCostValuation priceoracle styles (end span) today (error "multiBalanceReport: did not expect amount valuation to be called ") c v  -- PARTIAL: should not happen
-    pvalue' c v span = postingApplyCostValuation priceoracle styles (end span) today c v
-    end = fromMaybe (error "multiBalanceReport: expected all spans to have an end date")  -- XXX should not happen
-        . fmap (addDays (-1)) . spanEnd
+    -- For a Posting: convert to cost, apply valuation, then strip prices if we don't need them (See issue #1507).
+    pvalue v span = maybeStripPrices . postingApplyCostValuation priceoracle styles (end span) today (cost_ ropts) v
+    -- For an Account: Apply valuation to both the inclusive and exclusive balances.
+    avalue v span a = a{aibalance = value (aibalance a), aebalance = value (aebalance a)}
+      where value = mixedAmountApplyValuation priceoracle styles (end span) today (error "multiBalanceReport: did not expect amount valuation to be called ") v  -- PARTIAL: should not happen
+
+    maybeStripPrices = if show_costs_ ropts then id else postingStripPrices
+    end = maybe (error "multiBalanceReport: expected all spans to have an end date")  -- PARTIAL: should not happen
+            (addDays (-1)) . spanEnd
     styles = journalCommodityStyles j
 
 -- tests
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -92,7 +92,7 @@ postingsReport rspec@ReportSpec{rsOpts=ropts@ReportOpts{..}} j = items
             reportPeriodOrJournalLastDay rspec j
 
       -- Posting report does not use prices after valuation, so remove them.
-      displaypsnoprices = map (\(p,md) -> (removePrices p, md)) displayps
+      displaypsnoprices = map (\(p,md) -> (postingStripPrices p, md)) displayps
 
       -- Posting report items ready for display.
       items =
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -91,8 +91,9 @@ postingsReport rspec@ReportSpec{rsOpts=ropts@ReportOpts{..}} j = items
             fromMaybe (error' "postingsReport: expected a non-empty journal") $  -- PARTIAL: shouldn't happen
             reportPeriodOrJournalLastDay rspec j
 
-      -- Posting report does not use prices after valuation, so remove them.
-      displaypsnoprices = map (\(p,md) -> (postingStripPrices p, md)) displayps
+      -- Strip prices from postings if we won't need them.
+      displaypsnoprices = map (\(p,md) -> (maybeStripPrices p, md)) displayps
+        where maybeStripPrices = if show_costs_ then id else postingStripPrices
 
       -- Posting report items ready for display.
       items =
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -76,7 +76,9 @@ postingsReport rspec@ReportSpec{rsOpts=ropts@ReportOpts{..}} j = items
       (precedingps, reportps) = matchedPostingsBeforeAndDuring rspec j reportspan
 
       -- We may be converting posting amounts to value, per hledger_options.m4.md "Effect of --value on reports".
-      pvalue periodlast = postingApplyCostValuation priceoracle styles periodlast (rsToday rspec) cost_ value_
+      -- Strip prices from postings if we won't need them.
+      pvalue periodlast = maybeStripPrices . postingApplyCostValuation priceoracle styles periodlast (rsToday rspec) cost_ value_
+        where maybeStripPrices = if show_costs_ then id else postingStripPrices
 
       -- Postings, or summary postings with their subperiod's end date, to be displayed.
       displayps :: [(Posting, Maybe Day)]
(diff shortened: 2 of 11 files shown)


hledger/test/journal/valuation.test
```diff
@@ -360,9 +360,9 @@ $ hledger -f- bal -N --value=2000-01-15 -b 2000
 $ hledger -f- bal -N --value=now -b 2000
                 12 B  a
 
-# 33. single column balance report valued at default date (same as --value=end)
-$ hledger -f- bal -N -V -b 2000
-                12 B  a
+# 33. single column balance report valued at cost and end (should be the same as --cost in this case)
+$ hledger -f- bal -N --cost --value=end -b 2000
+                21 B  a
```
simonmichael, line 366: What drove this change ? Is it related to the rest of the PR ? Perhaps just a minor fixup, but I'm struggling to grok things.

hledger/test/journal/transaction-prices.test
```diff
@@ -210,7 +210,19 @@ hledger -f - balance --no-total -E
                  -1Y  b
 >>>= 0
 
-# 18. the above with -B
+# 18. The above, but hides accounts with zero amount but non-zero cost
```
simonmichael, line 213: Or perhaps: 
Without -E, a should be hidden since its balance is zero - even though it has a non-zero cost.

hledger-lib/Hledger/Reports/ReportOptions.hs
```diff
@@ -118,6 +118,7 @@ data ReportOpts = ReportOpts {
     ,drop_           :: Int
     ,row_total_      :: Bool
     ,no_total_       :: Bool
+    ,show_costs_     :: Bool  -- ^ Whether to show costs for reports which normally don't show them
```
simonmichael, line 121: I haven't got my head around the larger effect of this, so I'll just ask questions. Will --show-costs now be understood by more commands ? Which ones ? Will/should it appear in more commands' help ? And in the manual ? Will some commands show it in help but ignore it ? Will it cause bad effects with any commands ?

hledger-lib/Hledger/Reports/PostingsReport.hs
```diff
@@ -92,7 +92,8 @@ postingsReport rspec@ReportSpec{rsOpts=ropts@ReportOpts{..}} j = items
             reportPeriodOrJournalLastDay rspec j
 
       -- Posting report does not use prices after valuation, so remove them.
```
simonmichael, line 94: The change seems to invalidate this comment - remove/change it ?

hledger-lib/Hledger/Reports/MultiBalanceReport.hs
```diff
@@ -558,15 +558,16 @@ cumulativeSum value start = snd . M.mapAccumWithKey accumValued start
...
   where
-    avalue' c v span a = a{aibalance = value (aibalance a), aebalance = value (aebalance a)}
-      where value = mixedAmountApplyCostValuation priceoracle styles (end span) today (error "multiBalanceReport: did not expect amount valuation to be called ") c v  -- PARTIAL: should not happen
-    pvalue' c v span = postingApplyCostValuation priceoracle styles (end span) today c v
-    end = fromMaybe (error "multiBalanceReport: expected all spans to have an end date")  -- XXX should not happen
-        . fmap (addDays (-1)) . spanEnd
+    pvalue' c v span = maybeStripPrices . postingApplyCostValuation priceoracle styles (end span) today c v
+    avalue' v span a = a{aibalance = value (aibalance a), aebalance = value (aebalance a)}
+      where value = mixedAmountApplyValuation priceoracle styles (end span) today (error "multiBalanceReport: did not expect amount valuation to be called ") v  -- PARTIAL: should not happen
+    end = maybe (error "multiBalanceReport: expected all spans to have an end date")  -- XXX should not happen
```
simonmichael, line 567: let's also change that XXX to our standard PARTIAL: marker

hledger-lib/Hledger/Reports/MultiBalanceReport.hs
```diff
@@ -558,15 +558,16 @@ cumulativeSum value start = snd . M.mapAccumWithKey accumValued start
...
-    _              -> (pvalue' (cost_ ropts) (value_ ropts), const id)
+    Just v@(AtEnd _) -> (pvalue' (cost_ ropts) Nothing, avalue' v)
+    _                -> (pvalue' (cost_ ropts) (value_ ropts), const id)
   where
-    avalue' c v span a = a{aibalance = value (aibalance a), aebalance = value (aebalance a)}
-      where value = mixedAmountApplyCostValuation priceoracle styles (end span) today (error "multiBalanceReport: did not expect amount valuation to be called ") c v  -- PARTIAL: should not happen
-    pvalue' c v span = postingApplyCostValuation priceoracle styles (end span) today c v
-    end = fromMaybe (error "multiBalanceReport: expected all spans to have an end date")  -- XXX should not happen
-        . fmap (addDays (-1)) . spanEnd
+    pvalue' c v span = maybeStripPrices . postingApplyCostValuation priceoracle styles (end span) today c v
```
simonmichael, line 564: I feel this function is getting harder to understand for code readers. Could you try to clarify what's going on with a comment - why we might be stripping prices here, eg. Feel free to mention issue numbers if you think it will help.

hledger/Hledger/Cli/Commands/Register.hs
```diff
@@ -93,8 +93,9 @@ postingsReportItemAsCsvRecord (_, _, _, p, b) = [idx,date,code,desc,acct,amt,bal
                              BalancedVirtualPosting -> wrap "[" "]"
                              VirtualPosting -> wrap "(" ")"
                              _ -> id
-    amt = wbToText . showMixedAmountB oneLine $ pamount p
-    bal = wbToText $ showMixedAmountB oneLine b
+    -- We should already have stripped prices, so no need to do it again
```
simonmichael, line 96: As elsewhere, it's a bit unclear for readers why/when prices were/should be stripped - perhaps you can add a little more context ?

hledger/Hledger/Cli/Commands/Register.hs
```diff
@@ -190,12 +191,16 @@ postingsReportItemAsText opts preferredamtwidth preferredbalwidth (mdate, mendda
...
     amt = showAmountsLinesB dopts . (\x -> if null x then [nullamt] else x) . amounts $ pamount p
     bal = showAmountsLinesB dopts $ amounts b
-    dopts = noPrice{displayColour=color_ . rsOpts $ reportspec_ opts}
+    dopts = registerOpts opts
     -- Since this will usually be called with the knot tied between this(amt|bal)width and
     -- preferred(amt|bal)width, make sure the former do not depend on the latter to avoid loops.
     thisamtwidth = maximumDef 0 $ map wbWidth amt
     thisbalwidth = maximumDef 0 $ map wbWidth bal
 
+registerOpts :: CliOpts -> AmountDisplayOpts
```
simonmichael, line 200: What's the purpose of this ?

hledger/Hledger/Cli/Commands/Close.hs
```diff
@@ -82,14 +82,9 @@ close CliOpts{rawopts_=rawopts, reportspec_=rspec} j = do
...
 
-    -- should we preserve cost information ?
-    normalise = case boolopt "show-costs" rawopts of
-                  True  -> normaliseMixedAmount
-                  False -> normaliseMixedAmount . mixedAmountStripPrices
-
     -- the balances to close
     (acctbals,_) = balanceReport rspec_ j
-    totalamt = maSum $ map (\(_,_,_,b) -> normalise b) acctbals
+    totalamt = maSum $ map (\(_,_,_,b) -> b) acctbals
```
simonmichael, line 87: Maybe no further comment needed here, but for my edification could you say briefly why we don't have to normalise here now ?

hledger/Hledger/Cli/Commands/Balance.hs
```diff
@@ -388,11 +388,11 @@ balance opts@CliOpts{reportspec_=rspec} j = case reporttype_ of
 balanceReportAsCsv :: ReportOpts -> BalanceReport -> CSV
 balanceReportAsCsv opts (items, total) =
   ["account","balance"] :
-  [[accountNameDrop (drop_ opts) a, wbToText $ showMixedAmountB oneLine b] | (a, _, _, b) <- items]
+  [[accountNameDrop (drop_ opts) a, wbToText $ showMixedAmountB (balanceOpts False opts) b] | (a, _, _, b) <- items]
```
simonmichael, line 391: Could you say a little more about the changes in this file ? What drove them ?

simonmichael (commented):
I'm having a hard time grokking this one so I've added a bunch of dumb questions.

@Xitian9 if you ever want to ping me on chat and/or a call, I might be able to give speedier realtime feedback or get up to speed quicker so we can get these merged quicker.

hledger/test/journal/valuation.test
Xitian9: This was because I initially had an implementation which passed all the tests, but on closer inspection would have failed this test. I noticed we didn't have any tests where both cost and valuation are done, so I added one. The old test here was really a duplicate of a previous one (same code accessed by different flags), so I thought replacing it was a better plan than renumbering all the tests.

simonmichael:
> This was because I initially had an implementation which passed all the tests, but on closer inspection would have failed this test. I noticed we didn't have any tests where both cost and valuation are done, so I added one. The old test here was really a duplicate of a previous one (same code accessed by different flags), so I thought replacing it was a better plan than renumbering all the tests.
> 
Cool. If it's not too much bother, maybe I'd separate this into its own commit, or mention it in the overall commit message. ("Also replaced a redundant balance test with a more useful one.")

hledger/test/journal/transaction-prices.test
Xitian9: Sounds reasonable.

hledger-lib/Hledger/Reports/ReportOptions.hs
Xitian9: As implemented here, it will not be understood by more commands: it will still only be understood by the close command. The only difference at this stage is that the corresponding Bool has been moved into `ReportOpts`, rather than living in `Close.hs`. This change obviously suggests that the flag could be understood more widely, but it would require some more changes to work properly. That's probably an issue for another PR, if it's even decided it's worthwhile at all (I'm sceptical at this stage).

hledger-lib/Hledger/Reports/PostingsReport.hs
Xitian9: Will do.

hledger-lib/Hledger/Reports/MultiBalanceReport.hs
Xitian9: Sure.

hledger/Hledger/Cli/Commands/Register.hs
Xitian9: Got it.

hledger/Hledger/Cli/Commands/Register.hs
Xitian9: This, along with the changes in Balance.hs, was an attempt to reduce code duplication. I'm not altogether satisfied with the result, so perhaps another approach would be better.

hledger/Hledger/Cli/Commands/Close.hs
Xitian9: `normalise` did two things:
1. Strip prices, which has already been done.
2. Normalise the mixed amount, which is done anyway when you call `maSum`.

hledger/Hledger/Cli/Commands/Balance.hs
Xitian9: An attempt to reduce code duplication and make sure the display opts used in different parts of the balance report don't get out of sync. I'm inclined to leave this one in, but let me know if you disagree.

hledger/test/journal/valuation.test
Xitian9: Moved into a separate PR.

Xitian9:
Also now addresses #1522.

PR #1513, opened 2021-03-24
Tolerate spaces in amount fields in CSV files

glguy:
This allows the sign simplification to ignore leading and trailing spaces. This is especially important for leading spaces where a leading space causes the eventual amountp parser to fail.

Example problem input now allowed: `"+ $1"`


simonmichael:
Thanks @glguy ! Could you possibly add a test, eg in hledger/test/csv.test ? Would help catch regressions in future etc.

PR #1515, opened 2021-03-26
Tweak JSON serialisation

Xitian9:
Removes an unnecessary increase in precision for JSON representations for some Quantity.
Adds more efficient `toEncoding` definitions for some custom `ToJSON` declarations.

### DIFF
--- a/hledger-lib/Hledger/Data/Json.hs
+++ b/hledger-lib/Hledger/Data/Json.hs
@@ -41,8 +41,8 @@ import           Data.Aeson
 import           Data.Aeson.Encode.Pretty (encodePrettyToTextBuilder)
 --import           Data.Aeson.TH
 import qualified Data.ByteString.Lazy as BL
-import           Data.Decimal
-import           Data.Maybe
+import           Data.Decimal (DecimalRaw(..), roundTo)
+import           Data.Maybe (fromMaybe)
 import qualified Data.Text.Lazy    as TL
 import qualified Data.Text.Lazy.IO as TL
 import qualified Data.Text.Lazy.Builder as TB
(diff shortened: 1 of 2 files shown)


simonmichael:
Are these regression fixes / improvements you noticed after working on #1510 ? Any user-visible changes compared to the 1.21 release ? Does the "more efficient toEncoding" speed up large json output ?

Xitian9:
It's an improvement I noticed after working on #1510. I was confused as to why the serialisation of 1.0 was `10^10` with 9 decimal places instead of the more obvious 10 with 1 decimal place. This makes it more intuitive to understand. The user-visible change is only in the JSON output.

The “more efficient toEncoding” has no effect on our codebase, as we always use `aeson-pretty`, which needs the extra structure provided by `toJSON` rather than `toEncoding`. But it's simple to provide the `toEncoding` on top of the `toJSON`, and it may speed up others' code.

simonmichael:
Fair enough. Would you mind just posting some draft changelog text I can add later ?

Also, a heads up for @real-dam. I'll deploy this to demo.hledger.org when merged.

Xitian9:
CHANGELOG:

The JSON serialisation of `Decimal`/`Quantity` now has the same precision as the `Decimal`/`Quantity`, up to a maximum of 10 decimal places. Previously, it always had a precision of exactly 10 decimal places.

real-dam:
Thanks for the heads-up, @simonmichael. I noticed the merge and tried MoLe with demo.hledger.org. There were no surprises, JSON parsed OK.  

(I assume that the merge is part of the code behind demo.hledger.org now.)

simonmichael:
Yes it is deployed on demo.hledger.org, thanks for the update. (Aside: it would be great if hledger-web and the rest reported a git hash and date along with their version.)

PR #1518, opened 2021-04-03
lib,cli: Fix ledgerDateSpan, so that in takes transaction dates into account.

### LINKED ISSUE #772: a posting date can cause incorrect stats
The [posting date](http://hledger.org/manual.html#posting-dates) causes [stats](http://hledger.org/manual.html#stats) to not count this transaction:
```
2018/1/1
    (a)   1  ; date:2018/1/2
```
or
```
2018/1/1
    (a)  1 ; [2018/1/2]
```
result:
```
$ hledger -f a.j stats
Main file                : /Users/simon/src/PLAINTEXTACCOUNTING/hledger/a.j
Included files           : 
Transactions span        : 2018-01-02 to 2018-01-03 (1 days)
Last transaction         : none
Transactions             : 0 (0.0 per day)
Transactions last 30 days: 0 (0.0 per day)
Transactions last 7 days : 0 (0.0 per day)
Payees/descriptions      : 0
Accounts                 : 0 (depth 0)
Commodities              : 0 ()
```

Xitian9:
Fixes #772.

There may be a conversation to be had here on the value of `Hledger.Data.Ledger`. Aside from `ledgerFromJournal`, these functions are only used in two places in our codebase. As seen here, this kind of little-used functionality can lead to bugs. We could just trim out all functions aside from `ledgerFromJournal`, but this will affect the API.

### DIFF
--- a/hledger-lib/Hledger/Data/Ledger.hs
+++ b/hledger-lib/Hledger/Data/Ledger.hs
@@ -31,9 +31,7 @@ import Text.Printf
 import Hledger.Utils.Test
 import Hledger.Data.Types
 import Hledger.Data.Account
-import Hledger.Data.Dates (daysSpan)
 import Hledger.Data.Journal
-import Hledger.Data.Posting (postingDate)
 import Hledger.Query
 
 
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks!

I'm fine with trimming little-used functions from Hledger.Data.Ledger.

simonmichael:
..at least, if/when they give more trouble. If it ain't broke..

PR #1521, opened 2021-04-06
;lib,cli: Add Functor and Foldable instances for Journal and other types

kaol:
This redefines `Journal`, `Transaction`, `Posting` and `MixedAmount` as synonyms `Journal' Transaction`, `Transaction' Posting`, `Posting' MixedAmount` and `MixedAmount' Amount`, respectively.

`Journal'` and `Transaction'` get `Foldable` and `Functor` instances, `Posting'` gets `Functor` and `MixedAmount'` gets `Foldable`.

This allows using much of Haskell's base libraries for manipulating journals. Get all transactions? `toList`. Get all postings in a transaction? `toList`. Transform a Posting's amount? `fmap`. Get the list of amounts in the whole journal? `join . map (toList . pamount) . (toList <=< toList)`. Well okay, that last one isn't that practical.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -1,11 +1,13 @@
 {-# LANGUAGE PackageImports #-}
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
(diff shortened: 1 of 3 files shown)


simonmichael:
Thanks for exploring this @kaol!

This complicates the types a little, and I assume would be an incompatible API change. What real benefit do you see from doing this right now ? I think `toList <=< toList` isn't exactly something most people are looking for.

Xitian9:
Though I love these sorts of tricks, mucking up the types would be an annoying price to pay. An alternative which would provide a similar benefit, but avoid complicating the types, would be to provide lenses (or rather, traversals) for these fields instead.

Get all transactions? `j ^.. transactions`
Get all postings in a transaction? `t ^.. postings`
Transform a posting's amounts? `over postings`
Get the list of amounts in the whole journal? `j ^.. transactions . postings . amounts`

The main issues to resolve with a lens-based solution would be what naming scheme to use for the lenses, and what lens library to use (lens, microlens, optics).

simonmichael:
Closing.

PR #1528, opened 2021-04-13
add a Show instance for AmountDisplayOpts, WideBuilder

simonmichael:
For debug logging.


Xitian9:
Ha! I didn't realise there was a Show instance for Builder, but I guess the obvious solution of ‘convert to Text and then show’ works just fine.

simonmichael:
Whether this will produce usable debug output is another question.. I should test with something big.

Xitian9:
I just ran into an instance of hledger complaining about a lack of Show instance for WideBuilder. I would support merging this.

PR #1529, opened 2021-04-14
control color in debug output

simonmichael:
- lib: useColor, colorOption helpers usable anywhere
- lib: debug output now respects --color/NO_COLOR/ANSI support
- lib: debug output checks for color support on stderr, not stdout

I was capturing debug output into files for comparison, and couldn't turn off the ANSI codes. Debug output is part of "output" and users expect to control color there too. 

This adds a few more functions using unsafePerformIO to read IO state, similar to `debugLevel` but a little more adventurous - these ones check existence of an environment variable, and I suppose not necessarily at program startup. I think this usage is safe enough though. Also, I blindly copied the no-cse and NOINLINE pragmas, hopefully in a useful way.

Related: https://hledger.org/hledger.html#general-options -> `--color`

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -48,15 +48,12 @@ where
 
 import Control.Applicative ((<|>))
 import Data.List.Extra (nubSort)
-import Data.Maybe (fromMaybe, isJust, mapMaybe)
+import Data.Maybe (fromMaybe, mapMaybe)
 import qualified Data.Text as T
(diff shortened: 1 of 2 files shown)


simonmichael:
Perhaps I'll add
```
-- Caveats:
-- Existence of the NO_COLOR variable, and whether the output handle supports ANSI color,
-- might not be checked at program startup, but rather when this is (first?) evaluated.
-- When running code in GHCI, this module must be reloaded to see a change.
```

simonmichael:
Actually, env vars are read as they were at program startup. I see no significant caveats here.

Issue #1526, opened 2021-04-11
empty table cells in multi-period balance reports [since 1.20]


lestephane:
```
$ hledger bal -f tmp.journal --yearly Income:Gains
Balance changes in 2018-01-01..2021-12-31:

              || 2018  2019  2020  2021
==============++========================
 Income:Gains ||    0                 0
--------------++------------------------
              ||    0                 0
$ hledger bal -f tmp.journal -p 2019 Income:Gains
-93.56707511000000000000000000 EUR  Income:Gains
--------------------
-93.56707511000000000000000000 EUR
$ hledger bal -f tmp.journal -p 2020 Income:Gains
-1094.29560548000000000000000000 EUR  Income:Gains
--------------------
-1094.29560548000000000000000000 EUR
```

I don't understand what is going on here. The `--yearly` report has always worked for me, and that stopped recently. 
Now I cannot decide whether my understanding is wrong, or hledger has regressed. In any case, it is surprising.

lestephane:
Forgot to mention that it's on 1.21.

simonmichael:
Thanks for the report. That's unfortunate! Would you be able to boil it down to an example journal ?

lestephane:
> Thanks for the report. That's unfortunate! Would you be able to boil it down to an example journal ?

I'll try, but In the event that I'm unsuccessful, will you have a way to receive the tmp.journal securely?
gpg, onionshare, keybase, whatever...

lestephane:
I use a lot of piped `hledger print -f - ` commands, and `commodity` directives don't seem to be preserved in the output of those `print` commands. So `tmp.journal` does not have a `commodity` directive. Adding it back makes everything magically work.

```
(echo "commodity 1000.00 EUR"; cat tmp.journal) | hledger bal -f - --yearly Income:Gains
Balance changes in 2018-01-01..2021-12-31:

              || 2018        2019          2020  2021
==============++======================================
 Income:Gains ||    0  -93.57 EUR  -1094.30 EUR     0
--------------++--------------------------------------
              ||    0  -93.57 EUR  -1094.30 EUR     0
```

So the failure mode seems formatting related.

lestephane:
There's a rather large count of digits after the decimal point for most amounts in the `tmp.journal`, since it came from unrealized gains calculations computed in beancount.

```
[...]

PR #1530, opened 2021-04-14
lib: When displaying amounts on a single line, always display at leas…

Xitian9:
…t one amount, even if that would exceed the requested maximum width. Addresses #1526.


simonmichael:
Looks good to me, thanks!

Issue #1478, opened 2021-02-05
Show account names when many currencies


artkpv:
Thanks for this great tool! I have many currencies / tickets under one account in subaccounts. When I decrease depth or move all to one account (at the start of the year) then hledger-ui truncates account names on the left to show amounts in different currencies. Expected: configure account name width or show one account on multiple lines. Pls see image:

![image](https://user-images.githubusercontent.com/3657998/106999543-72485e00-6797-11eb-89bd-f3d37e5b9a19.png)

Example:
```
2021-02-05 
    asset:a             1000.00 A
    disbalance

2021-02-05 
    asset:a             1000.00 B
    disbalance

2021-02-05 
    asset:c             1000.00 C
    disbalance

2021-02-05 
    asset:a             1000.00 DDDD
    disbalance

2021-02-05 
    asset:a             1000.00 EEEE
    disbalance

2021-02-05 
    asset:a             1000.00 FFFF
    disbalance

2021-02-05 
    asset:a             1000.00 GGG
    disbalance

2021-02-05 
    asset:a             1000.00 HHH
    disbalance

2021-02-05 
    asset:a             1000.00 III
    disbalance

2021-02-05 
    asset:a             1000.00 JJJ
    disbalance

2021-02-05 
    asset:a             1000.00 KKK
    disbalance

2021-02-05 
    asset:a             1000.00 LLL
    disbalance

2021-02-05 
    asset:a             1000.00 MMM
    disbalance

2021-02-05 
    asset:a             1000.00 NNN
    disbalance

2021-02-05 
    asset:a             1000.00 OOO
    disbalance

```

simonmichael:
Thanks for the report. I agree something should be done. Multiple lines per item though is problematic, to implement with the brick library at least, I believe.

simonmichael:
Actually I would expect it to elide some of those amounts. Can you try the latest version (1.20.x) if you haven't already.

Xitian9:
I'm also getting this bug in master. It's easy enough to tell hledger-ui to elide to a given width, the question is how to we keep this adjustable so that you can see everything if you want.

Options:
- Elide to given width by default, but accept a `--no-elide` flag and/or toggle to disable this.
- Control elision width directly somehow

But even with this, there's still another problem to solve: the list of commodities may be so long that it's not possible to display everything. It may be worth considering whether anything should be done in these cases (multi-line, horizontal scrolling, option to open a dialogue which displays it in full?).

[...]

Issue #1474, opened 2021-01-31
Too many commodities (eg. stocks) make the tree view in hledger-ui unusable


mrkajetanp:
Hello!
I'm loving hledger so far but I've just stumbled onto a pretty annoying issue.
Namely, if I add too many different commodities (in my case it's >20 different stocks), whenever I switch to tree view hledger-ui becomes unusable because the long list of stocks accumulated in the parent Investment account fills out the whole screen and removes even all the account names from the view.
Would it be possible to somehow make the ui trim that list if it gets too long? I'm sure it'd be a great usability improvement for such cases. The functionality already seems to be implemented in hledger because "hledger bs" generates a balance sheet that trims this list and just says "20 more..".
I'd implement it myself but sadly I've never used Haskell so I'm hoping someone here could help with that.
Even more ideal would be if there was an option to be added when declaring a commodity that determines whether it can be trimmed or not (I'd like to trim stocks if it gets too long but keep traditional currencies in the front).
As an aside, is there some way to change the ordering of commodities/currencies when they're accumulated in tree view? I'd love to be able to make my currencies appear before my stocks and other assets. Currently the order seems to be pretty random.

simonmichael:
Hello @mrkajetanp, thanks for the report and sorry for the late reply. We think #1531 mostly resolves this. Commodities should be displayed in alphabetic order (not random). There's currently no way to change this, but we could imagine controlling it by the order of commodity directives in the journal, similar to how display order of accounts is controlled. I'd be glad to test a PR for that.

PR #1531, opened 2021-04-14
ui: In account screen, elide amounts which go past the specified width.

Xitian9:
Also leave at least 15 spaces for account names.

This addresses #1474 , and goes some way to addressing #1478.

### DIFF
--- a/hledger-ui/Hledger/UI/AccountsScreen.hs
+++ b/hledger-ui/Hledger/UI/AccountsScreen.hs
@@ -121,10 +121,10 @@ asDraw UIState{aopts=_uopts@UIOpts{cliopts_=copts@CliOpts{reportspec_=rspec}}
               ,aScreen=s@AccountsScreen{}
               ,aMode=mode
               } =
-  case mode of
-    Help       -> [helpDialog copts, maincontent]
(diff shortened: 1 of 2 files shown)


simonmichael:
This seems much better!

I think I see what you mean by "balance column's jumping around" - probably the accounts screen should keep balances right aligned with the right edge ? That could be a separate task, either.

I was thinking that there is some value to a fixed, constant layout - the eye always knows where to look, and it's easy to compare data in multiple hledger-ui instances by flipping between their windows/tabs. But I think this adaptive scheme will be the most usable.

Xitian9:
That ‘balance column jumping around’ comment actually wasn't from me: it was there before.

I can see arguments both in favour of and against right-aligning the balances. My impression is that it was done this was so that there weren't huge amounts of wasted white space in the middle making it difficult to determine which balance goes with which account. Granted this problem still exists in many situations, so we can't really say this is a complete solution.

An option would be to do some sort of even-odd row colouring differences. But I agree this is probably a separate issue.

simonmichael:
Oops, sorry. I agree that right-aligning could make it harder to see the rows if you run hledger-ui in a really wide terminal window. Let's continue to ignore that comment, I'll merge this.

PR #1537, opened 2021-04-22
Clean up some String/Text utility code.

Xitian9:
This cuts a few unnecessary conversions out, and allows us to remove/clean up a large amount of redundant utility code.

### DIFF
--- a/hledger-lib/Hledger/Utils/Text.hs
+++ b/hledger-lib/Hledger/Utils/Text.hs
@@ -6,61 +6,48 @@
 
 module Hledger.Utils.Text
   (
- -- -- * misc
- -- lowercase,
- -- uppercase,
- -- underline,
- -- stripbrackets,
+  -- * misc
+  -- lowercase,
+  -- uppercase,
+  -- underline,
+  -- stripbrackets,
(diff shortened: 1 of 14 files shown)


hledger-lib/Hledger/Utils/Text.hs
```diff
@@ -207,71 +194,6 @@ textConcatBottomPadded :: [Text] -> Text
 textConcatBottomPadded = TL.toStrict . renderRow def{tableBorders=False, borderSpaces=False}
                        . Group NoLine . map (Header . textCell TopLeft)
 
--- -- | Join multi-line strings horizontally, after compressing each of
--- -- them to a single line with a comma and space between each original line.
--- concatOneLine :: [String] -> String
--- concatOneLine strs = concat $ map ((intercalate ", ").lines) strs
```
simonmichael, line 213: These were reminders+templates for doing a full port of this collection of text utilities, on the theory that they might be useful to others even if our usage fluctuates, but it hasn't happened yet, ok.

simonmichael (approved):

simonmichael:
Thanks!

PR #1540, opened 2021-04-23
Allow yesod-form 1.7

felixonmars:
Builds fine and all tests pass here.

### DIFF
--- a/hledger-web/hledger-web.cabal
+++ b/hledger-web/hledger-web.cabal
@@ -196,7 +196,7 @@ library
     , yaml
     , yesod >=1.4 && <1.7
     , yesod-core >=1.4 && <1.7
-    , yesod-form >=1.4 && <1.7
+    , yesod-form >=1.4 && <1.8
     , yesod-static >=1.4 && <1.7
     , yesod-test
   if (flag(dev)) || (flag(library-only))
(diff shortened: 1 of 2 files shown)


juhp:
Could you please do a revision on Hackage or a new release of hledger-web?

This is blocking yesod-form-1.7 for Stackage Nightly: commercialhaskell/stackage#5983

simonmichael:
I have published a revision - let me know of any problem.

juhp:
Thanks Simon!

I am still trying to work out why stackage can't see your revision yet somehow.
Maybe it will work tomorrow if the hackage cache gets updated.

simonmichael:
@juhp, I wonder if https://github.com/commercialhaskell/stackage/blob/master/build-constraints.yaml#L6614 (no-revisions) is biting here ? It was to prevent some badness which I don't remember entirely, related to:

> Revision 1 was good for cabal installers with old GHC, but bad for
> stackage nightly. This should fix the latter. I might continue this
> policy - use revisions to patch up the cabal install experience,
> but avoid them in stackage.

If you want to remove the hledger entries there, I'm ok with it.

juhp:
Thanks, good catch!

Issue #1514, opened 2021-03-24
hledger-lib-1.21 failed to build with  Amount.hs:  Not in scope: ‘M.lookupMin’   error


agryaznov:
Failed to build  hledger-lib-1.21
```
cabal: Entering directory '/tmp/cabal-tmp-7594/hledger-lib-1.21'                                                                                                                              
Configuring hledger-lib-1.21...                                                                                                                                                               
Building hledger-lib-1.21...                                                                                                                                                                  
Preprocessing library hledger-lib-1.21...                                                                                                                                                     
[ 1 of 51] Compiling Text.WideString  ( Text/WideString.hs, dist/build/Text/WideString.o )                                                                                                    
[ 2 of 51] Compiling Text.Tabular.AsciiWide ( Text/Tabular/AsciiWide.hs, dist/build/Text/Tabular/AsciiWide.o )                                                                                
[ 3 of 51] Compiling Text.Megaparsec.Custom ( Text/Megaparsec/Custom.hs, dist/build/Text/Megaparsec/Custom.o )                                                                                
[ 4 of 51] Compiling Paths_hledger_lib ( dist/build/autogen/Paths_hledger_lib.hs, dist/build/Paths_hledger_lib.o )                                                                            
[ 5 of 51] Compiling Hledger.Utils.UTF8IOCompat ( Hledger/Utils/UTF8IOCompat.hs, dist/build/Hledger/Utils/UTF8IOCompat.o )                                                                    
[ 6 of 51] Compiling Hledger.Utils.Tree ( Hledger/Utils/Tree.hs, dist/build/Hledger/Utils/Tree.o )                                                                                            
[ 7 of 51] Compiling Hledger.Utils.Regex ( Hledger/Utils/Regex.hs, dist/build/Hledger/Utils/Regex.o )                                                                                         
[ 8 of 51] Compiling Hledger.Data.Types ( Hledger/Data/Types.hs, dist/build/Hledger/Data/Types.o )                                                                                            
[ 9 of 51] Compiling Hledger.Utils.Parse ( Hledger/Utils/Parse.hs, dist/build/Hledger/Utils/Parse.o )                                                                                         
[10 of 51] Compiling Hledger.Utils.Debug ( Hledger/Utils/Debug.hs, dist/build/Hledger/Utils/Debug.o )                                                                                         
[...]

PR #1543, opened 2021-04-26
cli,ui: Include full binary name and git hash in program version.

Xitian9:
One possible solution to #1539.

### DIFF
--- a/hledger-ui/Hledger/UI/UIOptions.hs
+++ b/hledger-ui/Hledger/UI/UIOptions.hs
@@ -13,15 +13,14 @@ import System.Environment
 import Hledger.Cli hiding (progname,version,prognameandversion)
 import Hledger.UI.Theme (themeNames)
 
-progname, version :: String
+progname, version, prognameandversion :: String
 progname = "hledger-ui"
 #ifdef VERSION
 version = VERSION
 #else
 version = ""
 #endif
-prognameandversion :: String
-prognameandversion = progname ++ " " ++ version :: String
+prognameandversion = versiondescription progname
 
 uiflags = [
   -- flagNone ["debug-ui"] (setboolopt "rules-file") "run with no terminal output, showing console"
--- a/hledger-web/Hledger/Web/WebOptions.hs
+++ b/hledger-web/Hledger/Web/WebOptions.hs
@@ -27,7 +27,7 @@ version = VERSION
 version = ""
 #endif
 prognameandversion :: String
-prognameandversion = progname ++ " " ++ version :: String
+prognameandversion = versiondescription progname
 
 webflags :: [Flag RawOpts]
 webflags =
--- a/hledger/Hledger/Cli/Main.hs
+++ b/hledger/Hledger/Cli/Main.hs
@@ -138,7 +138,6 @@ main = do
     isExternalCommand    = not (null cmd) && cmd `elem` addons -- probably
     isBadCommand         = not (null rawcmd) && null cmd
     hasVersion           = ("--version" `elem`)
-    hasDetailedVersion   = ("--version+" `elem`)
     printUsage           = putStr $ showModeUsage $ mainmode addons
     badCommandError      = error' ("command "++rawcmd++" is not recognized, run with no command to see a list") >> exitFailure  -- PARTIAL:
     hasHelpFlag args     = any (`elem` args) ["-h","--help"]
@@ -170,9 +169,6 @@ main = do
       | not (isExternalCommand || hasHelpFlag args || hasInfoFlag args || hasManFlag args)
         && (hasVersion args) --  || (hasVersion argsaftercmd && isInternalCommand))
                                  = putStrLn prognameandversion
-      | not (isExternalCommand || hasHelpFlag args || hasInfoFlag args || hasManFlag args) 
-        && (hasDetailedVersion argsbeforecmd)  --  || (hasDetailedVersion argsaftercmd && isInternalCommand))
-                                 = putStrLn prognameanddetailedversion
       -- \| (null externalcmd) && "binary-filename" `inRawOpts` rawopts = putStrLn $ binaryfilename progname
       -- \| "--browse-args" `elem` args     = System.Console.CmdArgs.Helper.execute "cmdargs-browser" mainmode' args >>= (putStr . show)
       | isNullCommand            = dbgIO "" "no command, showing commands list" >> printCommandsList addons
--- a/hledger/Hledger/Cli/Version.hs
+++ b/hledger/Hledger/Cli/Version.hs
@@ -1,4 +1,5 @@
-{-# LANGUAGE CPP, TemplateHaskell #-}
+{-# LANGUAGE CPP             #-}
+{-# LANGUAGE TemplateHaskell #-}
 {-
 Version number-related utilities. See also the Makefile.
 -}
@@ -7,26 +8,29 @@ module Hledger.Cli.Version (
   progname,
   version,
   prognameandversion,
-  prognameanddetailedversion,
-  binaryfilename
+  versiondescription,
+  binaryfilename,
 )
 where
-import System.Info (os, arch)
-import Text.Printf
 
+#if MIN_VERSION_githash(0,1,4)
+import GitHash (giDescribe, tGitInfoCwdTry)
+#else
+import GitHash (giHash, tGitInfoCwdTry)
+#endif
+import System.Info (os, arch)
 import Hledger.Utils
 
 
 -- package name and version from the cabal file
-progname, version, prognameandversion, prognameanddetailedversion :: String
+progname, version, prognameandversion :: String
 progname = "hledger"
 #ifdef VERSION
 version = VERSION
 #else
 version = "dev build"
 #endif
-prognameandversion = progname ++ " " ++ version
-prognameanddetailedversion = printf "%s %s" progname version
+prognameandversion = versiondescription progname
 
 -- developer build version strings include PATCHLEVEL (number of
 -- patches since the last tag). If defined, it must be a number.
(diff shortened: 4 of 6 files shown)


Xitian9:
Well I didn't notice it at the time, but it looks like this PR removes the `--version+` flag, which was undocumented and completely equivalent to `--version` otherwise. If you would prefer to maintain this flag, let me know and I can change this, either restoring it as equivalent to `--version`, or having it output this more detailed version information, and maintaining `--version` as the old behaviour.

simonmichael:
More dependence on Template Haskell is always a problem for bleeding edge platforms (web assembly ?), but I think we can't avoid it entirely. Precise version output is something we have needed for along time, so +1 for this. But can you replace gitrev with https://hackage.haskell.org/package/githash ? I think it is the successor package, uploaded 2020 instead of 2017 and used by eg stack.

simonmichael:
Old:

    hledger 1.21.99

New:

    hledger-1.21.99-mac-x86_64 f820f427b8dfdc99c53ddb7f3829411a2f741efd

The new info is all useful but I feel it could be formatted in a more readable way. I guess this is the standard format from gitrev. Maybe it'll be different in githash, and I don't know what would be ideal exactly, but some notes just for the record:

- The hyphenation obfuscates the name of the program/package a little bit.
- The hyphenation logic is unclear - why are all parts but the last hyphenated.
- The platform + architecture appears between the two parts of the version info, separating them.

simonmichael:
How do other tools do it - a few random examples:
```
$ stack --version
Version 2.7.1, Git revision 8afe0c2932716b0441cf4440d6942c59568b6b19 x86_64 hpack-0.34.4

$ git --version
git version 2.30.2

$ curl --version
curl 7.64.1 (x86_64-apple-darwin20.0) libcurl/7.64.1 (SecureTransport) LibreSSL/2.8.3 zlib/1.2.11 nghttp2/1.41.0
Release-Date: 2019-03-27
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS GSS-API HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM NTLM_WB SPNEGO SSL UnixSockets

$ pandoc --version
pandoc 2.11.4
Compiled with pandoc-types 1.22, texmath 0.12.1, skylighting 0.10.2,
citeproc 0.3.0.7, ipynb 0.1.0.1
User data directory: /Users/simon/.local/share/pandoc or /Users/simon/.pandoc
Copyright (C) 2006-2021 John MacFarlane. Web:  https://pandoc.org
This is free software; see the source for copying conditions. There is no
warranty, not even for merchantability or fitness for a particular purpose.

$ brew --version
Homebrew 3.1.9
Homebrew/homebrew-core (git revision a589318ffa; last commit 2021-06-03)
Homebrew/homebrew-cask (git revision 497625d493; last commit 2021-06-02)

$ gh --version
gh version 1.10.3 (2021-05-21)
https://github.com/cli/cli/releases/tag/v1.10.3

$ rg --version
ripgrep 12.1.1
-SIMD -AVX (compiled)
```

Xitian9:
I was just reusing the existing hyphenation logic that was implemented in hledger 12 years ago. I'm happy to change that to something with fewer hyphens.

simonmichael:
Following stack's style, showing the full git hash, not trying to fit in one 80-character line:
```
hledger 1.21.99, git revision f820f427b8dfdc99c53ddb7f3829411a2f741efd, mac-x86_64
```
A little shorter, more jargony, but more likely to fit in one line:
```
hledger 1.21.99, git hash f820f427b8dfdc99c53ddb7f3829411a2f741efd, mac-x86_64
```

There's still room for improvement. If we always show the git hash, it's not clear whether it represents a release build or not. Except for the ".99" hint. I don't know how hard this would be, but would it be ideal to also show a count of commits since last release, like git describe ?
```
$ git describe
1.21-93-gf820f427b
```
Could we ultimately do without the .99 convention ? I suspect not, it'll still be needed to help stack/cabal know when to rebuild.


PS we should do this for hledger-web --version also.

simonmichael:
PS: but porting to githash and implementing for hledger-web are the only merge blockers I'd say.

simonmichael:
PPS gitrev/githash don't spell out exactly how they impact building, will it be more fragile - requiring git ? Not working on windows ? Not a blocker, but something we'll find out.

Xitian9:
It looks like `gitHash` has `git describe` functionality built-in, but only since version 1.4, which is included in stackage versions since 15.7. This unfortunately means it's only supported by our `stack.yaml`, `stack8.8.yaml` and `stack9.0.yaml`. This leaves `stack8.6.yaml` which we still support which doesn't have it.

simonmichael:
I see "WARNING: None of this will work in a git repository without any commits.". Will it break non git builds, eg folks doing "cabal/stack unpack hledger" to do a quick fix ? Maybe we'd want to automatically disable it rather than break the build.

Xitian9:
I think `gitHash` includes two ways of getting this info: one which fails when not in a git environment, and one which returns an either. The second one sounds safer, but I'm not sure if that caveat about the git repository without any commits still holds.

Xitian9:
Would you prefer returning the `git describe` info if we're using a new enough version of `gitHash`, and falling back to something like `hledger 1.21.99, git revision f820f427b8dfdc99c53ddb7f3829411a2f741efd, mac-x86_64` if we're not, or would you rather it be uniform?

simonmichael:
> It looks like gitHash has git describe functionality built-in, but only since version 1.4, which is included in stackage versions since 15.7. This unfortunately means it's only supported by our stack.yaml, stack8.8.yaml and stack9.0.yaml. This leaves stack8.6.yaml which we still support which doesn't have it.
> 

Ok, we can punt on describe for now (or use it only when new enough githash is detected..)

I have forgotten which GHCs we officially support, do you remember ? According to hledger/package.yaml GHC 8.2 is the oldest. I should probably remove older stack*.yaml files..

simonmichael:
> Would you prefer returning the git describe info if we're using a new enough version of gitHash, and falling back to something like hledger 1.21.99, git revision f820f427b8dfdc99c53ddb7f3829411a2f741efd, mac-x86_64 if we're not, or would you rather it be uniform?
> 

Reliable uniform behaviour is good, but in this case I vote it'd be harmless enough to show the extra info when we can.

Xitian9:
Last word I heard on supported stack versions was here: https://github.com/simonmichael/hledger/issues/1514#issuecomment-835461657

simonmichael:
Right! Thanks. So GHC 8.6 and up.

(AFK)

Xitian9:
Changes made. The build failure is because `githash` is not present in the stackage used by `stack8.2.yaml`. We have dropped support for that, but it is still being tested in the CI.

simonmichael:
Thank you! Merged.

Xitian9:
I can confirm that this continues to successfully build outside of a git environment. In particular, nix does not fetch the `.git` directory when building, so I get the fallback version `hledger 1.21.99, linux-x86_64`. Maybe it's worth finding a solution which is more robust to build method, but I'm happy to leave this as is for now.

simonmichael:
I have been updating CI. cli.test 1 and 2 pass locally, but are failing in CI due to this strange version output:
```
:hledger/test/cli/cli.test:1: [Failed]
198
Command:
199
hledger --version
200
Expected stdout: 
201
/^hledger [0-9]/
202
Got stdout:      
203
hledger eeb23c2, linux-x86_64
204

205
:hledger/test/cli/cli.test:2: [Failed]
206
Command:
207
hledger balance --version
208
Expected stdout: 
209
/^hledger [0-9]/
210
Got stdout:      
211
hledger eeb23c2, linux-x86_64
```

Xitian9:
And maybe what version of git the CI runs with. This might be due to different `git describe` output on different platforms. In particular, one possibility is that CI uses special annotated tags internally, which might explain output like this.

Xitian9:
Hmm, I think I see the issue. The output of `giDescribe` is actually not the content of `git describe`, but of `git describe --always`, where `--always:  shows uniquely abbreviated commit object as fallback`. Since it's reporting the git hash, it must be using this fallback. The documentation doesn't make it explicit in which cases it uses the fallback, but I'm guessing that the CI somehow does not fetch the full tag history. And so, being unable to locate an appropriate ancestor tag to prefix, it uses the fallback.

I'm not sure what the appropriate resolution here is, if this is indeed the problem. The simplest is probably just to drop those two tests. More complicated would be to make sure CI fetches all the information it needs so `git describe` doesn't use the fallback.

Xitian9:
Perhaps relevant: https://github.com/empirical-soft/empirical-lang/issues/105

simonmichael:
Ha ha, we considered a git checkout and a non git checkout, but not "unusual" git checkouts. Thanks, I'll fix our CI checkout or something.

simonmichael:
Fixed by cloning the full repo history and tags on each CI run (instead of just the latest commit). That is a pretty slow operation on my laptop, but on a github runner seems to increase checkout time only from 1s to 7s.

simonmichael:
I have removed the *1.21.99 tags from master - if you have these locally, please delete them, I think they are spurious.

simonmichael:
More notes for possible future work:

git describe by default does not show a hash if you are at a tag. So if I tag 1.22 and immediately build it, --version looks like `hledger 1.22, mac-x86_64`. git describe's --long flag helps ensure the hash is shown, though the 0 patchlevel is ugly: `1.22-0-g03fe5df6b`. I don't know if githash supports --long yet. Another desirable git describe flag might be --dirty, for the truly paranoid.

simonmichael:
https://github.com/snoyberg/githash/pull/21

Issue #1177, opened 2020-01-16
stricter/more correct handling of unbalanced multicommodity transactions


psb777:
It is suggested in the docs at https://hledger.org/1.14/journal.html under the "Transaction Prices" section that one can record an FX transaction thusly in the journal:

```
2009/1/1
  assets:euros     €100          ; one hundred euros purchased
  assets:dollars  $-135          ; for $135
```

and this works. But it ought not to. The balance sheet no longer balances because neither commodity posting has a contra. hledger should never allow this as the basic double entry check that one's accounts are in order is that the balance sheet sums to zero in all currencies (and securities and commodities.) Anyone doing FX in this way loses this most basic of accountancy checks, that every transaction balances in every currency.

The correct way to post the above is something like the following:

```
2009/1/1
  assets:euros     €100          ; one hundred euros purchased
  assets:dollars  $-135          ; for $135
  control:fx
```

Then the balancing postings automatically get done for you by hledger, and the balance sheet continues to balance. 

If one believes that the invention of double entry bookkeeping was a great thing then hledger should prevent an imbalance being created in the balance sheet for FX transations just as it does for all other transactions.

Possibly helpful details: 

- a minimal reproducible example: see above
- related documentation: referenced above
- my hledger version: 1.14
- installation: apt-get 
- platform: devuan (i.e. vanilla debian but without systemd)

simonmichael:
Thanks for the report and clear advice. We also have #758 open for this. I would be fine with this personally, but it would be a breaking change in very long-established behaviour so I think a clear consensus and favourable community discussion of this are required.

psb777:
I know see this has been reported long ago. It is the same issue as #115 and #86. I very much like the solution proposed in #115 and would like to re-emphasise that the basic rules of double entry book-keeping should be enforced by any good accounting system. The alternative of memorandum style (single entry) book-keeping is available to "accountants" but note that it is not allowed by hledger on transactions with postings only in one currency. Every transaction must balance in every commodity, and if this rule is not followed then the balance sheet will not sum to zero, the most standard basic check that one's books are in order. Repeating myself, I know.

edkedk99:
@psb777 , may I disagree with your suggestion. I consider they way hledger works now in fact balance and respect the double entry principle.

It is better said that in your example, the transaction doesn't **explicit** balance, but if you consider it is implicit that one EUR equal 1.35 USD, it does indeed balance, at least in the default currency.

[...]

Issue #115, opened 2013-05-30
With --require-balanced=true, Hledger should not allow multi-currency(or commodity) transactions, where the units don't zero out.


tphyahoo:
Hledger allows transactions such as 

2009/1/1
    assets:cows  COW1   ; one cow at 100USD
    assets:usd     USD-100

where the units don't zero out, which violates the accounting equation.

Suggested fix:

The above transaction should be rejected by hledger, and instead users would be directed to enter multi-currency/commodity transactions as something like

2009/1/1
    assets:cow_account      COW1   ; one cow at 100USD
    equity:cow_account      -COW1
    assets:bank_account     USD-100
    equity:bank_account     USD+100

so that the units do zero out.

If it's important for backwards compatibility reasons that hledger allows transactions where the units don't zero out, perhaps there should be a flag that enables this mode.
  hledger --require-balanced-transactions would enable this mode, with the more permissive mode allowed by default
  or 
  hledger --allow-unbalanced-transactions would enable the current behavior, with the stricter behavior turned on by default.

My preference would be to do the second option, and have the stricter behavior be the default.

lestephane:
> This is how I am currently using hledger for multi currency accounting

I know this is an old post, but can you share how you're doing today (hopefully with hledger, hopefully not manual)? I'm at a loss as to how to implement the 4.2 logic myself. I have some automation capability as I'm using hledger-flow, and a bunch of awk scripts, but i could use some pointers from someone who's already gone through this. cheers

Issue #86, opened 2013-04-08
Auto-balancing between multiple currencies can hide errors, should be warned/preventable


simonmichael:
_Original author: rocon...@r6.ca (May 08, 2012 16:09:46)_

<b>What steps will reproduce the problem?</b>
1. create a file called /tmp/hledger.test with the following data:

2012/01/01 Initial
    accounts:bank           100 EUR
    equity:initial

2012/04/21 Test
    expenses:test1        1.78 EUR
    expenses:test2      $44.00
    accounts:bank       -45.78 EUR
1. execute hledger bal -f /tmp/hledger.test

<b>What is the expected output? What do you see instead?</b>

I expect there to be an error because the Test transaction does not balance to 0.  Instead I made an error where I entered $44 dollars instead of 44 EUR.

Instead the command succeeds and I'm oblivious to error in the entry. 

What version of the product are you using? 

hledger-0.16.1

On what operating system?

linux

<b>Please provide any additional information below.</b>

Some people on IRC were arguing that there is nothing that can be done to catch this error.  However I disagree.  Any time that sum of all the currencies of a transaction do not sum to 0 the entry is in error.  The fact that all the ledger entries sum to 0 is the fundamental property of double entry bookkeeping that makes it so effective at catching errors.  Removing this property in any situation makes the ledger program useless for catching mistakes.  For currency conversions a &quot;trading&quot; account should be used as described in: http://www.mscs.dal.ca/~selinger/accounting/tutorial.html

_Original issue: http://code.google.com/p/hledger/issues/detail?id=86_
## 

----
Related: #115, #758, #1177

simonmichael:
Also related: https://www.mathstat.dal.ca/~selinger/accounting/tutorial.html#4.2

dstrelnikov:
I do agree that the transactions of this form
```
2020/01/01 Purchase in USD
    Expenses:Food          50 USD
    Assers:Bank:Checking  -42 EUR
```
should raise at least a warning.

A more neat way to record this transaction could be
```
2020/01/01 Purchase in USD
    Expenses:Food          50 USD @@ 42 EUR
    Assers:Bank:Checking
```

However, I realize that the solution isn't easy. For example someone might want to record it in the following way
```
2020/01/01 Purchase in USD
    Expenses:Food          50 USD
    Assers:Bank:Checking  -42 EUR
    Equity:Trading
```

A more neat way would be
```
2020/01/01 Purchase in USD
    Expenses:Food          50 USD
[...]

PR #1544, opened 2021-04-26
Add option to disable automatic price inferral in unbalanced multicommodity transactions.

Xitian9:
This can be used on its own with the --no-infer-prices option, but is also enabled by --strict.

This at least partially addresses #86, #115, and #1177.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -722,7 +722,7 @@ journalModifyTransactions d j =
 -- | Check any balance assertions in the journal and return an error message
 -- if any of them fail (or if the transaction balancing they require fails).
 journalCheckBalanceAssertions :: Journal -> Maybe String
-journalCheckBalanceAssertions = either Just (const Nothing) . journalBalanceTransactions True
+journalCheckBalanceAssertions = either Just (const Nothing) . journalBalanceTransactions def
 
 -- "Transaction balancing", including: inferring missing amounts,
 -- applying balance assignments, checking transaction balancedness,
@@ -822,13 +822,14 @@ updateTransactionB t = withRunningBalance $ \BalancingState{bsTransactions}  ->
 --
 -- This does multiple things at once because amount inferring, balance
 -- assignments, balance assertions and posting dates are interdependent.
-journalBalanceTransactions :: Bool -> Journal -> Either String Journal
-journalBalanceTransactions assrt j' =
+journalBalanceTransactions :: BalancingOpts -> Journal -> Either String Journal
+journalBalanceTransactions bopts' j' =
   let
     -- ensure transactions are numbered, so we can store them by number
     j@Journal{jtxns=ts} = journalNumberTransactions j'
     -- display precisions used in balanced checking
     styles = Just $ journalCommodityStyles j
+    bopts = bopts'{commodity_styles_=styles}
     -- balance assignments will not be allowed on these
     txnmodifieraccts = S.fromList $ map paccount $ concatMap tmpostingrules $ jtxnmodifiers j
   in
@@ -845,7 +846,7 @@ journalBalanceTransactions assrt j' =
         -- and leaving the others for later. The balanced ones are split into their postings.
         -- The postings and not-yet-balanced transactions remain in the same relative order.
         psandts :: [Either Posting Transaction] <- fmap concat $ forM ts $ \case
-          t | null $ assignmentPostings t -> case balanceTransaction styles t of
+          t | null $ assignmentPostings t -> case balanceTransaction bopts t of
               Left  e  -> throwError e
               Right t' -> do
                 lift $ writeArray balancedtxns (tindex t') t'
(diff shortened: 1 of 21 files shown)


simonmichael:
I think it would be more consistent to name this check and let the check command run it individually, like the other https://hledger.org/hledger.html#strict-checks. Then the new flag wouldn't be needed.

`pricebalanced`/`fullybalanced` were some ideas from https://hledger.org/checking-for-errors.html#todo--maybe.

Xitian9:
That sounds eminently reasonable to me.

Xitian9:
I have removed the `--no-infer-prices` flag. You can now check that all transactions are balanced without auto-inferring prices either with the `--strict` flag, or by calling `hledger check pricebalanced`.

simonmichael:
Nice. Here's another test journal, showing the three kinds of "balanced" with the names I was thinking of:
```
; test https://github.com/simonmichael/hledger/pull/1544's strict mode

account a
account b
account e

commodity A
commodity B

; fullybalanced
2021-01-01
  a   1A
  e  -1A
  e   1B
  b  -1B

; pricebalanced
2021-01-02
  a   1A @@ 1B
  b  -1B

; autobalanced
2021-01-03
  a   1A
  b  -1B
```

I'm not super happy with these names, they are a bit obscure and potentially confusing. Would more verbose names be clearer ? Eg:
```
hledger check balancedwithinferredrates  # laxest, current default
hledger check balancedwithexplicitrates
hledger check balanced  # strictest
```
I'm trying "rates" here (think "conversion rate" or "exchange rate") as an alternative to "transaction prices"; the word "price" is currently a bit overloaded (transaction price vs market price).

I'm not sure if all three kinds of balanced checks will be needed in practice, but perhaps it's worth naming them all at least.

There was some talk of always automatically converting autobalanced/pricebalanced entries into the fullybalanced form, making them just syntactic sugar. If that happens would we still need these checks ? I suppose so.

Xitian9:
In fact, #1554 is a PR which proposes automatically converting to fully balanced form, so this is not just a theoretical concern. I think we would still need the `pricebalanced`/`balancewithexplicitrates` check; even if we can convert example 3 to fully-balanced form, we might want to stop it from doing so to avoid various errors, just as with the current situation. I would say this is sufficient:
```
hledger check balancedwithinferredrates  # laxest, current default
hledger check balanced  # strictest, requires explicit prices
```

simonmichael:
Of these two:

- autobalanced, pricebalanced, fullybalanced
- balancedwithinferredrates, balancedwithexplicitrates, balanced ?

should we use the second ? It would mean renaming `autobalanced` to `balancedwithinferredrates` in docs, and adding the `balanced` check.

simonmichael:
Another possibility: leave `autobalanced` as is, and add `balanced`.

simonmichael:
Also: I originally meant the third check (`fullybalanced`/`balanced`) to pass only if there are no inferred rates and no explicit (@/@@) rates. Does the new check you have added allow explicit rates ? (I can't quite see it..)

simonmichael:
Still brainstorming names: here are even more hypothetical balancing checks, in increasing order of strictness:

- autobalanced 
- balancednoinferredrates
- balancednorates
- balancednomissingamounts
- balancednoassignments
- balanced

Xitian9:
Yes, the new check allows explicit rates; it just turns off generating inferred rates. In your examples (included here), the first two would pass but the third would not.

```
; test https://github.com/simonmichael/hledger/pull/1544's strict mode

account a
account b
account equity:conversion

commodity A
commodity B

; fullybalanced
2021-01-01
  a   1A
  equity:conversion  -1A
  equity:conversion   1B
  b  -1B

; pricebalanced
2021-01-02
  a   1A @@ 1B
  b  -1B

; autobalanced
2021-01-03
  a   1A
  b  -1B
```

It may or may not be necessary to add a third check for which only the first example will pass. #1554, if merged, would change behaviour so that example #2 is converted to (something like) example #1, meaning that the two become (mostly) indistinguishable. If we were planning to merge that, it probably wouldn't be worth going through the effort of introducing a new check.

simonmichael:
Actually inferred amounts and assigned amounts complicate it - better for this set of checks to deal just with commodity conversions. What about:

1. balancedwithautoconversion
2. balancednoautoconversion
3. balanced   # (ie, balancednoconversion)

I think I like those, and right now I'd be fine with exposing 1 and 2, 1 and 3, or 1, 2 and 3.

simonmichael:
Would it be easier to postpone this till 1554 lands ?

Xitian9:
That sounds good to me. I'm happy for exposing just 1 and 2 right now. Potentially 3 can be something separate which is implemented later, or just a synonym for 2.

I'm happy to postpone this if you'd like, but I think we could also just proceed. I think the only thing here that actually depends on #1554 is whether check 3 is worth implementing, and the biggest effect I can see it would have is whether to rename 2 to ‘balanced’.

simonmichael:
Great, let's go ahead exposing 1 and 2 then, and we'll possibly add 3 or rename things if/when the time comes.

Xitian9:
Alright, I've renamed those two checks.

simonmichael:
Thank you!

PR #1545, opened 2021-04-28
Introduce lenses for ReportOpts, ReportSpec, CliOpts, etc.

Xitian9:
With so many nested options, it would greatly improve developer ergonomics to have some lenses in there.

I've put together a few commits to this end, but I don't think the API is ideal yet. Since the options all have names ending in an underscore, I've generated the lenses by dropping the underscore from the end. The main problem with this is that ReportOpts has `color_`, `drop_`, `empty_`, and `transpose_`, for which dropping the underscore would result in name clashes. My proposal would be to give these slightly longer names to avoid the clash, for example `color_ -> showcolor_`, `drop_ -> dropaccount_` , `empty_ -> showempty_`, `transpose_ -> transposetable_`, or something like that. We might also want to rename `watch_` and `change_` in UIOpts.

I would be interested in hearing opinions on this.

### DIFF
--- a/bin/hledger-balance-as-budget.hs
+++ b/bin/hledger-balance-as-budget.hs
@@ -43,4 +43,4 @@ main = do
       opts@CliOpts{reportspec_=rspec} <- getHledgerCliOpts' cmdmode args
       d <- getCurrentDay
       (report,j) <- withJournalDo opts $ \j -> return (multiBalanceReport rspec j, j)
-      return (rsOpts rspec,j,report)
+      return (reportopts_ rspec,j,report)
--- a/bin/hledger-combine-balances.hs
+++ b/bin/hledger-combine-balances.hs
@@ -65,7 +65,7 @@ main = do
   (_,report1) <- mbReport report1args
   (rspec2,report2) <- mbReport report2args
   let merged = appendReports report1 report2
-  TL.putStrLn $ multiBalanceReportAsText (rsOpts rspec2) merged
+  TL.putStrLn $ multiBalanceReportAsText (reportopts_ rspec2) merged
   where
     mbReport args = do
       opts@CliOpts{reportspec_=rspec} <- getHledgerCliOpts' cmdmode args
--- a/bin/hledger-smooth.hs
+++ b/bin/hledger-smooth.hs
@@ -57,21 +57,21 @@ _FLAGS
 main :: IO ()
 main = do
   copts@CliOpts{reportspec_=rspec, rawopts_} <- getHledgerCliOpts cmdmode
-  let ropts = rsOpts rspec
+  let ropts = reportopts_ rspec
       copts' = copts{
         -- One of our postings will probably have a missing amount; this ensures it's
         -- explicit on all the others.
         rawopts_ = setboolopt "explicit" rawopts_
         -- Don't let our ACCT argument be interpreted as a query by print
-        ,reportspec_ = rspec{rsOpts=ropts{querystring_=[]}}
+        ,reportspec_ = rspec{reportopts_=ropts{querystring_=[]}}
         }
   withJournalDo copts' $ \j -> do
     today <- getCurrentDay
     let
       menddate = reportPeriodLastDay rspec
-      q = rsQuery rspec
+      q = query_ rspec
       acct = headDef (error' "Please provide an account name argument") $ querystring_ ropts
-      pr = postingsReport rspec{rsQuery = And [Acct $ accountNameToAccountRegexCI acct, q]} j
+      pr = postingsReport rspec{query_ = And [Acct $ accountNameToAccountRegexCI acct, q]} j
 
       -- dates of postings to acct (in report)
       pdates = map (postingDate . fourth5) pr
--- a/bin/hledger-swap-dates.hs
+++ b/bin/hledger-swap-dates.hs
@@ -33,7 +33,7 @@ main = do
    \j -> do
     d <- getCurrentDay
     let
-      q = rsQuery rspec
+      q = query_ rspec
       ts = filter (q `matchesTransaction`) $ jtxns $ journalApplyValuationFromOpts rspec j
       ts' = map transactionSwapDates ts
     mapM_ (T.putStrLn . showTransaction) ts'
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -881,7 +881,7 @@ balanceTransactionAndCheckAssertionsB (Right t@Transaction{tpostings=ps}) = do
   ps' <- mapM (addOrAssignAmountAndCheckAssertionB . postingStripPrices) ps
   -- infer any remaining missing amounts, and make sure the transaction is now fully balanced
   styles <- R.reader bsStyles
-  case balanceTransactionHelper balancingOpts{commodity_styles_=styles} t{tpostings=ps'} of
+  case balanceTransactionHelper defbalancingopts{commodity_styles_=styles} t{tpostings=ps'} of
     Left err -> throwError err
     Right (t', inferredacctsandamts) -> do
       -- for each amount just inferred, update the running balance
--- a/hledger-lib/Hledger/Data/RawOptions.hs
+++ b/hledger-lib/Hledger/Data/RawOptions.hs
@@ -24,6 +24,7 @@ module Hledger.Data.RawOptions (
   maybeintopt,
   maybeposintopt,
   maybecharopt,
+  defrawopts,
   overRawOpts
 )
 where
@@ -39,7 +40,11 @@ import Hledger.Utils
 newtype RawOpts = RawOpts { unRawOpts :: [(String,String)] }
   deriving (Show)
 
-instance Default RawOpts where def = RawOpts []
+instance Default RawOpts where def = defrawopts
+
+-- | Empty RawOpts.
+defrawopts :: RawOpts
+defrawopts = RawOpts []
 
 overRawOpts f = RawOpts . f . unRawOpts
 
--- a/hledger-lib/Hledger/Data/Transaction.hs
+++ b/hledger-lib/Hledger/Data/Transaction.hs
@@ -9,11 +9,12 @@ tags.
 
 {-# LANGUAGE FlexibleContexts  #-}
 {-# LANGUAGE LambdaCase        #-}
+{-# LANGUAGE NamedFieldPuns    #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE Rank2Types        #-}
 {-# LANGUAGE RecordWildCards   #-}
+{-# LANGUAGE TemplateHaskell   #-}
 
-{-# LANGUAGE NamedFieldPuns #-}
 module Hledger.Data.Transaction (
   -- * Transaction
   nulltransaction,
@@ -30,7 +31,8 @@ module Hledger.Data.Transaction (
   balancedVirtualPostings,
   transactionsPostings,
   BalancingOpts(..),
-  balancingOpts,
+  HasBalancingOpts(..),
+  defbalancingopts,
   isTransactionBalanced,
   balanceTransaction,
   balanceTransactionHelper,
@@ -77,6 +79,7 @@ import qualified Data.Map as M
 import Safe (maximumDef)
 
 import Hledger.Utils
+import Hledger.Utils.TH (makeClassyLensesTrailing)
 import Hledger.Data.Types
 import Hledger.Data.Dates
 import Hledger.Data.Posting
@@ -360,10 +363,10 @@ data BalancingOpts = BalancingOpts
   , commodity_styles_  :: Maybe (M.Map CommoditySymbol AmountStyle)  -- ^ commodity display styles
   } deriving (Show)
 
-instance Default BalancingOpts where def = balancingOpts
+instance Default BalancingOpts where def = defbalancingopts
 
-balancingOpts :: BalancingOpts
-balancingOpts = BalancingOpts
+defbalancingopts :: BalancingOpts
+defbalancingopts = BalancingOpts
   { ignore_assertions_ = False
   , infer_prices_      = True
   , commodity_styles_  = Nothing
@@ -661,6 +664,10 @@ transactionFile Transaction{tsourcepos} =
   case tsourcepos of
     GenericSourcePos f _ _ -> f
     JournalSourcePos f _   -> f
+
+-- lenses
+
+makeClassyLensesTrailing ''BalancingOpts
 
 -- tests
 
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -199,7 +199,7 @@ data InputOpts = InputOpts {
                                                 --   by a filename prefix. Nothing means try all.
     ,mrules_file_       :: Maybe FilePath       -- ^ a conversion rules file to use (when reading CSV)
     ,aliases_           :: [String]             -- ^ account name aliases to apply
-    ,anon_              :: Bool                 -- ^ do light anonymisation/obfuscation of the data
+    ,anonymise_         :: Bool                 -- ^ do light anonymisation/obfuscation of the data
     ,new_               :: Bool                 -- ^ read only new transactions since this file was last read
     ,new_save_          :: Bool                 -- ^ save latest new transactions state for next time
     ,pivot_             :: String               -- ^ use the given field's value as the account name
@@ -215,7 +215,7 @@ definputopts = InputOpts
     { mformat_           = Nothing
     , mrules_file_       = Nothing
     , aliases_           = []
-    , anon_              = False
+    , anonymise_         = False
     , new_               = False
     , new_save_          = True
     , pivot_             = ""
@@ -230,7 +230,7 @@ rawOptsToInputOpts rawopts = InputOpts{
      mformat_           = Nothing
     ,mrules_file_       = maybestringopt "rules-file" rawopts
     ,aliases_           = listofstringopt "alias" rawopts
-    ,anon_              = boolopt "anon" rawopts
+    ,anonymise_         = boolopt "anon" rawopts
     ,new_               = boolopt "new" rawopts
     ,new_save_          = True
     ,pivot_             = stringopt "pivot" rawopts
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -23,13 +23,16 @@ Some of these might belong in Hledger.Read.JournalReader or Hledger.Read.
 {-# LANGUAGE Rank2Types          #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskell     #-}
 {-# LANGUAGE TupleSections       #-}
 {-# LANGUAGE TypeFamilies        #-}
 
 --- ** exports
 module Hledger.Read.Common (
-  Reader (..),
-  InputOpts (..),
+  Reader(..),
+  InputOpts(..),
+  HasInputOpts(inputOpts, mformat, mrules_file, aliases, anonymise,
+               new, new_save, pivot, auto, strict),
   definputopts,
   rawOptsToInputOpts,
 
@@ -153,6 +156,8 @@ import Text.Megaparsec.Custom
   (FinalParseError, attachSource, customErrorBundlePretty,
   finalErrorBundlePretty, parseErrorAt, parseErrorAtRegion)
 
+import Hledger.Utils.TH (makeClassyLensesTrailing)
+
 import Hledger.Data
 import Hledger.Utils
 import Text.Printf (printf)
@@ -207,6 +212,10 @@ data InputOpts = InputOpts {
     ,balancingopts_     :: BalancingOpts        -- ^ options for balancing transactions
     ,strict_            :: Bool                 -- ^ do extra error checking (eg, all posted accounts are declared, no prices are inferred)
  } deriving (Show)
+
+makeClassyLensesTrailing ''InputOpts
+
+instance HasBalancingOpts InputOpts where balancingOpts = balancingopts
 
 instance Default InputOpts where def = definputopts
 
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -74,6 +74,7 @@ import Text.Megaparsec hiding (match, parse)
 import Text.Megaparsec.Char (char, newline, string)
 import Text.Megaparsec.Custom (customErrorBundlePretty, parseErrorAt)
 import Text.Printf (printf)
+import Lens.Micro (set)
 
 import Hledger.Data
 import Hledger.Utils
@@ -116,7 +117,7 @@ parse iopts f t = do
               -- apply any command line account aliases. Can fail with a bad replacement pattern.
               in case journalApplyAliases (aliasesFromOpts iopts) pj' of
                   Left e -> throwError e
-                  Right pj'' -> journalFinalise iopts{balancingopts_=(balancingopts_ iopts){ignore_assertions_=True}} f t pj''
+                  Right pj'' -> journalFinalise (set ignore_assertions True iopts) f t pj''
 
 --- ** reading rules files
 --- *** rules utilities
--- a/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
+++ b/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
@@ -79,7 +79,7 @@ type AccountTransactionsReportItem =
   )
 
 accountTransactionsReport :: ReportSpec -> Journal -> Query -> Query -> AccountTransactionsReport
-accountTransactionsReport rspec@ReportSpec{rsOpts=ropts} j reportq thisacctq = items
+accountTransactionsReport rspec@ReportSpec{reportopts_=ropts} j reportq thisacctq = items
   where
     -- a depth limit should not affect the account transactions report
     -- seems unnecessary for some reason XXX
--- a/hledger-lib/Hledger/Reports/BalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/BalanceReport.hs
@@ -78,7 +78,7 @@ balanceReport rspec j = (rows, total)
 -- tests
 
 Right samplejournal2 =
-  journalBalanceTransactions balancingOpts
+  journalBalanceTransactions defbalancingopts
     nulljournal{
       jtxns = [
         txnTieKnot Transaction{
--- a/hledger-lib/Hledger/Reports/BalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/BalanceReport.hs
@@ -104,7 +104,7 @@ tests_BalanceReport = tests "BalanceReport" [
 
   let
     (rspec,journal) `gives` r = do
-      let opts' = rspec{rsQuery=And [queryFromFlags $ rsOpts rspec, rsQuery rspec]}
+      let opts' = rspec{query_=And [queryFromFlags $ reportopts_ rspec, query_ rspec]}
           (eitems, etotal) = r
           (aitems, atotal) = balanceReport opts' journal
           showw (acct,acct',indent,amt) = (acct, acct', indent, showMixedAmountDebug amt)
@@ -130,7 +130,7 @@ tests_BalanceReport = tests "BalanceReport" [
        mixedAmount (usd 0))
 
     ,test "with --tree" $
-     (defreportspec{rsOpts=defreportopts{accountlistmode_=ALTree}}, samplejournal) `gives`
+     (defreportspec{reportopts_=defreportopts{accountlistmode_=ALTree}}, samplejournal) `gives`
       ([
         ("assets","assets",0, mamountp' "$0.00")
        ,("assets:bank","bank",1, mamountp' "$2.00")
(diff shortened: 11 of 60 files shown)


Xitian9:
@simonmichael Do you have any opinions on the renaming of these option records? Do you think it's likely to cause much breakage?

simonmichael:
Xitan9, sorry. This is one I have been putting off because I have mixed feelings and it will take some work to have a useful opinion. This has been something to figure out for years and it's great that you have made a start. Why mixed  feelings ? I think some use of lenses will be great for experienced developers, I worry that they will obfuscate the code further for new contributors and for future porters/maintainers; it seems like a hard balance, such that we'll have to choose one or the other. Then simply the adoption process seems a big task - lots of churn and potential for confusion with some of the code lensified and some not, unless we figure out a clear policy and plan. Also which lens package - lens, microlens, optics..
I will try to review and give more feedback here soon.

Xitian9:
Not a problem. I've used microlens here since that was already in use in the code base – somebody had put it in hledger-ui at some point. I've done minimal refactoring to use the new lenses here, but I'm happy to remove those places I have put it in if you think they add more in indirection than they save in clarity. I think the most obvious place where they clear things up is in Hledger.UI.UIState, where the messes of nested pattern matching and record updating can be significantly simplified.

simonmichael:
Great. Yes indeed, hledger-ui is probably the most urgent need. I guess we already use microlens somewhere. I expect all will be well.

simonmichael:
I did a first pass on this.

Likely to cause much breakage - do you mean, in other users of hledger-lib/hledger libs, like hledger-iadd, and custom scripts ? Not sure but I'm guessing they will all break.

I see you add Data.Default in a few places, making the code more concise. I used to use this more but it seems to be out of favour in the community, and I think I agree. It breaks code navigation in IDEs, and makes the code a little harder to understand for the reader. I think perhaps we should avoid it.

Lenses. It feels like they both giveth and taketh away, especially for non-experts. Code can be much more concise, and (with luck) also easier to understand superficially, but at the same time it becomes much harder to understand completely. Generating them with TH also breaks IDE navigation, and reputedly makes compilation slower (I don't know if that's still true or true in all cases). I guess we could avoid both of those issues by defining them explicitly, at the cost of huge boilerplate. Perhaps worth a try to see how it looks though.

The typeclasses I guess are an optional extra, making it possible to use the lenses even more concisely. They too seem to make coding easier and full understanding harder, probably worth it though.

Perhaps we can discuss this more and run through the changes in chat next ? I'll look out for you, or feel free to suggest a time or two and we could schedule it.

Xitian9:
Re our discussions, I have made the following changes.
1. I have reduced usage of `def`
2. I have renamed the fields of `ReportOpts` mostly as suggested above (exception was that I changed `drop_` to `droplevels_` instead of `dropaccount_`
3. I have renamed `watch_` from `UIOpts` to `watchfiles_`
4. I have refactored the code so that `change_` is not needed at all, and removed it from `UIOpts`.
5. I have written out the classy lenses for `ReportSpec` manually in the last commit.

Writing the lenses out manually is not so bad, and is very doable for all classes except possibly `ReportOpts`. `ReportOpts` will be a pain, but not a disaster. Thoughts?

Xitian9:
Some more comments:
- `UIOpts` is now pretty anæmic; it's just a `CliOpts` and a `watchfiles_ :: Bool`, which is used in exactly one place (in Hledger.UI.Main). This could easily be pulled out and we could do away with `UIOpts` altogether, doing away with one layer of complication in the options tower.
- If we're defining the lenses manually, we can do some more custom things. In particular, instead of having `reportopts :: Lens' ReportSpec ReportOpts`, we could try somthing like `reportopts :: Lens ReportSpec (Either String ReportSpec) ReportOpts ReportOpts`, defined using some generalisation of `updateReportSpecWith`. This would mean that lenses would automatically update the `ReportSpec` query, removing a common source of bugs. I would have to do some investigations to see what this would mean for lens composibility, but it's tempting.

Xitian9:
Alternately, instead of eliminating UIOpts, we could make it bigger, and pull in all the ui options pulled from rawopts in an ad-hoc way. These include `theme`, `register`, and `status-toggles`.

Xitian9:
Regarding my second point above, it looks like this can be done with the following definition:
```
class HasReportSpec a where
    reportSpec :: Lens' a ReportSpec

    reportopts :: Lens a (Either String a) ReportOpts ReportOpts
    reportopts f = getCompose . reportSpec (Compose . reportopts f)
    …

instance HasReportSpec ReportSpec where
    reportSpec = id
    reportopts f rspec = reportOptsToSpec (reportday_ rspec) <$> f (reportopts_ rspec)
    …
```
The price we pay is that we can no longer have a useful typeclass `HasReportOpts`, and will instead have to manually compose `reportopts . average` instead of just calling `average`, but this seems like a reasonable trade-off.

Xitian9:
Actually, we could have our cake and eat it too through a more careful choice of the lens definitions. We could have normal ReportOpts lenses for anything which cannot possibly result in an error, and these fancier lenses only for `querystring_` and the bare `reportopts_`. Let me put some thought into the best design for that.

simonmichael:
I haven't yet reviewed the latest, but a quick comment: in master, let's move in small steps and be conservative, and please don't add more complexity to this PR since I have a small brain. :)

Xitian9:
Sure thing. Do you mean I should undo the refactoring to remove `change_`, or do you mean I should not implement any of the refactorings or fancier lens options I proposed in my above comments?

simonmichael:
No need to undo anything (unless you want to); I'll look at all of it when I can. Perhaps I shouldn't have tackled this PR next, as bugfixes, release, regression testing are probably more urgent. We can go two ways, try to limit the scope of this PR with hopes of merging it soon, or leave it open for longer and keep exploring lensification ideas.

Xitian9:
If you'd like, #1576 and #1564 are relatively simple bugfix PRs, and #1578 is a relatively simple regression fix (a major regression in the transaction screen of hledger-ui).

Xitian9:
I think it's worth getting the lenses right, and if we can further improve ergonomics by getting rid of the need to call ‘updateReportSpec`, that's a big gain. With that in mind holding off on merging this until we finish exploration is reasonable. Maybe you can take a quick peek at 2923b3e; if it doesn't look like too much boilerplate to you, I can go ahead and replace the TH with handcrafted lens definitions and incorporate it into this PR.

simonmichael:
I think we could accept that boilerplate if it provides the expected benefits (preserving IDE code navigation and compilation speed). But  someone in #haskell-language-server told me that they can navigate to definition even with TH-generated lenses. More research needed.

Xitian9:
Whether or not in provides those benefits, writing at least some of the definitions by hand would be necessary to implement lenses like what I propose in https://github.com/simonmichael/hledger/pull/1545#issuecomment-866865420.

Xitian9:
There are a few things done in this PR that are independent of the final shape of the lenses:
1. Renaming the fields of ReportSpec
2. Renaming some fields of ReportOpts and InputOpts
3. Refactoring to eliminate UIOpts

I think (3) is simple and quite valuable independent of this PR. (1) and (2) are good ideas, but their motivation presuposes that we'll be merging this PR eventually. I could break these out into a separate PR to simplify this, if you think that's valuable, or rebase to move those commits to the beginning of this PR.

Xitian9:
I have reworked this PR to make each commit more atomic, with a clearer flow between the commits. To avoid causing confusion I won't push until you give the okay.

Xitian9:
Relevant discussion: https://stackoverflow.com/questions/68116436/what-is-the-appropriate-abstraction-for-a-lens-which-can-fail-as-a-setter

Xitian9:
I've resolved a bunch of conflicts and cleaned up the commit history. Will it cause confusion if I force push?

simonmichael:
No, please go for it.

Xitian9:
Here's a summary of my investigations into fancy lenses that get around the need to call `updateReportSpec`. I had hoped that it would be as simple as defining a lens `querystring` for which `view querystring rspec` would give the correct element, and for which `set querystring rspec` would update the query string, then update the query, and return `Either String ReportSpec`. The overall thrust is that while it won't be quite as simple or free as I had hoped, we can get most of the way there.

There are three different ways that we can update `ReportOpts`, which need to have certain effects on `ReportSpec`:
1. Most options can be updated without requiring any changes to `ReportSpec`.
2. Some options – namely `real_`, `depth_`, `date2_`, `period_`, and `statuses_` – require updating the query in `ReportSpec`, but cannot possibly fail to do so. These can be handled as well as I had hoped above.
3. Updating `querystring_` or doing a fully general update of `reportopts_` requires updating the query in `ReportSpec`, and can fail to compile the regular expressions. This can be handled as below with some fancy lens work, courtesy of [Carl at stackoverflow](https://stackoverflow.com/a/68157315/7694443).

The result is that we can define a lens `querystring` for which `view querystring rspec` gives the correct element, `set querystring "Blah" rspec` will update the ReportSpec if possible, and return some default ReportSpec if not possible (which could be either the ReportSpec you started with, or something like `error "You forgot to call setWithReport you twit."`). We could then define some new functions `setWithReport` and `overWithReport` which will update the ReportSpec and return an `Either String ReportSpec`.

So in the end we still need to remember to call a special function when updating (`setWithReport` instead of `updateReportSpec`), but the consequences of forgetting to do so are less. Instead of the query not being updated silently, it would update the query if possible, and if not possible would complain at runtime (or silently not update, if we wanted that for some reason).

Xitian9:
My inclination is that this is a worthwhile change. It would require writing out the lens definition for `ReportSpec` by hand, and probably also for `ReportOpts`. The others could still be auto-derived, or they could be written out by hand.

Xitian9:
Just a quick question. Since this PR is quite large, it frequently requires rebases to avoid conflicts. The biggest culprits here are the first two commits which rename a bunch of fields. If you find those two commits beneficial/inoffensive enough, I could break them apart into a separate PR which could be merged immediately, and it will save rebasing work later on.

simonmichael:
I will prioritize this one, thanks.

simonmichael:
Looking at the first two rename commits. The first renames
```
data ReportSpec = ReportSpec
  { rsOpts      :: ReportOpts  -- ^ The underlying ReportOpts used to generate this ReportSpec
  , rsToday     :: Day         -- ^ The Day this ReportSpec is generated for
  , rsQuery     :: Query       -- ^ The generated Query for the given day
  , rsQueryOpts :: [QueryOpt]  -- ^ A list of QueryOpts for the given day
```
to
```
data ReportSpec = ReportSpec
  { reportopts_ :: ReportOpts  -- ^ The underlying ReportOpts used to generate this ReportSpec
  , reportday_  :: Day         -- ^ The Day this ReportSpec is generated for
  , query_      :: Query       -- ^ The generated Query for the given day
  , queryopts_  :: [QueryOpt]  -- ^ A list of QueryOpts for the given day
```
"to be consistent with ReportOpts, CliOpts, etc."

So the goal is to rename to flatcase or snake_case, rather than camelCase ? For consistency and maybe for easier/more idiomatic lens generation ?

(camelCase (typeinitialsFieldName) is the convention I have been moving towards, as a more scalable one that avoids clashes from two types having similarly named fields. Older/widespread types have the easier-to-type flatcase/snake_case convention, for historical/familiarity reasons and because the slightly reduced boilerplate/easier typing is perhaps justified for heavily used types.)

Xitian9:
Yes, the main objective was for more consistent lens generation. We could also accomplish this by renaming the fields of ReportOpts and CliOpts. If you're wanting to move towards camel case I would support that.

Since it's looking like we'll be defining at least some of the lenses manually rather than with template haskell, it's less important to rename for consistency. We can also give the historical easy-to-type names to the lenses, since those will be applicable to any options structure in the tower. But since we're on the topic now we can decide what course of action we should take on the names.

- Have inconsistent naming convention? (no effort)
- Would you prefer to rename the fields of ReportSpec? (consistent, least effort)
- Rename the fields of ReportOpts and CliOpts? (consistent, best practice, more effort)

simonmichael:
I'm ok with low effort / maintain current inconsistency / baby steps to start with. These particular proposed names look nice to use, but I would worry that no other type or local variable could be named query or reportopts, eg.

simonmichael:
There is some great forward exploration in this PR. Can we reduce the scope now to a first step that's easy to merge, postponing some of the further cleanups and fancy uses ? This seems a good sequence to me:

- Figure out some simple first goals/policies/idiom lens use in the hledger codebase. Eg, what things do we want to use lens for, and where ? Is there any restriction on what kinds of optics we will use ?
- Document these in CONTRIBUTING, if needed.
- Ensure basic lens libs and imports are available in all packages.
- Provide basic lenses for the most important/deeply nested core types, but don't use them yet.
- Evaluate the effect on build performance, consider reworking code structure (eg splitting up modules, separating TH..)
- Start using basic lenses in specific places (a few scripts, readers, reports, commands..)
- Experiment with more aggressive lensification in hledger-ui

simonmichael:
OR... I can save a bunch of time and just merge all this PR as-is, and let things just work out as they will. Any reviewers, you can have influence here.

simonmichael:
Also, what are all the different kinds of lenses/lens construction we can choose from ? I think:

1. Hand-written lenses - avoids possible TH compilation/portability issues
2. TH-generated lenses - avoids lots of boilerplate code and maintenance as types change
3. Hand-written classy lenses - allows more convenient, more implicit style of coding
4. TH-generated classy lenses

Is that about right ? The classy lenses add some complexity, is it a big readability/usability improvement ? Any other downsides ?

Xitian9:
That's about right. The fanciest classy lenses (that implicitly call `updateReportSpec`) would require hand-written definitions.

I think the classy lenses are a big readability improvement. Consider the following code snippet, and let's just concern ourselves with setting `accountlistmode_` to `ALFlat`.

https://github.com/simonmichael/hledger/blob/a89815cea02da0b25744ad094345d2d2aba4245d/hledger/Hledger/Cli/Commands/Close.hs#L104-L107

With ordinary lenses, this would be written as
```
rspec_ = set (rsOptsL . accountlistmode) ALFlat rspec
(acctbals',_) = balanceReport rspec_ j 
```

With classy lenses, this would be written
```
rspec_ = set accountlistmode ALFlat rspec
(acctbals',_) = balanceReport rspec_ j 
```

In fact, `rspec` is extracted from a `CliOpts`, so if we wanted to make sure `rspec` is kept consistent between all uses and not updated separately, this could be rewritten
```
copts = set accountlistmode ALFlat copts'
(acctbals',_) = balanceReport (reportspec_ copts) j 
```

It also allows for easy refactoring if we change the options structure (creating or destroying `BalancingOpts`, eliminating `UIOpts`). Contained within this PR is a commit 1032b2bc493f60e533a84a9e2ffe831d6df6ca4b which eliminates `UIOpts`. This is pretty simple when we use classy lenses, but otherwise updating all the code in `Hledger.UI.UIState` would be a nightmare.

The only downside that I can think of is the usual type inference problem with typeclasses. But I doubt this will arise in practice since we always use concrete types for `*Opts`.

simonmichael:
Great example, thanks. That's some kind of magic and it does seem like magic we will want.  So if it doesn't cause nasty type errors in practice, I'm for it.

Xitian9:
I'll address your first dot point above.

Restricting to `hledger-lib` and `hledger` for the moment, I think the example in the post above is pretty representative of the main use case. Each report takes some options, modifies them in report-specific ways, and then proceeds from there. This results in reports that have `rspec`, `rspec'`, `rspec''`, `rspec_`, `ropts`, `ropts'`, `ropts''`, and so on. have definitely encountered (and created) bugs where the wrong `ropts` with a not-yet-updated value. Add to this the need to call `updateReportSpec` when you update the embedded `ReportOpts` and it creates a lot of mess and places where bugs can creep in.

My goal would be to have every report be supplied with a single options argument, say `copts'`, which is then modified at a single point and given a bare name with no apostrophes or underscores, say `copts = set balanceaccum Historical $ set accountlistmode ALFlat copts'`. `copts'` would then never be referred to anywhere else, so all options would always be in sync.

A second lesser benefit may be to avoid clutter when reading from deeply nested options (`accountlistmode_ . rsOpts $ reportspec_ copts` -> `copts ^. accountlistmode`), but I wouldn't advocate for replacing things like `accountlistmode_ ropts` with `ropts ^. accountlistmode`.

Xitian9:
I'll add to the first step above: decide on a naming system for lenses.

For `ReportOpts` and `CliOpts` we have a consistent naming system with trailing underscores. For `ReportSpec` we have camel case with a prefix of `rs`, e.g. `rsOpts`. Ideally we would like the lens names to be consistent among themselves and with the bare record name.

An obvious choice is to just drop the trailing underscores. This creates a few name conflicts for which we have discussed proposed solutions earlier. If we are not renaming `ReportSpec` fields then we would need a different system, perhaps `rsQuery <-> query`, `rsOpts <-> reportopts`, `rsQueryOpts <-> queryopts`, `rsToday <-> reportday`.

If we are using classy lenses it probably doesn't make sense to keep a prefix for the type around, since the lens will be applicable to any options type in which the base type is nested.

alerque:
As a **non** Haskeller (I can only hack my way around using copy/paste) I reviewed a fair sampling of the refactoring here. Even though I could not give a clear explanation of what a "Lens" is, if anything it is easier for me to guess and surmise what the code is doing than in the originals.

From @simonmichael in IRC:

> It's time to make a decision on whether and how much to adopt lenses/optics in the hledger codebase. It will probably unavoidably be a trade off, making the code more compact and clearer for experts but more inscrutable for new haskellers.

I don't think it is is more inscrutable. The reduced clutter at least makes it easier to see the overall result and work backwards to how it got there. Mind you I still find Haskell pretty inscrutable, I just don't think this is a step backwards.

(On the other hand mixing whitespace and other style cleanup in refactor commits makes reviews like this much harder than they need to be!)

Xitian9:
Sorry for the whitespace and style cleanups! I need to be more disciplined about keeping those things segregated.

simonmichael:
Thanks for the feedback @alerque. I agree it looks/can look easier on the surface, and at the same time it is more difficult to understand in depth. That may be something we have to embrace to scale up further.

simonmichael:
> I'll add to the first step above: decide on a naming system for lenses.

Good point and our current blocker, I have added it to the list.

The usual rule is that field-accessing lenses are named very much like the fields they access, but differently. Some of our field names end with an underscore, I believe this just a convention to distinguish them as field accessors and not eg local variables, for both humans and GHC. 

With the typeinitialsFieldName scheme this became unnecessary. 

> If we are using classy lenses it probably doesn't make sense to keep a prefix for the type around, since the lens will be applicable to any options type in which the base type is nested.

Interesting. In normal development, is it better for us not to know/not care/forget which exactly type each field belongs to ?

I feel it's useful to have one or the other of these naming conventions to keep it clear when we are dealing with fields. 

`fieldname_` is easy on the eye and hands. It makes field names recognisable. It hides which type a field belongs to, requiring extra effort from new devs. Some care and an occasional sacrifice is required to pick memorable and unique names (two types can't have the same field name). There is an obvious corresponding lens name (`fieldname`) with the same pros and cons, plus an added reasonably high chance of clashing with local variables and imports from other packages. (Any other lens name conventions ? `fieldnameL` ?) 

`tiFieldName` is more systematic. It scales without limit, except when types have the same initials, which is easy enough to work around. It avoids collisions with other types, local variables and imports. It makes field names pretty recognisable. It makes pretty obvious which type they belong to. It's more effort for humans to read and type. There's no obvious corresponding lens name (`tiFieldNameL` ?). It exposes which type a field belongs to.

Xitian9:
I think `tiFieldNameL` is a bit ugly, but doable. The usual solution is to use `_tiFieldName` for the underlying field name, and `tiFieldName` for the lens name. Since we don't use the `tiFieldName` convention for any option type except `ReportSpec`, we could just adopt it for the lens names, e.g. `accountlistmode_ <-> roAccountListMode`. This makes the connection between the field name and the lens name less obvious, but we could decide to rename the field names later if we wanted.

This would require renaming the fields of `ReportSpec` though. Perhaps we could just use the usual trick of prepending an underscore, i.e. `rsOpts -> _rsOpts` with lens name `rsOpts`.

simonmichael:
Right you are, prepending _ is the usual way in the haskell world. Perhaps moving towards that is a good idea.

Whether it's _fieldname or _tiFieldName, I don't feel too strongly about at the moment. I'd be fine with allowing both for now, and even keeping some of the current field names around a bit longer to postpone churn and give us a chance to get more experience. Is it feasible to support all of these temporarily ?

simonmichael:
Not all for a single datatype, but allowing different types to use different conventions for the moment, for expedience and unhurried migration. At least the fieldname_, _fieldname, _tiFieldName conventions, with tiFieldName fields getting an underscore prepended. "Field names currently follow several conventions, but always begin or end with an underscore."

simonmichael:
Also I'm not against experimenting with `fieldname_` or `_fieldname` field having a `tiFieldName` lens, as you mentioned. I'm just a bit concerned about having multiple names for things, as in some of the proposed commits where the field name no longer matches the CLI option.

simonmichael:
How about:

- hledger field names always begin or end with an underscore.
- Fields are spelled `_fieldname` (lens: `fieldname`) when we can get away with it (not too clashing, not too hard to remember).
- Otherwise/when necessary, fields are spelled `_tiFieldName` (lens: `tiFieldName`) (ti = a unique prefix for the type, usually the lowercase initial(s) of the type name).
- Many fields currently have the legacy `fieldname_` spelling (lens: `fieldname`). We may migrate these to `_fieldname` later.

Xitian9:
That sounds like a reasonable proposal. This means we will currently leave field names for `ReportOpts` and `CliOpts` unchanged, but will have to migrate `ReportSpec`. To be clear, should we be migrating to `_rsOpts` or to `_reportopts` (i.e. is this a situation where we can ‘get away with it’?

simonmichael:
We are already using tiFieldName there, and the `reportopts` and `query` names seem a bit clashy, with local variables (probably not in the main hledger code but maybe in scripts), or with potential other types with similar fields. But you are close to this code, if you think it's worth the benefit then I support it.

Xitian9:
We will still need to rename some of the `ReportOpts` and `InputOpts` fields to avoid clashes, as in 22071c7820ac545328220536ee894af1a585f45b. I presume these naming choices are acceptable?

In that case, I will propose the following renaming for `ReportSpec`
- `rsOpts` -> `_rsReportOpts`, lens name `rsReportOpts` (this should rarely be needed in practice with classy lenses)
- `rsToday` -> `_rsDay`, lens name `rsDay` (it really just means the report day, which is usually, though not necessarily, today)
- `rsQuery` -> `_rsQuery`, lens name `rsQuery`
- `rsQueryOpts` -> `_rsQueryOpts`, lens name `rsQueryOpts`

As a first step, I will implement the above renamings, and hand-write classy lenses for `ReportOpts` and `ReportSpec`. This will use the fancy lens type proposed in https://github.com/simonmichael/hledger/pull/1545#issuecomment-872714192 where necessary. I will not use these lenses anywhere yet, just implement them.

Does this sound good?

Xitian9:
And I think I'll implement this in a new branch/PR for now, due to almost all the code being changed.

simonmichael:
22071c7 are the kind of renames I don't love, creating a new set of names for things. I'm thinking about alternatives. The ReportSpec renames are fine.

simonmichael:
They are the kind of clashy names where the policy say we resort to tiFieldName.  We could avoid renaming everything, instead just doing minimal renames on the problem fields - rocolor, rodrop, roempty...

Xitian9:
Okay, so we would have a mix of naming styles for `ReportOpts` fields. That is slightly aesthetically displeasing, but reasonable.

Would we use camel case for prefixed fields? My understanding of your proposal is that those would be called `roColor`, `roDrop`, `roEmpty`, etc..

simonmichael:
> Okay, so we would have a mix of naming styles for ReportOpts fields. That is slightly aesthetically displeasing, but reasonable.
> 
Yes it's ugly but it's just two extra characters on a couple of fields - I'd rather try it as a possibly temporary next step, than adding a new set of names.
> Would we use camel case for prefixed fields? My understanding of your proposal is that those would be called roColor, roDrop, roEmpty, etc..
> 
That would be the proper `tiFieldName` spelling, but I'm tentatively proposing something in between: `tifieldname`. It'll be less jarring than mixing `fieldname` and `tiFieldName` in the type don't you think.

simonmichael:
Or we go all the way and switch ReportOpts to tiFieldName but it seems a pity. Your call.

Xitian9:
> That would be the proper `tiFieldName` spelling, but I'm tentatively proposing something in between: `tifieldname`. It'll be less jarring than mixing `fieldname` and `tiFieldName` in the type don't you think.

Hmm… I don't think it's significantly less jarring than `tiFieldName`, and it would necessitate a further migration step if we rename the other fields.

> Or we go all the way and switch all ReportOpts fields to tiFieldName but it seems a pity ? Your call.

This would be my inclination, but I think experience has shown that I'm more willing to make breaking changes. I think these are pretty safe. If it's my call I'd go with this, but I'm happy for you to rein me in.

If we do go this route it will cause a lot of conflicts with other PRs. I would recommend merging any other PRs which are ready or close to it, then creating a PR implementing just the renamings and merging it quickly.

simonmichael:
ReportOpts is a special case because:

- It's a large type, with many fields
- it's used everywhere, so clarity, brevity, and avoiding unnecessary churn are more important than usual
- many of the fields correspond to UI flags, and we'd like to preserve that correspondence, so renaming is less desirable than usual

Importing qualified is an alternative to renaming, but probably too cumbersome. 

If we must rename I prefer it to be a consistent predictable change  rather than different names. Like prepending type initials, or appending two underscores, or appending a not-too-distracting suffix like flag (emptyflag, dropflag, colorflag..)

Let's sleep on it.

Xitian9:
There are several good options there. If we're not renaming everything, appending `flag` to the relevant options would be a good choice (except that `drop` is not a flag, but I'm not fussed about that).

simonmichael:
As general question, is it good or bad for (a) fields and (b) lenses to be visually distinct from other kinds of bindings (functions, "variables") ? 

I would guess it's good, because fields and lenses are used in specific ways, and it's helpful for comprehension and to avoid collisions if they follow some consistent, low-noise naming convention. I don't have enough lens experience to really know, though.

If it's good, here are some possibilities (`field` and `lens` are the same name here):

- `field_`, `lens__`
- `_field`, `lens_`
- `_field`, `lensl`/`lensL`

PS isn't `--drop` a flag ? (Personally, I call them options if they have an argument and flags if not.)

Xitian9:
> As general question, is it good or bad for (a) fields and (b) lenses to be visually distinct from other kinds of bindings (functions, "variables") ?
> 
> If it's good, here are some possibilities (`field` and `lens` are the same name here, they could be lowercase, camelCase or whatever):
> 
>     * `field_`, `lens__`
> 
>     * `_field`, `lens_`
> 
>     * `_field`, `lensl`/`lensL`

The only widely-used convention that I'm aware of is `_field` and `lens`. Here are some [justifications for that convention by Edward Kmett](https://www.reddit.com/r/haskell/comments/4lakpi/lens_consistency/d3lzees?utm_source=share&utm_medium=web2x&context=3). I think the idea is that when they exist, lenses are used much more commonly than bare functions, and so have the more idiomatic name.

> PS isn't `--drop` a flag ? (Personally, I call them options if they have an argument and flags if not.)

`--drop` takes an argument, an integer specifying how many levels to drop from the beginning of the account names.

simonmichael:
Great link! Ok then let's follow that advice. So how about:

- preferred naming is either _field/lens, or when necessary/preferred, _tiFieldName/tiLensName. Chosen per type. 

- when lens names are too clashy, we add a _ suffix.

- ReportOpts has its own convention for the moment: field_/lens. When lens is too clashy, we add a double __ suffix. Later, we'll migrate it to the preferred scheme.

- minimal renaming of existing things at this stage.

Right you are about --drop. I was using "flag" loosely. "opt" could be a more generic suffix for xOpts fields. I think we can get by without it though.

Xitian9:
That sounds good. Just to make sure I understand, for the clashy `ReportOpts` names, are we appending `flag` to the field name, or appending two underscores to the lens name?

- `empty_` -> `empty_`/`empty__`
- `empty_` -> `emptyflag_` / `emptyflag`

Agreed, I'm happy to play fast and loose with the term flag.

simonmichael:
I think `empty_`/`empty__` for now. 

 Later we'll  most likely rename to `_empty`/`empty_` I expect, or use `opt` or `flag`.

Xitian9:
Great. I think we've decided on the scope for the next PR. I'll provide a link to that when I've completed it.

Since the scope still includes renaming the fields of `ReportSpec`, it may still be wise to hold off on this until we merge some of the other PRs, then do a quick PR which is just the renaming, to avoid conflicts.

simonmichael:
Great! Thanks Stephen.

Xitian9:
Just recording this for my own information.

The only PR which seems to have significant conflicts with a renaming of the fields of `ReportSpec` is #1579.  #1593 has a very minor conflict, but as that PR may not be in final form it may get worse.

Xitian9:
I have submitted #1624 to rename the fields of `ReportSpec`.

I just realised, based on `InputOpts` and `CliOptions` not being explicitly exempted from the proposed naming scheme, should I rename those fields as well? e.g. `output_format_` -> `_output_format` or `coOutputFormat`, `ignore_assertions_` -> `_ignore_assertions` or `ioIgnoreAssertions`?

simonmichael:
Generally let's do the minimum renaming at this step. We can keep the legacy spelling for these types as well for now.

Xitian9:
With the merging of #1647, and the submission of #1676, all of the material from this PR is now merged or in other submitted PRs. There remains documentation which refers to discussion here, but this PR can now be closed.

PR #1554, opened 2021-05-04
Generate conversion postings for priced amounts.

Xitian9:
A potential implementation of the proposals in #1177. This is almost certainly not the final behaviour, but I'm putting this up to solicit feedback on a solid proposal. Let me know your thoughts!

This means there will no longer be unbalanced transactions, but will be
offsetting conversion postings to balance things out. For example.

```
2000-01-01
  a   1 AAA @@ 2 BBB
  b  -2 BBB
```

When converting to cost, this is treated the same as before.
When not converting to cost, this is now treated as:

```
2000-01-01
  a                           1 AAA
  b                          -2 BBB
  equity:conversion:AAA:BBB  -1 AAA
  equity:conversion:BBB:AAA   2 BBB
```

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -80,6 +80,7 @@ module Hledger.Data.Amount (
   amountUnstyled,
   showAmountB,
   showAmount,
+  showAmountPrice,
   cshowAmount,
   showAmountWithZeroCommodity,
   showAmountDebug,
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -80,6 +80,7 @@ module Hledger.Data.Journal (
   journalLiabilityAccountQuery,
   journalEquityAccountQuery,
   journalCashAccountQuery,
+  journalConversionAccount,
   -- * Misc
   canonicalStyleFrom,
   nulljournal,
@@ -120,9 +121,10 @@ import Hledger.Utils
 import Hledger.Data.Types
 import Hledger.Data.AccountName
 import Hledger.Data.Amount
+import Hledger.Data.Posting
 import Hledger.Data.Transaction
 import Hledger.Data.TransactionModifier
-import Hledger.Data.Posting
+import Hledger.Data.Valuation
 import Hledger.Query
 
 
@@ -395,7 +397,8 @@ letterPairs _ = []
 -- queries for standard account types
 
 -- | Get a query for accounts of the specified types in this journal. 
--- Account types include Asset, Liability, Equity, Revenue, Expense, Cash.
+-- Account types include:
+-- Asset, Liability, Equity, Revenue, Expense, Cash, Conversion.
 -- For each type, if no accounts were declared with this type, the query 
 -- will instead match accounts with names matched by the case-insensitive 
 -- regular expression provided as a fallback.
@@ -505,6 +508,13 @@ journalProfitAndLossAccountQuery  :: Journal -> Query
 journalProfitAndLossAccountQuery j = Or [journalRevenueAccountQuery j
                                         ,journalExpenseAccountQuery j
                                         ]
+
+-- | The 'AccountName' to use for automatically generated conversion postings.
+journalConversionAccount :: Journal -> AccountName
+journalConversionAccount =
+    headDef (T.pack "equity:conversion")
+    . M.findWithDefault [] Conversion
+    . jdeclaredaccounttypes
 
 -- Various kinds of filtering on journals. We do it differently depending
 -- on the command.
@@ -870,10 +880,11 @@ postingInferredmarketPrice p@Posting{pamount} =
 
 -- | Convert all this journal's amounts to cost using the transaction prices, if any.
 -- The journal's commodity styles are applied to the resulting amounts.
-journalToCost :: Journal -> Journal
-journalToCost j@Journal{jtxns=ts} = j{jtxns=map (transactionToCost styles) ts}
-    where
-      styles = journalCommodityStyles j
+journalToCost :: ConversionOp -> Journal -> Journal
+journalToCost cost j@Journal{jtxns=ts} =
+    j{jtxns=map (transactionToCost (journalConversionAccount j) styles cost) ts}
+  where
+    styles = journalCommodityStyles j
 
 -- -- | Get this journal's unique, display-preference-canonicalised commodities, by symbol.
 -- journalCanonicalCommodities :: Journal -> M.Map String CommoditySymbol
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -81,7 +81,7 @@ import Data.Foldable (asum)
 import qualified Data.Map as M
 import Data.Maybe (fromMaybe, isJust)
 import Data.MemoUgly (memo)
-import Data.List (foldl')
+import Data.List (foldl', sort)
 import qualified Data.Set as S
 import Data.Text (Text)
 import qualified Data.Text as T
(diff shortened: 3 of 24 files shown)


crocket:
I think this looks good.

Xitian9:
I've rebased on current master. This is currently failing a lot of tests, which is because the default way of handling costs has changed. Fixing this depends on making a decision about what the interface for this new feature is.

Previously, costing had two options: use the primary cost or the cost basis. Certain special cases, e.g. the entries report, had the option of displaying both in the standard format `1 USD @ 1 CAD`.

This introduces a new option, which will use the primay basis, but add conversion postings to leave things fully balanced.
- Use just the primary amount. This will produce unbalanced transactions. This is the current default behaviour.
- Use the primary amount, with conversion postings. This will produce balanced transactions. This is new behaviour.
- Use just the cost basis. This will produce balanced transactions. This is the behaviour with `--cost`.
- Do no transformations, and keep the raw amount around. This is used by entries reports.

What do we want to be the default behaviour? What do we want the UI to be to access the new behaviour?

Xitian9:
One proposal: extend the `--cost` flag to an option.
- Default/`--cost=conversion`: Use the primary amount, with conversion postings.
- `--cost`/`--cost=cost`: Use the cost basis.
- `--cost=nocost`: Use just the primary amount, with no conversion postings.

Xitian9:
Keeping the price around is reasonable, and IIRC what is done in this proposal. The only thing to watch out for is that cost conversion would not be idempotent, so we would need to make sure we don't do it twice.

Correct me if I'm wrong, but it seems that your proposal basically amounts to enforcing that the auto-generated postings instead be manually entered. This would be the principled way to go about it, but would be very verbose (I've tried this entry style in the last year; I made mistakes more often than not), and also non-backwards compatible.

What is your objection to generated postings in this case? If I can understand that, it would help tweaking the proposal.

la-wu:
> Keeping the price around is reasonable, and IIRC what is done in this proposal. The only thing to watch out for is that cost conversion would not be idempotent, so we would need to make sure we don't do it twice.

I see that prices are being stripped https://github.com/simonmichael/hledger/pull/1554/files#diff-291f93684d033af89a8fb3cc406fca888bcdc2b89a6df621e5fe015765ed177aR350.

I think this would be an easily spotted bug if we did apply it twice in the code. But I agree that workflow wise if the conversion postings get copied into the source journal somehow we might run into issues. This is part of the reason why I'm not a fan of auto-generated postings.

> This would be the principled way to go about it, but would be very verbose (I've tried this entry style in the last year; I made mistakes more often than not), and also non-backwards compatible.

Yeah I think the enforcement should be optional for now (I'm open for `--strict` or a new flag). And I agree that as a standalone it introduces room for errors. I think we should be able to enumerate most of those and do extra validation though. Which mistakes did you find most common?

> What is your objection to generated postings in this case? If I can understand that, it would help tweaking the proposal.

Generally I view the journal/ledger file as more a source of truth and that the tool (hledger) could support syntactic sugar (balance inference) but should not create new entries (I'm also somewhat on the fence about periodic expressions for that reason but they are at least require explicit input in the journal file).

A minor point is that we might want conversions from different transactions to be placed in different accounts (`equity:currency` vs `equity:stocks` vs `equity:crypto`?)

crocket:
I think it's okay for `bse` report to automatically generate conversion postings.

Xitian9:
> > Keeping the price around is reasonable, and IIRC what is done in this proposal. The only thing to watch out for is that cost conversion would not be idempotent, so we would need to make sure we don't do it twice.
> 
> I see that prices are being stripped https://github.com/simonmichael/hledger/pull/1554/files#diff-291f93684d033af89a8fb3cc406fca888bcdc2b89a6df621e5fe015765ed177aR350.

Aha, you are correct. My mistake.

> > This would be the principled way to go about it, but would be very verbose (I've tried this entry style in the last year; I made mistakes more often than not), and also non-backwards compatible.
> 
> Yeah I think the enforcement should be optional for now (I'm open for `--strict` or a new flag). And I agree that as a standalone it introduces room for errors. I think we should be able to enumerate most of those and do extra validation though. Which mistakes did you find most common?

Getting the signs wrong on the conversion postings. This was why I implemented the ‘no automatic price inferal’ behaviour of `--strict`: otherwise it would auto-generate prices.

> > What is your objection to generated postings in this case? If I can understand that, it would help tweaking the proposal.
> 
> Generally I view the journal/ledger file as more a source of truth and that the tool (hledger) could support syntactic sugar (balance inference) but should not create new entries (I'm also somewhat on the fence about periodic expressions for that reason but the at least require explicit input in the journal file).

I completely agree, but I actually don't think this is a big problem. I view this as precisely syntactic sugar, a shortcut way of writing the conversion postings without having to type them all out. 

The automatically generated postings would be clearly marked with a `generated-posting:` tag, and would only make their way back into the journal if you somehow overwrote the journal with the output of a print command. Given that we already have auto-postings, this doesn't introduce any new gotchas that don't exist already.

> A minor point is that we might want conversions from different transactions to be placed in different accounts (`equity:currency` vs `equity:stocks` vs `equity:crypto`?)

True. I think we should leave this as is for now, but if enough people request this feature we could make this configurable with a tag on the amount, e.g.
```
2021-01-01  Buy stuff
    assets         1 stock @ $1  ; conversionacct:equity:stocks
    assets:cash  -$1
```

la-wu:
> Getting the signs wrong on the conversion postings. This was why I implemented the ‘no automatic price inferal’ behaviour of `--strict`: otherwise it would auto-generate prices.

Could you provide an example in how this would be reflected in a conversion account? e.g the following balancing without error?

```
2021/08/09 Buy AAPL
  assets:trading                             -5 AAPL
  assets:trading                         -750.00 USD
  equity:conversion
```

> The automatically generated postings would be clearly marked with a `generated-posting:` tag, and would only make their way back into the journal if you somehow overwrote the journal with the output of a print command. Given that we already have auto-postings, this doesn't introduce any new gotchas that don't exist already.

I see. I wasn't familiar with that feature so thanks for pointing it out. After reading a bit about auto-postings, I'm still not a fan of them either but I suppose there isn't a new paradigm if it already exists. In general it feels to me that both auto postings and this conversion account would be better built into a workflow at data-entry time and not at report-generation time.

I would personally be satisfied if this auto-posting-like behavior was optional (or optionally disabled) and simultaneously we (optionally?) balance and infer transactions irrespective of lot basis (i.e `pricebalanced` in https://github.com/simonmichael/hledger/pull/1544 is no longer balanced).

crocket:
I had to spend a few hours to insert equity:conversion:CUR1:CUR2 into currency conversion.s

I'd very much like to see the insertion of conversion postings automatied.

crocket:
> I would personally be satisfied if this auto-posting-like behavior was optional (or optionally disabled) and simultaneously we (optionally?) balance and infer transactions irrespective of lot basis (i.e pricebalanced in #1544 is no longer balanced).

I think currency conversion can be enabled by a directive in a journal.

Xitian9:
> > Getting the signs wrong on the conversion postings. This was why I implemented the ‘no automatic price inferal’ behaviour of `--strict`: otherwise it would auto-generate prices.
> 
> Could you provide an example in how this would be reflected in a conversion account? e.g the following balancing without error?

I dislike the auto-balancing feature, so I generally enter the amounts explicitly. I would accidentally do things like the following, usually from copy-pasting and forgetting the switch the signs
```
2021/08/09 Buy AAPL
  assets:trading                            5 AAPL
  assets:trading                         -750.00 USD
  equity:conversion                         5 AAPL
  equity:conversion                      -750.00 USD
```

If you don't use strict mode, this will be automatically treated as
```
2021/08/09 Buy AAPL
  assets:trading                            5 AAPL @ 150.00 USD
  assets:trading                         -750.00 USD
  equity:conversion                         5 AAPL @ 150.00 USD
  equity:conversion                      -750.00 USD
```

Which balances, but is not what you want.

> I would personally be satisfied if this auto-posting-like behavior was optional (or optionally disabled) and simultaneously we (optionally?) balance and infer transactions irrespective of lot basis (i.e `pricebalanced` in #1544 is no longer balanced).

The strategy we seem to have adopted is that if you want more strict checks on your journals than provided by default, that should be handled with the `--strict` flag and/or a journal check command. The `fullybalanced` check which was proposed in #1544 but has not been implemented seems to do what you want.

Xitian9:
Rebased. I'm not sure why this is complaining about the commit message; it seems to match to me.

la-wu:
Looks like `Merge pull request #1715` doesn't match https://github.com/simonmichael/hledger/blob/master/bin/commitlint#L58

simonmichael:
I gave this a quick try and haven't seen any user-visible difference yet, eg in print -x or bal -B. Should I ?

simonmichael:
Nice. Could (should) we implement it deeper, so that it affects all commands, such as print ? I can see equity postings with `reg`, but not with `print`, which would be unusual.

Xitian9:
I chose not to do conversion postings for `print` because `print` already prints out the prices in `@/@@` notation, which is shorthand for the conversion postings. My understanding is that `print` should as closely cleave to the original journal entry as possible (e.g. not filling in auto-inferred amounts unless specifically requested with `--explicit`), so conversion postings should not be displayed unless explicitly entered in the journal.

simonmichael:
Agreed, but perhaps --explicit should show them ? Or if that's problematic, a new flag ?

simonmichael:
Though it will still be odd if print requires a special flag to see what other commands see by default. I'm not sure how to get around that, let's think more.

Xitian9:
We used to have just two cost modes – `Cost` and `NoCost` – in which case a flag was sufficient. We now have three – `Cost`, `NoCost`, and `ConversionCost`. The flag `--cost` will still access `Cost` mode, but there is no way to select between `NoCost` and `ConversionCost`. `NoCost` is used by `print` and `ConversionCost` is used by all other commands. I think we need more fine-grained control, for example `--cost=cost`/`--cost`, `--cost=conversion`, or `--cost=nocost`.

I think it's not unusual that print shows something different: it's already different in that it displays costs at all, while all other commands will strip them.

Xitian9:
Showing conversion postings with --explicit seems reasonable too.

Xitian9:
> We used to have just two cost modes – `Cost` and `NoCost` – in which case a flag was sufficient. We now have three – `Cost`, `NoCost`, and `ConversionCost`. The flag `--cost` will still access `Cost` mode, but there is no way to select between `NoCost` and `ConversionCost`. `NoCost` is used by `print` and `close` and `ConversionCost` is used by all other commands. I think we need more fine-grained control, for example `--cost=cost`/`--cost`, `--cost=conversion`, or `--cost=nocost`.

Any thoughts on this proposal?

simonmichael:
I think it needs draft docs ? That would help me assess it.

Xitian9:
I've added the enhanced --cost option and some docs in a new commit. Let me know what you think.

simonmichael:
Maybe this would be better as a separate flag ?  I think it's not variations of how to calculate cost, but rather a choice of whether to generate and show conversion postings, independent of whether reporting cost or not ?

simonmichael:
From trying to write feedback, I see this is a complex topic, such that I think we need to flesh out the specification/draft doc, here on this issue because #1177 is huge. Apologies if I have forgotten past discussion/context. It looks like my last coherent thoughts on this were [here](https://github.com/simonmichael/hledger/issues/1177#issuecomment-648215555). Am I right in thinking this PR is aimed at the second bullet point, "Let's implement conversion rate inference from a balanced @-free transaction - it seems possible and would allow -B and --infer-value to work without the need for @ notation" ?

simonmichael:
No, I think that was not right; this PR doesn't address that bullet point. 

Instead I think it aims to help people who are using @ notation, or implicit @ (autoconversion), to produce reports where the equity balance is correct and the accounting equation can balance (by generating the required equity postings where they are missing).

Would that be accurate ? Let's nail down 

Goals: 
User-visible changes:
Impact/compatibility:
Interactions:

Xitian9:
> Instead I think it aims to help people who are using @ notation, or implicit @ (autoconversion), to produce reports where the equity balance is correct and the accounting equation can balance (by generating the required equity postings where they are missing).
> 
> Would that be accurate ?

Yes, that's correct. Current behaviour without `-B/--cost` will generate unbalanced postings; this automatically generates equity postings so that @/@@ transactions are balanced. 

> Let's nail down
> 
> Goals: User-visible changes: Impact/compatibility: Interactions:

Goals:
- Transactions involving transaction prices can be balanced (preferably by default or by explicit choice)
- Users can continue to get previous `--cost` behaviour if desired
- Users can get previous unbalanced transactions if desired (for simplicity)

Impact/compatibility:
- If default behaviour is changed then there will be extra postings generated. They will be in special equity accounts, so this should cause minimal interference.
- If default behaviour is not changed, then there should be no noticeable effect without explicitly calling for it.

Interactions:
- None currently seen, as we have to deal with costing anyway. Perhaps some weird interaction with `--gain` that I'm not thinking of.

Xitian9:
> Maybe this would be better as a separate flag ? I think it's not variations of how to calculate cost, but rather a choice of whether to generate and show conversion postings, independent of whether reporting cost or not ?

I think this is not quite correct. If we're reporting cost, then generating conversion postings will make transactions unbalanced, exactly the problem we were trying to avoid.

simonmichael:
I have posted an attempted summary of this proposal, with some open questions, at https://hledger.org/conversion2.html#xitians-1554 .

simonmichael:
See #hledger chat log for much other discussion. Now in master - thanks a lot @Xitian9 !

PR #1560, opened 2021-05-13
Clean up valuation functions, and make clear which to use where.

Xitian9:
Doing costing and valuation for amounts is a bit confusing; every report handles it differently. This was fine when valuation was simpler, but now that we have `--value=then`, `--value=end`, `--valuechange`, and soon `--gain`, this means a lot of duplicated code handling special cases.

This PR simplifies things by defining one master valuation function which should be used in most cases: `journalApplyValuationFromOpts`. It does the following based on the `ReportSpec`:
- Convert to cost
- Strip prices if able
- Apply valuation

The only place this function is not sufficient is when performing `--value=end` valuation on balance reports that require you to sum amounts, such as the historical balance report. To handle these cases, we introduce `mixedAmountApplyValuationFromOptsWith`, which returns the valuation function to be used in those cases by `multiBalanceReport`. It also defines the function `valuationAfterSum`, which tests a `ReportOpts` to see if we need to do this second kind of valuation.

I have then pruned a bunch of helper functions which are no longer so helpful.

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -29,6 +29,8 @@ module Hledger.Reports.ReportOptions (
   simplifyStatuses,
   whichDateFromOpts,
   journalSelectingAmountFromOpts,
+  journalApplyValuationFromOpts,
+  journalApplyValuationFromOptsWith,
   intervalFromRawOpts,
   forecastPeriodFromRawOpts,
   queryFromFlags,
@@ -47,6 +49,7 @@ module Hledger.Reports.ReportOptions (
 where
 
 import Control.Applicative ((<|>))
+import Control.Monad ((<=<))
 import Data.List.Extra (nubSort)
 import Data.Maybe (fromMaybe, mapMaybe)
 import qualified Data.Text as T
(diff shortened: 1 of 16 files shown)


hledger-lib/Hledger/Reports/ReportOptions.hs
```diff
@@ -488,13 +491,56 @@ flat_ = not . tree_
...
 journalSelectingAmountFromOpts :: ReportOpts -> Journal -> Journal
-journalSelectingAmountFromOpts opts = case cost_ opts of
+journalSelectingAmountFromOpts ropts = maybeStripPrices . case cost_ ropts of
     Cost   -> journalToCost
     NoCost -> id
+  where
+    maybeStripPrices = if show_costs_ ropts then id else journalMapPostingAmounts mixedAmountStripPrices
+
+-- | Convert this journal's postings' amounts to cost using their transaction
+-- prices and apply valuation, if specified by options (-B/--cost). Strip prices
```
simonmichael, line 496: Would something like this be clearer: "Convert this journal's postings' amounts to cost and/or to value, if specified by options (-B/--cost/-V/-X/--value etc.) ..." 
I think I saw similar wording in some other new/updated functions as well, it had me thinking they just converted to cost.

simonmichael: Otherwise, LGTM.

Xitian9: I've updated the wording of that comment.

simonmichael:
Awesome!

PR #1561, opened 2021-05-14
Mention expansion of report start and end dates in report intevals documentation.

Xitian9:
Somebody in the Google group expressed surprise that their quarterly report ended at the end of the quarter, even though they specified an explicit end date in the middle of the of the quarter. It seems that the manual on report intervals does not mention this fact: it is left to later in the period expressions section. I've added a note about this.

### DIFF
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -477,6 +477,9 @@ selected with one of `-D/--daily`, `-W/--weekly`, `-M/--monthly`,
 specified with a [period expression](#period-expressions).
 Report intervals can not be specified with a [query](#queries).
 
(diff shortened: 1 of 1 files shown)


pkel:
This someone is me, hi.

I'm on version 1.18.1. The following might not apply the current version.

After doing `man hledger` and searching for `-e` I get the following.
```
       -e --end=DATE
              include postings/txns before this date
```

The special interpretation for periodic reports should be mentioned here too.

simonmichael:
Thanks. I have merged this, done some rewriting, and mentioned it in the flag help and a few more places.

simonmichael:
EG: https://hledger.org/dev/hledger.html#report-intervals

Issue #1563, opened 2021-05-19
When comparing amounts in different commodities, commodity symbol is compared first, even when amounts have different signs


Xitian9:
The problem of sorting amounts expressed in different commodities is a tough one. Do you convert them to some reference commodity when there's price information, or just compare as is.

Though there's certainly a more detailed conversation to have on this topic, I think there's one aspect of the current behaviour which can clearly be improved: negative numbers in one commodity should always less than positive numbers in another commodity. hledger does not currently respect this: see the following example.

```
$ cat sort.j
P 2021-01-01 YYY XXX1

2021-01-01 Post in XXX
    (a)   100 XXX

2021-01-01 Post in XXX
    (b)  -100 XXX

2021-01-02 Post in XXX
    (c)   100 XXX

2021-01-02 Post in YYY
    (d)  -100 YYY

$  hledger -f sort.j bal -DN --sort -V
Balance changes in 2021-01-01..2021-01-02, valued at period ends:

   || 2021-01-01  2021-01-02 
===++========================
 a ||    100 XXX           0 
 c ||          0     100 XXX 
 b ||   -100 XXX           0 
 d ||          0    -100 XXX 


$  hledger -f sort.j bal -DN --sort
Balance changes in 2021-01-01..2021-01-02:

   || 2021-01-01  2021-01-02 
===++========================
 d ||          0    -100 YYY 
 a ||    100 XXX           0 
 c ||          0     100 XXX 
 b ||   -100 XXX           0

**Comment by Xitian9** [comment]
That said, it looks like the simplest way to solve this problem involves coming up with a better `Ord` instance for `MixedAmount`, so we may have to address the more general problem. My proposal for comparing two mixed amounts:

1. Strip prices.
1. Convert each to their list of `Amount`s.
2. Sort those amounts by `aquantity`.
3. Compare the resulting lists.

It's not perfect, but should perform reasonably in most common cases.

**Comment by simonmichael** [comment]
Some more testing to help me understand this. Using your example, only slightly simplified.:
```
$ hledger bal -N  # Sorting by account is the default
               100 X  a
              -100 X  b
               100 X  c
              -100 Y  d
$ hledger bal -N --sort  # 1.21's sorting by amount. Hard to see the logic.
              -100 Y  d
               100 X  a
               100 X  c
              -100 X  b
$ hledger.1564 bal -N --sort  # #1564's sorting by amount. Numbers look a bit more sorted, at least in this example.
               100 X  a
               100 X  c
[...]

PR #1564, opened 2021-05-19
more useful sort order for multi-commodity MixedAmounts: treat missing commodities as 0

Xitian9:
Addresses #1563. <s>When comparing MixedAmounts, only the list of Quantities is considered</s>. Now substitutes an amount of quantity 0 for any missing commodities when doing comparisons.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -93,6 +93,7 @@ module Hledger.Data.Amount (
   -- * MixedAmount
   nullmixedamt,
   missingmixedamt,
+  isMissingMixedAmount,
   mixed,
   mixedAmount,
   maAddAmount,
@@ -540,6 +541,10 @@ nullmixedamt = Mixed mempty
 missingmixedamt :: MixedAmount
 missingmixedamt = mixedAmount missingamt
 
+-- | Whether a MixedAmount has a missing amount
+isMissingMixedAmount :: MixedAmount -> Bool
+isMissingMixedAmount (Mixed ma) = amountKey missingamt `M.member` ma
+
 -- | Convert amounts in various commodities into a mixed amount.
 mixed :: Foldable t => t Amount -> MixedAmount
 mixed = maAddAmounts nullmixedamt
@@ -637,13 +642,12 @@ maIsNonZero = not . mixedAmountIsZero
 --
 amounts :: MixedAmount -> [Amount]
 amounts (Mixed ma)
-  | missingkey `M.member` ma = [missingamt]  -- missingamt should always be alone, but detect it even if not
-  | M.null nonzeros          = [newzero]
-  | otherwise                = toList nonzeros
+  | isMissingMixedAmount (Mixed ma) = [missingamt]  -- missingamt should always be alone, but detect it even if not
+  | M.null nonzeros                 = [newzero]
+  | otherwise                       = toList nonzeros
   where
     newzero = fromMaybe nullamt $ find (not . T.null . acommodity) zeros
     (zeros, nonzeros) = M.partition amountIsZero ma
-    missingkey = amountKey missingamt
 
 -- | Get a mixed amount's component amounts without normalising zero and missing
 -- amounts. This is used for JSON serialisation, so the order is important. In
@@ -966,7 +970,14 @@ tests_Amount = tests "Amount" [
 
   ,tests "MixedAmount" [
 
-     test "adding mixed amounts to zero, the commodity and amount style are preserved" $
+     test "comparing mixed amounts compares based on quantities" $ do
+       let usdpos = mixed [usd 1]
+           usdneg = mixed [usd (-1)]
+           eurneg = mixed [eur (-12)]
+       compare usdneg usdpos @?= LT
+       compare eurneg usdpos @?= LT
+
+     ,test "adding mixed amounts to zero, the commodity and amount style are preserved" $
       maSum (map mixedAmount
         [usd 1.25
         ,usd (-1) `withPrecision` Precision 3
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -27,8 +27,8 @@ module Hledger.Data.Types
 where
 
 import GHC.Generics (Generic)
-import Data.Decimal
-import Data.Default
+import Data.Decimal (Decimal)
+import Data.Default (Default(..))
 import Data.Functor (($>))
 import Data.List (intercalate)
 import Text.Blaze (ToMarkup(..))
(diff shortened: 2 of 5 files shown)


Xitian9:
One ugly thing about this PR: the new Ord instance for MixedAmount can have `compare a b == EQ` when `a /= b`.  I could refine the instance so it's a bit more subtle, at the cost of extra code.

Xitian9:
Note that this differs from my proposal in #1563. It is much less ambitious in what it does, and as a result is more efficient: it does not require sorting the lists of amounts whenever doing a comparison.

Xitian9:
A consequence of this PR is that `nullmixedamt == mixedAmount nullamt{acommodity="XXX"}`, and more generally that `a == b` if and only if `maIsZero (maMinus a b)`

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -236,7 +236,27 @@ data Amount = Amount {
...
+                 EQ -> compareQuantities (Just x) (Just y) <> go xs ys
+                 LT -> compareQuantities (Just x) Nothing  <> go xs yss
+                 GT -> compareQuantities Nothing  (Just y) <> go xss ys
+    go ((_,x):xs) [] = compareQuantities (Just x) Nothing  <> go xs []
+    go [] ((_,y):ys) = compareQuantities Nothing  (Just y) <> go [] ys
+    go []         [] = EQ
+    compareQuantities = comparing (maybe 0 aquantity) <> comparing (maybe 0 totalprice)
+    totalprice x = case aprice x of
+                        Just (TotalPrice p) -> aquantity p
+                        _                   -> 0
```
simonmichael, line 259: So I think sorting is done:

1. based on quantities in all commodities, with alphabetically first commodities being more significant and missing commodities assumed to be zero

2. if all quantities are equal, then based on the quantity of the total prices (@@), if any, treating a missing total price as zero, and ignoring their commodities (?)

3. if all those are equal, the sort order is undefined

Did I understand it ?

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -970,7 +974,17 @@ tests_Amount = tests "Amount" [
...
-     test "adding mixed amounts to zero, the commodity and amount style are preserved" $
+     test "comparing mixed amounts compares based on quantities" $ do
+       let usdpos = mixedAmount $ usd 1
+           usdneg = mixedAmount $ usd (-1)
+           eurpos = mixedAmount $ eur 12
+           eurneg = mixedAmount $ eur (-12)
+       compare usdneg usdpos @?= LT
+       compare eurneg usdpos @?= LT
+       compare usdneg eurpos @?= LT
+       compare nullmixedamt (mixedAmount $ usd 0) @?= EQ
```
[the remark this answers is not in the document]
Xitian9: Yes, that is correct. I have been running this for a while and haven't encountered any issues, but there are lots of opportunities for tiny problems to sneak in. If you want to be conservative we could leave the typeclass instances unchanged, and make sure we call `sortBy maCompare` in all the appropriate places.

hledger-lib/Hledger/Data/Types.hs
Xitian9: Okay, let's write this out in full detail.

A mixed amount is a list of amounts, and when we're comparing two mixed amounts we step through them sequentially. Each amount has a *key* which determines its precedence in the comparison. An amount's key is determined as follows:
1. First, look at the amount's commodity. Amounts with alphabetically earlier commodities have higher precedence. If two amounts have the same commodity, then move on to the next step.
2. Second, look at the commodity of the amount's price, if it has any. Amounts with no price have higher precedence than amounts with a price; if they do have a price, amounts with alphabetically earlier price commodities have higher precedence. If two amounts have both the same commodity and the same price commodity (or lack of price commodity), then move on to the next step.
3. Next look at whether it's a total price or a unit price. Amounts with total price have precedence over amounts with unit price. If they have a unit price, then they are considered in order of this price.

All that information goes in to determining the key of each Amount in the mixed amount. You now compare the two mixed amounts by stepping through all the amount keys in order of precedence. If a key is present in both mixed amounts, you just compare the quantities of those amounts. If a key is missing from one of the amounts, treat it as being present with quantity 0. If the keys have the same quantity, compare the quantity of their total price, if present, and treat that quantity as 0 if there is no price or if there is a unit price.

Phew…

simonmichael: Phew is right... thanks for writing it out.

It would be good if we can present a simpler model to the user. Either different rules, or a simpler way to explain the above rules. So that it's not too difficult for a human to look at a sorted report and know why it is sorted that way.

simonmichael: PS since it touches our core data type I think this is not something for 1.22, so we can think this over.

simonmichael: As I think you said elsewhere, we could omit some of this detail in user docs. Amount "keys" are a combination of commodity symbol and price, but generally balance reports aggregate differently-priced amounts and don't show prices, so it's good enough to just talk about commodity symbols. Except for the case where all quantities are the same and prices become significant. But probably even that can be glossed over, I doubt it's going to arise in a confusing way in practice. 

And of course we're on the fringe of a fringe feature that no-one has really asked for. I think our main goal here is some kind of sensible/intuitive/explainable behaviour when people use --sort with multiple commodities. And secondly, cleaner/more principled internal design.

Xitian9: I'm happy to have this wait for after 1.22.

Most reports don't actually display prices, so the sort order only needs to look at the commodity symbol and quantity. I'm not sure if any reports both display amounts with prices and sort by amount.

Xitian9:
Actually, a potential issue here that I need to look into. Now that we again only strip prices on display, we might get some weird sort orders based on prices which do not show up in the final display (c.f. #1526). We will have to make sure that sorting happens based on stripped prices where appropriate. I should come up with a test for sorting which would catch this sort of thing.

hledger-lib/Hledger/Data/Types.hs
Xitian9: It's actually worth noting that this is not actually more complicated than the sorting behaviour in master. The only difference is that the behaviour in master is automatically implemented by the compiler, while our custom behaviour has to be implemented manually.

Existing behaviour:

A mixed amount is a list of amounts, and when we're comparing two mixed amounts we step through them sequentially. Each amount has a key which determines its precedence in the comparison. An amount's key is determined as follows:

1. First, look at the amount's commodity. Amounts with alphabetically earlier commodities have higher precedence. If two amounts have the same commodity, then move on to the next step.
2. Second, look at the commodity of the amount's price, if it has any. Amounts with no price have higher precedence than amounts with a price; if they do have a price, amounts with alphabetically earlier price commodities have higher precedence. If two amounts have both the same commodity and the same price commodity (or lack of price commodity), then move on to the next step.
3. Next look at whether it's a total price or a unit price. Amounts with total price have precedence over amounts with unit price. If they have a unit price, then they are considered in order of the quantity of the unit price.

All that information goes in to determining the key of each Amount in the mixed amount. You now compare the two mixed amounts by stepping through the amount keys of each mixed amount. This is where the behaviour starts to differ.

1. Look at the amount with highest precedence key in each mixed amount. If one of those amounts has higher precedence than the other, this determines the comparison of the keys. If the two keys have the same precedence, then compare the quantities of the two amounts. If those are equal, then compare whether the amounts are multipliers. If those are equal, then compare the styles. If those are equal, compare the quantity of the prices. (Plus some more comparisons on the prices which we don't care about).
2. If step 1 was equal, then repeat it with the amounts with second highest precedence in each mixed amount, and so on.

Xitian9:
Actually it looks like that is already tested for! I've enhanced that test to do both flat and tree mode, and added a new test for the proposed behaviour.

hledger-lib/Hledger/Data/Types.hs
Xitian9: The only place sorting by amount is used is in `MultiBalanceReport`, where it explicitly strips prices before sorting. So from a user-visible perspective, the price based stuff is unnecessary.

Xitian9:
How does this sound for a manual entry?

With `-S/--sort-amount`, accounts with the largest (most positive) balances are shown first. Eg: `hledger bal expenses -MAS` shows your biggest averaged monthly expenses first. When more than one commodity is present, they will be sorted by the alphabetically earliest commodity first, and then by subsequent commodities (if an amount is missing a commodity, it is treated as 0).

Revenues and liability balances are typically negative, however, so `-S` shows these in reverse order. To work around this, you can add `--invert` to flip the signs. (Or, use one of the higher-level reports, which flip the sign automatically. Eg: `hledger incomestatement -MAS`).

simonmichael:
It sounds good!

Xitian9:
Let me know if there's anything else I can do to get this ready for merge. And if you're just taking a break from code review that's fine too ^^.

simonmichael:
Will do. Yes I'm rationing out large-PR-code-review time, it is costly. Sorry for the delay.

simonmichael:
PS other devs can help by contributing code review! (I'm very pleased that this is starting to happen.)

simonmichael:
There was some review and discussion with ninewise on chat.
Code, tests, docs, commitlint are looking good. 
When #1597 is fixed let's check that chart with this PR.

The basic goal of this change is to provide more predictable and intuitive behaviour when using -S/--sort-amount with multiple commodities, by implementing a custom Ord (and Eq) instance for MixedAmount rather than relying on the derived ones. As a consequence, all the ways of representing zero with a MixedAmount ([], [A 0], [A 0, B 0, ...]) are now considered Eq-ual.

simonmichael:
(When you get a chance, please add the above summary or something similar to the imp: commit message, will be handy for changelog.)

Xitian9:
Commit message amended.

simonmichael:
Looking good, thank you!

PR #1571, opened 2021-06-15
Static builds for Linux on x64 (amd64) and arm32v7 architectures

thecount2a:
This PR will add github actions that do static builds for both the 64-bit intel platform and arm32v7 architectures. This means that hledger becomes accessible to people on many versions of Linux and many different types of devices without ANY complex dependencies and regardless of the glibc version of their Linux distro. This is huge, since the current Linux binary build that is offered is currently compiled on the latest version of Ubuntu, which means it cannot be used on most other Linuxes or on older versions of Ubuntu. 

These github actions will automatically build both static Linux builds when any tags of the format [numeric].[numeric]-any-modifying-text or [numeric].[numeric].[numeric]-any-modifying-text are pushed. "-any-modifying-text" can be omitted of course but may be included so that a release candidate build can be pushed before doing an actual release, just to make sure static builds are still working.

The x64 build is linked with the musl library (alternative to glibc). This makes it unlikely to ever try to dynamically load ANY other libraries, regardless of circumstances. This is the gold standard when it comes to generating static builds. This build generates the core hledger program, along with hledger-ui, and hledger-web. This build uses the alpine light-weight operating system as the underlying build environment and takes advantage of its support for the musl library as the default and makes static builds easy. Cabal is used for building, due to its support for static builds.

The ARM build is linked with with the glibc library and currently only builds the main hledger binary, rather than building all main hledger binaries. The reason for this is that glibc is known for trying to dynamically load the glibc library on the platform if you use networking features of glibc or do TTY manipulation, even if the original binary was built statically against glibc. This defeats the purpose of static builds, since it will look for a recent version of glibc. hledger-ui does major TTY manipulation (using ncurses) and hledger-web uses glibc networking features. Therefore, until a musl-linked GHC build becomes available on the ARM platform, we will only be able to compile and link hledger with glibc, excluding hledger-ui and hledger-web due to the fact that static builds of those binaries will be likely to misbehave. The hledger binary alone is sufficient for many people's workflow.  For more context on why GHC is not available in a form linked with the musl library, see this discussion on the alpine project: https://github.com/alpinelinux/aports/pull/4255 GHC used to be available on the ARM version of alpine but it was removed when they upgraded to GHC version 8.4.2 due to the difficulty in maintaining recent versions of GHC paired with musl. Regardless, glibc-linked hledger works great and works fine even on very old versions of glibc (I tested it with a version of glibc from year 2011, version 2.13 and it worked just fine).

Current build times that I've observed for each build are as follows:

x64: About 24 minutes to build all dependencies and the 3 hledger binaries

arm32v7: Around 5 hours (it can take a little less than 5 hours and it sometimes takes 5 and a half hours). This extended build time is due to the github action using qemu to emulate the ARM architecture. If github actions begins supporting native ARM builds at some later date, this may become much faster in the future. Many other CI providers are beginning to offer native ARM builds these days so github actions may do this soon to catch up. Right now, github actions allows self-hosted ARM runners but this would involve additional expense and automation, since we would have to pay for an ARM VM on the AWS cloud or somebody would have to maintain an "always up" ARM machine for building.


simonmichael:
Fantastic work @thecount2a. I will review and give more input when I can. I'm fine with a huge slow build if we don't have to run it often, so I'm planning to merge this and provide these as official binaries as you suggest.

simonmichael:
@thecount2a could you rebase this against master when you get a chance.

simonmichael:
For easier review now and in future, and a quicker merge: would you be able to rewrite and simplify this history at least a bit ? Eg rebase on master, avoid nuking the existing scripts, perhaps squash the minor fixups ? 

Could you also make these run on push to a special branch (ci-linux-static & ci-linux-static-arm32 ?), like linux/mac/windows.yml. I would maybe comment out the tag-based triggers for now, until I get the hang of this.

simonmichael:
Who maintains/controls changes to the new docker/* and softprop/ actions ?

thecount2a:
Yes, I'm happy to make these changes and rebase/cleanup history. Sorry about that. Hopefully I can get to it this evening.

The docker organization controls the docker github actions, so I feel that this is relatively safe. It is not "certified by github" but I don't think any of the github certified actions provides any supported approach for running docker buildx builds which we need for the ARM qemu approach to building. I could definitely reformulate the x64 static build to not use anything from docker if you would prefer that. I had it that way earlier in my development process. I just decided to make both of the static build actions similar to make the build process easy to follow and easy to compare to each other.

Softprop is managed by somebody by the name of Doug Tangren and it seems to be the preferred github action to use for automatically uploading release files. Github themselves used to maintain a github action for this but it fell unmaintained so github "archived" their action and updated their repo to point toward the third-party ones, since they were better maintained. Here is the original github one that was archived: https://github.com/actions/create-release The softprop action seems to be the most active one, has the most recent releases and the greatest number of maintainers. I can easily remove our usage of this github action if you prefer to push to a CI branch and generate artifacts that way anyway, then we don't have to worry about this. Then you will need to manually download them and upload the binaries to the actual release yourself but I assume this is your normal process anyway. I just wanted to automate it as much as possible during my testing.

Let me know if you want me to remove docker from the Intel Linux build, it shouldn't be difficult.

simonmichael:
Thanks Garret. No hurry.

Ordinarily the auto-upload would be good as a separate PR, but it's ok as is. The actions research is helpful. I hope we'll never find our CI is mining crypto or DDOS'ing hospitals or doing something sneakier. I want these conveniences, so I guess we'll hope that Docker and softprop are popular enough that such problems would be noticed quickly. 

Yes, if it's not too bothersome, and doesn't hurt maintainability too much, let's avoid docker except where its necessary - less complexity, less ways to fail.

thecount2a:
I was able to get all the history cleaned up (with some force-pushes), merge branch rebased against simonmichael/hledger, Linux build switched to use the non-docker approach, CI builds triggered by tags commented out and build triggered by special branch names added. Sorry about so many force-pushes. I don't usually push untested code so many times during my development process but since this is a unique type of development where I must push in order to test the github actions, I must cleanup messy git history by squashing commits and using force-pushes. Recent changes have all been tested on my branches. Linux x64 build took 24 minutes, Linux ARM build took 5 hours and 36 minutes. Thanks for taking a look at this!

simonmichael:
Yes, testing CI scripts is a pain. Thank you very much, looks great! I'll squash and merge.

simonmichael:
About the addition of softprops/action-gh-release - this looks super useful. I guess I'll find out, but is it right for multiple workflows to call this - will it create multiple releases, or will the first one create it and the others upload to that one ?

thecount2a:
In my experience, releases are created automatically at the moment a new tag is created and pushed. There are default files that are attached to the "auto-created" releases which are just .zip and .tar.gz source code bundles. action-gh-release will simply add files as new attachments to the release. You still have the opportunity to go in from the github.com interface and customize the release as usual by adding release text, adding or removing attached files and so on.

simonmichael:
Hi @thecount2a, some things I'm wondering about, in case you have any ideas:

- Why did this arm32 run fail ? https://github.com/simonmichael/hledger/actions/runs/997438576 

- Why did all my runs with the softprop release action seem to skip it, eg https://github.com/simonmichael/hledger/runs/2981249148 ? 

- Why did all my attempts to trigger on pushing tags fail to trigger, even after adopting your better patterns ? (I have added `workflow_dispatch` to all workflows so I can at least run them manually.)

thecount2a:
Unfortunately the arm32 build failed because the particular build server was heavily loaded and it took a full 6 hours to build, which is the limit. After building for 6 hours, GitHub kills builds. I had that happen to me one time but then the same build finished in under 6 hours at least half a dozen other times. I think it's just chance whether you happen to get assigned to a heavily loaded server or not.

I haven't had a chance to check into exactly why your tag pushes weren't working but when I click on your link, it says that particular workflow was triggered by a push to "master" which is supposed to be skipped since it's a pure master push. I had to specifically push tags using command-line git to get it to make releases and upload files. The blue highlighted part of the GitHub actions interface would always show me the tag name when it triggered based off of tag pushes.

simonmichael:
Thanks!

simonmichael:
A heads up, I'm doing a little renaming in these workflows (Static Binaries -> hledger-linux-static-x64, static-linux -> linux-static generally); hope it doesn't break your nextcloud flow.

thecount2a:
Nope, no problem at all

PR #1572, opened 2021-06-16
Add support for kakoune editor

crocket:
I tested this with hledger-ui. hledger-ui opened kakoune at the desired lines when I pressed `E`.

I don't yet know how to make hledger-ui open a file at the last line on kakoune.


simonmichael:
Sounds good to me. Thanks for your contribution!

PR #1574, opened 2021-06-16
Improve editor support

crocket:
I tested it with emacs(client), nano, vscode, kakoune, nvim, and vis.

### DIFF
--- a/hledger-ui/Hledger/UI/Editor.hs
+++ b/hledger-ui/Hledger/UI/Editor.hs
@@ -11,6 +11,7 @@ where
 import Control.Applicative ((<|>))
 import Data.List (intercalate)
 import Data.Maybe (catMaybes)
+import Data.Bifunctor (bimap)
 import Safe
 import System.Environment
 import System.Exit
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks! Could you make it the reason for the change a bit clearer in the commit message ? I'm not sure what it is.

crocket:
I just made the commit message more precise.

simonmichael:
The message looks quite informative now, but I'm still unclear -

> Documentation for editFileAtPositionCommand is made more
> precise.

Sounds good!

> This commit makes hledger-ui open an editor at the last line
> if the editor supports opening a file at the last line.
>
> It also makes hledger-ui open an editor without any argument
> if no information about cursor position is given.

I thought it did both of those already, did you find a problem ?

> In other words, editFileAtPositionCommand is made to behave
> with more precision.

What will we need to tell users in the release notes ?

> I tested it with emacs(client), nano, vscode, kakoune,
> nvim, and vis.

Excellent.

crocket:
> I thought it did both of those already, did you find a problem ?

`editFileAtPositionCommand` used to give `+` as an argument to editors.
It wasn't precise with arguments it passed to editors.

crocket:
I just improved commit message. Can you check it again?

simonmichael:
Looks great, thanks a lot!

Issue #1556, opened 2021-05-08
Don't write debug.log to $HOME when pressing `V` or `B` in hledger-ui.


crocket:
There is no option to disable writing debug.log to $HOME when I press `V` or `B` in hledger-ui.

simonmichael:
hledger-1.21 was writing `./debug.log` when you press V, thanks for the report. This is fixed in master.

Xitian9:
Sorry to resurect this, but there's another instance of this bug which arises when you press `g` in the transaction screen. I've put a PR for this in #1576.

simonmichael:
(Not actually a regression, the V key has being doing this since the feature was introduced.)

PR #1576, opened 2021-06-21
Don't write debug.log when reloading the journal in hledger-ui's transaction screen

Xitian9:
[Another instance of #1556.]

A couple of issues raised from this.

1. `plog` is no longer used anywhere in the code base. It's not clear to me what purpose it is supposed to server over `dbg`, and every instance of its use in the code caused some sort of bug. Should we just remove it?
2. `clearCostValue` is only used in one place: when reloading the journal file in the transaction screen. Why that code is implemented the way it is is very confusing to me (and I'm not the only one judging by the code comments), but there are some strange things in there likely to be bugs. In particular, `regenerateTransactions` uses the old `rspec`, not the new one that results from `clearCostValue`. If anybody knows more about what should happen here and what I should be looking for, it would help with debugging.
https://github.com/simonmichael/hledger/blob/61a3fbc0192d19348746e77850b4e3b3866e30f5/hledger-ui/Hledger/UI/TransactionScreen.hs#L165-L176

### DIFF
--- a/hledger-lib/Hledger/Utils/Debug.hs
+++ b/hledger-lib/Hledger/Utils/Debug.hs
@@ -87,9 +87,6 @@ module Hledger.Utils.Debug (
   ,dbg7IO
   ,dbg8IO
   ,dbg9IO
-  -- ** Trace to a file
-  ,plog
-  ,plogAt
   -- ** Trace the state of hledger parsers
   ,traceParse
   ,dbgparse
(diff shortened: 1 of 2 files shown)


Xitian9:
This now also clears up the second issue raised in #1577, and pressing `g` once is sufficient to clear both the Value and Cost settings.

simonmichael:
Re 1: yes plog seems redundant ("pretty log"), we can drop it.

Re 2: yes the state management in hledger-ui gets confusing and it's very easy for bugs to hide in there. Cleanup is welcome but we should tread/test extremely carefully here, at least until we figure out automated testing.

simonmichael:
(We can dig into that code some time but if it ain't yet broke or being replaced I won't.)

simonmichael:
Can you move the other two commits to their own PRs, they can be tackled separately.

Xitian9:
Second two commits moved to their own PR.

PR #1583, opened 2021-06-26
csv: convert datetimes with time zones to local date, mostly (WIP)

simonmichael:
Just sharing this.. it's an attempt to fix this CSV limitation I just noticed, dates from other timezones can be off by one:

fc364cd8c origin/master ;doc: csv: note a limitation with datetimes in other zones

I have abandoned it for now, because (in a non-UTC timezone) dates with no time/timezone all get adjusted, as they are parsed as 00:00:00 UTC. Perhaps with a bit more work this could be useful (somehow use local timezone as the default when parsing ? Convert to local date only when timezone-related strptime codes (%z, %ez, %Z, %eZ) appear in date-format  ?

Related: https://github.com/haskell/time/issues/172. ~This PR doesn't handle the daylight-savings inaccuracy mentioned there, but I think it would still improve the date in most cases (off by one errors would still be possible, but rather rare for most people). Adding the `tz` package as a dependency could allow "perfect" accuracy, but only for past dates, as far as I understand it.~ <- a mistake ? Hopefully.

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -60,7 +60,7 @@ import qualified Data.Text.Encoding as T
 import qualified Data.Text.IO as T
 import qualified Data.Text.Lazy as TL
 import qualified Data.Text.Lazy.Builder as TB
-import Data.Time.Calendar (Day)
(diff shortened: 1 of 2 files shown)


simonmichael:

Issue #806, opened 2018-06-08
Info files can't be added to the Info root


DamienCassou:
For hledger.info to be added to the Info root node, something like this must be done on all info files:

```diff
diff -u /home/cassou/Documents/projects/ledger/hledger/hledger-1.9.1/hledger.info.bak /home/cassou/Documents/projects/ledger/hledger/hledger-1.9.1/hledger.info
--- a/hledger.info.bak	2018-06-08 16:35:36.244331263 +0200
+++ b/hledger.info	2018-06-08 16:29:09.730763422 +0200
@@ -1,5 +1,10 @@
 This is hledger.info, produced by makeinfo version 6.5 from stdin.
 
+INFO-DIR-SECTION User Applications
+START-INFO-DIR-ENTRY
+* HLedger: (hledger).           Command-Line Accounting
+END-INFO-DIR-ENTRY
+
 
 File: hledger.info,  Node: Top,  Next: EXAMPLES,  Up: (dir)

```

DamienCassou:
I've been bitten by this again yesterday and wanted to report an issue. GitHub found out I already did 3 years ago :-). Can you please tell me what I should modify where? I will be happy to do a PR.

simonmichael:
Thanks! 

The `infomanuals` recipe in Shake.hs is where info manuals are generated. You could add the extra text there.

But perhaps it's easier to add it in the manual source files - *.m4.md - wrapped in the `_info_` macro, similar to the `_web_` and `_man_` calls you'll see there.

DamienCassou:
I'm currently blocked by #1565.

DamienCassou:
I tried adding the following to hledger.m4.md:

```m4
_info_({{
INFO-DIR-SECTION User Applications
START-INFO-DIR-ENTRY
* HLedger: (hledger).           Command-Line Accounting
END-INFO-DIR-ENTRY
}})
```

The end result is not satisfying though:

```
-DIR-SECTION User Applications START-DIR-ENTRY * HLedger: (hledger).
Command-Line Accounting END-DIR-ENTRY
```

As you can see:
1. all lines got merged
2. the `INFO` word got eaten on every lines where it appears

I don't have plans to learn m4 anytime soon :-).

simonmichael:
[...]

PR #1585, opened 2021-06-26
Add dir section to info file

DamienCassou:
This should solve #806. Thanks to @simonmichael for all the help.

### DIFF
--- a/hledger-ui/hledger-ui.info
+++ b/hledger-ui/hledger-ui.info
@@ -6,6 +6,11 @@ File: hledger-ui.info,  Node: Top,  Next: OPTIONS,  Up: (dir)
 hledger-ui(1)
 *************
 
+INFO-DIR-SECTION User Applications
+START-INFO-DIR-ENTRY
+* HLedger-UI: (hledger-ui).  Command-Line Accounting (TUI)
+END-INFO-DIR-ENTRY
+
 hledger-ui is a terminal interface (TUI) for the hledger accounting
 tool.  This manual is for hledger-ui 1.21.99.
 
--- a/hledger-ui/hledger-ui.m4.md
+++ b/hledger-ui/hledger-ui.m4.md
@@ -2,6 +2,11 @@
 % _author_
 % _monthyear_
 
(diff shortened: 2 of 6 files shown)


simonmichael:
Next quibble. "Command line accounting" is t really accurate for ui and web. Perhaps better to use one of the official blurbs for the respective tool, that could be more informative and reduced the number of unique doc strings to keep updated.

simonmichael:
Also, in what situations / where does one see the benefits of this ? 

If I view the info file, eg with emacs -> dired ->`I`, or with `hledger help`, I see the new metadata inline. See screenshot - old on the left, new on the right. Is this normal ?

Screen Shot 2021-06-27 at 2 50 01 PM

simonmichael:
I hope you don't mind, I went ahead and merged this and worked on it some more. Wow.. Info is hard. I think we are in a better place now, but I welcome any review on the above commit.

Now we finally can see, and are more likely to see, hledger/hledger-ui/hledger-web appearing Info's top-level directory. Thanks for pushing this along!

simonmichael:
Summarising:

Users or packagers can now add the hledger manuals to Info's top-level table of contents using install-info. Eg:
```
$ install-info hledger/hledger.info /usr/local/share/info/dir
$ install-info hledger-ui/hledger-ui.info /usr/local/share/info/dir
$ install-info hledger-web/hledger-web.info /usr/local/share/info/dir
$ info
```

Developers can do it in another way: by adding the package directories to INFOPATH (and running Shake once to generate dir files). Eg:
```
$ ./Shake.hs
$ ./Shake manuals  # or just infomanuals
$ export INFOPATH=~/src/hledger/hledger:~/src/hledger/hledger-ui:~/src/hledger/hledger-web:$INFOPATH
$ info
```

simonmichael:
A few more fixes: developers who want to see the most up to date Info directory links from the hledger source tree (me) can just add the hledger source to their `INFOPATH` (no need to run Shake); and the links now work in the Emacs info viewer too.

PR #1586, opened 2021-06-28
;update manuals

DamienCassou:
Remove duplicated DIR section in info manuals


simonmichael:
Thanks! I think I took care of this already.

PR #1588, opened 2021-07-01
In hledger-ui TransactionScreen, set the current transaction based on RegisterScreen.

Xitian9:
This fixes a bug where, when reloading journal files by pressing `g` or using the `--watch` command, the `cost_` and `value_` flags would not be reset without pressing `g` a second time.

### DIFF
--- a/hledger-ui/Hledger/UI/TransactionScreen.hs
+++ b/hledger-ui/Hledger/UI/TransactionScreen.hs
@@ -173,11 +173,7 @@ tsHandle ui@UIState{aScreen=s@TransactionScreen{tsTransaction=(i,t)
           ej <- liftIO $ journalReload copts
           case ej of
             Left err -> continue $ screenEnter d errorScreen{esError=err} ui
-            Right j' -> do
-              continue $
-                regenerateScreens j' d $
-                clearCostValue $
-                ui
+            Right j' -> continue $ regenerateScreens j' d ui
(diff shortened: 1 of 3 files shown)


simonmichael:
I think ideally they wouldn't be reset at all, could we do that instead ? I looked for the doc saying they are reset but haven't found it.

simonmichael:
Ack, I see #1589 changes that. So should both of these be merged ?

hledger-ui/Hledger/UI/TransactionScreen.hs
```diff
@@ -43,16 +45,20 @@ transactionScreen = TransactionScreen{
...
-  --       ++ "?"
-  --       ++" and first commodity is")
-  --       (acommodity$head$amounts$pamount$head$tpostings$snd$tsTransaction)
-  --      `seq`
-  ui
+    ui{aScreen=s{tsTransaction=(i',t'),tsTransactions=nts'}}
+  where
+    i' = maybe 0 (toInteger . (+1)) . elemIndex t' $ map snd nts'
+    -- If the previous screen was RegisterScreen, use the listed and selected items as
+    -- the transactions. Otherwise, use the provided transaction and list.
```
simonmichael, line 55: This seem more complicated than before - now there's two ways to initialise a transaction screen, one of them unused.

But if we no longer need regenerateTransactions, that's great, and I'm ready to merge this.

simonmichael:
And actually I think they conflict. Want to merge #1589 into this one ?

Xitian9:
#1589 merged.

simonmichael:
Thanks!

Issue #1577, opened 2021-06-21
Cannot toggle cost in hledger-ui transaction screen


Xitian9:
Well, you can, but it's difficult. Consider the following journal:

```
$ cat t.j
D 1. C

P 2021-06-01 A  1 C
P 2021-06-01 B  2 C

2021-06-01 Test
  (a)  1 A @ 1 B
```

There are four possible outputs I'd like to see based on cost and valuation:
1. No cost and no valuation: 1 A
2. No cost and valuation: 1 C
3. Cost and no valuation: 1 B
4. Cost and valuation: 2 C

- Open `hledger-ui -f t.j` and move to the transaction screen (press -> twice). It should display 1 A as expected.
- Pressing `V` toggles valuation, switching between 1 A and 1 C.
- Pressing `B` does not toggle cost, it remains at 1 A.

However, pressing `g` to reload the file can allow you to work around these, in unintuitive ways:
- Pressing `B` then `g` will switch to cost, showing 1 B. Pressing `V` at this point will toggle between 1 B and 2 C.
- Pressing `g` twice will get rid of valuation, but have no effect on cost.

To conclude:
- `V` works as expected.
-  `B` can be made to work if you press `g` afterwards.
-  Pressing `g` twice will unset valuation, but have no effect on toggle.

It's clear that pressing `B` will set the appropriate cost option, but the journal won't be updated until you reload it with `g`. `B` should work without updating the journal. I'm not sure what effect `g` is supposed to have on valuation/cost. Is it supposed to clear both valuation and cost? Have no effect on either? Clear only valuation but not cost? I can try to debug this when I know what the correct behaviour is.

Xitian9:
git bisect says this one is my fault: 53611be6e

Xitian9:
I think there are two issues here: B not working, and incorrect behaviour of `g`. I think this issue should be restricted to the former; the second can be dealt with elsewhere.

simonmichael:
`g` is supposed to just reload the data, not change your viewing mode - it shouldn't affect the B or V state.

simonmichael:
Great catch, thanks!

Xitian9:
Okay, then the bug is in the documentation: the code which `g` calls says that it resets both the cost and valuation mode, and actually did reset the value_ mode.

simonmichael:
Ack.. hmm. We'll discuss/investigate more elsewhere, as you rightly said.

Xitian9:
[...]

PR #1589, opened 2021-07-01
ui: Do not reset cost and value flags on reload.

Xitian9:
Discussion in #1577, indicates that cost and valuation flags should not be reset when reloading the data. This implements that.


Xitian9:
Merged into #1588.

PR #1591, opened 2021-07-02
lib: Remove aismultiplier from Amount.

Xitian9:
In Amount, aismultiplier is a boolean flag that will always be False,
except for in TMPostingRules, where it indicates whether the posting
rule is a multiplier. It is therefore unnecessary in the vast majority
of cases. This PR pulls this flag out of Amount and puts it into
TMPostingRule, so it is only kept around when necessary.

This changes the parsing of journals somewhat. Previously you could
include an `*` before an amount anywhere in a Journal, and it would
happily parse and set the aismultiplier flag true. This will now fail
with a parse error: `*` is now only acceptable before an amount within an
auto posting rule.

This changes the JSON output of Amount, as it will no longer include
aismultiplier.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -218,7 +218,7 @@ instance Num Amount where
 
 -- | The empty simple amount.
 amount, nullamt :: Amount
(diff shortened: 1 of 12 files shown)


simonmichael:
Much better. 👍️

Xitian9:
Since there's been a change to the JSON serialisation, we should make sure this is in hledger-demo so that @real-dam can make sure it doesn't break anything.

Issue #1065, opened 2019-07-12
Remove generated build artifacts from git repository


alerque:
This is more of a note to self rather than a request for anybody else to do this work so feel free to assign this issue to me.

There is a ton of duplicated content committed to the repository in files that are just format conversions of each-other (largely using Pandoc). These should ideally be stripped and the build process should make them as a matter of course. This may include generating a source tarball generator that includes these initial build steps (in addition to one that has fully built binaries) or setting up CI artifacts somewhere predictable with the extra formats available for packagers that don't want to have the whole build toolchain.

schoettl:
But please do not remove the (Make-generated) shell completion scripts (under `shell-completion`).

> I propose that the Makefile is not run at every built but rather manually when the CLI changes.

Because it takes a noticable time to generate the completions. hledger is executed with `--help` multiple times (e.g. for all sub-commands).

Also, they are not generated from the root Makefile.

alerque:
@schoettl Please excuse me if I'm being dense, but in my experience with make (which is actually quite a bit) the entire goal is to have a chain of rules that understand exactly when something needs to be regenerated (and when it doesn't) based on what source files have changed vs. what objects are created in the end.

I don't understand either why having or not having generated artifacts committed to the git repository would have any effect at all on the situation your describing.

schoettl:
The only make rule, I can think of, that takes dependencies into account, is "Generate the completion scripts whenever hledger binary has changed."

But actually (as quoted above), the scripts should only be generated, when the CLI changes (i.e. the options and sub-commands). Because this step takes so much time, it should not be executed on every hledger binary change.

For me, generating the completion scripts is more like a version bump.

And `shell-completion/Makefile` also uses external programs (hledger, and GNU tools, that are maybe not POSIX compliant or installed). So if you really have to delete the *all* artifacts, than shell completion needs much attention!

Otherwise, which make target do you propose?

alerque:
@schoettl As best I can make out I think you're conflating two different issues. I don't think what I'm proposing here will have a substantial impact on the issue you are raising. I'm not exactly sure what piece you're missing, but maybe some of these comments will help sort out what is what.

* Having generated artifacts in the git repository or not will not change what needs to be built when. The rules for what files need to be regenerated when other files change would not be affected at all by this change. Even if a prebuilt artifact exists in the repository now, it still goes out of date and needs to be regenerated when other files change. The build process does not take into account whether a file is tracked by the repository or not, only its state on disk.

[...]

PR #1592, opened 2021-07-04
[WIP] Distribute shell completions / add shell completions to release process

schoettl:
As @simonmichael suggested in #1067, I added the shell completions to the list in `package.yaml` to have it included in the tar.gz archive.

Then I did `( cd hledger && cabal sdist )` to generate the .tar.gz.

The problem is that the completion files are outside of the hledger directory, so

```
$ tar -tf dist-newstyle/sdist/hledger-1.22.tar.gz |grep complet
tar: Removing leading `hledger-1.22/../' from member names
hledger-1.22/../shell-completion/
hledger-1.22/../shell-completion/hledger-completion.bash
```

How to best fix this?

The other thing is that the shell completions are not yet automatically generated in the build process (see #1065 ). After @alerque 's refactoring, the generation of the completions is very fast, so it could be added to the main Makefile's install and build target?

-----

## TODO

- [x] package shell completion script into the sdist .tar.gz
- [ ] hook shell completion generation in the build process
- [ ] add shell completion bash file to .gitignore and delete it from git repo
- [ ] provide an install script or a Makefile to tell distributors how/where to install the completion script?


simonmichael:
Thanks! Indeed these are not integrated in the release process, I'll fix that somehow.

hledger/hledger.cabal
```diff
@@ -45,6 +45,7 @@ extra-source-files:
     embeddedfiles/hledger-web.1
     embeddedfiles/hledger-web.txt
     embeddedfiles/hledger-web.info
+    ../shell-completion/hledger-completion.bash
```
simonmichael, line 48: I am suspicious, I bet .. paths are problematic here.

simonmichael: There are completions for hledger-ui and hledger-web as well, I guess ?

Would it make sense to put each tool's completions in the respective package, or is it better to keep them all together ?

simonmichael: Yes, cabal check warns about such paths and hackage rejects them.

schoettl: > Yes, cabal check warns about such paths and hackage rejects them.

Thanks for checking.

> There are completions for hledger-ui and hledger-web as well, I guess ?

I just looked into `hledeger-completion.bash`. It provides basic completion for "extensions" hledger-ui and hledger-web. But it's one bash completion script and Make process for all hledger binaries.

> Would it make sense to put each tool's completions in the respective package, or is it better to keep them all together ?

I guess we could put the whole folder `shell-completion` into the hledger package. @alerque, what do you think?

simonmichael: FYI I'll do a minor release soon.

alerque: @schoettl I guess that depends on what "package" you are talking about? A source dist package? If so then yes.

schoettl: @alerque I'm talking about a haskell package, i.e. a folder containing a package.yaml. So I would move the shell-completion folder into the hledger folder to get it packaged in hledger-1.22.tar.gz.

alerque: Then yes — given that tarball is the source package used by at least some distros to compile their packaging, yes it needs to have the whole kit & kaboodle to make completions so that they make their way to end users.

schoettl:
```
$ make clean
$ ( cd hledger && cabal sdist )
$ tar -tf dist-newstyle/sdist/hledger-1.22.tar.gz |grep complet
hledger-1.22/shell-completion/
hledger-1.22/shell-completion/hledger-completion.bash
```

Now this looks OK.

But the completions are still not generated automatically.

I'm hesitating to touch the main Makefile because I don't know the build process.

My idea was to insert
```
.PHONY: shell-completions
shell-completions:
	make -C hledger/shell-completion/ clean-all all
```
in the Makefile. But a) the release process depends on shell-completions and b) shell-completions depends on a compiled hledger binary.

simonmichael:
Great! I'm happy to look at the process integration, it is in flux anyway.

simonmichael:
Merged as https://github.com/simonmichael/hledger/compare/861c34484ed93b43f84e08aa4c5cda948af8e12d...9701e19169c6615742590d16a358d1e05e583ffc (commit messages reworded to satisfy new conventions, `make shellcompletions` added). Thanks!

Issue #1162, opened 2020-01-02
command line option to override display precision in reports


Whil-:
Hi,

I've been searching the web and the documentation for some time now but cannot find any mention of rounding options in balance or register reports. At least in the balance report I think it can be valuable, but in register reports as well. The use case is to remove unnecessary clutter. For example a balance report for monthly expenses. I don't need all decimals in the report, even though all transactions in my ledger is written at that precision.

Have I missed something in the documentation? Or am I looking for something that still doesn't exist?

I've seen how commodity directives can modify the output precision but find modifying the global state of commodities in my ledger files is a bit hard to swallow when I'm simply wishing for some of my reports to be less cluttered and easier on the eye.

simonmichael:
Commodity directives are the way - we assume there’s one standard precision you’d like to see always. 

But I agree it might be nice to override that with an easy command line flag once in a while..

Whil-:
Ah ok, got it. A shame though ;-)

Your decision if you want to keep this wish as open or not. It's still open from my perspective. But I won't and can't "demand" it, as I'm not a paying customer. Thanks, and take care!

Simaex:
I am experiencing connected issue. I have set precision to one decimal using journal-wide commodity directive but if I enter commodity value display precision in reports and hledger-web is changed according to number of decimals used in statment.
P 2020/1/31 EUR 1.1034
Also I have used the next format with the same result - values with unnecessary 4 decimals.
P  2020/1/31 EUR 1.1034$
P  2020/1/31 EUR $1.1034

P  2020/1/31 EUR 1.10337 results in 5 decimals in reports and hledger-web interface 

Please let me know: is it an intended behaviour or do I miss something obvious.

simonmichael:
@Simaex this is not clear enough, would you mind opening as new issue with a reproducible example, or explain to me on #hledger. Thanks!

simonmichael:
@Whil- I'm leaving this open as I think a command line flag to override the display precision (for all commodities, and maybe selected commodities) is a nice-to-have feature.

I don't quite understand "modifying the global state of commodities in my ledger files is a bit hard to swallow when I'm simply wishing for some of my reports to be less cluttered and easier on the eye.", could you explain more ? Why don't you set up your files to display the right number of decimal places always ?

Whil-:
Hi again!

What I mean is that I might want the default one way, but for certain reports to  override the default. Having to modify the global config and then reset it after the report has been created isn't optimal.

[...]

PR #1593, opened 2021-07-04
Added output-commodity-style option

alang76:
First attempt at adding an extra command line option to enable overriding output commodity styles. This should also cover setting precision as requested here: https://github.com/simonmichael/hledger/issues/1162

### DIFF
--- a/hledger/test/cli/commodity-style.test
+++ b/hledger/test/cli/commodity-style.test
@@ -1,0 +1,49 @@
+# Test whether only the style without a symbol is changed
+<
+2021-07-09 no symbol
+    (a)   1234
+
+2021-07-09 Euro
+    (a)    EUR 1.234,56
+
+2021-07-09 Dollar
+    (a)      $ 1,234.56
+
+$ hledger -f- print -c '10 00'
+>
+2021-07-09 no symbol
+    (a)           12 34
+
+2021-07-09 Euro
+    (a)    EUR 1.234,56
+
+2021-07-09 Dollar
+    (a)      $ 1,234.56
+
+>= 0
+# Test whether setting the style of multiple symbols work
+<
+2021-07-09 Euro
+    (a)    EUR 1,234.56
+
+2021-07-09 Dollar
+    (a)      $ 1.234,56
+$ hledger -f- print -c 'EUR 1.000,00' -c '$ 1,000.00'
+>
+2021-07-09 Euro
+    (a)    EUR 1.234,56
+
+2021-07-09 Dollar
+    (a)      $ 1,234.56
+
+>= 0
+# When setting the same symbol multiple times, the last one is in effect
+<
+2021-07-09 Euro
+    (a)    EUR 1234
+$ hledger -f- print -c 'EUR 1.000,00' -c 'EUR 1,000.00'
+>
+2021-07-09 Euro
+    (a)    EUR 1,234.00
+
+>= 0
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -730,7 +730,7 @@ journalModifyTransactions d j =
 -- | Check any balance assertions in the journal and return an error message
 -- if any of them fail (or if the transaction balancing they require fails).
 journalCheckBalanceAssertions :: Journal -> Maybe String
-journalCheckBalanceAssertions = either Just (const Nothing) . journalBalanceTransactions def
+journalCheckBalanceAssertions = either Just (const Nothing) . journalBalanceTransactions balancingOpts
 
 -- "Transaction balancing", including: inferring missing amounts,
 -- applying balance assignments, checking transaction balancedness,
@@ -1415,7 +1415,7 @@ journalApplyAliases aliases j =
 --     liabilities:debts  $1
 --     assets:bank:checking
 --
-Right samplejournal = journalBalanceTransactions def $
+Right samplejournal = journalBalanceTransactions balancingOpts $
          nulljournal
          {jtxns = [
            txnTieKnot $ Transaction {
@@ -1558,7 +1558,7 @@ tests_Journal = tests "Journal" [
   ,tests "journalBalanceTransactions" [
 
      test "balance-assignment" $ do
-      let ej = journalBalanceTransactions def $
+      let ej = journalBalanceTransactions balancingOpts $
             --2019/01/01
             --  (a)            = 1
             nulljournal{ jtxns = [
@@ -1569,7 +1569,7 @@ tests_Journal = tests "Journal" [
       (jtxns j & head & tpostings & head & pamount & amountsRaw) @?= [num 1]
 
     ,test "same-day-1" $ do
-      assertRight $ journalBalanceTransactions def $
+      assertRight $ journalBalanceTransactions balancingOpts $
             --2019/01/01
             --  (a)            = 1
             --2019/01/01
@@ -1580,7 +1580,7 @@ tests_Journal = tests "Journal" [
             ]}
 
     ,test "same-day-2" $ do
-      assertRight $ journalBalanceTransactions def $
+      assertRight $ journalBalanceTransactions balancingOpts $
             --2019/01/01
             --    (a)                  2 = 2
             --2019/01/01
@@ -1598,7 +1598,7 @@ tests_Journal = tests "Journal" [
             ]}
 
     ,test "out-of-order" $ do
-      assertRight $ journalBalanceTransactions def $
+      assertRight $ journalBalanceTransactions balancingOpts $
             --2019/1/2
             --  (a)    1 = 2
             --2019/1/1
--- a/hledger-lib/Hledger/Data/Transaction.hs
+++ b/hledger-lib/Hledger/Data/Transaction.hs
@@ -360,8 +360,6 @@ data BalancingOpts = BalancingOpts
   , commodity_styles_  :: Maybe (M.Map CommoditySymbol AmountStyle)  -- ^ commodity display styles
   } deriving (Show)
 
-instance Default BalancingOpts where def = balancingOpts
-
 balancingOpts :: BalancingOpts
 balancingOpts = BalancingOpts
   { ignore_assertions_ = False
@@ -856,7 +854,7 @@ tests_Transaction =
     , tests "balanceTransaction" [
          test "detect unbalanced entry, sign error" $
           assertLeft
-            (balanceTransaction def
+            (balanceTransaction balancingOpts
                (Transaction
                   0
                   ""
@@ -871,7 +869,7 @@ tests_Transaction =
                   [posting {paccount = "a", pamount = mixedAmount (usd 1)}, posting {paccount = "b", pamount = mixedAmount (usd 1)}]))
         ,test "detect unbalanced entry, multiple missing amounts" $
           assertLeft $
-             balanceTransaction def
+             balanceTransaction balancingOpts
                (Transaction
                   0
                   ""
@@ -888,7 +886,7 @@ tests_Transaction =
                   ])
         ,test "one missing amount is inferred" $
           (pamount . last . tpostings <$>
-           balanceTransaction def
+           balanceTransaction balancingOpts
              (Transaction
                 0
                 ""
@@ -904,7 +902,7 @@ tests_Transaction =
           Right (mixedAmount $ usd (-1))
         ,test "conversion price is inferred" $
           (pamount . head . tpostings <$>
-           balanceTransaction def
+           balanceTransaction balancingOpts
              (Transaction
                 0
                 ""
@@ -922,7 +920,7 @@ tests_Transaction =
           Right (mixedAmount $ usd 1.35 @@ eur 1)
         ,test "balanceTransaction balances based on cost if there are unit prices" $
           assertRight $
-          balanceTransaction def
+          balanceTransaction balancingOpts
             (Transaction
                0
                ""
@@ -939,7 +937,7 @@ tests_Transaction =
                ])
         ,test "balanceTransaction balances based on cost if there are total prices" $
           assertRight $
-          balanceTransaction def
+          balanceTransaction balancingOpts
             (Transaction
                0
                ""
@@ -958,7 +956,7 @@ tests_Transaction =
     , tests "isTransactionBalanced" [
          test "detect balanced" $
           assertBool "" $
-          isTransactionBalanced def $
+          isTransactionBalanced balancingOpts $
           Transaction
             0
             ""
@@ -976,7 +974,7 @@ tests_Transaction =
         ,test "detect unbalanced" $
           assertBool "" $
           not $
-          isTransactionBalanced def $
+          isTransactionBalanced balancingOpts $
           Transaction
             0
             ""
@@ -994,7 +992,7 @@ tests_Transaction =
         ,test "detect unbalanced, one posting" $
           assertBool "" $
           not $
-          isTransactionBalanced def $
+          isTransactionBalanced balancingOpts $
           Transaction
             0
             ""
@@ -1009,7 +1007,7 @@ tests_Transaction =
             [posting {paccount = "b", pamount = mixedAmount (usd 1.00)}]
         ,test "one zero posting is considered balanced for now" $
           assertBool "" $
-          isTransactionBalanced def $
+          isTransactionBalanced balancingOpts $
           Transaction
             0
             ""
@@ -1024,7 +1022,7 @@ tests_Transaction =
             [posting {paccount = "b", pamount = mixedAmount (usd 0)}]
         ,test "virtual postings don't need to balance" $
           assertBool "" $
-          isTransactionBalanced def $
+          isTransactionBalanced balancingOpts $
           Transaction
             0
             ""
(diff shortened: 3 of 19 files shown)


Xitian9:
I don't think there's a need to create a new `OutputCommodityStyleOpts` datatype which is stored in `CliOpts`. This can just be put in the existing `commodity_styles_` option in `BalancingOpts` (in `InputOpts`, in `CliOpts`).

simonmichael:
Thanks for the PR.

Does it make sense to put this in "BalancingOpts" though ? I'd rather find the most logical place to put it.

simonmichael:
(Renaming things might be enough to make that logical place ?)

Xitian9:
It could very easily be pulled out somewhere else. I created `BalancingOpts` just because there were two Boolean opts being passed to the balancing functions, with lots of possibility for conclusion. I bundled `commodity_styles_` with those because it was always used alongside them, but could be pulled up to `InputOpts` (or higher).

alang76:
What is a little confusing is that `output_commodity_styles_` is essentially an output option, more akin to output-file and output-format. If we would pull it out from `InputOpts`, then we would have to pass the entire CliOpts to the balancing functions, just so it could use this other field which happens to share the same type. Could it be an alternative to wrap the balancing booleans with the needed styles map before the balancing functions are called because it's not currently specified as a command line parameter anyway?

simonmichael:
Sorry, I haven't fully re-grokked all our options structures but @alang76's approach does seem to make sense to me. Unless there's a way to simplify further, which of course I'd be interested in.

Some UI-ish thoughts. 

`-c` is a valuable short flag but I agree with using it for this.

The flag name is long. I was thinking of consistent "--output-X" options, but we have many other options which also affect output. How about just `--commodity-style` ? Shorter and makes the `-c` more mnemonic. (But read on..)

Out of scope for this PR, but perhaps related: we would also like to be able to rewrite commodity symbols ("commodity aliases"). Simplest would be an independent option, say --commodity-alias, similar to --alias for accounts. Perhaps more convenient for users though, should it be combined it with the commodity style option ? As in, say,
```
-c/--commodity-style=[SYMBOL=](AMOUNT|SYMBOL)
```
Examples:
1. `-c 'USD 1,000.00'`      set the style for USD, as above
2. `-c '$=USD 1,000.00'`   rewrite $ to USD and set the style
3. `-c '$=USD'`   rewrite $ to USD (and use the style of USD if any, otherwise keep the style of $)
4. `-c '$='`   remove $ symbols (and use the style of "" if any, otherwise keep the style of $)
5. `-c '=$'`   add $ symbol to no-symbol amounts (and use the style of $ if any, otherwise keep the style of "")
```
Even a simpler form of this could be quite useful. If we did combine these functionalities, we might want to revisit the "--commodity-style" long flag name ? Some possibilities:

- `--commodity-style`
- `--commodity-display`
- `--commodity-as`
- `--commodity`

**Comment by simonmichael** [comment]
This is incomplete of course; I forgot about the plan of setting style for all commodities when there's no commodity symbol. We'd need to drop that or integrate with that somehow.

**Comment by simonmichael** [comment]
Testing: how come this fails ?
```
ghci> :main -f examples/sample.journal print --output-commodity-style='EUR1.000,00'
*** Exception: ParseErrorBundle {bundleErrors = TrivialError 12 (Just (Tokens ('\'' :| ""))) (fromList [Tokens (' ' :| ""),Label ('d' :| "igit"),Label ('e' :| "xponent"),Label ('l' :| "edger-style lot date"),Label ('l' :| "edger-style lot price"),Label ('t' :| "ransaction price"),EndOfInput]) :| [], bundlePosState = PosState {pstateInput = "'EUR1.000,00'", pstateOffset = 0, pstateSourcePos = SourcePos {sourceName = "", sourceLine = Pos 1, sourceColumn = Pos 1}, pstateTabWidth = Pos 8, pstateLinePrefix = ""}}
```
Error message: instead of 
```
ghci> :main -f examples/sample.journal print -c 'test'  
*** Exception: ParseErrorBundle {bundleErrors = TrivialError 6 (Just EndOfInput) (fromList [Tokens ('+' :| ""),Tokens ('-' :| ""),Label ('n' :| "umber")]) :| [], bundlePosState = PosState {pstateInput = "'test'", pstateOffset = 0, pstateSourcePos = SourcePos {sourceName = "", sourceLine = Pos 1, sourceColumn = Pos 1}, pstateTabWidth = Pos 8, pstateLinePrefix = ""}}
```
it would be good to show something like
```
--output-commodity-style could not parse "test", please use amount syntax
https://hledger.org/hledger.html#amounts
```

**Comment by simonmichael** [comment]
We'll want a test or two; either unit or functional, the latter will be easier probably.

**Comment by simonmichael** [comment]
And (as last step) we'll want to follow the [new commit conventions scheme](https://hledger.org/CONTRIBUTING.html#commit-messages), eg adding the `feat: ` prefix.

**Comment by Xitian9** [comment]
> What is a little confusing is that `output_commodity_styles_` is essentially an output option, more akin to `output_file_` and `output_format_` If we would pull it out from `InputOpts`, then we would have to pass the entire `CliOpts` to the balancing functions, just so it could use this other field which happens to share the same type. 

I agree that where the styles map lives is a bit confusing. It does seem more like an output option than an input option, but it plays two important roles in InputOpts:
1. It styles any inferred amounts (i.e. amounts left blank in the journal)
2. It determines whether amounts look zero (because we consider them balanced if they look zero; they don't need to actually be zero to arbitrary precision).

> Could it be an alternative to wrap the balancing booleans with the needed styles map before the balancing functions are called because the balancing styles map is not currently specified as a command line parameter anyway?

I don't think I fully understand how this differs from the current situation. Could you give an example?

**Comment by alang76** [comment]
Thanks @simonmichael  and @Xitian9 for your comments. I need some more time to work on the recommendations. Some upfront comments:

@simonmichael:

> UI-ish thoughts

Commodity-style looks better indeed. I like the idea of combining it the optional alias. Agree with out of scope for this PR.

> Testing: how come this fails ?

Regarding the first error, is ghci doing something funny with the quotes? The compiled binary doesn't have any issues it seems.

> Error message: instead of

I'll add a proper error message. Since you can pass the option multiple times to supply the style for multiple commodities, I guess it would be nice to list all the problematic ones

> We'll want a test or two; either unit or functional, the latter will be easier probably.

Agreed.

@Xitian9:

> It does seem more like an output option than an input option, but it plays two important roles in InputOpts:

When mentioning the output option, I was more referring to the new option for overriding the style (ie the one added in this PR). I agree that the location of the original one inside `BalancingOpts` was probably fine to be placed there. My point was that there appear  to be two different use cases: one serving as type to hold inference results in the balancing function and the other for overriding the output styles. If both use cases consumed the same option value from `CliOpts` that could be a source of confusion.

> I don't think I fully understand how this differs from the current situation. Could you give an example?

I was suggesting that the potential issue of having two use cases for the same option as mentioned above might be solved by removing the one from balancing opts altogether from the CliOpts, since it is not set from the command line at all if I am correct? If its only there to support the inference process in the balancing function, could it not be passed along to that function at a later stage, after command options have been parsed? On the other hand I guess there is no real downside to having multiple options with different purposes, which just happen to have the same type. Or maybe I'm not fully understanding :)

**Comment by simonmichael** [comment]
Oh you're probably right about ghci quoting, thanks. 

Good idea, to show the option value that failed to parse. Probably just showing the first failure is enough, that's our usual MO.

**Comment by simonmichael** [comment]
The new "check commit messages" CI step always fails on force pushes I think. A regular push of a new commit might get CI working.

The changes are looking good. We should figure out a good place to document the new option

**Comment by simonmichael** [comment]
A new section under hledger.m4.md -> OUTPUT, perhaps

**Comment by alang76** [comment]
> A regular push of a new commit might get CI working.

Ah right was wondering about that.

> A new section under hledger.m4.md -> OUTPUT, perhaps

Sounds good, let me try and propose a description.

@Xitian9, in the end I figured the semantics of the original parameter was pretty much the same as you suggested. It's pulled out at InputOpts level now.

**Comment by Xitian9** [comment]
Sounds great! At this point I think `BalancingOpts` can just be eliminated and its two remaining booleans put into `InputOpts`. If you want to do that you're welcome to, otherwise I'll submit a PR once this one has been merged.

In `CliOptions.hs` some of the fields of `CliOpts`, its default declaration, and the record update in `rawOptsToCliOpts` have gained some extra spacing that's no longer necessary.

**Comment by simonmichael** [comment]
CI should be ok with force pushes now.

Where/when to mention this in docs is a little tricky. I'm thinking

## OUTPUT

### Commodity styles

*1. new section, describing --commodity-style option as briefly as we can get away with and linking to...*

## JOURNAL FORMAT

### Amounts

#### Commodity display style

*2. At end, mention --commodity-style for overriding, link to 1*

### Declaring commodities

*3. At end, mention --commodity-style for overriding, link to 1*

I can add this if you prefer.

**Comment by alang76** [comment]
@Xitian9:  Balancing opts is gone now. Does it look OK? Was a bit unsure about the location of the defaults.

@simonmichael: Feel free to change the text to be more in line with the general style

**Comment by Xitian9** [comment]
@alang76 Sorry, by flattening `BalancingOpts` into `InputOpts`, I meant that `InputOpts` should be passed around wherever `BalancingOpts` previously were. Passing two separate booleans as arguments to the balancing functions will give us a bad case of boolean blindness, especially when the default values of each are different (e.g. what does `journalBalanceTransactions True False` do, without looking up the definitions in code?).

I would just get rid of `defaultIgnoreAssertions` and `defaultInferPrices`. It's sufficient to just put the defaults in `defaultInputOpts`.

**Comment by apauley** [comment]
Thanks for adding this @alang76, @simonmichael and @Xitian9 - this can solve an issue that an hledger-flow user have reported when converting Paypal CSV files to journals (here is a [simplified summary](https://github.com/simonmichael/hledger/issues/1375#issuecomment-876748013) for background context).

I have built this branch and tested it using `hledger print --commodity-style '€1 000.00'` on a test file - it works great like that.

Not sure if this is too big of an ask, but would it be possible to allow the commodity styles to be specified using an external file?

hledger-flow uses a file named `directives.journal` and then automatically includes it in the main journal if it exists.

This is what my directives.journal looks like:
```
commodity $1,000.00
commodity 1,000.00 USD
P 2000-01-01 USD $1

commodity €1,000.00
commodity 1,000.00 EUR
P 2000-01-01 EUR €1
```

Internally hledger-flow runs `hledger print` to convert CSV files to hledger journals, and it would be convenient to provide the same file as an option there.

Alternatively I can also extract all the commodity directives from that file, and specify them as multiple `--commodity-style` options - this can then avoid over-complicating this PR if not many people will find an external file useful.

**Comment by Xitian9** [comment]
That looks good! @simonmichael has mentioned that he'd like to reduce the use of `def` in the code base due to it interfering with IDE and inference tools, so maybe those `def :: InputOpts` could be replaced by `definputopts`.

**Comment by alang76** [comment]
@Xitian9 Yeah that's a good idea. I was a bit unfamiliar with Data.Default but according to [this](https://markkarpov.com/post/data-default.html) there are also other reasons why they are not a great idea.

@apauley Hmm that could be done I guess, but that would be really close to the hledger import behaviour wouldn't it?

**Comment by apauley** [comment]
> @apauley Hmm that could be done I guess, but that would be really close to the hledger import behaviour wouldn't it?

True, I can also switch to using `hledger import`. It behaves a bit differently, so I'll have to test that the `.latest.bank.csv` behaves well with the rest of the hledger-flow assumptions.

**Comment by alang76** [comment]
@apauley Ah yeah that's true and in fact one of the reasons I prefer print in my own workflow. Curious to know what @simonmichael and @Xitian9 think of the proposal. In the interest of gettings things done ™ maybe we could take this on in a followup PR?

**Comment by apauley** [comment]
> In the interest of gettings things done ™ maybe we could take this on in a followup PR?

Sure, a follow-up PR sounds good - assuming it should be done at all.

**Comment by simonmichael** [comment]
I squashed the current commits into one and suddenly everything became very clear - can we do that ?

Could you add the hledger-lib.cabal change in package.yaml also. (I would add it to package.yaml first and do a stack build to be certain sure all the right places in the cabal file get updated.)

I think this is looking great. We are passing more unused options in some places but the simplification seems worth it.

Is InputOptions used more in Hledger.Data or in Hledger.Read ? If the latter, how hard would it be to move it there ?

**Comment by simonmichael** [comment]
Thanks for the input @apauley. Yes let's discuss this separately (chat could be good too). A separate file normally runs into https://hledger.org/hledger.html#directives-and-multiple-files limitations, and a special case doesn't sound so good.

**Comment by apauley** [comment]
> Thanks for the input @apauley. Yes let's discuss this separately (chat could be good too).

I think I should test `hledger import`, it does what I need when it comes to the directives, and I think there are ways to ensure that the `.latest.bank.csv` plays nicely with the rest of hledger-flow. We can also chat if there are more options that you think I can consider.

**Comment by alang76** [comment]
> Is InputOptions used more in Hledger.Data or in Hledger.Read ? If the latter, how hard would it be to move it there ?

As discussed there is a situation now where stuff is used for reading, which should be in Hledger.Read, but with a dependency from the lower level Hledger.Data. Might need to take a second look at which types do what.

**Comment by simonmichael** [comment]
Thanks. Yes from chat I'm clearer that we moved InputOpts, primarily Read-related, into Data. Let's take a little more time to think on the best factoring/naming/placement here.

**Comment by Xitian9** [comment]
The only thing we need `Hledger.Data` for in the new `InputOpts` module is for
```
type StorageFormat = String
type CommodityStyles = Maybe (M.Map CommoditySymbol AmountStyle)
defcommoditystyles = Nothing
```

`defcommoditystyles` isn't pulling a lot of weight, and can probably just be replaced with `Nothing` everywhere.

I'm personally not a big fan of type synonyms (and I'm [not the only one](https://www.reddit.com/r/haskell/comments/64ui11/reasons_i_dislike_type_synonyms/)), but if we just substitute the actual types in the definition of `InputOpts` then we require no imports whatsoever and can put that module wherever we like.

**Comment by Xitian9** [comment]
Oh, I see. The problem is that `InputOpts` is now used inside `Hledger.Data.Transaction`.

Well, it may not be aesthetically pleasing with the hierarchical naming scheme, but so long as `InputOptions` is a leaf of the dependency graph, it can be imported anywhere without causing problems. There's nothing stopping us putting it in `Hledger.Read.InputOptions` and importing that in both `Hledger.Data.Transaction` and the root `Hledger.Read` module.

**Comment by alang76** [comment]
Yes that could work practically I think. I wonder though, wouldn't it be conceptually strange if Data.Transaction (which is a core module) would depend on a data structure which is part of the shell around it to expose it the UI? I keep thinking that the original `BalancingOpts` wasn't such a bad idea, except maybe that having the `InputOpts` include `BalancingOpts` mixes types defined in `Hledger.Data` and `Hledger.Read`, but that is perhaps purely an aesthetic concern.

**Comment by Xitian9** [comment]
Well, another option would be to take `transactionCheckBalanced` and `isTransactionBalanced` out of `Hledger.Data.Transactions` and `journalBalanceTransactions` and `journalCheckBalanceAssertions` out of `Hledger.Data.Journal`. They could perhaps be put somewhere like `Hledger.Read.Balancing`.

**Comment by alang76** [comment]
Hmm, maybe for balancing the conceptual separation between Read and Data isn't as large as I initially thought. I will meditate on this during my holiday :)

Thanks for the hint regarding type synonyms btw.

**Comment by Xitian9** [comment]
Enjoy your holiday, and try to find more pleasant things to meditate upon!

**Comment by Xitian9** [comment]
Hi @alang76. Is there anything I can do to help get this out the door?

**Comment by alang76** [comment]
Hi @Xitian9, going to try give this a push today.

**Comment by alang76** [comment]
Disregard above.

**Comment by alang76** [comment]
- Moved `InputOpts` to `Hledger.Read`
- Removed type synonym for the commodity styles
- Reverted to the original `BalancingOpts` (I think this is better then flattening into `InputOpts` since that requires `Hledger.Data.Transaction` to have a dependency on `InputOpts` which is now in `Hledger.Read`)
- Removed default instances for `BalancingOpts` and `InputOpts`

File: hledger/test/cli/commodity-style.test [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+2021-07-09 Euro
+    (a)    EUR 1.234,56
+
+2021-07-09 Dollar
+    (a)      $ 1,234.56
+
+$ hledger -f- print -c '1 0'
+>
+2021-07-09 no symbol
+    (a)           1 2 3
```
Xitian9, line 15: I think I might need some clarification with what's going on here. Here we're changing to digit groupings of size one with a space for a ‘thousands’ separator? Odd, but reasonable.

Xitian9:
Looks great! There are a couple of missing newlines at the end of files, but I don't think that's a blocker for merging.

hledger/test/cli/commodity-style.test
Xitian9: To be clear, a space for a thousands separator is the only reasonable option ^^. I meant the digit groupings of sizse one was odd.

alang76: Yeah, although an interesting corner case, this is probably not the place to test it.

alang76:
> Looks great! There are a couple of missing newlines at the end of files

Where are you seeing them? I'll fix it up

Xitian9:
Github will show you with a little red circle with a line through it in the PR. Check the end of `commodity-style.test` under Files changed.

alang76:
@Xitian9 Should be fixed now.

simonmichael:
Looks great to me, rebased and merged as 3426030a. Thank you very much!

PR #1595, opened 2021-07-04
;imp: added systemd and nginx files to examples dir

harleypig:
This commit contains example files for a hledger-web systemd service and the basics for an nginx reverse proxy.

### DIFF
--- a/examples/systemd/hledger-web.service
+++ b/examples/systemd/hledger-web.service
@@ -1,0 +1,24 @@
+[Unit]
+Description=hledger-web -- plain text accounting, made easy
+Documentation=https://hledger.org/index.html
+After=network.target
+
+[Service]
+# Your account environment is not available if you are not logged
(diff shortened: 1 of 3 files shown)


simonmichael:
Thanks! This is the first PR to which we'll apply the [new commit conventions scheme](https://hledger.org/CONTRIBUTING.html#commit-messages). Would you mind just adding a prefix to the commit message, oh let's say `;imp: doc: ` ("a user-visible improvement in the area of documentation, don't bother with CI").

I'm fine with merging this as is for now, but perhaps the nginx config won't end up under `systemd/`.

harleypig:
Done. And don't expect the nginx file to stay there, or even unchanged. :) I didn't take the time to create a real template, just deleted my identifying info from it.

harleypig:
fix above comment: And I don't expect the nginx file to stay ...

simonmichael:
Good enough, thanks for the contribution!

PR #1601, opened 2021-07-09
;cln: lib: Do not double quote when reporting on missing journal file.

Xitian9:


Xitian9:
I've put the first character of the commit message as a `;`, but it seems to be building the CI anyway.

PR #1602, opened 2021-07-09
run commitlint to check commits as part of CI

simonmichael:
Working on adding [commitlint](https://github.com/simonmichael/hledger/blob/master/bin/commitlint) to all of our CI workflows.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -87,6 +87,23 @@ jobs:
       with: 
         fetch-depth: 0        
 
+    #- name: Print debug output
+    #   env:
+    #     GITHUB_CONTEXT: ${{ toJson(github) }}
+    #   run: |
+    #     echo "$GITHUB_SHA"
+    #     echo "$GITHUB_REF"
+    #     echo "$GITHUB_HEAD_REF"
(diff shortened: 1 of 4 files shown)


simonmichael:
I'm trying to run commitlint in our CI workflows; needless to say it's not easy.

In https://github.com/simonmichael/hledger/pull/1602/checks?check_run_id=3034665524 -> Print debug output, it is git log-ing the just-pushed commits - just one in this case, c34d26e4ef97b6b91d4b71aa322c24164959f6fc. However we can see github has used a merge commit, disrupting my linting. Why is this and can I prevent it ?

simonmichael:
Never mind.. it seems github always adds an interim merge commit at the end of a PR. That one fails to pass commitlint, so I'm trying to exclude it with `origin/master..HEAD^` or `origin/master..HEAD~1`. These work locally but in the github action environment they list no commits.

simonmichael:
This PR's history is messed up. Summary: I fiddled with it as much as I could stand and merged as f7f88a4a9 and 01f9c7054, which hopefully runs commitlint on the right commits in most of our CI scenarios.

Issue #1597, opened 2021-07-08
web: register: chart breaks with multiple commodities & transaction prices, showing nothing (since 1.22)


simonmichael:
In hledger-web 1.22's register screen, when viewing an account containing multiple commodities with transaction prices, chart data can be misformatted (due to multiline amount rendering ?) causing a javascript error.

Xitian9:
Git bisect seems to show that this broke with the merging of the haskellerz talk notes. I can't tell precisely which commit because there are no stack files within the branch that was merged.

Xitian9:
Correction: it seems to break with commit 5e7b6935, where the internal representation of `MixedAmount` changed. This makes much more sense.

simonmichael:
Thanks! Here's a repro (viewing the x account):
```
commodity $ 0.00
commodity A 0.000000

2021-02-02
    (x:a)                                       A -0.173025 @ $0.499537 
    (x:b)                                        A 0.173025 @ $0.401414
    (x:c)                                         $0.016
```

Xitian9:
Found it. Prices were not being properly stripped, and so it was trying to render on more than one line.

simonmichael:
Merged, thanks for the quick fix.

simonmichael:
I think there's still a problem, either this issue or another. Compare the chart for the above example with hledger-web 1.21. With master (which now includes f3f821b), it's different. The blue A point is being plotted at ~0.17, even though the tooltip correctly says it is 0.

simonmichael:
20 minute rule - I'm going to unpush that commit from master to take the pressure off, please excuse the disruption.

simonmichael:
Here's another request, for the commit message: fix: is good, and since that denotes an end-user visible change, let's write (at least the first part of) the text for end users, ready for release notes. Maybe "fix: web: register chart with multiple commodities/prices, broken in 1.22 (#1597)" or something like that.

Xitian9:
Noted. I'll look into this and see what the problem is.

Xitian9:
Ah yes, prices are still not being stripped properly. They're now stripped for the display, but not for the plotting. I'll fix that shortly.

simonmichael:
This brings up the "how would we test for such bugs" and as usual with UIs it's a bit awkward. The best I can think of right now is run selenium tests and check for JS errors.

PR #1603, opened 2021-07-10
fix: web: Make sure to strip prices from amounts before displaying in hledger-web chart. (#1597)

Xitian9:
Version 2, now with more price stripping!


simonmichael:
Merged with tweaked commit message as 3380190. Thanks!

PR #1610, opened 2021-07-15
ref: bal: rename some report types to clarify/sync with docs

simonmichael:
Some renaming that seemed helpful as I contemplated working on something (named budgets). Types now match the terminology at https://hledger.org/hledger.html#balance-report-types, constructors are shorter, and there's no longer a constructor with the same name as the unrelated `BudgetReport` type. Any feedback welcome.

### DIFF
--- a/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
+++ b/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
@@ -107,8 +107,8 @@ accountTransactionsReport rspec@ReportSpec{rsOpts=ropts} j reportq thisacctq = i
       $ journalApplyValuationFromOpts rspec j
 
     startbal
-      | balancetype_ ropts == HistoricalBalance = sumPostings priorps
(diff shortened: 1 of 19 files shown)


simonmichael:
@Xitian9 I'll leave this here till you have a chance to glance at it. 

I wondered why I couldn't request your review in Github's sidebar - it seems you weren't configured as a repo collaborator. Fixed!

Xitian9:
It looks reasonable to me. It'll take a bit of getting used to, but shouldn't be too bad.

If you're working on budget reports, one issue that bothers me slightly. `Hledger.Reports.BudgetReport` does a mix of things which, for the balance report, are split between `Hledger.Reports.MultiBalanceReport` and `Hledger.Cli.Commands.Balance`. It may be more consistent to move `budgetReportAsText`, `budgetReportAsTable`, and `budgetReportAsCsv` out of `hledger-lib` and into `hledger`.

simonmichael:
Thanks! Noted.

Issue #1607, opened 2021-07-13
ui: transaction screen: can't navigate through transactions with up/down (since 1.22)


simonmichael:
While viewing one transaction of many in an account, the up/down keys no longer change which transaction you are viewing . 

A surprising one!

simonmichael:
Oh, 307bd9366 I guess.

Xitian9:
I'll look into this (since I did the breaking).

PR #1611, opened 2021-07-15
fix: ui: Make sure that you can scroll between matching transactions in

Xitian9:
TransactionScreen. (#1607)

Make sure that the previous register screen is updated whenever we
scroll in TransactionScreen.


simonmichael:
Fast work, thanks! Passes tests. I rebased and merged as 5f53da87, usually that closes the PR but not this time.

PR #1613, opened 2021-07-16
feat: bal: named budgets

simonmichael:
This makes it possible to keep multiple named budgets in one journal, and select the one you want with --budget's argument.

More precisely, you can select the subset of periodic transactions rules which contain a certain case-insensitive substring. (A regex or query would be nicer and more consistent, but requires quite a bit more work.) Only one such --budget argument is supported, the last one on the command line takes precedence.

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -117,8 +117,11 @@ data ReportOpts = ReportOpts {
     -- for account transactions reports (aregister)
     ,txn_dates_      :: Bool
     -- for balance reports (bal, bs, cf, is)
-    ,balancecalc_    :: BalanceCalculation
-    ,balanceaccum_   :: BalanceAccumulation
+    ,balancecalc_    :: BalanceCalculation  -- ^ What to calculate in balance report cells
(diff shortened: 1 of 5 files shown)


hledger-lib/Hledger/Reports/ReportOptions.hs
```diff
@@ -117,8 +117,11 @@ data ReportOpts = ReportOpts {
     -- for account transactions reports (aregister)
     ,txn_dates_      :: Bool
     -- for balance reports (bal, bs, cf, is)
-    ,balancecalc_    :: BalanceCalculation
-    ,balanceaccum_   :: BalanceAccumulation
+    ,balancecalc_    :: BalanceCalculation  -- ^ What to calculate in balance report cells
+    ,balanceaccum_   :: BalanceAccumulation -- ^ How to accumulate balance report values over time
+    ,budgetpat_      :: T.Text  -- ^ If non-empty, a case-insensitive description substring
```
Xitian9, line 122: Do you think it would make more sense for this to be `Maybe T.Text`? It would keep it a bit cleaner conceptually to maintain the distinction between ‘no pattern provided’ and ‘null pattern provided’.

simonmichael:
Yes! Was just thinking that. And rename it budget_.

simonmichael:
Merged as d2b799a7.

Issue #1609, opened 2021-07-14
Strange ROI reporting


ohmyohmyohmy:
I'm using hledger 1.22-0-g8d237869d, windows-x86_64

```
test.ledger:
2021/01/01 Opening Balance
 Assets:Test 101:Opening  $ 1000
 Assets:Test 102:Opening  $ 1000
 Equity:Opening Balance

2021/03/11 Dividends
 Assets:Test 101:Dividend  $ 101
 Equity:Test 101:Dividend

2021/03/11 Dividends
 Assets:Test 102:Dividend  $ 102
 Equity:Test 102:Dividend

2021/03/11 Deposit
 Assets:Test 101:Deposit  $ 1000
 Assets:Test 102:Deposit  $ 1000
 Assets:Cash

2021/04/22 Deposit
 Assets:Test 101:Deposit  $ 1000
 Assets:Test 102:Deposit  $ 1000
 Assets:Cash

```

When I run queries, I get some funny results:
```
hledger -f test.ledger roi -Y --inv "Assets:Test" --pnl "Equity:Test"
+---++------------+------------++---------------+----------+-------------+-------++-------+-------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) |   PnL ||   IRR |   TWR |
+===++============+============++===============+==========+=============+=======++=======+=======+
| 1 || 2021-01-01 | 2021-12-31 ||             0 |   $ 6000 |      $ 6203 | $ 203 || 4.06% | 6.84% |
+---++------------+------------++---------------+----------+-------------+-------++-------+-------+
Strange, but I guess it's acceptable.
```

```
hledger -f test.ledger roi -Y --inv "Assets:Test 101" --pnl "Equity:Test 101"
+---++------------+------------++---------------+----------+-------------+-------++-------+-------+
|   ||      Begin |        End || Value (begin) | Cashflow | Value (end) |   PnL ||   IRR |   TWR |
+===++============+============++===============+==========+=============+=======++=======+=======+
| 1 || 2021-01-01 | 2021-12-31 ||             0 |   $ 6000 |      $ 6102 | $ 102 || 2.04% | 5.10% |
+---++------------+------------++---------------+----------+-------------+-------++-------+-------+
Unexpected, getting totals of Test 101 AND Test 102, not just Test 101.
PnL of Test 101 has been switched with Test 102
```

```
hledger -f test.ledger roi -Y --inv "Assets:Test 102" --pnl "Equity:Test 102"
[...]

PR #1616, opened 2021-07-18
Provide more verbose advise on complex quoting use-cases highlighted in #1609

adept:
I think this should cover the topics we discussed in #1609 (and thus close it)

### DIFF
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -197,6 +197,10 @@ or:
 $ hledger register credit\ card
 ```
 
+Windows users should keep in mind that `cmd` treats single quote as a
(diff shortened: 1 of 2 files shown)


hledger/hledger.m4.md
```diff
@@ -197,6 +197,10 @@ or:
 $ hledger register credit\ card
 ```
 
+Windows users should keep in mind that `cmd` treats signle quote as a
```
Xitian9, line 200: s/signle/single/

adept: Argh. Pushed the fix. Thanks!

simonmichael:
Great, thanks,

PR #1618, opened 2021-07-21
imp: json: The keys of JSON objects are now displayed in alphabetical…

Xitian9:
… order, making it stable across different systems and compilers.


simonmichael:
Hmm, a pity. Ok, let's go with alphabetical. I'll just merge this and not mix in an indent change right now.

PR #1623, opened 2021-07-23
Implement gain report for balance reports.

Xitian9:
This is an adaptation of @chvp's PR #1432 to the new valuation framework. (Sorry for submitting a new PR).

Before we merge we need more tests, and in particular we need to decide on:
- the behaviour when there's no cost but changing valuation
- the behaviour with different valuation modes
- some tests for compound balance reports

### DIFF
--- a/hledger/Hledger/Cli/Commands/Balance.hs
+++ b/hledger/Hledger/Cli/Commands/Balance.hs
@@ -288,9 +288,11 @@ balancemode = hledgerCommandMode
       , "transactions. With a DESCPAT argument (must be separated by = not space),"
       , "use only periodic transactions with matching description"
       , "(case insensitive substring match)."
-      ])    
+      ])
     ,flagNone ["valuechange"] (setboolopt "valuechange")
-      "show change of value of period-end historical balances"
+      "show total change of value of period-end historical balances (caused by deposits, withdrawals, market price fluctuations)"
+    ,flagNone ["gain"] (setboolopt "gain")
+      "show unrealised capital gain/loss (historical balance value minus cost basis)"
     ,flagNone ["change"] (setboolopt "change")
       "accumulate amounts from column start to column end (in multicolumn reports, default)"
     ,flagNone ["cumulative"] (setboolopt "cumulative")
@@ -639,6 +641,9 @@ multiBalanceReportAsText ropts@ReportOpts{..} r = TB.toLazyText $
     mtitle = case (balancecalc_, balanceaccum_) of
         (CalcValueChange, PerPeriod  ) -> "Period-end value changes"
         (CalcValueChange, Cumulative ) -> "Cumulative period-end value changes"
+        (CalcGain,        PerPeriod  ) -> "Incremental gain"
+        (CalcGain,        Cumulative ) -> "Cumulative gain"
+        (CalcGain,        Historical ) -> "Historical gain"
         (_,               PerPeriod  ) -> "Balance changes"
         (_,               Cumulative ) -> "Ending balances (cumulative)"
         (_,               Historical)  -> "Ending balances (historical)"
--- a/hledger/Hledger/Cli/Commands/Balance.md
+++ b/hledger/Hledger/Cli/Commands/Balance.md
@@ -35,6 +35,7 @@ Many of these work with the higher-level commands as well.
 - or actual and planned balance changes ([`--budget`](#budget-report))
 - or value of balance changes ([`-V`](#valuation-type))
 - or change of balance values ([`--valuechange`](#balance-report-types))
+- or unrealised capital gain/loss ([`--gain`](#balance-report-types))
 
 ..in..
 
@@ -419,7 +420,9 @@ It is one of:
 - `--sum` : sum the posting amounts (**default**)
 - `--budget` : like --sum but also show a goal amount
 - `--valuechange` : show the change in period-end historical balance values
-<!-- - `--gain` : show the change in period-end historical balances values caused by market price fluctuations -->
+  (caused by deposits, withdrawals, and/or market price fluctuations)
+- `--gain` : show the unrealised capital gain/loss, (the current valued balance
+  minus each amount's original cost)
 
 **Accumulation type:**\
 Which postings should be included in each cell's calculation.
@@ -445,7 +448,7 @@ It is one of:
 - no valuation, show amounts in their original commodities (**default**)
 - `--value=cost[,COMM]`       : no valuation, show amounts converted to cost
 - `--value=then[,COMM]`       : show value at transaction dates
-- `--value=end[,COMM]`        : show value at period end date(s) (**default with `--valuechange`**)
+- `--value=end[,COMM]`        : show value at period end date(s) (**default with `--valuechange`, `--gain`**)
 - `--value=now[,COMM]`        : show value at today's date
 - `--value=YYYY-MM-DD[,COMM]` : show value at another date
 
--- a/hledger/Hledger/Cli/CompoundBalanceCommand.hs
+++ b/hledger/Hledger/Cli/CompoundBalanceCommand.hs
@@ -60,7 +60,9 @@ compoundBalanceCommandMode CompoundBalanceCommandSpec{..} =
    ([flagNone ["sum"] (setboolopt "sum")
       "show sum of posting amounts (default)"
    ,flagNone ["valuechange"] (setboolopt "valuechange")
-      "show change of value of period-end historical balances"
+      "show total change of period-end historical balance value (caused by deposits, withdrawals, market price fluctuations)"
+    ,flagNone ["gain"] (setboolopt "gain")
+      "show unrealised capital gain/loss (historical balance value minus cost basis)"
    ,flagNone ["budget"] (setboolopt "budget")
       "show sum of posting amounts compared to budget goals defined by periodic transactions\n "
 
@@ -123,18 +125,23 @@ compoundBalanceCommand CompoundBalanceCommandSpec{..} opts@CliOpts{reportspec_=r
         -- "2008/01/01-2008/12/31", not "2008").
         titledatestr = case balanceaccumulation of
             Historical -> showEndDates enddates
-            _                 -> showDateSpan requestedspan
+            _          -> showDateSpan requestedspan
           where
             enddates = map (addDays (-1)) . mapMaybe spanEnd $ cbrDates cbr  -- these spans will always have a definite end date
             requestedspan = reportSpan j rspec
 
         -- when user overrides, add an indication to the report title
         -- Do we need to deal with overridden BalanceCalculation?
-        mtitleclarification = flip fmap mbalanceAccumulationOverride $ \case
-            PerPeriod | changingValuation -> "(Period-End Value Changes)"
-            PerPeriod                     -> "(Balance Changes)"
-            Cumulative                    -> "(Cumulative Ending Balances)"
-            Historical                    -> "(Historical Ending Balances)"
+        mtitleclarification = case (balancecalc_, balanceaccumulation, mbalanceAccumulationOverride) of
+            (CalcValueChange, PerPeriod,  _              ) -> Just "(Period-End Value Changes)"
+            (CalcValueChange, Cumulative, _              ) -> Just "(Cumulative Period-End Value Changes)"
+            (CalcGain,        PerPeriod,  _              ) -> Just "(Incremental Gain)"
+            (CalcGain,        Cumulative, _              ) -> Just "(Cumulative Gain)"
+            (CalcGain,        Historical, _              ) -> Just "(Historical Gain)"
+            (_,               _,          Just PerPeriod ) -> Just "(Balance Changes)"
+            (_,               _,          Just Cumulative) -> Just "(Cumulative Ending Balances)"
+            (_,               _,          Just Historical) -> Just "(Historical Ending Balances)"
+            _                                              -> Nothing
 
         valuationdesc =
           (case cost_ of
@@ -149,9 +156,9 @@ compoundBalanceCommand CompoundBalanceCommandSpec{..} opts@CliOpts{reportspec_=r
                Nothing                 -> "")
 
         changingValuation = case (balancecalc_, balanceaccum_) of
-            (CalcValueChange, PerPeriod)     -> True
+            (CalcValueChange, PerPeriod)  -> True
             (CalcValueChange, Cumulative) -> True
-            _                                     -> False
+            _                             -> False
 
     -- make a CompoundBalanceReport.
     cbr' = compoundBalanceReport rspec{_rsReportOpts=ropts'} j cbcqueries
--- a/hledger-lib/Hledger/Data/Valuation.hs
+++ b/hledger-lib/Hledger/Data/Valuation.hs
@@ -20,6 +20,8 @@ module Hledger.Data.Valuation (
   ,mixedAmountToCost
   ,mixedAmountApplyValuation
   ,mixedAmountValueAtDate
+  ,mixedAmountApplyGain
+  ,mixedAmountGainAtDate
   ,marketPriceReverse
   ,priceDirectiveToMarketPrice
   -- ,priceLookup
@@ -114,28 +116,24 @@ amountToCost NoCost _      = id
 amountToCost Cost   styles = styleAmount styles . amountCost
 
 -- | Apply a specified valuation to this amount, using the provided
--- price oracle, reference dates, and whether this is for a
--- multiperiod report or not. Also fix up its display style using the
--- provided commodity styles.
+-- price oracle, and reference dates. Also fix up its display style
+-- using the provided commodity styles.
 --
 -- When the valuation requires converting to another commodity, a
--- valuation (conversion) date is chosen based on the valuation type,
--- the provided reference dates, and whether this is for a
--- single-period or multi-period report. It will be one of:
+-- valuation (conversion) date is chosen based on the valuation type
+-- and the provided reference dates. It will be one of:
 --
--- - a fixed date specified by the ValuationType itself
---   (--value=DATE).
+-- - the date of the posting itself (--value=then)
 --
 -- - the provided "period end" date - this is typically the last day
 --   of a subperiod (--value=end with a multi-period report), or of
 --   the specified report period or the journal (--value=end with a
 --   single-period report).
 --
--- - the provided "report end" date - the last day of the specified
---   report period, if any (-V/-X with a report end date).
+-- - the provided "today" date (--value=now).
 --
--- - the provided "today" date - (--value=now, or -V/X with no report
---   end date).
+-- - a fixed date specified by the ValuationType itself
+--   (--value=DATE).
 --
 -- This is all a bit complicated. See the reference doc at
 -- https://hledger.org/hledger.html#effect-of-valuation-on-reports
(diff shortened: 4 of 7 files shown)


chvp:
Thanks for picking this up. I realize I haven't been very active (my time for open-source contributions comes in short bursts), so it's definitely no problem if you want to finish this.

Xitian9:
A question: what is the correct behaviour when you buy a commodity in a different currency than you are reporting in?

```
<
D A

P 2000/01/01 stock  2 A
P 2000/02/01 stock  3 A

P 2000/01/01 B      5 A
P 2000/02/01 B      6 A

2000/01/01
  (assets:b)        1 stock @ 3 B
```

My expectation would be as follows:
```
$ hledger -f- bal -M --gain --no-total
Gain in 2000-01-01..2000-02-29, valued at period ends:

            ||  2000-01  2000-02 
============++===================
 assets:b   ||    -13 A      1 A 
```

What do others think about this?

Xitian9:
**Question 2**

```
D A

P 2000/01/01 B      5 A
P 2000/02/01 B      6 A

2000/01/01
  (assets:b)        3 B
```

```
$ hledger -f- bal -M --gain --no-total
Gain in 2000-01-01..2000-02-29, valued at period ends:

            ||  2000-01  2000-02 
============++===================
 assets:b   ||        0      3 A 
```
or should there only be zeros in this report?

**Question 3**
```
D A

P 2000/01/01 stock   5 A
P 2000/02/01 stock  10 A

P 2000/01/01 B      2 A
P 2000/02/01 B      3 A

2000/01/01
  (assets:b)        1 stock @ 1B
```

```
$ hledger -f- bal -M --gain --no-total --value=2000-02-01
Gain in 2000-01-01..2000-02-29, valued at period ends:

            ||  2000-01  2000-02 
============++===================
 assets:b   ||      8 A        0 
```
or
```
$ hledger -f- bal -M --gain --no-total --value=2000-02-01
Gain in 2000-01-01..2000-02-29, valued at period ends:

            ||  2000-01  2000-02 
============++===================
 assets:b   ||      7 A        0 
```

chvp:
For question one I would expect -3 A in february. Question two seems correct to me. For question three I would expect 7A.

Xitian9:
So can I summarise this by saying that you think the cost should be valued at the same time as the primary amount?

I'm actually not sure where you get the 3 A from for question 1. Here are two arguments I could see being made.

1. I have a stock which I paid 3 B for in January. In February, the value of the stock increased from 2 A to 3 A, an increase of 1 A. However, the 3 B that I paid for it back in January has gone from a value of 15 A to 18 A, meaning that the amount I paid for it a month ago would have been worth 3 A more if I had paid for it in February. So the 1 A increase in value of the stock is offset by the 3 A more I would have paid for it today, and so overall I lose 2 A.

2. I have a stock which I paid 3 B for in January. In February, the value of the stock increased from 2 A to 3 A, in increase of 1 A. Though the 3 B I paid for it would cost more A if I bought it today, that is irrelevant as I already paid back in January. So the overall increase in gain is just 1 A.

chvp:
> So can I summarise this by saying that you think the cost should be valued at the same time as the primary amount?

That's definitely a fair summarization. I think the idea behind it for me is that I would expect the commodity to be sold in the commodity in which it was bought, in which case its valuation matters for the gain calculation. I think it's a fair assumption, but maybe not everyone will agree.


> I'm actually not sure where you get the -3 A from for question 1. Here are two arguments I could see being made.
> 
>     1. I have a stock which I paid 3 B for in January. In February, the value of the stock increased from 2 A to 3 A, an increase of 1 A. However, the 3 B that I paid for it back in January has gone from a value of 15 A to 18 A, meaning that the amount I paid for it a month ago would have been worth 3 A more if I had paid for it in February. So the 1 A increase in value of the stock is offset by the 3 A more I would have paid for it today, and so overall I lose 2 A.
> 
>     2. I have a stock which I paid 3 B for in January. In February, the value of the stock increased from 2 A to 3 A, in increase of 1 A. Though the 3 B I paid for it would cost more A if I bought it today, that is irrelevant as I already paid back in January. So the overall increase in gain is just 1 A.

I did the calculation in my head, must have made a mistake somewhere. I meant the -2A scenario.

the-solipsist:
Hi, @Xitian9. Off-topic, but I just wanted to make a request to use commodity symbols like `USD` / `EUR` / `INR` / `AAPL` / `$` / `€` / `₹` in place of commodity symbols like `A` and `B`. It wouldn't be more effort while writing the examples, but would make a big difference for folks like me. For some reason, conceptualizing the `A`/`B` examples takes far more neurons for me, and based on my experience bringing this up on IRC once, this is true at least for some other people as well.  Thanks.

Xitian9:
Here is a summary of the design space. Consider the example:

```
2000/01/01
  (assets:usd)        1 stock @ 1 USD
```

For an amount with a cost, we have two amounts which need valuation
1. The primary amount (stock), which will be valued according to `--value=blah`
2. The costed amount (USD), for which we can either:
    * value according to `--value=blah`
    * value at posting time

If we are valuing in the commodity of the costed amount (USD), then these methods are identical. This is the most common case. However, if we want to value in another commodity (e.g. AUD), then they can produce different answers.

Another option is to by default always value in the commodity of the costed amount, unless otherwise requested. This will essentially mean that the default journal commodity is replaced by a per-posting default based on the commodity of the costed amount.

chvp:
I got the chance to test this just now, and it works great. I get the same results as with my earlier PR.

Xitian9:
I think the only thing needed for this to be ready for merge is to decide on the most sensible behaviour for when the cost is in a different commodity than the final valuation. Do we:
1. value at posting time, or
2. at the same time as the primary commodity

@chvp has given one vote for (2). (1) makes the most intuitive sense to me, but I don't have a use case in mind, so I wouldn't put too much stock in that.

The current implementation in this PR is (2), so we could just merge it as is and change it if/when somebody runs into it and complains, I suppose.

@the-solipsist, @simonmichael: Do you have any opinions on the matter?

chvp:
One thing that I noticed after some more testing is that this doesn't interact with the `--valuechange` argument. It can give useful results though. It can give a clearer image of the performance of a stock month over month. Say the value grows $5 every month, this PR will now report a gain of $5, $10, $15, ... I think the $5, $5, $5, ... report can be interesting as well though.

Xitian9:
Yes, you are correct that this is how it is implemented. Sussing this out was basically the content of question 2 above.

As implemented `--gain` and `--valuechange` are mutually exclusive; along with `--change` and `--budget`, these are the four Calculation types.

I think the basic gain report here will give you what you want.


```
$ cat gain.j
P 2000/01/01 stock    $10
P 2000/02/01 stock    $15
P 2000/03/01 stock    $20
P 2000/04/01 stock    $25

2000/01/01
  (assets)      1 stock @ $12

$ hledger -f gain.j bal -M --gain
Gain in 2000-01-01..2000-04-30, valued at period ends:

        || Jan  Feb  Mar  Apr 
========++====================
 assets || $-2   $5   $5   $5 
--------++--------------------
        || $-2   $5   $5   $5 

$ hledger -f gain.j bal -M --gain -H
Historical gain in 2000-01-01..2000-04-30, valued at period ends:

        || 2000-01-31  2000-02-29  2000-03-31  2000-04-30 
========++================================================
 assets ||        $-2          $3          $8         $13 
--------++------------------------------------------------
        ||        $-2          $3          $8         $13 
```

chvp:
Ah yes, I was using the `balancesheet` report in combination with `--gain` which is historical by default, which is why I didn't get the numbers I expected.

Xitian9:
Perhaps this can be clarified with a better choice of titles for the report. I've gone with Gain, Cumulative Gain, and Historical Gain, but maybe Incremental Gain, Cumulative Gain, and (Historical) Gain would be better?

chvp:
Sounds like a good idea to me.

Xitian9:
Okay, it seems that nobody else feels strongly enough to suggest an opinion on this issue. With that in mind, I propose we merge the current behaviour, document it, and put instructions in the manual for people to get in contact if they want other behaviour.

la-wu:
FWIW, I agree with valuing at the same time as primary commodity.

In the Question 1 example, if we're attempting to evaluate in `AUD` we are really putting on 2 separate trades: a currency trade from `AUD` -> `USD` and then a stock trade from `USD` -> `stock`. Then the gains are really something along the lines of

```
                  ||    2000-01  2000-02 
==================++=====================
 assets:currency  ||          0   -3 AUD 
 assets:stock     ||    -13 AUD    1 AUD 
```

Xitian9:
Thanks @la-wu. That is what we get with the current implementation (except that there's only one account, so the sum of those is what we see).

simonmichael:
I haven't looked at this at all, will try to catch up with y'all soon.

hledger/Hledger/Cli/CompoundBalanceCommand.hs
```diff
@@ -61,6 +61,8 @@ compoundBalanceCommandMode CompoundBalanceCommandSpec{..} =
       "show sum of posting amounts (default)"
    ,flagNone ["valuechange"] (setboolopt "valuechange")
       "show change of value of period-end historical balances"
+    ,flagNone ["gain"] (setboolopt "gain")
+      "show gain (present valuation minus cost basis) of period-end historical balances"
```
simonmichael, line 65: I wonder if we can explain the difference between these more clearly. They are not easy to describe briefly. What would you think of:

  valuechange : show total change of value of period-end historical balance (caused by deposits, withdrawals, market price fluctuations)
  
  gain: show period-end unrealised capital gain/loss (period-end historical balance value minus cost basis)

IIRC I intended both of these to work with period-end dates, as the only really meaningful dates in practice. But did I see that they can be used with other dates (then/now/DATE) ? Am I wrong to say only "period-end" ?

hledger/Hledger/Cli/CompoundBalanceCommand.hs
```diff
@@ -129,11 +131,16 @@ compoundBalanceCommand CompoundBalanceCommandSpec{..} opts@CliOpts{reportspec_=r
...
         -- when user overrides, add an indication to the report title
         -- Do we need to deal with overridden BalanceCalculation?
-        mtitleclarification = flip fmap mbalanceAccumulationOverride $ \case
-            PerPeriod | changingValuation -> "(Period-End Value Changes)"
-            PerPeriod                     -> "(Balance Changes)"
-            Cumulative                    -> "(Cumulative Ending Balances)"
-            Historical                    -> "(Historical Ending Balances)"
+        mtitleclarification = case (balancecalc_, balanceaccumulation, mbalanceAccumulationOverride) of
+            (CalcValueChange, PerPeriod,  _              ) -> Just "(Period-End Value Changes)"
+            (CalcValueChange, Cumulative, _              ) -> Just "(Period-End Value Changes)"
```
simonmichael, line 136: Should the above titles not be different from each other ?

simonmichael (commented):
Some doc notes that seemed to make sense to me right now. Excuse any mistakes or possibilities I've missed, it's not possible to keep the details of this stuff in your head unless you're using it often. 

I made some edits, I'm not sure if those will be included in the email notification.

simonmichael:
> and the valuation of the cost of the amount

"The valuation of the cost" will be confusing at first read. I guess it's to allow reporting capital gain in a different currency ?

Re using familiar currency symbols in examples +1 to that in cases where it helps, but I'll continue my practice of minimal symbols in some situations when it helps encourage the mindset of removing all non-essential detail (you get used to them).

hledger/Hledger/Cli/Commands/Balance.md
```diff
@@ -35,6 +35,7 @@ Many of these work with the higher-level commands as well.
 - or actual and planned balance changes ([`--budget`](#budget-report))
 - or value of balance changes ([`-V`](#valuation-type))
 - or change of balance values ([`--valuechange`](#balance-report-types))
+- or changes of gain ([`--gain`](#balance-report-types))
```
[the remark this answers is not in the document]
Xitian9: Sounds good. I'll update.

hledger/Hledger/Cli/Commands/Balance.md
```diff
@@ -354,7 +355,12 @@ It is one of:
 - `--sum` : sum the posting amounts (**default**)
 - `--budget` : like --sum but also show a goal amount
 - `--valuechange` : show the change in period-end historical balance values
-<!-- - `--gain` : show the change in period-end historical balances values caused by market price fluctuations -->
+- `--gain` : show the gain, the difference between the valuation of the primary amount and the cost basis
```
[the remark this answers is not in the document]
Xitian9: Sounds good.

hledger/Hledger/Cli/Commands/Balance.md
```diff
@@ -354,7 +355,12 @@ It is one of:
...
+
+A note on gain reports. If the commodity valued in is not the same as the cost
+basis, both are valued at the same date.
+If this is not what you want, for example if you want the primary amount valued
+at period end and the cost basis valued at posting time, please submit a request.
```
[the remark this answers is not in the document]
Xitian9: Should I just remove this?

hledger/Hledger/Cli/CompoundBalanceCommand.hs
Xitian9: `--valuechange` can only be used with period-end, but `--gain` can be used with any valuation. Period end is most likely, but other valuation methods are reasonable too (I want to know the capital gains at a specific point in time, e.g. tax-filing deadline)

hledger/Hledger/Cli/Commands/Balance.md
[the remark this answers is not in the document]
simonmichael: Good idea, it may not be needed much at least for a while. You could leave a note in the code.

hledger/Hledger/Cli/CompoundBalanceCommand.hs
Xitian9: Yes, they should.

Xitian9:
Changes made. Let me know if you see any other improvements to be made.

Xitian9:
I've rebased on master, but still seem to be getting the build failure.

simonmichael:
Pushed another possible fix.. I'll keep tweaking, no need worry about it for this PR.

hledger/Hledger/Cli/Commands/Balance.hs
```diff
@@ -291,6 +291,8 @@ balancemode = hledgerCommandMode
       ])    
     ,flagNone ["valuechange"] (setboolopt "valuechange")
       "show change of value of period-end historical balances"
+    ,flagNone ["gain"] (setboolopt "gain")
+      "show gain (present valuation minus cost basis) of period-end historical balances"
```
simonmichael, line 295: I think I missed this one; do you think this is better ?

show unrealised capital gain/loss (period-end historical balance value minus each amount's original cost)

simonmichael:
> 
> --valuechange can only be used with period-end, but --gain can be used with any valuation. Period end is most likely, but other valuation methods are reasonable too (I want to know the capital gains at a specific point in time, e.g. tax-filing deadline)


Check. Thanks!

hledger/Hledger/Cli/Commands/Balance.hs
Xitian9: Done.

simonmichael:
Thanks @Xitian9, @chvp, and reviewers!

simonmichael:
@Xitian9 , could you help me understand this better for the changelog:

> -- If the commodity we are valuing in is not the same as the commodity of the
> -- cost, this will value the cost at the same date as the primary amount. This
> -- may not be what you want; for example you may want the cost valued at the
> -- posting date. If so, let us know and we can change this behaviour.

Posting date is the date the amount was acquired (specifically, the date it posted to the asset account) ?

What's the date of the primary amount ?

Also, what happens when the amount is being disposed of (removed from the asset account) ?

Xitian9:
> Posting date is the date the amount was acquired (specifically, the date it posted to the asset account) ?

Yes.

> What's the date of the primary amount ?

The date of the primary amount is just the date you are asking to value the currency at, according to the usual `--value` option. 

> Also, what happens when the amount is being disposed of (removed from the asset account) ?

Currently nothing special. If you sell it your total gain is reported at that point and remains unchanged after that.

```
P 2021-01-01 stock  $0.50
P 2021-02-01 stock  $1.00
P 2021-04-01 stock  $2.00

2021-01-01  Buy stock
  assets:stock    1 stock @ $1
  assets:cash   -$1

2021-03-01  Sell stock
  assets:stock  -1 stock @ $2
  assets:cash   $2
```

```
$ hledger -f- bal -M --gain -H
Historical gain in 2021-01-01..2021-04-30, valued at period ends:

              || 2021-01-31  2021-02-28  2021-03-31  2021-04-30 
==============++================================================
 assets:stock ||     $-0.50           0       $1.00       $1.00 
--------------++------------------------------------------------
              ||     $-0.50           0       $1.00       $1.00 
```

chvp:
Note that I would do that sale differently to avoid creating money out of thin air. I would write
```
2021-03-01
    assets:stock    -1 stock @ $1
    assets:cash     $2
    equity:realized gain
```

You sell the stock at the cost you bought it at, reducing the (unrealized) gain to 0 at that point, since it has been converted to realized gain.

Xitian9:
Noted, good thinking.

simonmichael:
> On Aug 19, 2021, at 6:52 PM, Stephen Morgan ***@***.***> wrote:
> Posting date is the date the amount was acquired (specifically, the date it posted to the asset account) ?
> 
> Yes.
> 
> What's the date of the primary amount ?
> 
> The date of the primary amount is just the date you are asking to value the currency at, according to the usual --value option.
> 
Got it. Valuation date, I usually call that.

> Also, what happens when the amount is being disposed of (removed from the asset account) ?
> 
> Currently nothing special. If you sell it your total gain is reported at that point and remains unchanged after that.
> 
> P 2021-01-01 stock  $0.50
> P 2021-02-01 stock  $1.00
> P 2021-04-01 stock  $2.00
> 
> 2021-01-01  Buy stock
>   assets:stock    1 stock @ $1
>   assets:cash   -$1
> 
> 2021-03-01  Sell stock
>   assets:stock  -1 stock @ $2
>   assets:cash   $2
> $ hledger -f- bal -M --gain -H
> Historical gain in 2021-01-01..2021-04-30, valued at period ends:
> 
>               || 2021-01-31  2021-02-28  2021-03-31  2021-04-30 
> ==============++================================================
>  assets:stock ||     $-0.50           0       $1.00       $1.00 
> --------------++------------------------------------------------
>               ||     $-0.50           0       $1.00       $1.00 
Thanks for the example. My understanding of capital gain does not seem to match it. I would have said there is capital gain/loss only when there is a change of market value of an asset you are holding. So I expected to see some gain on 2/1 (because market price doubled) and on 4/1 (ditto), only. Am I wrong ?

Aside: it's very useful to have a way to see more detail, to explain and verify any given report, especially as we get into more powerful and complex reports. I have been wondering how to do it for the recent balance report types, and just now used these to mimic register (possibly useful for troubleshooting/explaining):

hledger bal --gain -D --transpose assets:stock

hledger bal --valuechange -D --transpose assets:stock

simonmichael:
Also, with Charlotte's modification the report changes. Is it correct ?
```
P 2021-01-01 stock  $0.50
P 2021-02-01 stock  $1.00
P 2021-04-01 stock  $2.00

2021-01-01  Buy stock
  assets:stock    1 stock @ $1
  assets:cash   -$1

2021-03-01 Sell stock
    assets:stock    -1 stock @ $1
    assets:cash     $2
    equity:realized gain

```
```
$ hledger bal -M --gain -H
Historical gain in 2021-01-01..2021-04-30, valued at period ends:

              || 2021-01-31  2021-02-28  2021-03-31  2021-04-30 
==============++================================================
 assets:stock ||     $-0.50           0           0           0 
--------------++------------------------------------------------
              ||     $-0.50           0           0           0 
```

chvp:
In the example the stock is bought at a higher price than the market value. Selling before 02-01 would result in a loss, which is the -0.5 in the first column. From 02-01 to 02-28, selling would result in 0 or loss or profit. The gain increases by 0.5, but it's a historical report, so a 0 is shown. 

In march, the expected profit rises to 1. We sell at market price in march, so the realized gain is 1 from that point on. The unrealized gain can't change anymore, since we dont hold any stock anymore. (Which is why we don't see any changes anymore from that point on.) The change I made does indeed have an impact on the report, since the stock is sold at the same cost basis it was bought at. This makes the gain valuation of that transaction a perfect offset from the initial transaction, resulting in zeroes from that point on.

I hope that explanation makes things a bit clearer 😅

simonmichael:
It sure does, thanks!

We probably will need example + walk through docs for these reports, just like the above, to help folks understand them and reduce our future support load.

Is it true to say that the correct entry for a commodity exchange, using equity, is currently not supported by --gain ? (Similar to --cost).

chvp:
> We probably will need example + walk through docs for these reports, just like the above, to help folks understand them and reduce our future support load.

I can draft a walkthrough using an example if you want.

> Is it true to say that the correct entry for a commodity exchange, using equity, is currently not supported by --gain ? (Similar to --cost).

`--gain` needs the same info as `--cost`, so yes, that's true.

simonmichael:
That would be great. 👍🏻

Issue #1559, opened 2021-05-12
balance reports: one currency per row, separate currency field


simonmichael:
It is a common request to have the amount quantity and the currency symbol separated in CSV output, rather than combined in the one field, for easier consumption by other tools. As one example, it could be useful to colour code amounts by currency in https://hledger.org/charts.html.

This is problematic in transaction/register reports, where any amount can be a multicommodity amount (especially the running totals in a register report). We don't want to have a varying number of columns, nor to give up the correspondence of rows to transactions or postings.

But it shouldn't be a problem for balance reports (`balance` and friends). There, for an account with a multicommodity balance we could just generate multiple rows, one per commodity. And then we could also put the commodity symbol and amount quantity in separate columns. 

Questions: Any problems with this ? Does it work with multicolumn reports ? Does it work with tree mode ? Should it also be an option for the other output formats, including terminal reports ? Could it help with the problem of too-wide/too-elided multicommodity amounts ? Do it as an option, or always ?

simonmichael:
nMockups: instead of the current `bal -O csv` output:
```
"account","balance"
"Assets:US:ETrade","70.00 GLD, 17.00 ITOT, 5120.50 USD, 36.00 VEA, 294.00 VHT"
"total","70.00 GLD, 17.00 ITOT, 5120.50 USD, 36.00 VEA, 294.00 VHT"
```

it could be:
```
"account","balance"
"Assets:US:ETrade","70.00 GLD"
"Assets:US:ETrade","17.00 ITOT"
"Assets:US:ETrade","5120.50 USD"
"Assets:US:ETrade","36.00 VEA"
"Assets:US:ETrade","294.00 VHT"
"total","70.00 GLD"
"total","17.00 ITOT"
"total","5120.50 USD"
"total","36.00 VEA"
"total","294.00 VHT"
```

or:
```
"account","commodity","amount"
"Assets:US:ETrade","GLD","70.00"
"Assets:US:ETrade","ITOT","17.00"
"Assets:US:ETrade","USD","5120.50"
"Assets:US:ETrade","VEA","36.00"
"Assets:US:ETrade","VHT","294.00"
"total","GLD","70.00"
"total","ITOT","17.00"
"total","USD","5120.50"
"total","VEA","36.00"
"total","VHT","294.00"
```

With multiple periods:
```
"account","2012","2013","2014"
[...]

PR #1626, opened 2021-07-23
bal: Show commodity symbol as a subaccount / in a column

la-wu:
This is a potential implementation of proposals (6) and (10) from https://github.com/simonmichael/hledger/issues/1559. These commits handle the `balance` family of commands but should be extendable to `compound-balance` family as well.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -152,7 +152,7 @@ import Data.Foldable (toList)
 import Data.List (find, foldl', intercalate, intersperse, mapAccumL, partition)
 import Data.List.NonEmpty (NonEmpty(..), nonEmpty)
 import qualified Data.Map.Strict as M
-import Data.Maybe (fromMaybe, isNothing)
+import Data.Maybe (fromMaybe, isNothing, isJust)
 import Data.Semigroup (Semigroup(..))
 import qualified Data.Text as T
 import qualified Data.Text.Lazy.Builder as TB
@@ -175,6 +175,9 @@ data AmountDisplayOpts = AmountDisplayOpts
   , displayOneLine       :: Bool       -- ^ Whether to display on one line.
   , displayMinWidth      :: Maybe Int  -- ^ Minimum width to pad to
   , displayMaxWidth      :: Maybe Int  -- ^ Maximum width to clip to
+  -- | Display amounts in this order (without the commodity symbol) and display
+  -- a 0 in case a corresponding commodity does not exist
+  , displayOrder         :: Maybe [CommoditySymbol]
   } deriving (Show)
 
 -- | Display Amount and MixedAmount with no colour.
@@ -186,8 +189,9 @@ noColour = AmountDisplayOpts { displayPrice         = True
                              , displayColour        = False
                              , displayZeroCommodity = False
                              , displayOneLine       = False
-                             , displayMinWidth      = Nothing
+                             , displayMinWidth      = Just 0
                              , displayMaxWidth      = Nothing
+                             , displayOrder         = Nothing
                              }
 
 -- | Display Amount and MixedAmount with no prices.
@@ -429,14 +433,15 @@ showAmountB :: AmountDisplayOpts -> Amount -> WideBuilder
 showAmountB _ Amount{acommodity="AUTO"} = mempty
 showAmountB opts a@Amount{astyle=style} =
     color $ case ascommodityside style of
-      L -> c' <> space <> quantity' <> price
-      R -> quantity' <> space <> c' <> price
+      L -> showC c' space <> quantity' <> price
+      R -> quantity' <> showC space c' <> price
   where
     quantity = showamountquantity a
     (quantity',c) | amountLooksZero a && not (displayZeroCommodity opts) = (WideBuilder (TB.singleton '0') 1,"")
                   | otherwise = (quantity, quoteCommoditySymbolIfNeeded $ acommodity a)
     space = if not (T.null c) && ascommodityspaced style then WideBuilder (TB.singleton ' ') 1 else mempty
     c' = WideBuilder (TB.fromText c) (textWidth c)
+    showC l r = if isJust (displayOrder opts) then mempty else l <> r
     price = if displayPrice opts then showAmountPrice a else mempty
     color = if displayColour opts && isNegativeAmount a then colorB Dull Red else id
 
@@ -820,13 +825,16 @@ showMixedAmountLinesB :: AmountDisplayOpts -> MixedAmount -> [WideBuilder]
 showMixedAmountLinesB opts@AmountDisplayOpts{displayMaxWidth=mmax,displayMinWidth=mmin} ma =
     map (adBuilder . pad) elided
   where
-    astrs = amtDisplayList (wbWidth sep) (showAmountB opts) . amounts $
+    astrs = amtDisplayList (wbWidth sep) (showAmountB opts) . orderedAmounts opts $
               if displayPrice opts then ma else mixedAmountStripPrices ma
     sep   = WideBuilder (TB.singleton '\n') 0
-    width = maximum $ fromMaybe 0 mmin : map (wbWidth . adBuilder) elided
+    width = maximum $ map (wbWidth . adBuilder) elided
 
-    pad amt = amt{ adBuilder = WideBuilder (TB.fromText $ T.replicate w " ") w <> adBuilder amt }
-      where w = width - wbWidth (adBuilder amt)
+    pad amt
+      | Just mw <- mmin =
+          let w = (max width mw) - wbWidth (adBuilder amt)
+           in amt{ adBuilder = WideBuilder (TB.fromText $ T.replicate w " ") w <> adBuilder amt }
+      | otherwise = amt
 
     elided = maybe id elideTo mmax astrs
     elideTo m xs = maybeAppend elisionStr short
@@ -843,7 +851,7 @@ showMixedAmountOneLineB opts@AmountDisplayOpts{displayMaxWidth=mmax,displayMinWi
     . max width $ fromMaybe 0 mmin
   where
     width  = maybe 0 adTotal $ lastMay elided
-    astrs  = amtDisplayList (wbWidth sep) (showAmountB opts) . amounts $
+    astrs  = amtDisplayList (wbWidth sep) (showAmountB opts) . orderedAmounts opts $
                if displayPrice opts then ma else mixedAmountStripPrices ma
     sep    = WideBuilder (TB.fromString ", ") 2
     n      = length astrs
@@ -865,6 +873,15 @@ showMixedAmountOneLineB opts@AmountDisplayOpts{displayMaxWidth=mmax,displayMinWi
 
     -- Add the elision strings (if any) to each amount
     withElided = zipWith (\num amt -> (amt, elisionDisplay Nothing (wbWidth sep) num amt)) [n-1,n-2..0]
+
+orderedAmounts :: AmountDisplayOpts -> MixedAmount -> [Amount]
+orderedAmounts AmountDisplayOpts{displayOrder=ord} ma
+  | Just cs <- ord = fmap pad cs
+  | otherwise = as
+  where
+    as = amounts ma
+    pad c = fromMaybe (amountWithCommodity c nullamt) . find ((==) c . acommodity) $ as
+
 
 data AmountDisplay = AmountDisplay
   { adBuilder :: !WideBuilder  -- ^ String representation of the Amount
--- a/hledger-lib/Text/Tabular/AsciiWide.hs
+++ b/hledger-lib/Text/Tabular/AsciiWide.hs
@@ -10,13 +10,16 @@ module Text.Tabular.AsciiWide
 , render
 , renderTable
 , renderTableB
+, renderTableByRowsB
 , renderRow
 , renderRowB
+, renderColumns
 
 , Cell(..)
 , Align(..)
 , emptyCell
 , textCell
+, textsCell
 , cellWidth
 ) where
 
@@ -30,7 +33,7 @@ import qualified Data.Text.Lazy as TL
 import Data.Text.Lazy.Builder (Builder, fromString, fromText, singleton, toLazyText)
 import Safe (maximumMay)
 import Text.Tabular
-import Text.WideString (WideBuilder(..), textWidth)
+import Text.WideString (WideBuilder(..), wbFromText, textWidth)
 
 
 -- | The options to use for rendering a table.
@@ -60,6 +63,10 @@ emptyCell = Cell TopRight []
 textCell :: Align -> Text -> Cell
 textCell a x = Cell a . map (\x -> WideBuilder (fromText x) (textWidth x)) $ if T.null x then [""] else T.lines x
 
+-- | Create a multi-line cell from the given contents with its natural width.
+textsCell :: Align -> [Text] -> Cell
+textsCell a = Cell a . fmap wbFromText
+
 -- | Return the width of a Cell.
 cellWidth :: Cell -> Int
 cellWidth (Cell _ xs) = fromMaybe 0 . maximumMay $ map wbWidth xs
@@ -86,20 +93,31 @@ renderTableB :: TableOpts       -- ^ Options controlling Table rendering
              -> (a -> Cell)   -- ^ Function determining the string and width of a cell
              -> Table rh ch a
              -> Builder
-renderTableB topts@TableOpts{prettyTable=pretty, tableBorders=borders} fr fc f (Table rh ch cells) =
+renderTableB topts fr fc f = renderTableByRowsB topts (fmap fc) (\(rh, as) -> (fr rh, fmap f as))
+
+-- | A version of renderTable that operates on rows (including the 'row' of
+-- column headers) and returns the underlying Builder.
+renderTableByRowsB :: TableOpts      -- ^ Options controlling Table rendering
+             -> ([ch] -> [Cell])     -- ^ Rendering function for column headers
+             -> ((rh, [a]) -> (Cell, [Cell])) -- ^ Rendering function for row and row header
+             -> Table rh ch a
+             -> Builder
+renderTableByRowsB topts@TableOpts{prettyTable=pretty, tableBorders=borders} fc f (Table rh ch cells) =
    unlinesB . addBorders $
      renderColumns topts sizes ch2
      : bar VM DoubleLine   -- +======================================+
      : renderRs (fmap renderR $ zipHeader [] cellContents rowHeaders)
  where
+  renderR :: ([Cell], Cell) -> Builder
   renderR (cs,h) = renderColumns topts sizes $ Group DoubleLine
                      [ Header h
                      , fmap fst $ zipHeader emptyCell cs colHeaders
                      ]
 
-  rowHeaders   = fmap fr rh
-  colHeaders   = fmap fc ch
-  cellContents = map (map f) cells
+  rows         = unzip . fmap f $ zip (headerContents rh) cells
+  rowHeaders   = fmap fst $ zipHeader emptyCell (fst rows) rh
+  colHeaders   = fmap fst $ zipHeader emptyCell (fc $ headerContents ch) ch
+  cellContents = snd rows
 
   -- ch2 and cell2 include the row and column labels
   ch2 = Group DoubleLine [Header emptyCell, colHeaders]
@@ -161,6 +179,7 @@ renderColumns TableOpts{prettyTable=pretty, tableBorders=borders, borderSpaces=s
     padCell (w, Cell BottomLeft  ls) = map (\x -> wbBuilder x <> fromText (T.replicate (w - wbWidth x) " ")) ls
     padCell (w, Cell TopRight    ls) = map (\x -> fromText (T.replicate (w - wbWidth x) " ") <> wbBuilder x) ls
     padCell (w, Cell BottomRight ls) = map (\x -> fromText (T.replicate (w - wbWidth x) " ") <> wbBuilder x) ls
+
 
     -- Pad each cell to have the same number of lines
     padRow (Cell TopLeft     ls) = Cell TopLeft     $ ls ++ replicate (nLines - length ls) mempty
--- a/hledger-lib/Text/WideString.hs
+++ b/hledger-lib/Text/WideString.hs
@@ -34,7 +34,7 @@ instance Monoid WideBuilder where
 wbToText :: WideBuilder -> Text
 wbToText = TL.toStrict . TB.toLazyText . wbBuilder
 
--- | Convert a WideBuilder to a strict Text.
+-- | Convert a strict Text to a WideBuilder.
 wbFromText :: Text -> WideBuilder
 wbFromText t = WideBuilder (TB.fromText t) (textWidth t)
 
--- a/hledger-lib/Text/WideString.hs
+++ b/hledger-lib/Text/WideString.hs
@@ -8,7 +8,8 @@ module Text.WideString (
   -- * Text Builders which keep track of length
   WideBuilder(..),
   wbUnpack,
-  wbToText
+  wbToText,
+  wbFromText
   ) where
 
 import Data.Text (Text)
@@ -32,6 +33,10 @@ instance Monoid WideBuilder where
 -- | Convert a WideBuilder to a strict Text.
 wbToText :: WideBuilder -> Text
 wbToText = TL.toStrict . TB.toLazyText . wbBuilder
+
+-- | Convert a WideBuilder to a strict Text.
+wbFromText :: Text -> WideBuilder
+wbFromText t = WideBuilder (TB.fromText t) (textWidth t)
 
 -- | Convert a WideBuilder to a String.
 wbUnpack :: WideBuilder -> String
--- a/hledger-lib/Hledger/Data/StringFormat.hs
+++ b/hledger-lib/Hledger/Data/StringFormat.hs
@@ -136,7 +136,7 @@ formatfieldp = do
     char '('
     f <- fieldp
     char ')'
-    return $ FormatField (isJust leftJustified) (parseDec minWidth) (parseDec maxWidth) f
+    return $ FormatField (isJust leftJustified) (parseDec minWidth <|> Just 0) (parseDec maxWidth) f
     where
       parseDec s = case s of
         Just text -> Just m where ((m,_):_) = readDec text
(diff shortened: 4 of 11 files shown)


hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -961,16 +961,16 @@ accumulateMixedAmount :: AmountAccumulator -> MixedAmount -> AmountAccumulator
...
 
-    addAmounts l mp am
+    addAmounts mp am
       | T.null c = mp
       | c `M.member` mp = M.adjust ((:) am) c mp
       | otherwise = M.insert c (am : replicate l 0{acommodity = c}) mp
       where c = acommodity am
 
-    padAmounts r k a = if k `M.member` r then a else 0{acommodity = k} : a
+    padAmounts k a = if k `M.member` r then a else 0{acommodity = k} : a
```
Xitian9, line 978: We're trying to reduce the use of the `Num` instances for `Amount`. It would probably be clearer to use `amountWithCommodity k nullamt` instead of `0{acommodity = k}`.

simonmichael:
Rendered doc: https://github.com/simonmichael/hledger/blob/caaf799fc25463b2af8e85dcf91386faddaccfc6/hledger/Hledger/Cli/Commands/Balance.md#commodity-account

simonmichael:
This looks like an excellent start!

I tried to understand the new code just by reading the names and haddocks, and had trouble. Perhaps we can find ways to make some of those clearer.

The terminal output shows (fake) subaccounts, while the CSV output shows a separate commodity column. Should we have them both do the same thing ? Which is better ? Will the fake accounts be confusing/frustrating, eg will people try to filter by those accounts or view their register ? If the subaccounts are a good idea, would we want to implement them more deeply so they are visible in all reports ?

Xitian9:
I think a good approach would be to extend `showMixedAmountB` and `showMixedAmountLinesB` so that you can specify the order they display amounts in. That is a flexible approach which could be easily adapted to any of the proposed multi-line display methods.

I'm personally in favour of a commodities column instead of fake subaccounts, but if others prefer subaccounts I wouldn't object.

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -741,6 +749,18 @@ styleMixedAmount styles = mapMixedAmountUnsafe (styleAmount styles)
...
+-- this will only display the (padded) amount quantity
+showMixedAmountCA :: Bool -> (AmountDisplayOpts -> MixedAmount -> WideBuilder)
+showMixedAmountCA commodity_account
+  | not commodity_account = showMixedAmountB
+  | otherwise = \dopts -> pad dopts . showAmountQuantity . head . amounts
+  where
+    pad dopts amt = WideBuilder (TB.fromText $ T.replicate w " ") w <> amt
+      where w = width - wbWidth amt
+            width = max (fromMaybe 0 (displayMinWidth dopts)) (wbWidth amt)
+
```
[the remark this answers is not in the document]
la-wu: Would you suggest the display order be constructed within the `renderTableB` or for it to remain in the caller at table construction (e.g `balanceReportAsTable`)?

It seems like in the former case we need to introduce a row-based rendering function.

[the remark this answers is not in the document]
Xitian9: I would probably calculate the display order in `balanceReportTableAsText`. I think it should be straightforward to introduce a row-based rendering function. The current `renderTableB` can be re-implemented as a simple wrapper with the same rendering function for each row. Let me know if you'd like me to submit a PR for that.

simonmichael:
A column seems simpler and a more limited scope than an account name. However it doesn't work for tree-mode balance reports, am I right ?

Xitian9:
I think we have big problems for tree-mode balance reports with any method. For example, if we have `1 USD, 1 CAD` in an account `assets:bank`, how do we display the inclusive balances in `assets`? Splitting into accounts as in this PR will not work. A column would work even in tree mode, though it would make for very long reports.

simonmichael:
We have discussed three places to show commodity symbols.

1. Inline in each amount. Current behaviour.
2. In a separate column. Implemented by this PR for CSV output. Works only with reports in list mode. 
3. As a fake (generated) subaccount. Implemented by this PR for terminal output. Might or might not need deeper support of these fake subaccounts for other commands to avoid confusion.

If we will ultimately want all three of these, we should provide a three-way option, like `--show-commodity-in=amount|column|account`. 

It seems to me the simplest lowest-impact way to address #1559 is to implement the column mode for both terminal and CSV output. It should give an error when used with unsupported output formats (json...) or incompatible modes (tree). It would be ok to start with this and change our minds later I guess. 

Thoughts ?

la-wu:
I agree that tree mode would make this 'fake subaccount' view quite confusing. If we adopted the syntax of (9), i.e commodity in brackets after, that could resolve that issue.

However, I think that having a consistent column for csv and terminal output is probably the least surprising behavior. I'll rework this to take some of Xitian9's suggestions and the commodity column in both.

simonmichael:
Subaccounts should work just fine in tree mode, but let's do the simple thing for now.

Xitian9:
@simonmichael: I think we agree on the course of action to take, but I'm actually not able to picture how subaccounts would work in tree mode. Could you do a quick mockup of how this should appear in tree mode? Assume that we're calling it with `--no-elide` so that there is no boring parent elision.

```
$ hledger bal -T -Y --commodity-account
Balance changes in 2012-01-01..2014-12-31:

                       ||     2012    2013     2014    Total
=======================++====================================
 Assets:US:ETrade:GLD  ||        0   70.00        0    70.00
 Assets:US:ETrade:USD  ||   337.18  -98.12  4881.44  5120.50
-----------------------++------------------------------------
 total:GLD             ||        0   70.00        0    70.00
 total:USD             ||   337.18  -98.12  4881.44  5120.50
```

la-wu:
In the case of tree mode, should the account name be repeated? e.g

```
                   || Commodity              2012              2013              2014
===================++=================================================================
 Assets:US         || GLD                       0             70.00                 0
 Assets:US         || ITOT                  10.00             18.00            -11.00
 Assets:US         || RGAGX      139.973000000000  171.689000000000  178.295000000000
 Assets:US         || USD                 7785.76           -299.58          -1769.65
 Assets:US         || VACHR                120.12            120.12             97.02
 Assets:US         || VBMPX      111.933000000000  101.000000000000   97.017000000000
 Assets:US         || VEA                   12.00             10.00             14.00
 Assets:US         || VHT                  106.00             18.00            170.00
   BofA:Checking   || USD                 7448.62           -201.50          -6651.07
   ETrade          || GLD                       0             70.00                 0
   ETrade          || ITOT                  10.00             18.00            -11.00
   ETrade          || USD                  337.18            -98.12           4881.44
   ETrade          || VEA                   12.00             10.00             14.00
   ETrade          || VHT                  106.00             18.00            170.00
     Cash          || USD                  337.18            -98.12           4881.44
     GLD           || GLD                       0             70.00                 0
     ITOT          || ITOT                  10.00             18.00            -11.00
     VEA           || VEA                   12.00             10.00             14.00
     VHT           || VHT                  106.00             18.00            170.00
   Hoogle:Vacation || VACHR                120.12            120.12             97.02
   Vanguard        || RGAGX      139.973000000000  171.689000000000  178.295000000000
   Vanguard        || USD                   -0.04              0.04             -0.02
   Vanguard        || VBMPX      111.933000000000  101.000000000000   97.017000000000
     Cash          || USD                   -0.04              0.04             -0.02
     RGAGX         || RGAGX      139.973000000000  171.689000000000  178.295000000000
     VBMPX         || VBMPX      111.933000000000  101.000000000000   97.017000000000
```

vs

```
                   || Commodity              2012              2013              2014
===================++=================================================================
 Assets:US         || GLD                       0             70.00                 0
                   || ITOT                  10.00             18.00            -11.00
                   || RGAGX      139.973000000000  171.689000000000  178.295000000000
                   || USD                 7785.76           -299.58          -1769.65
                   || VACHR                120.12            120.12             97.02
                   || VBMPX      111.933000000000  101.000000000000   97.017000000000
                   || VEA                   12.00             10.00             14.00
                   || VHT                  106.00             18.00            170.00
   BofA:Checking   || USD                 7448.62           -201.50          -6651.07
   ETrade          || GLD                       0             70.00                 0
                   || ITOT                  10.00             18.00            -11.00
                   || USD                  337.18            -98.12           4881.44
                   || VEA                   12.00             10.00             14.00
                   || VHT                  106.00             18.00            170.00
     Cash          || USD                  337.18            -98.12           4881.44
     GLD           || GLD                       0             70.00                 0
     ITOT          || ITOT                  10.00             18.00            -11.00
     VEA           || VEA                   12.00             10.00             14.00
     VHT           || VHT                  106.00             18.00            170.00
   Hoogle:Vacation || VACHR                120.12            120.12             97.02
   Vanguard        || RGAGX      139.973000000000  171.689000000000  178.295000000000
                   || USD                   -0.04              0.04             -0.02
                   || VBMPX      111.933000000000  101.000000000000   97.017000000000
     Cash          || USD                   -0.04              0.04             -0.02
     RGAGX         || RGAGX      139.973000000000  171.689000000000  178.295000000000
     VBMPX         || VBMPX      111.933000000000  101.000000000000   97.017000000000
```

Also, in the normal `balance` mode with text rendering, do we agree that the current output is OK or should the commodity be aligned separately? e.g

```
         5120.50 USD  Assets:US:ETrade:Cash
           70.00 GLD  Assets:US:ETrade:GLD
          17.00 ITOT  Assets:US:ETrade:ITOT
           36.00 VEA  Assets:US:ETrade:VEA
          294.00 VHT  Assets:US:ETrade:VHT
```

vs

```
         5120.50  USD  Assets:US:ETrade:Cash
           70.00  GLD  Assets:US:ETrade:GLD
           17.00  ITOT Assets:US:ETrade:ITOT
           36.00  VEA  Assets:US:ETrade:VEA
          294.00  VHT  Assets:US:ETrade:VHT
```

Xitian9:
For the multi-column report, I think the second is more consistent with how the normal balance report handles things, and saves a lot of clutter. For the question of aligning columns I don't have a strong opinion, but maybe we should leave this question aside as a minor stylistic tweak until we've sorted the primary behaviour.

I think your tree example is an excellent demonstration of the fact that we probably shouldn't worry too much about how it looks in tree mode. It will be clunky and verbose in any case, and flat mode is the primary use case for commodity columns.

simonmichael:
Sorry, can't look in detail right now, perhaps I'm missing something about the subaccount scenario. But to be clear I'm suggesting we ignore it for now, just support column display, and error if it's combined with tree mode.

Xitian9:
I've run the branch locally for a few sample reports, and I have to say it's looking great so far! Very clean and concise. I'll give more in-depth feedback when I've had a chance to look in the code in detail, but the results are great.

Xitian9:
Is it convenient to squash some of your commits so that changes which are later overwritten disappear? That would make review a bit easier.

Xitian9:
We seem to have lost amount elision in ordinary single-line multi-column balance reports. In particular test/balance/multicommodity.test is now failing in tests 1 and 3.

la-wu:
Thanks for taking a look. I've fixed the regression and consolidated the git history as well.

Do we care to label everything as `commodity-column` now?

simonmichael:
> Do we care to label everything as commodity-column now?
> 

Yes --commodity-column sounds decent at the moment.

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -817,17 +823,25 @@ showMixedAmountB opts ma
...
+-- | Helper for showMixedAmountB to show a list of Amounts on multiple lines. This returns
+-- the list of WideBuilders: one for each Amount, and with elision but without
+-- padding.  This does not honour displayOneLine: all amounts will be displayed
+-- as if displayOneLine were False.
+showMixedAmountLinesB' :: (AmountDisplay -> Int -> AmountDisplay)
+                       -> AmountDisplayOpts -> MixedAmount -> [WideBuilder]
+showMixedAmountLinesB' pad opts@AmountDisplayOpts{displayMaxWidth=mmax, displayMinWidth=mmin} ma =
+    map (adBuilder . flip pad width) elided
+  where
+    astrs = amtDisplayList (wbWidth sep) (showAmountB opts) . orderedAmounts opts $
```
Xitian9, line 840: I might be missing something, but `showMixedAmountLinesB'` seems to just be `showMixedAmountLinesB` with the option to not provide any padding. Couldn't this just be achieved with `showMixedAmountLinesB opts{displayMinWidth=Nothing}`, which would not require introducing a new rendering function?

hledger-lib/Text/WideString.hs
```diff
@@ -33,6 +34,10 @@ instance Monoid WideBuilder where
 wbToText :: WideBuilder -> Text
 wbToText = TL.toStrict . TB.toLazyText . wbBuilder
 
+-- | Convert a WideBuilder to a strict Text.
```
Xitian9, line 37: “Convert a strict Text to a WideBuilder”, rather than the other way around.

simonmichael:
Quick test notes, if not out of date:

- --commodity-column help needs updating
- currently affects only csv output
- adds the commodity column, but does not yet remove the commodity symbol from the `balance` column

hledger-lib/Hledger/Data/Amount.hs
la-wu: We pad to the maximum of `displayMinWidth` (if it exists) and the width of the maximum `WideBuilder`. https://github.com/simonmichael/hledger/pull/1626/files/90ed5aa2b43b23eaef87d2248036930b333723c0#diff-a305cb83032e81d09c773ed3ab1fe919f06f482406e4e48aa022e3c8a4c6e83aR843. I didn't spend that much time thinking about a simpler alternative though.

Xitian9: I'd just like to avoid a proliferation of helper functions. I can go through and try to refactor after you submit, if you'd prefer.

la-wu:
> * currently affects only csv output

What do you mean by this?

> * `--commodity-column` help needs updating
> * adds the commodity column, but does not yet remove the commodity symbol from the `balance` column

Done, thanks!

simonmichael:
It should affect both csv and terminal output. When I tested I saw no change in terminal output.

Xitian9:
It was definitely affecting terminal output on the version I saw.

simonmichael:
What am I doing wrong ?
```
22:32:29 ~/src/hledger$ stack exec -- hledger --version
hledger 1.22-84-g41ce169ba, mac-x86_64
22:32:34 ~/src/hledger$ stack exec -- hledger bal -f examples/sample.journal
                  $1  assets:bank:saving
                 $-2  assets:cash
                  $1  expenses:food
                  $1  expenses:supplies
                 $-1  income:gifts
                 $-1  income:salary
                  $1  liabilities:debts
--------------------
                   0  
22:32:42 ~/src/hledger$ stack exec -- hledger bal -f examples/sample.journal --commodity-column
                  $1  assets:bank:saving
                 $-2  assets:cash
                  $1  expenses:food
                  $1  expenses:supplies
                 $-1  income:gifts
                 $-1  income:salary
                  $1  liabilities:debts
--------------------
                   0  
```

simonmichael:
Have I mentioned how much I love the new detailed --version ? I love it.

stack goes one better, and lists the versions of all deps...

Xitian9:
Since the single-column balance report always displays multi-line commodities, and the fact that there's only one column means we don't need to keep things aligned, I wouldn't expect any difference. The difference is is multi-column balance reports.

```
$ stack exec -- hledger --version
hledger 1.22-84-g41ce169ba, linux-x86_64
$ stack exec -- hledger bal -f examples/sample.journal -Y
Balance changes in 2008:

                    || 2008 
====================++======
 assets:bank:saving ||   $1 
 assets:cash        ||  $-2 
 expenses:food      ||   $1 
 expenses:supplies  ||   $1 
 income:gifts       ||  $-1 
 income:salary      ||  $-1 
 liabilities:debts  ||   $1 
--------------------++------
                    ||    0 
$ stack exec -- hledger bal -f examples/sample.journal -Y --commodity-column
Balance changes in 2008:

                    || Commodity  2008 
====================++=================
 assets:bank:saving || $             1 
 assets:cash        || $            -2 
 expenses:food      || $             1 
 expenses:supplies  || $             1 
 income:gifts       || $            -1 
 income:salary      || $            -1 
 liabilities:debts  || $             1 
--------------------++-----------------
                    || $             0 
```

simonmichael:
I see, thanks. We should improve the UX here. Minimising the number of special cases helps users, so ideally this would not be limited to periodic reports ? If it must be so, the help and manual should say so and it should give an error in that case, like with --tree mode - it's less confusing.

hledger-lib/Hledger/Data/Amount.hs
la-wu: I can make it an option on the `AmountDisplayOpts` if that's the preferred method.

la-wu:
> I see, thanks. We should improve the UX here. Minimising the number of special cases helps users, so ideally this would not be limited to periodic reports ?

I don't particularly mind either way we handle it. In case we do want to render the single-period commodity-column differently, would you prefer `%(total) %(commodity) %(account)`, or perhaps `%(account) %(commodity) %(total)`?

>  If it must be so, the help and manual should say so and it should give an error in that case, like with --tree mode - it's less confusing.

For tree mode, commodity-column still does the 'right thing' (see mockups https://github.com/simonmichael/hledger/pull/1626#issuecomment-887150576). I think the main issue is that it is verbose.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: Let me say the three use cases we currently have, to make sure I have my understanding correct. Is this the case?
1. `showMixedAmountLinesB opts{displayMinWidth=Nothing}`: pads each line to the maximum width of all amounts
2. `showMixedAmountLinesB opts{displayMinWidth=Just n}`: pads each line to the maximum width of all amounts, or n if that is greater
3. `showMixedAmountLinesB' const`: does no padding whatsoever

la-wu: Yes, I believe that is the current behavior

Xitian9: I think we can get the exact same behaviour without a helper function. In particular, case 1 can also be achieved with `opts{displayMinWidth=Just 0}`. If we change `opts{displayMinWidth=Nothing}` to do no padding whatsoever, that would encompass all our needs. If we change the `displayMinWidth` field of `noColour` to `Just 0` then nothing would break.

la-wu: Sounds good. I looked at the other usages of `displayMinWidth` and they seem OK with this change and I implemented this.

One behavioral change is with the `StringFormatComponent` parser in case widths are not specified. https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Data/StringFormat.hs#L141. Now the default will not pad anything as well.

Xitian9: Do any of the tests fail? The `renderRow` function will also pad everything out to the same width (previously we were doing this twice: a little inefficient), so I would expect that nothing would break with the change for `StringFormatComponent`

hledger-lib/Text/Tabular/AsciiWide.hs
```diff
@@ -86,20 +92,30 @@ renderTableB :: TableOpts       -- ^ Options controlling Table rendering
              -> (a -> Cell)   -- ^ Function determining the string and width of a cell
              -> Table rh ch a
              -> Builder
-renderTableB topts@TableOpts{prettyTable=pretty, tableBorders=borders} fr fc f (Table rh ch cells) =
+renderTableB topts fr fc f = renderTableB' topts (fmap fc) (\(rh, as) -> (fr rh, fmap f as))
+
+-- | A version of renderTable that operates on rows
+renderTableB' :: TableOpts      -- ^ Options controlling Table rendering
```
Xitian9, line 98: I think we should use a more understandable name here, something like `renderTableByRowsB'`.

hledger-lib/Hledger/Data/Amount.hs
la-wu: The tests pass but there is a difference in output.

With `Just 0` (old behavior when `Just 0 == Nothing`)

```
$ stack exec -- hledger bal -f examples/bcexample.hledger --commodity-column assets.*etrade --format "%(total) %(account)"
5120.50 USD Assets:US:ETrade:Cash
70.00 GLD Assets:US:ETrade:GLD
17.00 ITOT Assets:US:ETrade:ITOT
36.00 VEA Assets:US:ETrade:VEA
294.00 VHT Assets:US:ETrade:VHT
--------------------
  70.00 GLD
 17.00 ITOT
5120.50 USD
  36.00 VEA
 294.00 VHT
```

With `Nothing` (new behavior)

```
5120.50 USD Assets:US:ETrade:Cash
70.00 GLD Assets:US:ETrade:GLD
17.00 ITOT Assets:US:ETrade:ITOT
36.00 VEA Assets:US:ETrade:VEA
294.00 VHT Assets:US:ETrade:VHT
--------------------
70.00 GLD
17.00 ITOT
5120.50 USD
36.00 VEA
294.00 VHT
```

hledger-lib/Text/Tabular/AsciiWide.hs
Xitian9: And perhaps a more detailed haddock.
A version of renderTableB which uses a different rendering function for each row.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: Hmm. Neither of those is ideal output, but we should definitely fix the left-aligned totals field.

It should be sufficient to change `mmin` here to `mmin <|> Just 0`.
https://github.com/simonmichael/hledger/blob/f12ff0df3a194b388a428cd9fd344dcde7925481/hledger/Hledger/Cli/Commands/Balance.hs#L463

Xitian9:
I've taken a look at the remaining code, and it looks great! It's a bit disappointing that budget reports require so much extra code to handle, but they are a bit clunky in any case. I'm sure there's a more principled solution to budget reports in general which will require less cludgy code for many things.

If we feel that a separate commodity column should also be put in for single-column balance reports, I think your proposal looks good to me.
```
         5120.50  USD  Assets:US:ETrade:Cash
           70.00  GLD  Assets:US:ETrade:GLD
           17.00  ITOT Assets:US:ETrade:ITOT
           36.00  VEA  Assets:US:ETrade:VEA
          294.00  VHT  Assets:US:ETrade:VHT
```

hledger-lib/Hledger/Data/Amount.hs
la-wu: I would prefer to change it at parse site so it's not confusing if options are displayed at different places. e.g

```
-    return $ FormatField (isJust leftJustified) (parseDec minWidth) (parseDec maxWidth) f
+    return $ FormatField (isJust leftJustified) (parseMin minWidth) (parseDec maxWidth) f
     where
-      parseDec s = case s of
-        Just text -> Just m where ((m,_):_) = readDec text
-        _ -> Nothing
+      parseDec = fmap $ fst . head . readDec
+      parseMin = Just . fromMaybe 0 . parseDec
```

Thoughts?

Xitian9: Yeah, that sounds good too. Maybe `parseDec minWidth` -> `parseDec minWidth <|> Just 0` would do the job.

la-wu:
> It's a bit disappointing that budget reports require so much extra code to handle, but they are a bit clunky in any case. I'm sure there's a more principled solution to budget reports in general which will require less cludgy code for many things.

Agreed. I tried to get something working cell-wise (by having `total, left-bracket, percent, %-delim, budget, right-bracket` be separate cells) so that alignment / width would be solved by `renderTable*` , but relatively quickly ran into issues trying to align the header columns with that.

> If we feel that a separate commodity column should also be put in for single-column balance reports, I think your proposal looks good to me.

Sounds good, I'll take a look tomorrow. Thanks!

Xitian9:
> Agreed. I tried to get something working cell-wise (by having `total, left-bracket, percent, %-delim, budget, right-bracket` be separate cells) so that alignment / width would be solved by `renderTable*` , but relatively quickly ran into issues trying to align the header columns with that.

Yes, I tried to do that at some point too, and that was exactly the problem I ran into. The ideal solution would be to allow headers to have a column width argument, but this would require re-implementing even more of the `Text.Tabular` structure for our own purposes.

hledger-lib/Hledger/Data/Amount.hs
la-wu: OTOH, this would make it so that on the command-line we wouldn't be able to specify 'no padding' directly (which is consistent with current behavior). That is, `%(total)` is equivalent to `%0(total)`.

la-wu: There is no 'regression' with that but it seems a bit unfortunate.

simonmichael:
Thanks for all the work on this. I'm not following closely just now but in case it's useful to know, I see balance's --format as a bit of a niche feature - I guess it's useful occasionally but it's not something I use personally. Like the non periodic balance report itself, it works well enough that it's hard to justify dropping it. But it's always worth
keeping this option in mind if it's costing progress elsewhere.

la-wu:
For ease of implementation, the single-period balance report in `--commodity-column` does not respect custom formatting.

la-wu:
@Xitian9 @simonmichael What are your thoughts on the PR at this point? Would it be helpful if I rebased/squashed commits again?

simonmichael:
Thanks, I was prioritising other things, I'll check this again soon.

Xitian9:
It's overall looking very good. I'll do a proper code review later, but here are some quick suggestions:
- If it's not too hard, squash together any commits which involve making changes and then undoing them later (e.g. creating and then deleting `showMixedAmountLinesB`)
- Create a test for normal (non-multicolumn) balance report

Xitian9:
Looks good to me! I have no particular code comments.

simonmichael:
Thanks for the review. Looking good @la-wu! I have some minor comments about docs, if you don't mind I'm going to merge and just do them. 

I think this will be a valuable feature, unlocking charting for one thing. Thank you!

simonmichael:
Rendered doc: https://hledger.org/dev/hledger.html#commodity-column

simonmichael:
@la-wu : in the doc, the -T in the CSV example looks redundant. More importantly I couldn't reproduce your example, trying the below; what am I missing ?

```
$ hledger -f examples/bcexample.hledger bal -T -O csv etrade
"account","balance"
"Assets:US:ETrade:Cash","5120.50 USD"
"Assets:US:ETrade:GLD","70.00 GLD"
"Assets:US:ETrade:ITOT","17.00 ITOT"
"Assets:US:ETrade:VEA","36.00 VEA"
"Assets:US:ETrade:VHT","294.00 VHT"
"Income:US:ETrade:Gains","-373.34 USD"
"total","70.00 GLD, 17.00 ITOT, 4747.16 USD, 36.00 VEA, 294.00 VHT"
```

simonmichael:
Ah, found it: -T should be -3. Fixed.

la-wu:
All the examples were based on `-f examples/bcexample.hledger assets.*etrade -3` (which are the flags used in the mockups on the original issue).

Issue #1412, opened 2020-12-13
Auto-Postings don't work on auto-deduced remaining amounts of periodic transactions


nobodyinperson:
Hi, thanks again for `hledger`, it's awesome!

I noticed that [auto postings](https://hledger.org/1.20/journal.html#auto-postings) don't work with auto-deduced remaining amounts of [periodic transactions](https://hledger.org/journal.html#periodic-transactions).

## Steps to Reproduce

Take this `log.hledger` file:

```hledger
~ 2021-01-02   Periodic
    Checking            -10
    Costs

2021-01-01   Fixed
    Checking            -20
    Costs

= acct:Costs
    (Auto)     *1
```

Execute the following:

```bash
hledger --version
# hledger 1.20
hledger -f log.hledger print --forecast --auto --explicit
```

## Observed Output

```hledger
2021-01-01 Fixed  ; modified:
    Checking             -20
    Costs                 20
    (Auto)                20  ; generated-posting: = acct:Costs

2021-01-02 Periodic
    ; generated-transaction: ~ 2021-01-02, modified: 
    Checking             -10
    Costs                 10
    (Auto)                    ; generated-posting: = acct:Costs
                          ^^
                      nothing here
```

## Expected Output

```hledger
2021-01-01 Fixed  ; modified:
    Checking             -20
    Costs                 20
    (Auto)                20  ; generated-posting: = acct:Costs

2021-01-02 Periodic
    ; generated-transaction: ~ 2021-01-02, modified: 
    Checking             -10
    Costs                 10
    (Auto)                10  ; generated-posting: = acct:Costs
                          ^^
                      amount is used
```

## Notes

The empty generated amount also ”breaks” the `register` output. Take for example this `log.hledger` file:

```hledger
~ monthly from 2021-04   Periodic
    Checking            -10
    Costs

2021-03-02   Fixed Nr. 1
    Checking            -20
    Costs

2021-03-31   Fixed Nr. 2
    Checking            -30
    Costs

= acct:Costs
    (Auto)     *1
```
And run the following:

```bash
faketime '2020-12-13' hledger -f log.hledger reg --forecast=today- --auto  Auto
```
Output:

```
2021-03-02 Fixed Nr. 1          (Auto)                          20            20
2021-03-31 Fixed Nr. 2          (Auto)                          30            50
2021-04-01 Periodic             (Auto)                                          
2021-05-01 Periodic             (Auto)                                          
2021-06-01 Periodic             (Auto)   
[...]

PR #1627, opened 2021-07-26
Add balancing amounts in periodic transactions before applying auto postings

Xitian9:
Addresses #1412.

The second commit here is a minimal fix for the problem. The first was work I did while trying to understand where the problem is.

<s>I actually think that generating the forecast transactions should happen in journal finalisation, not as a separate step after the journal has already been finalised.</s> This would avoid having to re-balance the whole journal several times.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -711,9 +711,9 @@ journalUntieTransactions t@Transaction{tpostings=ps} = t{tpostings=map (\p -> p{
 -- relative dates in transaction modifier queries.
 journalModifyTransactions :: Day -> Journal -> Either String Journal
 journalModifyTransactions d j =
-  case modifyTransactions d (jtxnmodifiers j) (jtxns j) of
-    Right ts -> Right j{jtxns=ts}
-    Left err -> Left err
+    case modifyTransactions (journalCommodityStyles j) d (jtxnmodifiers j) (jtxns j) of
+      Right ts -> Right j{jtxns=ts}
+      Left err -> Left err
 
 -- | Check any balance assertions in the journal and return an error message
 -- if any of them fail (or if the transaction balancing they require fails).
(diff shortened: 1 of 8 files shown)


simonmichael:
journalTransform: can you add to the haddock some details of the error case - what can cause this to fail.

journalAddForecast: the old haddock says "...will raise an error". Is that still true ? Can the new code both raise a (program terminating) error,  and return an error message ? Let's add a mention of the error message case to the haddock here also.

Xitian9:
> journalTransform: can you add to the haddock some details of the error case - what can cause this to fail.

Added.

> journalAddForecast: the old haddock says "...will raise an error". Is that still true ?

No, that is not. I have updated the haddock.

simonmichael:
Thanks!

Not a problem for this PR, but is your new style:
```
return $ journalTransform opts =<< files
```
better than the old one ?
```
return $ journalTransform opts <$> files
```
The old one is quite common around the codebase, and to me it's easier to understand here (a chain of $-ish, versus a more complicated mixture of operators..

Xitian9:
> Not a problem for this PR, but is your new style:
> ```
> return $ journalTransform opts =<< files
> ```
> better than the old one ?
> ```
> return $ journalTransform opts <$> files
> ```

It's better in the sense that it works. That `return` is to lift into the `IO` monad, and `files` is of type `Either String Journal`. On the other hand,  the type of `journalTransform opts` has changed from `Journal -> Journal` to `Journal -> Either String Journal`. We therefore need monadic bind (`>>=` or its flipped version `=<<`) instead of `fmap`

simonmichael:
Ah of course, thanks.

I forgot I had some small questions about:

```
lib!: modifyTransactions now takes a Map of commodity styles, and will  …
style amounts according to that argument. journalAddForecast and
journalTransform now return an Either String Journal.

This improves efficiency, as we no longer have to restyle all amounts in
the journal after generating auto postings or periodic transactions.
Changing the return type of journalAddForecast and journalTransform
reduces partiality.
```

Just curious, this means a change in user functionality ? Users will see the standard styles applied where they weren't before, in.. auto posting amounts ? Inferred balancing amounts involving auto postings ?

And "improves efficiency" here is in "in principle", not "measured and noticeable in normal usage" I assume.

Xitian9:
> ```
> lib!: modifyTransactions now takes a Map of commodity styles, and will  …
> style amounts according to that argument. journalAddForecast and
> journalTransform now return an Either String Journal.
> 
> This improves efficiency, as we no longer have to restyle all amounts in
> the journal after generating auto postings or periodic transactions.
> Changing the return type of journalAddForecast and journalTransform
> reduces partiality.
> ```
> 
> Just curious, this means a change in user functionality ? Users will see the standard styles applied where they weren't before, in.. auto posting amounts ? Inferred balancing amounts involving auto postings ?

No change in user behaviour. In every case where these functions were used before, the default styles were applied afterwards by manually calling the function and iterating over the whole journal. Now this is done as part of generating the new postings/transactions, and doesn't need to be done separately. (Which also means it is only done for the new transactions, and not the whole journal.)
 
> And "improves efficiency" here is in "in principle", not "measured and noticeable in normal usage" I assume.

Well, I haven't tested, but I presume so. The main objective here was not to speed things up, but to avoid errors where you forget to re-style the journal.

simonmichael:
Lovely!

Issue #1630, opened 2021-07-28
reg: Related postings and valuation don't play well together.


Xitian9:
Closely related to #1625. Related postings will not be valued, and the current posting will be including without valuation.

```
P 2021-01-01 A  1000 B

2021-01-01
  (a)    1 A
  (b)   10 A @ 0.01 B
  (c)    1 B @ 1 A
  (d)    1 B
```
```
$ hledger reg -f- -V a -r
2021-01-01            (a)                  1 A           1 A
                      (b)                 10 A          11 A
                      (c)                  1 B          11 A
                                                         1 B
                      (d)                  1 B          11 A
                                                         2 B
```

simonmichael:
Thanks for the report. FWIW I never use `-r`, so I wouldn't want to delay #1625 for it, as long as we don't make things worse.

Xitian9:
The solution is simple. It's not delaying anything and I'll submit it along with #1625 when I've finished.

Issue #1629, opened 2021-07-28
reg: Showing related postings when multiple postings match creates many duplicates


Xitian9:
This problem will arise more generally whenever multiple postings in a transaction match the query. Related postings will be pulled in for each matching posting.

```
2021-01-01
  (a)    1 A
  (a)    2 A
  (b)    3 A
```

```
$ hledger reg -f - a -r
2021-01-01            (a)                  2 A           2 A
                      (b)                  3 A           5 A
                      (a)                  1 A           6 A
                      (b)                  3 A           9 A
```

simonmichael:
Wow, I guess this never worked right ? Nice fixes.

Issue #1625, opened 2021-07-23
regression 1.22: cur:/amt: now match after valuation, not before


julian-klode:
I'm trying to figure out how much I hold in a given currency, in my normal currency. This used to work in 1.21, but stopped working in 1.22 (using the docker images atm) - it now always produces 0.

For the given test.journal:
```
P 2021/07/23 test 50€

2021/07/23 buy 10 test
    assets   10 test
    income  -50€
```

Which value you use for `--value` does not matter, fwiw, well I tried `now`, `cost`, `end`.

```
$ hledger -f $PWD/test.journal  bal --value=now cur:test
--------------------
                   0  
$ hledger-1.21 -f $PWD/test.journal  bal --value=now cur:test
                500€  assets
--------------------
                500€  
$ hledger-1.21 -f $PWD/test.journal  bal 
             10 test  assets
                -50€  income
--------------------
             10 test
                -50€  
```

Xitian9:
Hmm. I believe the question is whether we should match currencies before valuation or after valuation. Based on this report, it looks like we used to match before valuation, and now match after.

I would actually expect to match after valuation rather than before, so it's not clear to me which is better. That said, if you are performing valuation, you probably already know which currency you're going to get, in most cases.

Thoughts?

simonmichael:
i'm not sure which is better behaviour either. I'll test a bit with both. 

Even if we decide to keep the new behaviour, I'm inclined to call this a regression since we didn't document it and it broke at least one user's reports. So qualifies for the https://hledger.org/regressionbounty !

julian-klode:
I guess I definitely need at least a switch to revert to the old behavior, otherwise I'd have no way of knowing how much I hold of a given currency (which really are securities here). Otherwise I have to also run `hledger prices` and parse that output and calculate myself which is annoying :)

simonmichael:
There are some other ways (prefilter with print; match by account instead) but they are less accurate, so I hear that. 

That said, it can be a bit confusing for new and forgetful users when cur:, which normally reliably limits the currencies you see, sometimes appears not to do so.  

I wonder about keeping the new cur: behaviour and adding ocur: (match old/original currency) for the old. I can imagine use cases for both kinds of matching, though I'm not sure how important they are. This might be an unnecessary feature. Also it means a change in behaviour, affecting folks like Julian. 

I think this issue. won't come up much with other kinds of query - valuation changes what's being matched which causes the order of operations to matter, but this is unusual.

[...]

PR #1631, opened 2021-07-29
Various fixes for combining transaction filtering with valuation.

Xitian9:
Addresses #1625, #1629, and #1630.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -31,9 +31,11 @@ module Hledger.Data.Journal (
   -- * Filtering
   filterJournalTransactions,
   filterJournalPostings,
+  filterJournalRelatedPostings,
   filterJournalAmounts,
   filterTransactionAmounts,
   filterTransactionPostings,
+  filterTransactionRelatedPostings,
   filterPostingAmount,
   -- * Mapping
   journalMapTransactions,
@@ -105,7 +107,7 @@ import Data.Foldable (toList)
 import Data.Function ((&))
 import qualified Data.HashTable.Class as H (toList)
 import qualified Data.HashTable.ST.Cuckoo as H
-import Data.List (find, foldl', sortOn)
+import Data.List ((\\), find, foldl', sortOn)
 import Data.List.Extra (nubSort)
 import qualified Data.Map.Strict as M
 import Data.Maybe (catMaybes, fromJust, fromMaybe, isJust, mapMaybe, maybeToList)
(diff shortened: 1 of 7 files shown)


simonmichael:
> note that aregister currently only tests queries other than
# "cur:" and the account query when given the --txn-dates option. Who knows why

I didn't recover the full detail (and have some small cleanups to push), but I bet it's something to do with "those are the only kinds of query that don't disturb the historical running balance".

simonmichael:
Looks good to me!

simonmichael:
@Xitian9, I'd like to cherry-pick these from master to 1.22-branch:

- 2da0e6714 * fix: Add balancing amounts in periodic transactions before applying auto-postings. (#1412)
- 9df574b3c * fix: reg: register --related should not create duplicate postings when more than one posting in a transaction matches. (#1629)
- 6528f2559 * fix: lib: Filter postings before performing valuation and costing in MultiBalanceReport, PostingsReport, EntriesReport, and AccountTransactionsReport. (#1625, #1630)

There are conflicts with 6528f2559 at least, I think it depends on some big refactors which I'd rather not take for 1.22-branch. Could you give it a try, and see if you can get these cherry-picked with reasonable confidence ?

Xitian9:
I'll give it a shot.

Xitian9:
I can cherry-pick 2da0e67 and 9df574b pretty easily, but give that 6528f25 still has problems that need to be sorted out with the aregister report, I would hold off on that one until I've sorted those out, if you don't mind.

Xitian9:
Or I can just cherry-pick the changes to non-aregister reports for now, and we can merge the aregister report changes when those are done.

Xitian9:
Well, I managed to cherry-pick all of them (#1635). But we'll still have the problems with aregister, which I presume we'll move over when they're done.

simonmichael:
Thanks! I thought 1.22-branch was free of the aregister breakage. I guess it got pulled in somehow ?

Xitian9:
Indeed, 1.22-branch seems to pass all the new aregister tests. I guess it's fine.

simonmichael:
1.22 branch gets the aregister problem when I merge your last commit:

- 2059d86ab * fix: lib: Filter postings before performing valuation and costing in MultiBalanceReport, PostingsReport, EntriesReport, and AccountTransactionsReport. (#1625, #1630)

Xitian9:
Aha, then the problem is with the fix itself.

Do you think merging #1636 into 1.22-branch is best, or would you rather a more minimal fix?

simonmichael:
Ideally, minimal code changes in 1.22-branch, reducing its scope and risk of additional issues. It's fine if 1.22.1's aregister still has some problems, as long as it's not worse that 1.21. 

Are we in a situation where it's not really feasible to fix just the 1.22 regressions in 1.22.1, without pulling in the extensive refactors/renames/etc. I was trying to avoid ? I can imagine we might have to pull them all in, in that case.

Xitian9:
I think it should be possible to merge #1636 without bringing any other refactors. However, #1636 itself is a big refactor of `accountTransactionsReport`.

If we're looking to avoid those, I would drop the aregister changes from #1635. We would still be left with cur: and amt: matching after valuation rather than before, but we wouldn't have the historical postings problem.

simonmichael:
Well.. matching after valuation is one of the changes we want to undo in 1.22.1, so we need that.

simonmichael:
(If the commits are just too interdependent, let me know and I'll have a look at merging "everything" into 1.22-branch.)

Xitian9:
I think just merge #1636 in and that will fix it. Nothing else should be required.

simonmichael:
> I think just merge #1636 <https://github.com/simonmichael/hledger/pull/1636> in and that will fix it.
> 
Easier said than done.. cherry picking has lots of conflicts. Literally merging merges everything from master - 112 commits and some conflicts.

Xitian9:
Really? That surprises me. Let me give it a shot.

Xitian9:
I've now cherry-picked the commits from #1636 and included them in #1635.

Issue #1517, opened 2021-03-28
New periodic expression request: ~ every weekday [$50]


makdisse:
### Use case

For anyone that does contract work or any type of business relationship that has a daily rate.
At the moment we can do the days of the week, eg: `~ every monday`, so it is possible to achieve what `~every weekday` would do.  But when you have to do it for multiple accounts it gets messy and hard to maintain. 

Having `weekday` available as a period would allow us to properly capture daily rates for months with a variable number of days and weekends.

My particular use case is to forecast the pay for contractor workers.

### Behaviour

Pretty much like `~ every day` but only considering `mon|tue|wed|thu|fri`

**Example:**

```
~ every weekday
  Income:Client1          -10 GBP
  Assets:Contractors:Contractor1
```

Executed with `hledger.exe -f .\test.ledger reg -b "2021-04-01" -e "2021-04-15" --forecast -w 120` should result in:

```
2021-04-01                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-02                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-05                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-06                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-07                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-08                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-09                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-12                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0
2021-04-13                                          Assets:Contractors:Contractor1                  10 GBP        10 GBP
                                                    Income:Client1                                 -10 GBP             0

```

### Bonus

Maybe getting `~ every weekend` will be a nice bonus for feature completeness.
It could be used by people who provide contracting work only on weekends I guess.

simonmichael:
Nice proposal!

[...]

PR #1632, opened 2021-07-30
New periodic expressions for every weekday et al

la-wu:
Possible implementation for https://github.com/simonmichael/hledger/issues/1517

### DIFF
--- a/hledger-lib/Hledger/Data.hs
+++ b/hledger-lib/Hledger/Data.hs
@@ -54,6 +54,7 @@ import Hledger.Utils.Test
 tests_Data = tests "Data" [
    tests_AccountName
   ,tests_Amount
+  ,tests_Dates
   ,tests_Journal
   ,tests_Ledger
   ,tests_Posting
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -73,6 +73,8 @@ module Hledger.Data.Dates (
   yearp,
   daysInSpan,
   maybePeriod,
+
+  tests_Dates
 )
 where
 
@@ -193,7 +195,7 @@ spansSpan spans = DateSpan (maybe Nothing spanStart $ headMay spans) (maybe Noth
 -- [DateSpan 2007-12-02..2008-01-01,DateSpan 2008-01-02..2008-02-01,DateSpan 2008-02-02..2008-03-01,DateSpan 2008-03-02..2008-04-01]
 -- >>> t (WeekdayOfMonth 2 4) 2011 01 01 2011 02 15
 -- [DateSpan 2010-12-09..2011-01-12,DateSpan 2011-01-13..2011-02-09,DateSpan 2011-02-10..2011-03-09]
--- >>> t (DayOfWeek 2) 2011 01 01 2011 01 15
+-- >>> t (DaysOfWeek [2]) 2011 01 01 2011 01 15
 -- [DateSpan 2010-12-28..2011-01-03,DateSpan 2011-01-04..2011-01-10,DateSpan 2011-01-11..2011-01-17]
 -- >>> t (DayOfYear 11 29) 2011 10 01 2011 10 15
 -- [DateSpan 2010-11-29..2011-11-28]
@@ -211,7 +213,19 @@ splitSpan (Quarters n)   s = splitspan startofquarter (applyN n nextquarter) s
 splitSpan (Years n)      s = splitspan startofyear    (applyN n nextyear)    s
 splitSpan (DayOfMonth n) s = splitspan (nthdayofmonthcontaining n) (nthdayofmonth n . nextmonth) s
 splitSpan (WeekdayOfMonth n wd) s = splitspan (nthweekdayofmonthcontaining n wd) (advancetonthweekday n wd . nextmonth) s
-splitSpan (DayOfWeek n)  s = splitspan (nthdayofweekcontaining n)  (applyN (n-1) nextday . nextweek)  s
+splitSpan (DaysOfWeek []) s = [s]  -- shouldn't happen in parser but for completeness
+splitSpan (DaysOfWeek days@(n:_)) ds
+  | DateSpan Nothing  (Just e)  <- ds = split (DateSpan (Just $ start e) (Just $ nextday $ start e))
+  | DateSpan (Just s) Nothing  <- ds = split (DateSpan (Just $ start s) (Just $ nextday $ start s))
+  | DateSpan (Just s) (Just e) <- ds =
+      if s == e then [ds] else split (DateSpan (Just $ start s) (Just e))
+  where
+    start = nthdayofweekcontaining n
+
+    wheel = (\x -> zipWith (-) (tail x) x) . concat . zipWith fmap (fmap (+) [0,7..]) . repeat $ days
+
+    split = splitspan' (repeat startofday) (fmap (flip applyN nextday) wheel)
+
 splitSpan (DayOfYear m n) s = splitspan (nthdayofyearcontaining m n) (applyN (n-1) nextday . applyN (m-1) nextmonth . nextyear) s
 -- splitSpan (WeekOfYear n)    s = splitspan startofweek    (applyN n nextweek)    s
 -- splitSpan (MonthOfYear n)   s = splitspan startofmonth   (applyN n nextmonth)   s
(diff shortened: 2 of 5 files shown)


simonmichael:
Nice! Quick thoughts:

- I'd prefer `every weekendday` to `every weekend` - though it looks weird I think it would be clearer, `every weekend` sounds like one event per week.
- Are both singular and plural forms accepted ? I see `weekdays` in the doc.
- Would it not be better to keep DayOfWeek in case we ever want to represent a single day, and work with [DayOfWeek] ?
- The `every <multiple disjoint selectors>`, like `every mon|thu` is interesting. I think that's new behaviour. Period expressions continue to grow a bit out of control - it might be ok to add it as a special case for now, but is it something we should think about supporting more generally ? `every Q1|Q3`, `every feb|aug`, etc.

la-wu:
> * I'd prefer `every weekendday` to `every weekend` - though it looks weird I think it would be clearer, `every weekend` sounds like one event per week.

Sure, I'll change that. Probably good to make it seem weird and obvious rather than use `weekend` with potential ambiguity.

> * Are both singular and plural forms accepted ? I see `weekdays` in the doc.

Yes it supports the 'compact' forms `~ weekdays` and `~ weekenddays` along with the `every <singular>`

> * Would it not be better to keep `DayOfWeek` in case we ever want to represent a single day, and work with `[DayOfWeek]` ?

I intended to represent a single day by having `DaysOfWeek` contain exactly 1 element. Since  `DaysOfWeek` should never be empty, potentially a non-empty list would be clearer?

> * The `every <multiple|disjoint|selectors>` (`every mon|wed|fri`) is interesting. I think that's a new capability ? Period expressions continue to grow a bit out of control - it might be ok to add it as a special case for now, but is it something we should think about supporting more generally ? `every Q1|Q3`, `every feb|aug`, `2020|2022|2024` (<- no `every`), etc. ?

Yes I thought it was a relatively straightforward parser to create and solves the issue mentioned in https://github.com/simonmichael/hledger/issues/1517#issuecomment-848592937 where we might have different spans represent the 'weekdays' vs 'weekenddays'. I think limiting the usage would be a reasonable start. Unless there are other compelling examples where it could be useful.

> * [Or contrarily, is it time to think about supporting more standard ISO recurring syntax ?]

I think that could be done independently of this particular feature. Is there a common use case for said syntax?

simonmichael:
> The periods will be built between specified days.

This wording confused me, "between" sounds like the span between the days would be included. I'll adjust it a little while resolving the conflict.

simonmichael:
https://hledger.org/hledger.html#period-expressions changed today, and I am reconciling this with it. Here's the WIP doc:

> Weekly on custom day(s):
> 
> - `every Nth day of week` (`th`, `nd`, `rd`, or `st` are all accepted after the number)
> - `every WEEKDAYNAMES[|WEEKDAYNAME|...]` (one or more full or three-letter english weekday names, case insensitive)
> - `every weekday` (same as `every mon|tue|wed|thu|fri`)
> - `every weekendday` (same as `every sat|sun`)
> 
> (The last three above are the only form that generate multiple dates per interval.)

Are the alternate `~ weekdays` and `~ weekenddays` forms worth it ? I'm inclined to omit them to keep things feeling a bit more consistent.

Note the final sentence. All other forms specify one date within each interval/subperiod. I feel there might be a bit of ambiguity here, which this doc also touches on:

>  `-p "every 2nd day of week"` | periods will go from Tue to Tue

Ie, most of the time we are interested in the intervals bounded by these dates, so eg `register -p 'every tue'` groups the postings in each tuesday..tuesday period. So what does `register -p 'every tue|wed|fri'` do ? I think it still reports all the postings in the week, but with the week further divided into three periods (tuesday, wednesday-thursday, and friday-monday). I guess that's ok.

I found myself wanting to write -p 'every tue,wed,fri' instead. This might be clearer and more shell friendly ?

simonmichael:
Just  to complete that thought: we interpret period expressions in two different ways:

1. With `-p`, the generated dates delimit report subperiods.
2. In `~` periodic transaction rules, with `--forecast` or `balance --budget`, each date generates a transaction.

simonmichael:
> So what does register -p 'every tue|wed|fri' do ?

Similarly, `register -p 'every weekday'` creates five periods per week: mon, tue, wed, thu, fri-sun:
```
~ every day
    (a)  1
```
```
$ hledger reg --forecast date:thismonth
2021-08-17                                 (a)                                        1             1
2021-08-18                                 (a)                                        1             2
2021-08-19                                 (a)                                        1             3
2021-08-20                                 (a)                                        1             4
2021-08-21                                 (a)                                        1             5
2021-08-22                                 (a)                                        1             6
2021-08-23                                 (a)                                        1             7
2021-08-24                                 (a)                                        1             8
2021-08-25                                 (a)                                        1             9
2021-08-26                                 (a)                                        1            10
2021-08-27                                 (a)                                        1            11
2021-08-28                                 (a)                                        1            12
2021-08-29                                 (a)                                        1            13
2021-08-30                                 (a)                                        1            14
2021-08-31                                 (a)                                        1            15
$ hledger reg --forecast date:thismonth -p 'every weekday'
2021-08-17              a                                                             1             1
2021-08-18              a                                                             1             2
2021-08-19              a                                                             1             3
2021-08-20..2021-08..   a                                                             3             6
2021-08-23              a                                                             1             7
2021-08-24              a                                                             1             8
2021-08-25              a                                                             1             9
2021-08-26              a                                                             1            10
2021-08-27..2021-08..   a                                                             3            13
2021-08-30              a                                                             1            14
2021-08-31              a                                                             1            15
```
(Using a periodic rule and --forecast here just to generate some transactions, then -p to group these into subperiods.)

simonmichael:
So is the above ok ? Alternative behaviour would be for `register -p 'every weekday'` to show five one-day-periods per week: mon, tue, wed, thu, fri, with sat-sun excluded from the report. This would be new behaviour: non-adjacent report periods, which might cause problems somewhere else; more testing would be needed.

This is a small niche feature at this stage, but we like to sweat the details (so users don't have to). Which `register -p` behaviour seems more intuitive here ?

lubieowoce:
> Alternative behaviour would be for `register -p 'every weekday'` to show five one-day-periods per week: mon, tue, wed, thu, fri, with sat-sun excluded from the report.

this is what i'd expect. the first behavior - gluing fri-sat-sun together into one period (AFAIU) - seems pretty counterintuitive to me – wouldn't that cause the report to include transactions that happened on sat/sun?

simonmichael:
Yes it does. I agree it's counterintuitive, at first. It is consistent though. 

If on the contrary `reg -p "every weekday"` generates five single-day periods per week and excludes sat/sun, then we'd expect `reg -p "every mon|tue"` to generate two single-day periods per week, and `reg -p every tue` to generate one single-day period per week... ? No, traditionally this last one generates a seven-day period per week. 

You could say there's a new rule: when multiple days are specified (|, weekday, weekendday), disjoint periods will be generated, one for each day. Is that intuitive and useful enough to be worthwhile ? Will it generalise to other, larger kinds of disjoint period we may add later ? Do we have enough practical uses that it's worth spending time on this now ?

Or, we can postpone that and just document the current behaviour, on the basis that it is intended for forecasting/budgeting and defining strange sub-week report periods isn't needed.

lubieowoce:
made some ascii diagrams to verify my understanding.
a `===>` symbolizes a period (right-exclusive) on a timeline. the periods above the number line are according to option 1, the periods below it are option 2.

```
every weekday = every mon|tue|wed|thu|fri

|==>|==>|==>|==>|==========>
1---2---3---4---5---6---7---
|==>|==>|==>|==>|==>

every tue|thu

===>|======>|===============
1---2---3---4---5---6---7---
    |==>    |==>

1. (above line) every x, start a new period

2. (below line) every x, start a new period lasting y
      in this case, y = 1 day.
      possible default: y = time resolution used in rule.
       every <day>   => y = 1d
       every <month> => y = 1mo
```

simonmichael:
That's correct. Currently, -p always generates adjacent report periods with no gaps between. Proposed option 2 would sometimes generates disjoint periods with gaps.

la-wu:
The implemented behavior was chosen since the report intervals are really defined by the start dates of each interval. Thus the period expressions where multiple days are specified can be thought of as starting a new sub-period every specified day. With that in mind, I don't think this interval behavior is any less intuitive than that `every monday` generates intervals that contain non-mondays (it generates week-long intervals beginning every monday). Perhaps we could indicate that the report intervals are specified as `beginning on every ...` with the syntax or documentation.

I do think it's awkward that intervals are no longer always directly comparable since they span a different number of days and I'm not sure how useful the reports will be when aggregated as such. I've never needed such granular breakdowns so I don't have a good use case for either aggregation behavior.

simonmichael:
I'd like get this merged without our spending much more time on it. @la-wu, I heard no objection so could you

- drop the `weekdays`/`weekenddays` alternative spellings
- use `,` instead of `|`

and we'll leave it working as it is now - generating adjacent but possibly uneven periods - for now, avoiding additional work and risk.

simonmichael:
@la-wu I meant to get rid of just `weekdays` and `weekenddays`, but to keep `every weekday` and `everyweekendday` - those are the original goal after all. Sorry for the confusion.

simonmichael:
Your doc was good but I couldn't resist trying to clarify things a bit more - sorry
(cf https://hledger.org/dev/hledger.html#periods-or-dates-). 
This is now merged, thank you!
(Also, --forecast=PERIOD now rejects a report interval, since I got confused wondering why `--forecast='every weekendday from today to next month'` didn't work.)

PR #1635, opened 2021-08-02
Cherry-picking #1631 fixes for 1.22

Xitian9:
[For #1631]


Xitian9:
Now also includes commits cherry-picked from #1636.

simonmichael:
Thanks! I was navigating the patch dependencies and making progress, but this is much better.

Issue #1634, opened 2021-07-31
aregister miscalculates initial balance (master)


simonmichael:
In current master (hledger 1.22-103-gc537f9426), `areg -b` is ignoring the begin date and doubling the initial running balance:

    2021-07-26
      (assets)  1000

    $ hledger areg -w80 assets -b 2021-07-31 -e 2021-08-05
    Transactions in assets and subaccounts:
    2021-07-26                      assets                        1000          2000

Xitian9:
As expected, the error is introduced in 6528f2559. I'll look into this.

aregister has very few tests in the suite, so this is probably a good opportunity to introduce more.

Xitian9:
Okay, here's the nub of the problem.

In order to have a proper running balance, `aregister` needs to hold off on filtering transactions until after it has calculated the running balance. But in order for `cur:` and `amt:` to match the pre-valuation currency and amount, rather than the post-valuation currency and amount, we need to filter before doing valuation. I guess a solution is to calculate the running balance with un-valued postings, then value it after all is said and done, but this does not work with `--value=then` valuation, which needs to value before adding all the numbers together.

This is a bit knotty. I think one problem is that I don't have a clear conception of what the running balance should include. It seems that it needs to include all changes to that account, even those that are not displayed in the report. Is this correct?

simonmichael:
IIRC, aregister's running balance should match what you'd see in the bank's online register. This would seem to permit very limited filtering, if any. And yet, https://hledger.org/dev/hledger.html#aregister says "Any additional arguments form a query which will filter the transactions shown.". The example shows only a `date:` query, which is non-problematic. I'll have to play around a bit more to find out.

simonmichael:
If we can "just" get this second test passing again as in 1.22, that would be a good step:
```
<
2021-01-01
  (a)  1

2021-01-02
  (a:aa)  10

2021-01-03
  (a:aa:aaa)  100

# aregister always includes transactions in subaccounts, ignoring a depth limit #1448
$ hledger -f- areg -w80 a -1
Transactions in a and subaccounts:
2021-01-01                      a                                1             1
2021-01-02                      a:aa                            10            11
2021-01-03                      a:aa:aaa                       100           111

# aregister is always in historical mode, showing balance from prior transactions
$ hledger -f- areg -w80 a -b 2021-01-02
Transactions in a and subaccounts:
2021-01-02                      a:aa                            10            11
2021-01-03                      a:aa:aaa                       100           111
```

[...]

PR #1636, opened 2021-08-02
Make sure aregister correctly identifies historical postings (#1634)

Xitian9:
<s>Explicitly avoid filtering by depth queries, perform valuation after all reportq filtering is done (in both the historical and recent postings), and make sure filtering is done at the correct point based on whether --txn-dates is set.</s> (#1634)

Make sure the reportq properly excludes depth, currency, and amount queries. Do not transform transactions (except as documented) unless we keep the original transaction.

### DIFF
--- a/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
+++ b/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
@@ -23,9 +23,10 @@ module Hledger.Reports.AccountTransactionsReport (
 )
 where
 
-import Data.List (mapAccumL, nub, partition, sortOn)
+import Data.List (mapAccumR, nub, partition, sortOn)
 import Data.List.Extra (nubSort)
 import Data.Maybe (catMaybes)
+import Data.Ord (Down(..))
 import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Time.Calendar (Day, addDays)
@@ -116,7 +117,7 @@ accountTransactionsReport rspec@ReportSpec{_rsReportOpts=ropts} j thisacctq = it
     -- sort by the transaction's register date, for accurate starting balance
     transactions =
         ptraceAtWith 5 (("ts4:\n"++).pshowTransactions)
-      . sortOn (transactionRegisterDate reportq thisacctq)
+      . sortOn (Down . transactionRegisterDate reportq thisacctq)
       . jtxns
       . ptraceAtWith 5 (("ts3:\n"++).pshowTransactions.jtxns)
       -- maybe convert these transactions to cost or value
@@ -154,7 +155,7 @@ accountTransactionsReport rspec@ReportSpec{_rsReportOpts=ropts} j thisacctq = it
         mstartdate = queryStartDate (date2_ ropts) reportq
         datelessreportq = filterQuery (not . queryIsDateOrDate2) reportq
 
-    items = reverse $ accountTransactionsReportItems reportq thisacctq startbal maNegate transactions
+    items = accountTransactionsReportItems reportq thisacctq startbal maNegate transactions
 
 pshowTransactions :: [Transaction] -> String
 pshowTransactions = pshow . map (\t -> unwords [show $ tdate t, T.unpack $ tdescription t])
@@ -166,7 +167,7 @@ pshowTransactions = pshow . map (\t -> unwords [show $ tdate t, T.unpack $ tdesc
 accountTransactionsReportItems :: Query -> Query -> MixedAmount -> (MixedAmount -> MixedAmount) -> [Transaction] -> [AccountTransactionsReportItem]
 accountTransactionsReportItems reportq thisacctq bal signfn =
     catMaybes . snd .
-    mapAccumL (accountTransactionsReportItem reportq thisacctq signfn) bal
+    mapAccumR (accountTransactionsReportItem reportq thisacctq signfn) bal
 
 accountTransactionsReportItem :: Query -> Query -> (MixedAmount -> MixedAmount) -> MixedAmount -> Transaction -> (MixedAmount, Maybe AccountTransactionsReportItem)
 accountTransactionsReportItem reportq thisacctq signfn bal torig = balItem
--- a/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
+++ b/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
@@ -96,8 +96,9 @@ accountTransactionsReport rspec@ReportSpec{_rsReportOpts=ropts} j thisacctq = it
   where
     -- a depth limit should not affect the account transactions report
     -- seems unnecessary for some reason XXX
-    reportq = simplifyQuery $ And [_rsQuery rspec, periodq, excludeforecastq (forecast_ ropts)]
+    reportq = simplifyQuery $ And [depthlessq, periodq, excludeforecastq (forecast_ ropts)]
       where
+        depthlessq = filterQuery (not . queryIsDepth) $ _rsQuery rspec
         periodq = Date . periodAsDateSpan $ period_ ropts
         -- Except in forecast mode, exclude future/forecast transactions.
         excludeforecastq (Just _) = Any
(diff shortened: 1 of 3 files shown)


Xitian9:
There is some differing behaviour here in terms of whether depth queries are ever respected. My reading of the docs is that they should not be, but I may be misinterpreting.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
```diff
@@ -94,63 +95,71 @@ triCommodityBalance c = filterMixedAmountByCommodity c  . triBalance
...
-      -- apply any cur:SYM filters in reportq
-      $ if queryIsNull symq then j else filterJournalAmounts symq j
+    -- Note that within this functions, we are only allowed limited
+    -- transformation of the transaction postings: this is due to the need to
+    -- pass the original transactions into accountTransactionsReportItem.
+    -- Generally, we either include a transaction in full, or not at all.
+    -- We only allow ourselves to:
+    --   - filter amounts (e.g. cur: or amt:)
+    --   - apply valuation or costing
+    --
```
simonmichael, line 120: I wasn't sure if this meant filtering postings within transactions or whole transactions. How about:
```
  - filter transactions by amt: and/or cur:
  - apply valuation or costing to transaction amounts
```
I'm still a little confused, why can we filter by cur:/amt: here but not desc: or status: ?

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
```diff
@@ -159,28 +168,27 @@ pshowTransactions = pshow . map (\t -> unwords [show $ tdate t, T.unpack $ tdesc
 -- using the provided user-specified report query, a query specifying
 -- which account to use as the focus, a starting balance, a sign-setting
 -- function and a balance-summing function.
```
simonmichael, line 170: I think "and a balance-summing function" can be dropped, right ?

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
```diff
@@ -159,28 +168,27 @@ pshowTransactions = pshow . map (\t -> unwords [show $ tdate t, T.unpack $ tdesc
 -- using the provided user-specified report query, a query specifying
```
simonmichael, line 168: "transactions" -> "transactions (along with their original dates)" ?
Why must we provide these dates, are they different from what's in the transactions ?

hledger/test/journal/amounts-and-commodities.test
```diff
@@ -327,7 +327,7 @@ $ hledger -f- register -V -B cur:A "amt:<5"
 # 24. aregister report cur: and amt: query matches currency before
 # valuation/cost (note that aregister currently only tests queries other than
 # "cur:" and the account query when given the --txn-dates option. Who knows why.
-$ hledger -f- aregister a -V -B cur:A "amt:<5" --txn-dates
+$ hledger -f- aregister a -V -B cur:A "amt:<5"
```
simonmichael, line 330: "aregister currently only tests queries other than cur: and the account query when given the --txn-dates option." - this can be dropped, right ? Seems wrong in all recent versions.

simonmichael (commented):
Thanks, this seems to be working great ! I have only some minor questions/notes for code readability.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
```diff
@@ -94,63 +95,71 @@ triCommodityBalance c = filterMixedAmountByCommodity c  . triBalance
...
         -- Except in forecast mode, exclude future/forecast transactions.
         excludeforecastq (Just _) = Any
         excludeforecastq Nothing  =  -- not:date:tomorrow- not:tag:generated-transaction
           And [ Not . Date $ DateSpan (Just . addDays 1 $ _rsDay rspec) Nothing
               , Not generatedTransactionTag
               ]
-    symq    = filterQuery queryIsSym reportq
-    realq   = filterQuery queryIsReal reportq
-    statusq = filterQuery queryIsStatus reportq
+    symq = filterQuery queryIsSymOrAmt $ _rsQuery rspec
```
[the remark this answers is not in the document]
Xitian9: Sure thing. I just continued to use the established naming system, but I agree that it's less understandable. I'll change it.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
```diff
@@ -94,63 +95,71 @@ triCommodityBalance c = filterMixedAmountByCommodity c  . triBalance
...
 accountTransactionsReport rspec@ReportSpec{_rsReportOpts=ropts} j thisacctq = items
   where
-    -- a depth limit should not affect the account transactions report
-    -- seems unnecessary for some reason XXX
-    reportq = simplifyQuery $ And [_rsQuery rspec, periodq, excludeforecastq (forecast_ ropts)]
+    -- a depth limit should not affect the account transactions report, nor
+    -- should cur or amt queries, as they are handled separately
+    reportq = simplifyQuery $ And [depthlessq, periodq, excludeforecastq (forecast_ ropts)]
       where
+        depthlessq = filterQuery (not . queryIsSymOrAmt) . filterQuery (not . queryIsDepth) $ _rsQuery rspec
```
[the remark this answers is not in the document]
Xitian9: Good point. Will change.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
Xitian9: These were chose based on going back to hledger-1.20 and documenting what was done there, rather than a systematic consideration of what would be appropriate. I could update the comment to give the rationale and list these as examples.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
```diff
@@ -94,63 +95,71 @@ triCommodityBalance c = filterMixedAmountByCommodity c  . triBalance
 accountTransactionsReport :: ReportSpec -> Journal -> Query -> AccountTransactionsReport
 accountTransactionsReport rspec@ReportSpec{_rsReportOpts=ropts} j thisacctq = items
   where
-    -- a depth limit should not affect the account transactions report
-    -- seems unnecessary for some reason XXX
-    reportq = simplifyQuery $ And [_rsQuery rspec, periodq, excludeforecastq (forecast_ ropts)]
+    -- a depth limit should not affect the account transactions report, nor
+    -- should cur or amt queries, as they are handled separately
```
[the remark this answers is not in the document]
Xitian9: They are handled separately because they need to be matched before valuation. They are then removed from the query, because otherwise it tries to match again after valuation, which will mean any cur: queries will fail if they have been converted.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
Xitian9: Probably true.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
Xitian9: Yes, it is the minimum of the posting dates which match both `thisacctq` and `reportq`, or `tdate` if there are no matching postings.

hledger/test/journal/amounts-and-commodities.test
Xitian9: Yes, I'll drop it.

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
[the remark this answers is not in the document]
simonmichael: Good. What do you think about
```
    -- A depth limit should not affect the account transactions report; it should show all transactions in/below this account.
    -- Queries on currency or amount are also ignored at this stage; they are handled earlier, before valuation.
```

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
simonmichael: Maybe this would be enough ?
```
-- Do some limited filtering and valuing of the journal's transactions:
-- filter them by the account query if any,
-- discard amounts not matched by the currency query if any,
-- then apply valuation if any.
```

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
```diff
@@ -94,63 +95,71 @@ triCommodityBalance c = filterMixedAmountByCommodity c  . triBalance
...
+    -- transformation of the transaction postings: this is due to the need to
+    -- pass the original transactions into accountTransactionsReportItem.
+    -- Generally, we either include a transaction in full, or not at all.
+    -- We only allow ourselves to:
+    --   - filter amounts (e.g. cur: or amt:)
+    --   - apply valuation or costing
+    --
+    -- accountTransactionsReportItem will keep transactions of any date which
+    -- have any posting inside the report period.
+    -- Should we also require that transaction date is inside the report period ?
```
simonmichael, line 123: This comment about dates might be obsolete, related to --txn-dates, and it's unclear - perhaps we should just drop it ?

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
simonmichael: What do you think about
```
-- | Generate transactions report items from a list of transactions,
-- using the provided user-specified report query, a query specifying
-- which account to use as the focus, a starting balance, and a sign-setting
-- function.
-- Each transaction is accompanied by the date that should be shown for it 
-- in the report, which is not necessarily the transaction date; it is
-- the earliest of the posting dates which match both thisacctq and reportq, 
-- otherwise the transaction's date if there are no matching postings.
```

hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
Xitian9: I agree. I couldn't find any code which specifically implements this, so I think it's no longer accurate.

PR #1639, opened 2021-08-04
Add forecast transactions at journal finalisation, rather than as a post-processing step.

### LINKED ISSUE #1638: v1.22.1 breaks auto-deduced amounts of forecasted balance assertions
Hi,

`hledger-1.22.1` introduces a regression preventing auto-deduced amounts of forecasted balance assertions of being calculated.

## How To Reproduce

Take this `log.hledger` file:

```ledger
2021-09-01   Normal Balance Assertion Works
    Checking   = -60
    Costs

~ 2021-10-01       explicit forecasted assertion
    Checking      = -100
    Costs             40

~ 2021-10-02       auto-deduced forecasted assertion
    Checking      = -100
    Costs
```

```bash
# Works with v1.22
➤  hledger-1.22 reg -f log.hledger --forecast 
2021-09-01 Normal Balance Assertion Works          Checking                  -60           -60
                                                   Costs                      60             0
2021-10-01 explicit forecasted assertion           Checking                  -40           -40
                                                   Costs                      40             0
2021-10-02 auto-deduced forecasted assertion       Checking                    0             0
                                                   Costs                       0             0
# Doesn't work with v1.22.1
➤  hledger-1.22.1 reg -f log.hledger --forecast
hledger-1.22.1: "" (lines 1-1)
could not balance this transaction:
can't have more than one real posting with no amount
(remember to put two or more spaces between account and amount)
2021-10-02 auto-deduced forecasted assertion
    ; generated-transaction: ~ 2021-10-02
    Checking                 = -100
    Costs

  error, called at ./Hledger/Cli/Utils.hs:125:27 in hledger-1.22.1-inplace:Hledger.Cli.Utils
```

Is this intended? Auto-deduced amounts for forecasted balance assertions are really handy to say e.g. ”I'll move all of the money from that account to another one at that time”.

Platform: up-to-date Manjaro x86_64 with `hledger` 1.22 and 1.22.1

Xitian9:
Fixes #1638.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -49,6 +49,8 @@ module Hledger.Read.Common (
   journalCheckAccountsDeclared,
   journalCheckCommoditiesDeclared,
   journalCheckPayeesDeclared,
+  journalAddForecast,
+  journalAddAutoPostings,
   setYear,
   getYear,
   setDefaultCommodityAndStyle,
@@ -135,7 +137,9 @@ import Data.Bifunctor (bimap, second)
 import Data.Char (digitToInt, isDigit, isSpace)
 import Data.Decimal (DecimalRaw (Decimal), Decimal)
 import Data.Default (Default(..))
+import Data.Either (lefts, rights)
 import Data.Function ((&))
+import Data.Functor ((<&>))
 import Data.Functor.Identity (Identity)
 import "base-compat-batteries" Data.List.Compat
 import Data.List.NonEmpty (NonEmpty(..))
@@ -144,7 +148,7 @@ import qualified Data.Map as M
 import qualified Data.Semigroup as Sem
 import Data.Text (Text)
 import qualified Data.Text as T
-import Data.Time.Calendar (Day, fromGregorianValid, toGregorian)
+import Data.Time.Calendar (Day, addDays, fromGregorianValid, toGregorian)
 import Data.Time.LocalTime (LocalTime(..), TimeOfDay(..))
 import Data.Word (Word8)
 import System.Time (getClockTime)
@@ -156,6 +160,8 @@ import Text.Megaparsec.Custom
   finalErrorBundlePretty, parseErrorAt, parseErrorAtRegion)
 
 import Hledger.Data
+import Hledger.Query (Query(..), filterQuery, parseQueryTerm, queryEndDate, queryIsDate, simplifyQuery)
+import Hledger.Reports.ReportOptions (ReportOpts(..), queryFromFlags, rawOptsToReportOpts)
 import Hledger.Utils
 import Text.Printf (printf)
 
@@ -228,6 +234,8 @@ definputopts = InputOpts
     , strict_            = False
     }
 
+-- | Parse an InputOpts from a RawOpts and the current date.
+-- This will fail with a usage error if the forecast period expression cannot be parsed.
 rawOptsToInputOpts :: RawOpts -> IO InputOpts
 rawOptsToInputOpts rawopts = do
     d <- getCurrentDay
@@ -250,16 +258,25 @@ rawOptsToInputOpts rawopts = do
       }
   where noinferprice = boolopt "strict" rawopts || stringopt "args" rawopts == "balancednoautoconversion"
 
--- | get period expression from --forecast option
+-- | Get period expression from --forecast option.
+-- This will fail with a usage error if the forecast period expression cannot be parsed.
 forecastPeriodFromRawOpts :: Day -> RawOpts -> Maybe DateSpan
-forecastPeriodFromRawOpts d opts =
-  case maybestringopt "forecast" opts
-  of
+forecastPeriodFromRawOpts d rawopts = case maybestringopt "forecast" rawopts of
     Nothing -> Nothing
-    Just "" -> Just nulldatespan
-    Just str ->
-      either (\e -> usageError $ "could not parse forecast period : "++customErrorBundlePretty e) (Just . snd) $ 
-      parsePeriodExpr d $ stripquotes $ T.pack str
+    Just "" -> Just forecastspanDefault
+    Just str -> either (\e -> usageError $ "could not parse forecast period : "++customErrorBundlePretty e)
+                       (\(_,requestedspan) -> Just $ requestedspan `spanDefaultsFrom` forecastspanDefault) $
+                  parsePeriodExpr d $ stripquotes $ T.pack str
+  where
+    -- "They end on or before the specified report end date, or 180 days from today if unspecified."
+    mspecifiedend = dbg2 "specifieddates" $ queryEndDate False datequery
+    forecastendDefault = dbg2 "forecastendDefault" $ addDays 180 d
+    forecastspanDefault = DateSpan Nothing $ mspecifiedend <|> Just forecastendDefault
+    -- Do we really need to do all this work just to get the requested end date? This is duplicating
+    -- much of reportOptsToSpec.
+    ropts = rawOptsToReportOpts d rawopts
+    argsquery = lefts . rights . map (parseQueryTerm d) $ querystring_ ropts
+    datequery = simplifyQuery . filterQuery queryIsDate . And $ queryFromFlags ropts : argsquery
 
 --- ** parsing utilities
 
@@ -338,6 +355,8 @@ parseAndFinaliseJournal' parser iopts f txt = do
 --
 -- - save misc info and reverse transactions into their original parse order,
 --
+-- - add forecast transactions,
+--
 -- - evaluate balance assignments and balance each transaction,
 --
 -- - apply transaction modifiers (auto postings) if enabled,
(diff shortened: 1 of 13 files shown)


Xitian9:
I'm not particularly happy with the implementation of `forecastPeriodFromRawOpts`.

Since we need the ‘requested report end date’, we need to generate the whole _rsQuery, so we can parse it. This includes handling any failed regular expression parsing, all just to get that single day. It also means we have to pull `Hledger.Reports.ReportOpts` into `Hledger.Read.Common`, which doesn't cause problems, but seems like it should be avoidable.

If you have any ideas as to how to avoid this I would love to hear them.

https://github.com/simonmichael/hledger/pull/1639/commits/c5a3df8360c14e8b7d79203ab547d6b2d5199a45#diff-fcee3c006b1d2e35b8c279c9daa9963ecdb5910bdc90099f7b3a37b8b529a65dR258-R272

Xitian9:
Note for self: I haven't yet checked that the forecast behaviour works well with hledger-ui and hledger-web.

Xitian9:
@simonmichael My knowledge of Yesod is basically non-existent, and I'm having trouble getting the hledger-web tests working. If you check commit fc6ca07, it is correctly generating the forecast journal (confirmed by writing out the journal) and correctly generates the webpage when run from the command line, but I'm not able to get the tests to pass. Do you have any ideas what's going wrong?

simonmichael:
I have learned this much: Test.hs "0 indicates a fileless (forecasted) txn" is no longer true; a traceShowId or dbg0 in transactionFragment shows they now are all 1-numbered (`transaction-1-1`, `transaction-1-2`, ...)

simonmichael:
It seems in transactionFragment tfileindex, journalFilePaths is a strange list of paths `[ "", "(string)" ]`, so the null file path `""` gets counted as file number 1.

hledger/test/forecast.test
```diff
@@ -235,3 +235,28 @@ hledger -f - reg -b 2021-01-01 -e 2021-01-05 --forecast
...
+
+>>>
+2021-09-01 Normal Balance As..  Checking                       -60           -60
+                                Costs                           60             0
+2021-10-01 explicit forecast..  Checking                       -40           -40
+                                Costs                           40             0
+2021-10-02 auto-deduced fore..  Checking                       -20           -20
+                                Costs                           20             0
+>>>2
+>>>=0
```
simonmichael, line 262: Maybe hledger print -x output would make a clearer test here ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -220,28 +227,47 @@ definputopts = InputOpts
...
-    ,anon_              = boolopt "anon" rawopts
-    ,new_               = boolopt "new" rawopts
-    ,new_save_          = True
-    ,pivot_             = stringopt "pivot" rawopts
-    ,auto_              = boolopt "auto" rawopts
-    ,balancingopts_     = def{ ignore_assertions_ = boolopt "ignore-assertions" rawopts
-                             , infer_prices_      = not noinferprice
-                             }
-    ,strict_            = boolopt "strict" rawopts
+rawOptsToInputOpts :: Day -> RawOpts -> InputOpts
```
simonmichael, line 236: I guess this was done for a reason - is it used somewhere ?

hledger-lib/Hledger/Read/Common.hs
```diff
@@ -220,29 +227,50 @@ definputopts = InputOpts
...
+    Just "" -> Just $ DateSpan Nothing mspecifiedend
+    Just str ->
+      either (\e -> usageError $ "could not parse forecast period : "++customErrorBundlePretty e) (Just . snd) $ 
+      parsePeriodExpr d $ stripquotes $ T.pack str
+  where
+    -- "They end on or before the specified report end date, or 180 days from today if unspecified."
+    mspecifiedend = dbg2 "specifieddates" . fmap (addDays 1) $ reportPeriodLastDay =<< rspec
+    -- Ugly hack. We just need the requested end date; do we really need to calculate the whole
+    -- rspec from scratch? Coerce errors to Maybe because any errors should be handled elsewhere,
+    -- but this is not the correct behaviour.
```
simonmichael, line 269: This all sounds rather obscure and possibly error-hiding (the last sentence). Maybe it's an old comment ? It sounds like something that needs work so let's mark it with XXX, and let's clarify the consequences if any, especially if user-visible

simonmichael (commented):
Looking good!

hledger-lib/Hledger/Read/Common.hs
Xitian9: Not very old, I just added it. I certainly don't want this merged as is, but am having trouble coming up with a more elegant solution.

hledger-lib/Hledger/Read/Common.hs
Xitian9: Sorry, can you clarify what you're referring to? Do you mean the addition of the `Day` argument? it is used in parsing any smart dates like `yesterday` in the forecast period.

hledger/test/forecast.test
Xitian9: Sure thing!

hledger-lib/Hledger/Read/Common.hs
Xitian9: I've found a better solution, though it's still not ideal. No more questionable error handling required.

Xitian9:
Probably a separate issue, but since it's come up.

In general, hledger will extend requested report start and end dates to fill out the whole period (e.g. week, month) it is reporting on. However, it does not seem to do this with forecast transactions. Should it?

Xitian9:
Okay, I think this now seems to pass all tests and work with both hledger-ui and hledger-web. I'd encourage those affected @simonmichael, @nobodyinperson, @jericdeleon to check out this branch and see if it has all the features you want and doesn't cause any unexpected problems.

nobodyinperson:
A quick check reveals that in my use case, everything (auto-postings on auto-deduced amounts of forecasted transations and balance assertions with auto-deduced amounts on forecasted transactions) works with `hledger 1.22.99, git revision cff66771bdc2b2c07167affcc82c46d889eb6819, linux-x86_64`. Well done and thank you very much 👍

hledger-lib/Hledger/Read/Common.hs
simonmichael: Yes I mean the addition of the Day argument. (I was trying to reference that commit/just the change of type signature, unsuccessfully). I just couldn't see where it was used but I guessed it was.

simonmichael:
> In general, hledger will extend requested report start and end dates to fill out the whole period (e.g. week, month) it is reporting on. However, it does not seem to do this with forecast transactions. Should it?

Perhaps it should, indeed. It feels like something we should hold for next major release, separate issue sounds good.

hledger-lib/Hledger/Reports/ReportOptions.hs
```diff
@@ -194,56 +192,51 @@ defreportopts = ReportOpts
...
-          ,show_costs_  = boolopt "show-costs" rawopts
-          ,sort_amount_ = boolopt "sort-amount" rawopts
-          ,percent_     = boolopt "percent" rawopts
-          ,invert_      = boolopt "invert" rawopts
-          ,pretty_tables_ = boolopt "pretty-tables" rawopts
-          ,color_       = useColorOnStdout -- a lower-level helper
-          ,forecast_    = forecastPeriodFromRawOpts d rawopts
-          ,transpose_   = boolopt "transpose" rawopts
-          }
+rawOptsToReportOpts :: Day -> RawOpts -> ReportOpts
```
simonmichael, line 198: I should have put a brief haddock here, let's add one now mentioning that this raises an error if there are problems with the options (unparseable --format, --valuechange combined with non-end valuation, etc.)

hledger/Hledger/Cli/CliOptions.hs
```diff
@@ -454,8 +454,9 @@ replaceNumericFlags = map replace
 -- Also records the terminal width, if supported.
 rawOptsToCliOpts :: RawOpts -> IO CliOpts
 rawOptsToCliOpts rawopts = do
-  let iopts = rawOptsToInputOpts rawopts
-  rspec <- rawOptsToReportSpec rawopts
+  d <- getCurrentDay
+  let iopts = rawOptsToInputOpts d rawopts
+  rspec <- either fail return $ rawOptsToReportSpec d rawopts
```
simonmichael, line 459: fail in IO is similar to `error` I guess, but what's the error message ? We might need to ensure it's a good one ?

hledger-lib/Hledger/Reports/ReportOptions.hs
simonmichael: And if you notice any other error-raising functions without a haddock, let's mention it for those too: forecastPeriodFromRawOpts, rawOptsToInputOpts, ...

simonmichael (commented):
A few more comments. Again, I couldn't find the place where this change was necessary. Is it a proactive refactoring for future need ? ("This increases purity, and allows us to have more control over which day we are generating options for.")

hledger/Hledger/Cli/CliOptions.hs
Xitian9: The error message is the string from the `Either String`

Xitian9:
> A few more comments. Again, I couldn't find the place where this change was necessary. Is it a proactive refactoring for future need ? ("This increases purity, and allows us to have more control over which day we are generating options for.")

It's not necessary. I can pull it out and submit it as part of a general clean-up PR, if you'd prefer.

simonmichael:
Ahh, I can understand that. Yes let's pull it out and consider separately/postpone till needed, avoiding unnecessary changes in this bugfix.

Xitian9:
My mistake, it actually is required now. Since `forecastPeriodFromRawOpts` now calls `rawOptsToReportOpts`, we either need to lift `forecastPeriodFromRawOpts` to the IO monad, or pass its Day argument to `rawOptsToReportOpts`. I'm inclined to do the latter.

simonmichael:
Let's not change a whole bunch of type signatures for no good reason. If it's just better, that's fair enough - I'll let you make the call.

Xitian9:
I changed just the type signature of `rawOptsToReportOpts`. 
* It requires the least change to the code base
* Only one type signature changes
* The change is to the function which is most ‘internal’, and least likely to be used by others

simonmichael:
Sounds great. I think we're good to go on this, am I right ?

InputOpts was originally only concerned with how to read files, and its scope seems to be creeping a bit into ReportOpts' area, but it's probably unavoidable. We'll keep an eye on it...

Xitian9:
Yeah, I agree that the lines between them are blurring. @alang76 was also commenting on this in #1593, as the new option `commodity_styles_` would have implications for both reading and displaying.

Yes, I think we're good to go here.

simonmichael:
Well.. @Xitian9 could you take a look at cherry picking these to 1.22-branch when you get a chance ? The first was easy for me but not these:

- 918c243fa * fix: forecast: Generate forecast transactions at journal finalisation, rather than as a postprocessing step.
   (#1638)
- c404800fb * ref!: forecast: Move forecast_ from ReportOpts to InputOpts.

Xitian9:
I think I've accomplished this in #1646.

simonmichael:
Thank you very much. Was it hard for you also ?

Xitian9:
Non-trivial, but having written the things in the first place I at least knew the solutions to each problem.

simonmichael:
Continuing the process chat here: I think this (cherry picking fixes to a long-lived release branch) is now the limiting factor on our release process, and I am wondering how to make it cheaper and more routine, even in the hard case where master is changing a lot and we need to push out many enmeshed fixes to stabilise the previous release. Any ideas welcome. I'm thinking perhaps when we think a fix will need to be in a minor release, we should make it simultaneously to both master and release branch (two pushes/merges/prs), when it's top of mind. And if that's not possible, prioritise fixing in the release branch, which is more time sensitive.

Issue #1642, opened 2021-08-06
aregister shows same-day transactions in wrong order (since 1.22.1)


simonmichael:
Another case for the areg test suite: 1.22.1 seems to be displaying same-day transactions in reverse order:
```
2021-07-31 first
    (a)           1

2021-07-31 second
    (a)           2

```
```
$ hledger-1.22 -f a.j areg a -w80
Transactions in a and subaccounts:
2021-07-31 first                a                                1             1
2021-07-31 second               a                                2             3
$ hledger-1.22.1 -f a.j areg a -w80
Transactions in a and subaccounts:
2021-07-31 second               a                                2             2
2021-07-31 first                a                                1             3
```

PR #1643, opened 2021-08-06
fix: areg: Make sure aregister sorts first by transaction register date,

Xitian9:
then by parse order to break any ties. (#1642)


simonmichael:
Thanks!

PR #1645, opened 2021-08-06
test: Add forecast transactions to aregister ordering test.

Xitian9:


Xitian9:
That build failure seems unrelated to the PR. It has something to do with haddock generation.

simonmichael:
Right you are, it was something I broke in master. Thanks.

PR #1647, opened 2021-08-08
Introduce lenses for InputOpts and CliOpts.

Xitian9:
Continuing the implementation discussed in #1545. This is the last of the lenses we will need.

<s>The `BalancingOpts` part of the hierarchy is collapsed into `InputOpts` in preparation for the elimination of `BalancingOpts` in #1593.</s> Since `BalancingOpts` has been left as-is, I've implemented `HasBalancingOpts` as you would normally expect.


simonmichael:
Looks good. Sorry if I asked this before - remind me what made us choose explicit lens implementations over using template haskell ? Did we measure a difference in build time, or was it just from my vague fears of adding more TH ?

If it was the former, let's mention it wherever we do this, to remind future devs of the rationale and that this is potentially removable boilerplate. 

If it was the latter, and build time is the same, maybe we should reconsider ?

simonmichael:
Oh, or reason 3 - are we doing it just in places where we need it to make non-standard field names work, temporarily ? If so we should note that too.

Xitian9:
We did not check a build time difference. It was a combination of vague fears of adding more TH (including for possible IDE introspection reasons), and to make non-standard field names work, temporarily.

Xitian9:
Beyond running `stack clean; time stack build hledger-lib`, is there an accepted way to test build times?

simonmichael:
No better way right now, sounds fine for a quick test.

Xitian9:
I've run a quick test of build times for `hledger-lib` under two circumstances:
1. Manual definition of lenses for `BalancingOpts` and `InputOpts`
2. Template haskell derivation of lenses for `BalancingOpts` and `InputOpts`

The results were roughly consistent with what [this](https://www.parsonsmatt.org/2019/11/27/keeping_compilation_fast.html) says: a modest increase in compilation time of under half a second.

1. `stack build hledger-lib  83.53s user 1.45s system 100% cpu 1:24.70 total`
2. `stack build hledger-lib  83.94s user 1.46s system 100% cpu 1:25.12 total`

This seems reasonable. We'll have to keep manual definitions of `ReportOpts` and `ReportSpec` if we want to keep our fancy lenses.

Xitian9:
You know, I think we can actually get template haskell to do almost all the work for us, even for the special lenses in `ReportOpts` and `ReportSpec`. We would only need to write the logic for the special parts of `ReportSpec`.

Xitian9:
I have managed to implement basically all the lenses using Template Haskell, which cuts down significantly on code, and greatly helps maintenance. It results in a roughly 1% increase in compilation times, well within random noise.

hledger-lib/Hledger/Utils/TH.hs
```diff
@@ -0,0 +1,43 @@
...
+import Data.List.Extra (uncons, unsnoc)
+import qualified Data.Set as Set
+import Language.Haskell.TH (mkName, nameBase)
+import Lens.Micro ((&), (.~))
+import Lens.Micro.TH (DefName(TopName), lensClass, lensField, makeLensesWith, classyRules)
+
+-- | Make lenses for fields with trailing underscores.
+-- Most lenses will be created by stripping the underscore from the name, but
+-- for certain common names it will instead add a second underscore.
+makeHledgerClassyLenses x = flip makeLensesWith x $ classyRules
```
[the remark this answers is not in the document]
Xitian9: Unfortunately they type signature is not very informative, and also not possible to put in, as far as I can tell. It should be `makeHledgerClassyLenses :: Name -> DescQ`, but `DescQ` is defined in an internal module in `Language.Haskell.TH` and is not exported.

simonmichael:
I think you have a typo there, makeHledgerClassyLenses :: Name -> DecsQ worked for me.

Xitian9:
> I think you have a typo there, makeHledgerClassyLenses :: Name -> DecsQ worked for me.

Ha! So it is! Ironic, given that my partner's name has a cs in it which everybody always mis-spells sc.

hledger-lib/Hledger/Reports/ReportOptions.hs
```diff
@@ -718,227 +691,79 @@ setWithReport l = overWithReport l . const
 type ReportableLens' s a = forall f. Reportable f String => (a -> f a) -> s -> f s
 
 -- | Lenses for ReportOpts.
--- Note that some of these are not true lenses, as they have a further restriction on
+
+-- Implement HasReportOptsNoUpdate, the basic lenses for ReportOpts.
+makeHledgerClassyLenses ''ReportOpts
+makeHledgerClassyLenses ''ReportSpec
+
+-- Special lenses for ReportOpts which also update the Query and QueryOpts in ReportSpec.
```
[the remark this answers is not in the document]
Xitian9: Corrected, and added examples/doctests.

hledger-lib/Hledger/Reports/ReportOptions.hs
```diff
@@ -719,227 +692,95 @@ setWithReport l = overWithReport l . const
...
+-- Acct (RegexpCI "assets")
+-- >>> _rsQuery $ set querystring ["(assets"] defreportspec
+-- *** Exception: Updating ReportSpec failed: try using overWithReport instead of over or setWithReport instead of set
+-- >>> _rsQuery $ set period (MonthPeriod 2021 08) defreportspec
+-- Date DateSpan 2021-08
+class HasReportOptsNoUpdate a => HasReportOpts a where
     reportOpts :: ReportableLens' a ReportOpts
-    reportOptsNoUpdate :: Lens' a ReportOpts
+    reportOpts = reportOptsNoUpdate
+    {-# INLINE reportOpts #-}
```
simonmichael, line 723: Maybe I asked this before: why the INLINE pragmas here, do we know they are worthwhile ?

hledger-lib/Hledger/Utils/TH.hs
```diff
@@ -0,0 +1,54 @@
...
+-- When run on X, it will create a typeclass named HasX (except for ReportOpts,
+-- which will be named HasReportOptsNoUpdate) containing all the lenses for that type.
+-- If the field name starts with an underscore, the lens name will be created
+-- by stripping the underscore from the front on the name. If the field name ends with
+-- an underscore, the field name ends with an underscore, the lens name will be
+-- mostly created by stripping the underscore, but a few names for which this
+-- would create too many conflicts instead have a second underscore appended.
+-- ReportOpts fields for which updating them requires updating the query in
+-- ReportSpec are instead names by dropping the trailing underscore and
+-- appending NoUpdate to the name, e.g. querystring_ -> querystringNoUpdate.
```
simonmichael, line 24: This is some tricky stuff we're adding, so I'm emphasising the docs. The above is excellent at explaining what this does. I think we could just add a sentence saying why we have these shenanigans - IIRC there are two reasons, 1. to temporarily accommodate legacy field names ending in underscore, postponing a big rename and 2. to assist with safely modifying ReportSpec, removing a source of bugs. And perhaps a link to the issue with the full discussion.

hledger-lib/Hledger/Reports/ReportOptions.hs
Xitian9: I put the INLINE pragmas in because the Template Haskell splices put them in. I'm not sure whether they are worthwhile, but they are also low-cost. I could remove them if you'd prefer.

simonmichael: Whatever you prefer, but I'd probably leave them out for now, on the basis that GHC quite likely can figure this out by itself, these probably aren't called in a hot loop, and we don't want to have to be annotating things if we can avoid it.

hledger-lib/Hledger/Reports/ReportOptions.hs
```diff
@@ -719,227 +692,95 @@ setWithReport l = overWithReport l . const
 type ReportableLens' s a = forall f. Reportable f String => (a -> f a) -> s -> f s
 
 -- | Lenses for ReportOpts.
--- Note that some of these are not true lenses, as they have a further restriction on
+
```
simonmichael, line 695: This comment refers to older code, but: in overWithReport and setWithReport, is "Report" referring to their ability to report an error message ? I found it a bit confusing to understand what's going on there, thinking it might be related to ReportOptions or ReportSpec. If so, would overOrError and setOrError be any clearer ?

simonmichael:
Accidentally pushed an unrelated commit here, force-pushed to remove it.

hledger-lib/Hledger/Reports/ReportOptions.hs
Xitian9: Yes, Report is referring to their ability to report an error message.

Perhaps `overOrEither` since they return an `Either`, rather than throwing an error.

hledger-lib/Hledger/Reports/ReportOptions.hs
Xitian9: Edward Kmett (author of the lens library) on using INLINE pragmas in lens instance declarations.
https://www.reddit.com/r/haskell/comments/cjkc3l/should_i_be_inlining_instance_implementations/

Based on this I'm inclined to leave it in.

Xitian9:
I've implemented the above suggestions:
- Adding more context to the docs for `makeHledgerClassyLenses`
- Changing `overWithReport`/`setWithReport` to `overOrEither`/`setOrEither`

I have left the INLINE pragmas in for now based on the thread I linked to.

I also moved `makeHledgerClassyLenses` from a separate module straight into `Hledger.Utils`. I don't think there's sufficient justification for adding a whole new module for it.

simonmichael:
Ok, sounds good to leave it in.

simonmichael:
> Perhaps overOrEither since they return an Either, rather than throwing an error.
> 
I don't love this - it's always an Either, and "error" can also mean error message... but it's ok.

Xitian9:
The reason I don't think `overOrError` is best is because `over` will throw an error if it fails, and `overWithReport` (or whatever we call it) will return a Left with an error message. Maybe we can come up with a better idea…

Xitian9:
Maybe `overAsEither`?

simonmichael:
> I don't love this - it's always an Either, and "error" can also mean error message... but it's ok.

overEither, setEither could also work ?

Xitian9:
How about `overAsEither`. `overEither` sounds like it operates on `Either`s, rather than just returning them.

simonmichael:
I know it makes less sense grammatically, but for a short and common operator like this I think we'll end up preferring the short version. There is precedent, things like https://hackage.haskell.org/package/base-4.15.0.0/docs/Text-Read.html#v:readEither in base. But if you feel I'm wrong, go for it. I think this is ready to merge after this.

Xitian9:
Okay, I'll do `overEither`/`setEither`.

Xitian9:
Done.

simonmichael (approved):

simonmichael:
Thanks, great work. Lens powers coming online!

PR #1651, opened 2021-08-12
cln: Derive more instances in Hledger.Data.Types to reduce orphans.

Xitian9:
Also clean up some outdated or inaccurate comments.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -40,7 +40,6 @@ exchange rates.
 
 -}
 
-{-# LANGUAGE BangPatterns       #-}
 {-# LANGUAGE OverloadedStrings  #-}
 {-# LANGUAGE RecordWildCards    #-}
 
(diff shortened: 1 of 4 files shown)


hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -247,8 +246,8 @@ amt @@ priceamt = amt{aprice=Just $ TotalPrice priceamt}
 -- Prices are ignored and discarded.
 -- Remember: the caller is responsible for ensuring both amounts have the same commodity.
 similarAmountsOp :: (Quantity -> Quantity -> Quantity) -> Amount -> Amount -> Amount
-similarAmountsOp op !Amount{acommodity=_,  aquantity=q1, astyle=AmountStyle{asprecision=p1}}
-                    !Amount{acommodity=c2, aquantity=q2, astyle=s2@AmountStyle{asprecision=p2}} =
+similarAmountsOp op Amount{acommodity=_,  aquantity=q1, astyle=AmountStyle{asprecision=p1}}
+                    Amount{acommodity=c2, aquantity=q2, astyle=s2@AmountStyle{asprecision=p2}} =
```
simonmichael, line 250: Just curious, how do you know these are redundant ?

simonmichael:
I think it's causing a warning in master somehow ?
https://github.com/simonmichael/hledger/runs/3391904284?check_suite_focus=true

Xitian9:
I think that warning was caused by 993d0b5c54e364b26fe78a8e6ff62bd626e42c1e, which introduced that import without using it anywhere.

hledger-lib/Hledger/Data/Amount.hs
Xitian9: hlint says so. Pattern matching is always strict. It needs to put it into whnf to see if the pattern matches.

Xitian9:
I think the situation is that `unlines` is already in Prelude, so importing it from `Data.String.Compat` is redundant.

simonmichael:
Cool, thanks. Sorry about the warning in master, my fault.

PR #1653, opened 2021-08-17
Various hlint cleanups

Xitian9:
I got a bit ahead of myself, but here are a bunch of hlint cleanups.


Xitian9:
This PR is unwieldy as is. I'll break it up into bite-sized chunks.

PR #1649, opened 2021-08-09
imp: balance: Allow using --commodity-column with compound balance reports.

Xitian9:
Also return a usage error if asked to use --commodity-column with --format.

PR #1654, opened 2021-08-17
fix: bal: fix commodity-column interaction with transpose and empty commodities

la-wu:
I was introducing `--commodity-column` into my personal reports and realized that the textual output didn't correctly handle `--transpose`  when both flags are set. While fixing that, I also realized that empty commodities were being silently filtered out (e.g https://github.com/simonmichael/hledger/blob/master/hledger/Hledger/Cli/Commands/Balance.hs#L717). This wasn't a regression per-se but obviously undesirable behavior.

There is some overlap with https://github.com/simonmichael/hledger/pull/1649 since handling transposition unfortunately involved some reworking of the compound call-site.

Perhaps the suite of flag interactions should be more formally considered?

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -100,6 +100,7 @@ module Hledger.Data.Amount (
   maAddAmounts,
   amounts,
   amountsRaw,
+  maCommodities,
   filterMixedAmount,
   filterMixedAmountByCommodity,
   mapMixedAmount,
@@ -152,6 +153,7 @@ import Data.Foldable (toList)
 import Data.List (find, foldl', intercalate, intersperse, mapAccumL, partition)
 import Data.List.NonEmpty (NonEmpty(..), nonEmpty)
 import qualified Data.Map.Strict as M
+import qualified Data.Set as S
 import Data.Maybe (fromMaybe, isNothing, isJust)
 import Data.Semigroup (Semigroup(..))
 import qualified Data.Text as T
@@ -661,6 +663,11 @@ amounts (Mixed ma)
 -- unit price from most negative to most positive.
 amountsRaw :: MixedAmount -> [Amount]
 amountsRaw (Mixed ma) = toList ma
+
+-- | Get the set of mixed amount commodities. Returns an empty set of no amounts
+maCommodities :: MixedAmount -> S.Set CommoditySymbol
+maCommodities = S.fromList . fmap acommodity . amounts'
+  where amounts' ma@(Mixed m) = if M.null m then [] else amounts ma
 
 normaliseMixedAmount :: MixedAmount -> MixedAmount
 normaliseMixedAmount = id  -- XXX Remove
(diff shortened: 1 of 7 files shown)


Xitian9:
I haven't done a full read-through of the code here, but I think we shouldn't worry about conflicts between PRs for merging this: I can easily rebase.

My bigger thought is that we seem to be jumping through a lot of hoops to work around the limitations of the table rendering code. If I understand correctly, the problem here is that we basically have two columns worth of row headers (account name, and commodity), while the rendering code can only handle one. So we are ad-doc throwing commodity columns in using purpose-designed functions.

Would it be reasonable to solve this problem with a transpose option in the rendering code? Perhaps to add `transposeTable` to `TableOpts`, which would flip the roles of the rows and columns in the rendering function?

Xitian9:
Thinking about it: you have a working implementation here. Let's work on merging this with the basic method used here. If we come up with a simpler way later, we can do the refactoring then and make sure it still passes all these tests.

la-wu:
> My bigger thought is that we seem to be jumping through a lot of hoops to work around the limitations of the table rendering code. If I understand correctly, the problem here is that we basically have two columns worth of row headers (account name, and commodity), while the rendering code can only handle one. So we are ad-doc throwing commodity columns in using purpose-designed functions.
>
> Would it be reasonable to solve this problem with a transpose option in the rendering code? Perhaps to add `transposeTable` to `TableOpts`, which would flip the roles of the rows and columns in the rendering function?

That is certainly one way to approach it. Are you imagining a situation where `renderTable` first creates the full table of strings and then does transposition and alignment after (`Cell` becomes a single `WideBuilder`)?

Either way we would also need to implement the header-column independence for the budget report that we talked about https://github.com/simonmichael/hledger/pull/1626#issuecomment-889585570 (Which might indicate a compelling use case). Budget was also the most annoying piece of code to change so I'm open to improvements there...


> Thinking about it: you have a working implementation here. Let's work on merging this with the basic method used here. If we come up with a simpler way later, we can do the refactoring then and make sure it still passes all these tests.

Sounds good to me. Would it be helpful if I split this into separate PRs for empty-commodity handling fixes and transpose-flag fixes?

Xitian9:
> Sounds good to me. Would it be helpful if I split this into separate PRs for empty-commodity handling fixes and transpose-flag fixes?

That would keep the commit history cleaner, but it's not crucial.

This PR seems to include enabling `--commodity-column` for compound balance reports. Does this render #1649 completely redundant?

la-wu:
> This PR seems to include enabling `--commodity-column` for compound balance reports. Does this render #1649 completely redundant?

I didn't touch the format specifier or any of the aggressive account name elisions so not completely covered. (The CSV and HTML output changes in the compound reports is handled by this and not 1649 though)

Xitian9:
In that case, I'll mark #1649 as draft and wait until this one is merged before deciding what in there is still useful.

Xitian9:
This build failure seems to be unrelated to the content of the PR.

simonmichael:
> This build failure seems to be unrelated to the content of the PR.

Pushed a fix that might help.

simonmichael:
...to master; if this PR gets rebased its CI may work again.

simonmichael:
Thanks for the followup testing and fixes!

PR #1657, opened 2021-08-19
fix!: register: Tighten up spacing around the date in register reports. (#1655)

### LINKED ISSUE #1655: reg: report intervals with custom start dates are hard to read
When showing https://hledger.org/dev/hledger.html#intervals-with-custom-start-date , in general `register` must display both the start and end date for each period, but it doesn't allow enough width in the first column. `aregister` might have the same problem.

Xitian9:
Fixes #1655.

As a side effect, this changes the Json representation of the PostingsReport. The maybe report end date is now replaced with a maybe period.

I think this will provide strictly more information, and so is a desirable change, but if it will break too much we can change it back.

### DIFF
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -25,8 +25,8 @@ import Data.List (nub, sortOn)
 import Data.List.Extra (nubSort)
 import Data.Maybe (fromMaybe, isJust, isNothing)
 import Data.Text (Text)
-import Data.Time.Calendar (Day, addDays)
-import Safe (headMay, lastMay)
+import Data.Time.Calendar (Day)
+import Safe (headMay)
 
 import Hledger.Data
 import Hledger.Query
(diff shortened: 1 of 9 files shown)


Xitian9:
We seem to be getting some early build failures due to one of the check rules regarding commit messages.

Xitian9:
The last commit message doesn't begin with a `;`, but it still seems to be skipping the CI tests.

simonmichael:
Yes, you can see the `git log` failing with
```
fatal: ambiguous argument 'origin/regalign': unknown revision or path not in the working tree.
```
That hard-coded `origin` seems not to be working for PRs (any more ? things were working somehow before). It's rather hard to see how to reliably list the new commits in github CI. We could open an issue, otherwise I'll just keep tweaking here and there.

simonmichael:
(And, pushed another fix. Thanks for pointing it out.)

hledger-lib/Hledger/Reports/PostingsReport.hs
```diff
@@ -38,27 +38,25 @@ import Hledger.Reports.ReportOptions
...
+                          ,MixedAmount   -- The running total after this posting, or with --average,
+                                         -- the running average posting amount. With --historical,
+                                         -- postings before the report start date are included in
+                                         -- the running total/average.
                           )
 
 -- | A summary posting summarises the activity in one account within a report
 -- interval. It is kludgily represented by a regular Posting with no description,
 -- the interval's start date stored as the posting date, and the interval's end
 -- date attached with a tuple.
```
simonmichael, line 58: My comment here looks a bit sketchy.. and out of date now, hopefully ?

hledger/test/cli/report-interval.test
```diff
@@ -13,22 +13,22 @@
 # The last report interval option takes precedence.
 
 $ hledger -f- register --weekly --monthly
-2019-01                 a                                        2             2
-2019-02                 a                                        1             3
+2019-01   a                                                      2             2
```
simonmichael, line 16: Any particular reason it's 3 spaces ? I think 2 is our normal minimum inter-field separator.

hledger-lib/Hledger/Reports/PostingsReport.hs
simonmichael: Sorry, I may have misclicked. I'm wondering if the "It is kludgily represented..." haddock needs updating.

hledger/test/cli/report-interval.test
Xitian9: I'll look at tightening that up further.

Xitian9: Ah, got it. It's actually 1 space between the date and the description, a 0 width empty description, then 2 spaces between the description and the account, for 3 spaces total separation in this case.

hledger-lib/Hledger/Reports/PostingsReport.hs
Xitian9: You are correct. I've updated it.

hledger/test/cli/report-interval.test
simonmichael: Ok. It looked a little weird to me but I think you're saying it's fine and good as is.

simonmichael:
Thanks!

PR #1663, opened 2021-08-22
web: base64 encode user controlled strings (#1525)

### LINKED ISSUE #1525: XSS vulnerability: malicious javascript can be stored via hledger-web add/edit
hledger-web's forms sanitise obvious javascript, but not obfuscated javascript (cf https://owasp.org/www-community/xss-filter-evasion-cheatsheet). This means hledger-web instances, especially anonymously-writable ones like demo.hledger.org, could be loaded with malicious javascript to be executed by subsequent visitors. Reported by Gaspard Baye and Hamidullah Muslih, who also provided a video and example of such js which I can share.

hledger-web processes forms with runFormPost, part of [yesod-form](https://hackage.haskell.org/package/yesod-form). I am guessing that makes use of [xss-sanitize](https://hackage.haskell.org/package/xss-sanitize) ? But note sanitising and escaping are different (https://github.com/yesodweb/haskell-xss-sanitize#readme). I get the impression it may be impossible to escape/sanitise with 100% reliability; if so, what should we do ? Tighten up sanitization as much as we can and document that anonymous-writable hledger-web's are unsafe and not recommended, and make demo.hledger.org read only ? 

I would appreciate any help diagnosing and resolving this. Questions to be answered: 

- which libraries are involved here - yesod-form, xss-sanitize, ... ?
- are we getting the most protection available from them ?
- should we open bugs against any of them ?
- is it ok for me to share an example of obfuscated js here ?
- are there any public hledger-web instances other than demo.hledger.org ?
- is reliable escaping/sanitisation possible ?
- if not, what should be done for hledger-web ?

ArsenArsen:
This fixes a reported Stored XSS vulnerability in toBloodhoundJson by
encoding the user-controlled values in this payload into base64 and
parsing them with atob.

In my exploration of the vulnerability with various payloads I and
others crafted, it would appear that this is the only available XSS in
hledger-web in relation to stored accounts and transaction details. If
there is other parts of the UI which may contain user-controlled data,
they should be examined for similar things. In this instance,
protections provided by yesod and other libraries worked fine, but in a
bit of code that hledger-web was generating, the user could insert a
</Script> tag (which is valid HTML and equivalent to </script> but not
caught by the T.Replace that existed in toBloodhoundJson) in order to
switch out of a script context, allowing the parser to be reset, and for
arbitrary JavaScript to run.

The real fix is a bit more involved, but produces much better results:
Content-Security-Policy headers should be introduced, and using
sha256-<hash of script> or a different algorithm, they should be marked
as trusted in the header. This way, if the (in-browser) parser and
hledger-web generator disagree on the source code of the script, the
script won't run. Note that this would still be susceptible to attacks
that involve changing the script by escaping from the string inside it
or something similar to that, which can be avoided additionally by using
either the method used in this commit, or a proper JSON encoder.

Test payload: ``</Script><svg onload=alert(1)//>``

Closes #1525

### DIFF
--- a/hledger-web/Hledger/Web/Widget/AddForm.hs
+++ b/hledger-web/Hledger/Web/Widget/AddForm.hs
@@ -13,14 +13,14 @@ module Hledger.Web.Widget.AddForm
 import Control.Monad.State.Strict (evalStateT)
 import Data.Bifunctor (first)
 import Data.Foldable (toList)
-import Data.List (dropWhileEnd, intercalate, unfoldr)
+import Data.List (dropWhileEnd, unfoldr)
 import Data.Maybe (isJust)
 import qualified Data.Set as S
 import Data.Text (Text)
 import Data.Text.Encoding.Base64 (encodeBase64)
 import qualified Data.Text as T
 import Data.Time (Day)
-import Text.Blaze.Internal (Markup, preEscapedString)
+import Text.Blaze.Internal (Markup, preEscapedText)
 import Text.Megaparsec (bundleErrors, eof, parseErrorTextPretty, runParser)
 import Yesod
 
@@ -91,11 +91,11 @@ addForm j today = identifyForm "add" $ \extra -> do
       -- This used to work, but since 1.16, it seems like something changed.
       -- toJSON ("a"::Text) gives String "a" instead of "a", etc.
       -- preEscapedString . escapeJSSpecialChars . show . toJSON
-      preEscapedString $ concat [
+      preEscapedText $ T.concat [
         "[",
-        intercalate "," $ map (
-          ("{\"value\":" ++).
-          (++"}").
+        T.intercalate "," $ map (
+          ("{\"value\":" <>).
+          (<> "}").
           -- This will convert a value such as ``hledger!`` into
           -- ``atob("aGxlZGdlciE=")``. When this gets evaluated on the client,
           -- the resulting string is ``hledger!`` again. The same data is
@@ -107,8 +107,8 @@ addForm j today = identifyForm "add" $ \extra -> do
         "]"
         ]
       where
-b64wrap :: Text -> String
-b64wrap = ("atob(\""++) . (++"\")") . T.unpack . encodeBase64
+b64wrap :: Text -> Text
+b64wrap = ("atob(\""<>) . (<>"\")") . encodeBase64
 
 validateTransaction ::
      FormResult Day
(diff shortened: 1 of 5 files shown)


simonmichael:
Thanks @ArsenArsen . 

The CI failure can be fixed by adding `- base64-0.4.2.3` to `extra-deps:` in `stack8.6.yaml`. 

Also that package seems to have quite a few deps, including ghc-byteorder which requires cabal-version (cabal file format) 2.2, which I believe means hledger would be no longer buildable with cabal-install <2.2. Debian oldstable has cabal 2.2, so I guess that would be ok.

There's also the base64-string package. Although older and probably unmaintained, it has no dependencies.

simonmichael:
Could you briefly summarise the limitations of this fix versus the "real fix" you mention ?

Your commit message passed bin/commitlint, but not quite all of our [guidelines](https://hledger.org/CONTRIBUTING.html#commmit-messages) - would you mind making it `fix: web: ...` to help the changelog time.

Also for the changelog, could you remind be briefly of how the vulnerability looked to users ? Eg, here's a draft:

- An XSS (cross-site scripting) vulnerability has been fixed in hledger-web's add form. 
  Previously, obfuscated javascript could be added via (the account name field ?), 
  and could be executed automatically by subsequent visitors viewing (hledger-web ?)

simonmichael:
PS if you can say without too much work which hledger version introduced the vulnerability, I'll mention that too.

ArsenArsen:
Would other GHC versions not also require the extra dep specified?

To quote the base64 readme: "There are no dependencies aside from those bundled with GHC, text-short, and the ghc-byteorder re-export".
I thought these would be rather minor. If you want, I can refactor the patch to use base64-strings, but I intended to refactor the rest of this function and its callers to use Text instead.

The second approach has the advantage of preventing further XSS, to the
extent specified above, in practice, a combination of both should be
used, b64 for embedded data and the CSP sha256-hash script-src over
everything else, which will eliminate all injected or malformed script
blocks (via CSP), in combination with eliminating any HTML closing tags
which might occur in stored data (via b64).

This vulnerability appears to have been first introduced when
autocompletion was added in hledger-web, git tag hledger-0.24, commit
hash: ec51d28839b2910eea360b1b8c72904b51cf7821

An appropriate changelog entry for this vulnerability would be:

- An XSS vulnerability has been fixed in hledger-web. Previously, JavaScript could be added to any autocomplete-able field and could be executed automatically by subsequent visitors viewing the journal.

It's not fair to call this obfuscated JavaScript, it just uses some HTML laxness to go around an insufficient prevention previously added to the add form. And this works in any field that can be autocompleted via Bloodhound.

simonmichael:
Great work, thanks for the vulnerability fix!

simonmichael:
Sure, is it needed ? The bug was fixed by 1.23. The install page may show which distros ship an older version. Is it something you can do ?

bayegaspard:
Yes, just to record version <1.23 was vulnerable to stored XSS. I will test the recent versions soon, when I get some free time.

bayegaspard:
We disclose vulnerabilities only when it is fixed. So  now we have it fixed, we can file for CVE.

PR #1664, opened 2021-08-22
Fix cmdline example typo, f- should be -f-

DanielG:
Just a quick typo fix,

```
hledger: command f- is not recognized, run with no command to see a list
```
is obviously not what the example is supposed to print :)


simonmichael:
Thanks!

PR #1667, opened 2021-08-23
Implement more intuitive forecast transaction rules.

### LINKED ISSUE #1648: imp: more intuitive forecast behaviour when there are no ordinary transactions ?
Reviewing #1538, I expected this to show some forecasted transactions, even without specifying an explicit --forecast=PERIOD, because there are no ordinary transactions that would suppress them:
```
~ every friday
    (foo)                                          1.00 USD
```
```
$ hledger reg --forecast -b 2021/4/1 -e 2021/5/1
```

It took me two careful readings of the [forecast period doc](https://hledger.org/hledger.html#forecasting-with-periodic-transactions) (`hledger help forecast`) to understand why it doesn't (forecast start date was defaulting to today):

>   Periodic transactions are generated within some forecast period.  By default, this
>
>   * begins on **the later of**
>        * the **report start date** if specified with -b/-p/date:
>        * the day after the latest normal (non-periodic) transaction in
>          the journal, or **today if there are no normal transactions**.

Wouldn't it would be more intuitive to show some transactions in the above case ? The rule could be:

- if there are normal transactions in the journal
  - begins on the later of
    - the day after the latest normal transaction
    - the report start date if specified
- if there are no normal transactions in the journal
  - begins on the report start date if specified, otherwise today

I think it is backward compatible and the user-visible behaviour is strictly more intuitive. However it now looks even more complicated. I wonder

- Is there a simpler way to describe it ? 
- Is there a simpler rule that would be worth changing to ?
-

### LINKED ISSUE #1665: forecast generates too many transactions when -b specified
Forecast transactions are only supposed to be generated from the report start date if present, but they will actually be generated from the last journal transaction if present, and only displayed from the report start date. However, they can be revealed with a historical report.

```
2021-01-01
  (a)      1000

~ daily
  (a)       1
```

```
$ hledger -f span.j reg --forecast -H -b 2021-02 -e 2021-02-05
2021-02-01                      (a)                              1          1031
2021-02-02                      (a)                              1          1032
2021-02-03                      (a)                              1          1033
2021-02-04                      (a)                              1          1034
```

If this were only generating forecast transactions from 2021-02, we would expect the total balance at the first step to be 1001.

Xitian9:
Fixes #1648. Fixes #1665.

This is an idea of what a solution to #1648 would look like, so we have something concrete to consider when discussing it.

Note that this probably breaks forecast transactions in hledger-ui and hledger-web, so should not be merged until that's been fixed.


Xitian9:
There are alternative implementations that are more reliable and less hacky. My favourite is to change the type of `forecast_` to `Maybe Day -> Maybe DateSpan`, where it takes the journal end date as an argument. This allows us to implement all the logic in one place (`rawOptsToInputOpts`), but we lose a simple `Show` instance for `InputOpts`.

PR #1669, opened 2021-08-25
Some changes to the module structure

Xitian9:
The general thrust of these changes is to try to flatten the module structure to improve compile times, and enable future changes for which the current module structure is causing import cycles. This generally follows the advice [here].(https://www.parsonsmatt.org/2019/11/27/keeping_compilation_fast.html)

A summary of the main changes:

- Functions related to balancing `Transaction` and `Journal`s have been moved out into a separate module `Hledger.Data.Balancing`. This shrinks the massive `Journal.hs` module, and groups related balancing functions together.
- The `Hledger.Read` and `Hledger.Reports` are now almost entirely independent. Since `Hledger.Read` is one of the big bottlenecks for complilation times, this should enable work on getting this down.

This PR is largely setting the stage for future work, but I don't want to create a massive and unwieldy PR, so I'm submitting this at this stage.

<s>The first commit is an unrelated change to rename `balancingOpts` -> `defbalancingopts` for consistency. Sorry for letting it sneak in here.</s> Rebased to remove this commit.

### DIFF
--- a/hledger-lib/Hledger/Data.hs
+++ b/hledger-lib/Hledger/Data.hs
@@ -12,6 +12,7 @@ module Hledger.Data (
                module Hledger.Data.Account,
                module Hledger.Data.AccountName,
                module Hledger.Data.Amount,
+               module Hledger.Data.Balancing,
                module Hledger.Data.Dates,
                module Hledger.Data.Journal,
                module Hledger.Data.Json,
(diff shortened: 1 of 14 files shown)


Xitian9:
For information, here is the module dependency graph before:
before

And here is the module dependency graph after:
after

Xitian9:
Unfortunately a disadvantage of these types of changes is that, even though they are simple in nature (copy-pasting functions between modules), they touch a lot of things, and so generate a lot of conflicts which are annoying to resolve. If you're not generally open to these sorts of changes let me know early and I'll abandon it, rather than trying to resolve these conflicts.

simonmichael:
Nifty! I'm interested in this sort of thing. We'll try to navigate the inter-PR dependencies with as little work as possible.

PR #1675, opened 2021-08-28
Remove some under-used modules to save on compilation time.

Xitian9:
- `Hledger.Utils.UTF8IOCompat` does nothing beyond define `error'` and `usageError`, which has
been moved to `Hledger.Utils`.
- `Hledger.Utils.Tree` is only used in one place in `Hledger.Data.Account`, so the code has been moved there.
- `Hledger.Utils.Color` is only used in one place in `Hledger.Data.Amount`. The code has been moved to `Hledger.Utils`.
- `Hledger.Data.Commodity` is very small, and has been moved into `Hledger.Data.Amount`.


Xitian9:
I'm doing some analysis of build times, and found a tool for visualising the build times of each module. https://github.com/codedownio/time-ghc-modules

I put this here not just so that others know about this tool, but because of the example they give.

simonmichael:
> I'm  slightly sad at the thought of dropping modules like Commodity which although small keeps the organisation clear - one module per major type
> and Utils.Color, which could easily grow and possibly be a candidate for extraction to a common package

> On my system these changes together save about as much time as adding template haskell for lenses added. Less than 1% build time, and within random noise, but noticable.

> ok, let's go with it

Thanks!

PR #1676, opened 2021-08-29
Cleanup UIOpts and UIState

Xitian9:
Pulls all the UI options into UIOpts, removes a redundant option, and uses the power of lenses to significantly simplify UIState.hs.


simonmichael:
Thanks!

PR #1677, opened 2021-08-29
test: cli: Add --today option to explicitly set the current date. (#1674)

### LINKED ISSUE #1674: need a --today option for testing
I committed some hledger-ui forecast tests yesterday which are sensitive to the current date, which was hard to to avoid. Within the next few months, a `--today=DATE` flag is needed to help make such date-sensitive tests repeatable.

It could be a hidden option, but there's probably no need. Perhaps it will have other uses.

It should be a general option, supported by all the hledger tools. 

The basic operation is most important. It may or may not interact sensibly with all other date-related options.

Xitian9:
Fixes #1674.

This is so that we can run tests containing future transactions that
won't fail as soon as ‘the future’ actually arrives.

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -259,6 +259,7 @@ rawOptsToInputOpts day rawopts =
                                , commodity_styles_  = rawOptsToCommodityStylesOpts rawopts
                                }
       ,strict_            = boolopt "strict" rawopts
+      ,_ioDay             = day
       }
 
 -- | Get the date span from --forecast's PERIODEXPR argument, if any.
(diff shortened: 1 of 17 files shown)


Xitian9:
Yet another instance of `forecast_` being in `InputOpts` causing scope creep. If we want to make sure a `--today` option works properly, we need to make sure that all uses of `getCurrentDay` are replaced with the date given to `--today`. The day is stored in `ReportSpec`, so we would need access to the `ReportSpec` to get the appropriate day.

Unfortunately the current day is needed in two places in `InputOpts`, which does not have access to `ReportSpec`.
1. Parsing smart dates in periodic transactions (These should really not be dates relative to a reference date. We could probably get away with replacing this use of `Day` with an error, but that would increase partiality)
2. Determining the fallback forecast dates when the forecast and report dates are unspecified.

Options:
- Hope this doesn't cause a problem and don't do anything about it.
- Package today's date into `InputOpts` as well as `ReportSpec`, and hope these don't get out of sync.
- Move forecasting to a post-processing step.
- I'm sure there are others… 

https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Read/Common.hs#L233-L235
https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Read/Common.hs#L367-L370

PR #1682, opened 2021-09-03
fix: roi: rare bug with PnL applied on the first day of investment

adept:
Recent discussion on the mailing list uncovered a rare corner-case bug in `roi` which will manifest itself when someone tries to apply the PnL to the investment at the same date as it is "opened" (initially funded).

Roi applies the PnL (changes the valuation of the investment) before applying fundings/withdrawals, when they fall on the same date. This does not work when investment is still empty and PnL transaction is applied prior to the initial funding transaction.

Fix makes sure that TWR computations for the empty investment never start from applying PnL.


simonmichael:
Thanks for the fix!

PR #1690, opened 2021-09-09
fix: balance: Show all empty accounts in flat balance reports. (#1688)

### LINKED ISSUE #1688: bal --no-elide (and -E, and hledger-ui zero display ?) not working as I expected
I'm not sure if there's another explanation, but I'm puzzled enough to note it as a possible bug:

```
$ hledger-1.18 -f avodonosov.journal  bal revenues -N -E --no-elide
           200.0 BYN  revenues
           200.0 BYN    forex gain
           200.0 BYN      unrealised
$ hledger -f avodonosov.journal  bal revenues -N -E --no-elide
           200.0 BYN  revenues:forex gain:unrealised
```

With any hledger version >1.18, I'm unable to show the two zero-balance parent accounts here. I expected that -E, certainly when combined with --no-elide, would show them and their 0 balance as separate line items.

Xitian9:
Previously we only showed empty leaves.

Fixes #1688.


simonmichael:
Thanks!

PR #1693, opened 2021-09-10
imp: csv: sql: Use . as a decimal point in sql output for Print

### LINKED ISSUE #1684: disable number styling in SQL output ?
It was reported in this Hacker News thread (https://news.ycombinator.com/item?id=28421267) that our SQL output can show styled numbers, eg with thousands separators, and suggested that this is never appropriate. This sounds right. We should investigate the effect of commodity styles on numbers in SQL and JSON output - perhaps it's true to say that these numbers should always be in "vanilla" ISO format, ie no digit group separators and period as decimal mark ?

I think CSV output is a special case, number styling is sometimes desired there.

Xitian9:
command.

Fixes #1684.


simonmichael:
Many countries use comma as their decimal mark, so I think we probably should allow it in CSV ?

I'm not finding discussion of any SQL dialect allowing decimal commas, so forcing decimal period seems the right call there.

Xitian9:
Okay, this now only affects sql output.

Issue #1683, opened 2021-09-07
prevent excessive memory usage


simonmichael:
# Or: Don't let a typo hang user's machine

Eduardo reported in chat:

> I have made a typo in a transaction. The year was 20211 instead of 2021. Later, I tried to generate a balance report with --daily. My CPU and memory usage went to the roof and my computer didn't freeze because I stopped it soon. Can we consider this a bug or only a user error?

We must consider it both I think - it's not very good if it's this easy to hang your machine with hledger. Here are some tasks that come to mind:

1. Reproduce and study (quantify) this case
2. Search for and catalog other ways to cause the same result
3. Optimise them so they perform well enough (& decide what that means), if possible
4. As a fallback, try to detect operations likely to exceed available memory/acceptable resource usage and fail with an explanation ? (Process no more than 1m transactions per 1G system ram, or similar ? There could also be an interactive warning prompt, but we don't like these by default..)
5. Allow this to be overridden with a --no-memory-check flag.

Any thoughts ?

Xitian9:
As I understand it, the crux of this error is that through a simple user error, we can generate so many computer-generated postings that the CPU hangs. There are three types of computer-generated postings.

1. Auto-postings
2. Periodic postings
3. Summary postings

In this case it looks like summary postings are the proximate cause, but any of these others will compound the problem, resulting in 1, 2, or 4 traversals over the whole list of postings/transactions.

I think that summary postings with no content probably don't need to actually be stored, but can be generated on-demand at display time. Is this the case now, or do they continue to live in memory?

Another problem might be the amount of time to display large amounts of text on the terminal. Is this the case here, or does it hang before that stage?

Otherwise, I think the performance considerations we've been discussing regarding periodic and auto-postings apply here as well.

simonmichael:
Terminology police - I'd say automatic postings (generated by --auto), periodic transactions (generated by --forecast/--budget), and summary transactions (generated by -D/-W/-M/-Q/-Y/--period).

We don't know the exact command yet - Eduardo posted briefly on 2021-09-03. I suspect you're right, it was probably `balance --daily`, with report end date defaulting to 20211-01-01.

simonmichael:
[...]

PR #1696, opened 2021-09-13
fix: journal: Disallow parsing years later than 10000. (#1683)

Xitian9:
Though there is no technical reason for this restriction, it prevents a
user-input error where someone will fat-finger a 5-digit year with daily
reporting intervals, causing their computer to hang for a long time.


simonmichael:
When I said we shouldn't do this, in case someone wants to model cosmic events... I meant it. :)

simonmichael:
Ie, I'd prefer not to do this, at least not as a universal hard-coded limitation. If there's really no other way to protect users from runaway resource usage, maybe it could be an optional check or something..

Xitian9:
Ah, sorry! I misinterpreted a serious statement as sarcasm. Closing.

PR #1697, opened 2021-09-14
fix: extend application of commodity style to prices (except precision)

alang76:
This PR:
- Changes the behavior of applying commodity styles (such as those set by the  the -c/--commodity-style option) to include applying it to amount prices (except for precision, which is not affected). 
- Changes -c/--commodity-style into a general (reporting) option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,10 +73,11 @@ old
 hledger/test/addons/hledger-*
 tools/generatejournal
 tools/simplebench
-/examples/10*.journal
+/examples/[1-9]0*.journal
 *.webmanual.md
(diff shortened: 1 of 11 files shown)


Xitian9:
Looks good to me.

simonmichael:
It looks great @alang76, could you just add "and make --commodity-style a general flag" to the commit description before we merge.

simonmichael:
(No hurry. I would do it but I'm afraid of messing up your authorship.)

alang76:
Done!

simonmichael:
Thanks!

PR #1699, opened 2021-09-16
fix: bal: Make sure boring parent accounts in compound balance reports

### LINKED ISSUE #1698: Combining --tree with --begin for balancesheet causes miscalculation
I noticed a difference in the final line between these two commands, even though the only difference is the `--tree` flag. Here is an example (anonymised):

```
$ hledger bs --no-elide --begin=2019-01-01
                                       ||                                                 2021-09-14 
=======================================++============================================================
 Assets                                ||                                                            
---------------------------------------++------------------------------------------------------------
 assets:parent:child1                  ||                                                  GBP 90.00 
 assets:parent:child2                  ||                                                GBP 2731.02 
---------------------------------------++------------------------------------------------------------
                                       || GBP 2821.02 
=======================================++============================================================
 Liabilities                           ||                                                            
---------------------------------------++------------------------------------------------------------
---------------------------------------++------------------------------------------------------------
                                       || 
=======================================++============================================================
 Net:                                  || GBP 2821.02

$ hledger bs --no-elide --begin=2019-01-01 --tree
Balance Sheet 2021-09-14

                         ||                                                2021-09-14 
=========================++===========================================================
 Assets                  ||                                                           
-------------------------++-----------------------------------------------------------
 assets                  || GBP 4121.98 
   parent                ||                                               GBP 4121.98 
     child1              ||                                                 GBP 90.00 
     child2              ||                                               GBP 2731.02 
-------------------------++-----------------------------------------------------------
                         || GBP 4121.98
=========================++===========================================================
 Liabilities             ||                                                           
-------------------------++-----------------------------------------------------------
-------------------------++-----------------------------------------------------------
=========================++===========================================================
 Net:                    || GBP 4121.98
```

I can't explain why the the two `Net` numbers differ, even though the only difference is the `--tree` flag. This change in calculation doesn't seem to happen when I don't pass `--begin=2019-01-01`.

I'm afraid I can't figure out a small reproducible example, that's why I shared an anonymous sample output, in which I removed most of the lines.

I'm using the latest hledger version on Ubuntu:

```
$ hledger --version
hledger 1.21
```

Xitian9:
include historical postings when account declarations have undeclared
parents. (#1698)

Fixes #1698.


simonmichael:
Merged, thanks for the fast diagnosis and fix.

PR #1701, opened 2021-09-16
lib: Allow multiline comments in csv rules

maralorn:
This change gives the ability to write `comment \nIBAN: %iban\nBIC:%bic` or `comment1 \nIBAN: %iban\nBIC:%bic` and have every tag on a new line.

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -920,10 +920,11 @@ transactionFromCsvRecord sourcepos rules record = t
               ]
     code        = maybe "" singleline $ fieldval "code"
     description = maybe "" singleline $ fieldval "description"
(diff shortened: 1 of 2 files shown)


maralorn:
I have to admit this could lead to unexpected behavior for people if there is an unwanted newline in a csv field that they want to splice into a comment. I don‘t know if that is something that happens.
Theoretically we could preserve most of the old behavior with `maybe "" (unescapeNewlines . singleline`. That way newlines from the csv get replaced by spaces by explicit masked newlines written in the rule would get interpreted as desired.

simonmichael:
Great! I'll try to use this a bit. Preserving newlines from the CSV, in hledger comments, sounds harmless.

simonmichael:
Well, I haven't tested this further. Trusting all is well, thanks!

PR #1703, opened 2021-09-17
ref: performance: Improve performance for some high-use functions.

Xitian9:
mixedAmount(Looks|Is)Zero now operate directly on the MixedAmount,
rather than converting them to a list of amounts first.

mixedAmountCost no longer reconstructs the entire MixedAmount when there
are amounts with no cost.

transactionCheckBalanced only checks if signs are okay if sums are not
okay. It also only traverses the list of postings once when picking real
and balanced virtual postings.

accountNamesFromPostings now uses a Set instead of nubSort.

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -639,13 +639,13 @@ isNegativeMixedAmount m =
 -- i.e. does it have zero quantity with no price, zero quantity with a total price (which is also zero),
 -- and zero quantity for each unit price?
 mixedAmountLooksZero :: MixedAmount -> Bool
-mixedAmountLooksZero = all amountLooksZero . amounts . normaliseMixedAmount
+mixedAmountLooksZero (Mixed ma) = all amountLooksZero ma
 
 -- | Is this mixed amount exactly zero, ignoring its display precision?
 -- i.e. does it have zero quantity with no price, zero quantity with a total price (which is also zero),
 -- and zero quantity for each unit price?
 mixedAmountIsZero :: MixedAmount -> Bool
-mixedAmountIsZero = all amountIsZero . amounts . normaliseMixedAmount
+mixedAmountIsZero (Mixed ma) = all amountIsZero ma
 
 -- | Is this mixed amount exactly zero, ignoring its display precision?
 --
@@ -759,7 +759,9 @@ mapMixedAmountUnsafe f (Mixed ma) = Mixed $ M.map f ma  -- Use M.map instead of 
 -- | Convert all component amounts to cost/selling price where
 -- possible (see amountCost).
 mixedAmountCost :: MixedAmount -> MixedAmount
-mixedAmountCost = mapMixedAmount amountCost
+mixedAmountCost (Mixed ma) =
+    foldl' (\m a -> maAddAmount m (amountCost a)) (Mixed noPrices) withPrices
+  where (noPrices, withPrices) = M.partition (isNothing . aprice) ma
 
 -- -- | MixedAmount derived Eq instance in Types.hs doesn't know that we
 -- -- want $0 = EUR0 = 0. Yet we don't want to drag all this code over there.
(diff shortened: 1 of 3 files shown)


Xitian9:
These functions are dominated by journal parsing functions, but if you want to remove as much of the journal processing overhead as possible you can use the following journal.

```
~ daily
  assets:my:stuff:here   1
  revenues:your:stuff:there
```

And then test with a command like:
```
hledger bal --forecast -e 3000 +RTS -P -s
```

hledger-lib/Hledger/Data/Amount.hs
```diff
@@ -639,13 +639,13 @@ isNegativeMixedAmount m =
 -- i.e. does it have zero quantity with no price, zero quantity with a total price (which is also zero),
 -- and zero quantity for each unit price?
 mixedAmountLooksZero :: MixedAmount -> Bool
-mixedAmountLooksZero = all amountLooksZero . amounts . normaliseMixedAmount
+mixedAmountLooksZero (Mixed ma) = all amountLooksZero ma
```
simonmichael, line 642: While this certainly looks more efficient, couldn't it change the user-visible behaviour ? Eg with an amount like Mixed [usd 1, usd -1].

Xitian9: Mixed amount is now normalised by design, so you cannot store `mixed [usd 1, usd -1]` in a way that differs from `mixed [usd 0]`.

simonmichael: Ah, right. MixedAmount is now always normalised. (And probably there's no reason for normaliseMixedAmount to still be in the codebase..)

simonmichael:
Cc'ing benchmark results from chat: this speeds up report calculations a little (it's most visible when there's minimal parsing and much calculation, as in this example):
```
$ cat 1703.j 
~ daily
  assets:my:stuff:here   1
  revenues:your:stuff:there
$ export PERIOD=0000..1000; quickbench -w hledger,hledger.1703 -N3 "hledger -f 1703.j print --forecast=$PERIOD" "hledger -f 1703.j reg -W --forecast=$PERIOD" "hledger -f 1703.j bal -W --forecast=$PERIOD"
Running 3 tests 1 times with 2 executables at 2021-09-18 15:51:28 HST:

Best times 1:
+----------------------------------------++---------+--------------+
|                                        || hledger | hledger.1703 |
+========================================++=========+==============+
| -f 1703.j print --forecast=0000..1000  ||   12.87 |        11.05 |
| -f 1703.j reg -W --forecast=0000..1000 ||   12.06 |         9.04 |
| -f 1703.j bal -W --forecast=0000..1000 ||   11.27 |         9.28 |
+----------------------------------------++---------+--------------+

Best times 2:
+----------------------------------------++---------+--------------+
|                                        || hledger | hledger.1703 |
+========================================++=========+==============+
| -f 1703.j print --forecast=0000..1000  ||   12.07 |        12.39 |
| -f 1703.j reg -W --forecast=0000..1000 ||   11.35 |         9.52 |
| -f 1703.j bal -W --forecast=0000..1000 ||   10.89 |         9.56 |
+----------------------------------------++---------+--------------+

Best times 3:
+----------------------------------------++---------+--------------+
|                                        || hledger | hledger.1703 |
+========================================++=========+==============+
| -f 1703.j print --forecast=0000..1000  ||   12.54 |        11.94 |
| -f 1703.j reg -W --forecast=0000..1000 ||   11.64 |         9.68 |
| -f 1703.j bal -W --forecast=0000..1000 ||   11.05 |         9.45 |
+----------------------------------------++---------+--------------+
```

hledger-lib/Hledger/Data/Amount.hs
Xitian9: Correct. We left it in to easy migration. Do you think we can take it out now? (Same for `mixedAmountSquashPricesForDisplay`)

PR #1704, opened 2021-09-20
imp: performance: Add strictness annotations for some mixed amount

Xitian9:
arithmetic.

This seems to result in an approximately 5-10% speedup for reports
involving a lot of arithmetic (register reports, balance reports).

```
Running 5 tests 1 times with 2 executables at 2021-09-20 11:39:20 AEST:

Best times:
+----------------------------------------------------++----------------------------+-------------------------------+
|                                                    || profiling/hledger-masterNP | profiling/hledger-strictamtNP |
+====================================================++============================+===============================+
| -f examples/10000x1000x10.journal print            ||                       0.61 |                          0.59 |
| -f examples/10000x1000x10.journal register         ||                       9.06 |                          8.52 |
| -f examples/10000x1000x10.journal balance          ||                       0.52 |                          0.54 |
| -f examples/1000x1000x10.journal balance --weekly  ||                       0.52 |                          0.45 |
| -f examples/10000x1000x10.journal balance --weekly ||                       6.00 |                          5.64 |
+----------------------------------------------------++----------------------------+-------------------------------+
```

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -40,6 +40,7 @@ exchange rates.
 
 -}
 
+{-# LANGUAGE BangPatterns       #-}
 {-# LANGUAGE OverloadedStrings  #-}
 {-# LANGUAGE RecordWildCards    #-}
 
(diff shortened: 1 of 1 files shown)


Xitian9:
I've tried repeating this a few times, and I'm less convinced there's any difference at all. I'll leave this here a few days to see if others can repeat my initial results.

Xitian9:
Looks like that was a fluke.

PR #1705, opened 2021-09-20
Replace --pretty-tables with --pretty, a general option.

Xitian9:


Xitian9:
I feel like the author of the second commit should be `sed`, not me.

simonmichael:
Needs an entry in the hledger manual I think, wherever color is mentioned. (Perhaps both should be mentioned at https://hledger.org/hledger.html#output as well ? Not urgent)

Xitian9:
Docs added. Let me know what you think.

fancsali:
I personally would prefer the ASCII to be default. There are so many different terminals around (think about stuff like Termux on Android), and ASCII is nice enough -- but something that falls apart by default is rather annoying.

simonmichael:
@fancsali I'm sympathetic to this, but there's some value in having things be attractive by default. I'd like to quantify the risk so as to make the right trade-off here. Currently I don't have any real examples of where this would happen, can you think of one ?

simonmichael:
Useful resources:
- https://en.wikipedia.org/wiki/Box-drawing_character

simonmichael:
Here's one case from Serge in chat: Iosevka font configured in terminal and vs code, box drawing chars look ok in terminal but not when viewing the saved output in vs code

image

image

Xitian9:
I've removed the change of default from this PR. The remaining changes should be inoffensive.

simonmichael:
Thank you! 

We decided not to make pretty the default just yet. It's still something we'd quite like to do, but we need to search more for examples of breakage like the above. I'll keep this open for now.

PR #1707, opened 2021-09-20
ref: auto: Add auto postings during the transaction balancing phase,

Xitian9:
rather than during a second balancing pass during journal finalisation.

Adding some auto postings and forecast transactions to 10000x1000x10.journal.

```
Running 4 tests 3 times with 2 executables at 2021-09-20 22:40:31 AEST:

Best times 1:
+------------------------------------------------------------------------------++--------------------------+------------------------+
|                                                                              || profiling/hledger-master | profiling/hledger-auto |
+==============================================================================++==========================+========================+
| -f examples/10000x1000x10.journal print --auto --forecast -e 2030            ||                     0.90 |                   0.87 |
| -f examples/10000x1000x10.journal register --auto --forecast -e 2030         ||                    18.00 |                  17.22 |
| -f examples/10000x1000x10.journal balance --auto --forecast -e 2030          ||                     0.70 |                   0.67 |
| -f examples/10000x1000x10.journal balance --weekly --auto --forecast -e 2030 ||                     5.96 |                   5.96 |
+------------------------------------------------------------------------------++--------------------------+------------------------+

Best times 2:
+------------------------------------------------------------------------------++--------------------------+------------------------+
|                                                                              || profiling/hledger-master | profiling/hledger-auto |
+==============================================================================++==========================+========================+
| -f examples/10000x1000x10.journal print --auto --forecast -e 2030            ||                     0.87 |                   0.85 |
| -f examples/10000x1000x10.journal register --auto --forecast -e 2030         ||                    17.52 |                  17.14 |
| -f examples/10000x1000x10.journal balance --auto --forecast -e 2030          ||                     0.69 |                   0.67 |
| -f examples/10000x1000x10.journal balance --weekly --auto --forecast -e 2030 ||                     6.11 |                   5.91 |
+------------------------------------------------------------------------------++--------------------------+------------------------+

Best times 3:
+------------------------------------------------------------------------------++--------------------------+------------------------+
|                                                                              || profiling/hledger-master | profiling/hledger-auto |
+==============================================================================++==========================+========================+
| -f examples/10000x1000x10.journal print --auto --forecast -e 2030            ||                     0.87 |                   0.86 |
| -f examples/10000x1000x10.journal register --auto --forecast -e 2030         ||                    17.68 |                  16.85 |
| -f examples/10000x1000x10.journal balance --auto --forecast -e 2030          ||                     0.70 |                   0.67 |
| -f examples/10000x1000x10.journal balance --weekly --auto --forecast -e 2030 ||                     5.65 |                   5.94 |
+------------------------------------------------------------------------------++--------------------------+------------------------+
```

### DIFF
--- a/hledger-lib/Hledger/Data/Balancing.hs
+++ b/hledger-lib/Hledger/Data/Balancing.hs
@@ -8,6 +8,7 @@ Functions for ensuring transactions and journals are balanced.
 {-# LANGUAGE PackageImports      #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell     #-}
+{-# LANGUAGE TupleSections       #-}
 
 module Hledger.Data.Balancing
 ( -- * BalancingOpts
@@ -32,6 +33,7 @@ import "extra" Control.Monad.Extra (whenM)
 import Control.Monad.Reader as R
 import Control.Monad.ST (ST, runST)
 import Data.Array.ST (STArray, getElems, newListArray, writeArray)
+import Data.Bifunctor (first, second)
 import Data.Foldable (asum)
 import Data.Function ((&))
 import qualified Data.HashTable.Class as H (toList)
@@ -45,6 +47,7 @@ import Data.Time.Calendar (fromGregorian)
 import qualified Data.Map as M
 import Safe (headDef)
 import Text.Printf (printf)
+import Text.Show.Functions ()
 
 import Hledger.Utils
 import Hledger.Data.Types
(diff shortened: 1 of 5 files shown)


simonmichael:
I see tests are passing, but I wonder how this works without changing behaviour ? From memory the two balancing passes are required, one to balance transactions as far as possible using their local information, and another to do any remaining balancing using global information (balance assignments).

Xitian9:
These two balancing passes still happen, just inside the `balanceTransactionHelper` function. You can see that there are two calls to `inferMissing` and two calls to `checkBalanced`.

That said, I think you should hold off on this until after the next release. I think it needs more testing.

I had actually managed to implement this in a way such that #893b passed, and changed it back so that it maintains current behaviour. Is that behaviour which was decided would be unhelpful?

Xitian9:
Issues of correctness aside, I'm also getting slightly higher maximum residency with this solution than the existing one. Let me see if that can be smoothed over.

Xitian9:
Argh, I'm really having trouble nailing down the magnitude of the speed improvement, if indeed there is any. There's too much noise in the data. If you have time could you check to see whether you're seeing meaningful improvements on your system?

I'm fond of this solution, but as is I'm thinking that the gains are relatively small, and the risk of introducing regressions relatively high. If you're not seeing big gain, I might favour dropping this.

simonmichael:
This sounds worth looking at carefully as a possible refactoring/functionality change, aside from performance. Perhaps let's leave it open for next release cycle.

Xitian9:
I think this PR was really a dead end. I'll withdraw this.

PR #1708, opened 2021-09-20
cln: Remove unnecessary normaliseMixedAmount(SquashPricesForDisplay)? functions.

Xitian9:


simonmichael:
I was thinking about leaving those in as noops to keep scripts working, but we don't have that many api users - let's clean up.

PR #1709, opened 2021-09-20
fix: ui: Do not display a screen full of .. when there are no transac…

### LINKED ISSUE #822: hledger-ui shows an empty transaction list as a screenful of  `..  ..`
With `hledger-ui` 1.9 (built a week ago using `stack install --resolver=lts hledger-lib-1.9 hledger-1.9 hledger-ui-1.9`) on OSX; attempting to show **Cleared** or **Pending** transactions when there are none to show causes `hledger-ui` to display `100/0` empty transactions `.. ..`.

To reproduce:

    $ less mwe.journal
    
    2018/06/18 Widget Co
        assets:checking              -20
        expenses:widgets              20

    $ hledger-ui --register checking -C -f mwe.journal

Output is (*long screenshot*):

mwe output

(In addition, navigating upwards through the 'transactions' decrements the number of transactions and also prevents navigating back down)

Checking the whole journal (as opposed to an account) has different behaviour, showing `1/0` and two spaces:

    $ hledger-ui -C -f mwe.journal

overall journal

Xitian9:
…tions. (#822)

Fixes #822.


simonmichael:
Cc'ing from chat: dots still visible for me, eg when doing:

- hledger-ui on any file, press 0, RIGHT
- hledger-ui on any file, RIGHT, / alksjdlfakjdslkfjaskldj

Xitian9:
I'm not seeing that. When I do what you say, I get a screen full of blank lines, with the bottom blank line selected, and a title bar saying `All transactions (100/0)`.

simonmichael:
We never did find the culprit, but it's working now. Thank you very much for fixing this! I have been ignoring it for years.

PR #1718, opened 2021-09-23
imp!: value: For register reports with no reporting interval and

Xitian9:
--value=end, historical reports should be valued at journal/report end
date.

I'm waiting for feedback on this change from the mailing list (https://groups.google.com/g/hledger/c/Rc-H4koWcbA), but presuming nobody objects here's the commit.


Xitian9:
There's been no objection on the mailing list. I think this is safe to change. If people object later we can always revert.

simonmichael:
Sounds ok to me - thanks.

PR #1723, opened 2021-09-27
Drop base-compat-batteries, and bump megaparsec

Xitian9:


Xitian9:
Maybe hold off on merging this until we are sure it passes CI. There's a non-zero chance that `base-compat-batteries` is still necessary in some cases.

Xitian9:
Passes!

simonmichael:
I guess all ghc/base versions we support now are compatible enough. Great!

Issue #1670, opened 2021-08-25
support `decimal-mark` in journal files also


simonmichael:
Cf https://github.com/simonmichael/hledger/issues/793#issuecomment-723694564: `decimal-mark` in CSV rules files declares which character is the decimal mark when parsing numbers. We'd like to be able to do this in journal files too. 

Currently `commodity` directive is used for this, but it also sets the decimal mark to be displayed. This dual function can lead to confusion, and needing to put the directive in just the right place. Eg from chat:

```journal
;commas.journal

2021-05-01 YANDEX.PLUS
    Assets:Yandex            -199,00 RUB
    Expenses:Unknown:RUB
```
```journal
;main.journal

; first amounts parsed set default display style, 1000,00 in this case
include commas.journal

; commodity directive must go here, to set desired 1000.00 display style without disrupting parsing of commas.journal
commodity 1000.00 RUB

2021-05-19 MTS Phone
    Assets:Yandex         -100.5 RUB
    Expenses:Phone         100.5 RUB
```

It's not hard to work around this once you understand what's happening, but it's hard to understand at first.

With `decimal-mark` declared in each file (or at least in commas.journal), `commodity`'s placement would not matter:
```journal
;commas.journal
decimal-mark ,
2021-05-01 YANDEX.PLUS
    Assets:Yandex            -199,00 RUB
    Expenses:Unknown:RUB
```
```journal
;main.journal
commodity 1000.00 RUB
include commas.journal
2021-05-19 MTS Phone
    Assets:Yandex         -100.5 RUB
    Expenses:Phone         100.5 RUB
```

`decimal-mark`'s area of effect would be all subsequently-parsed amounts, including in included files, until end of current file, or another decimal-mark directive. (So another decimal-mark in an included file would take precedence, until the end of that included file).

It's probably too disruptive to remove `commodity`'s dual function at this point, so that would continue, but a `decimal-mark` would override it for number parsing. This would allow a simpler guaranteed-to-work recipe in times of trouble: `decimal-mark` in all files to control number parsing, `commodity` somewhere to control display.

simonmichael:
Merged @la-wu's #1726.

PR #1726, opened 2021-09-28
Support decimal-mark directive in journal files

la-wu:
Implementation and sample behavioral tests for https://github.com/simonmichael/hledger/issues/1670


simonmichael:
Great! If this meets the spec in #1670, I think we just need to document it, in the https://hledger.org/hledger.html#directives table and in its own "Decimal mark" subsection.

simonmichael:
(in hledger.m4.md).

simonmichael:
Docs added, merged as fb817ffa5. Thank you @la-wu!

Issue #1689, opened 2021-09-08
subaccount-inclusive balance assignments should generate a posting for each subaccount ?


simonmichael:
In a recent example, to transfer all earnings to equity I wrote essentially this, naming each account separately:
```
1/30 temporary for reporting: retain earnings
    revenues                                     == 0
    revenues:forex gain                          == 0
    revenues:forex gain:unrealised               == 0
    equity:retained earnings
```
I actually wanted to write just:
```
1/30 temporary for reporting: retain earnings
    revenues                                      ==* 0
    equity:retained earnings
```
The problem with that seemed to be that it generated three postings to `revenues` - so a zero inclusive balance of `revenues` was achieved, but the subaccount balances were not zeroed out. I think instead it should generate one posting to each account, zeroing out each one.

la-wu:
I think this is pretty clear for a balance of `0 BYN` but what happens for non-zero inclusive assignments? e.g for

```
2020-09-27 temporary for reporting: open
    revenues                                -1 BYN
    revenues:forex gain                     -1 BYN
    revenues:forex gain:unrealised          -1 BYN
    assets:fudge                             3 BYN

2021-1-30 temporary for reporting: retain earnings
    revenues                                       ==* 1 BYN
    equity:retained earnings
```

Should we generate

```
2021-1-30 temporary for reporting: retain earnings
    revenues                                 2 BYN ==* 1 BYN
    revenues:forex gain                      1 BYN ==* 0 BYN
    revenues:forex gain:unrealised           1 BYN ==* 0 BYN
    equity:retained earnings                -4 BYN
```

or

```
2021-1-30 temporary for reporting: retain earnings
    revenues                                 2 BYN ==* 1 BYN
    revenues:forex gain                      2 BYN ==* 1 BYN
    revenues:forex gain:unrealised           2 BYN ==* 1 BYN
    equity:retained earnings                -6 BYN
```

or something else?

simonmichael:
Good question! 

I feel the zero case is intuitive enough, and the current behaviour non-useful enough, that we could allow it to work differently, as a special case. Any thoughts ?

PR #1727, opened 2021-09-28
Set subaccount balances to zero on inclusive assignments of zero

la-wu:
Implementation and sample behavioral tests for https://github.com/simonmichael/hledger/issues/1689

### DIFF
--- a/hledger-lib/Hledger/Data/Balancing.hs
+++ b/hledger-lib/Hledger/Data/Balancing.hs
@@ -474,7 +474,7 @@ balanceTransactionAndCheckAssertionsB (Right t@Transaction{tpostings=ps}) = do
   mapM_ checkIllegalBalanceAssignmentB ps
   -- for each posting, infer its amount from the balance assignment if applicable,
   -- update the account's running balance and check the balance assertion if any
-  ps' <- mapM (addOrAssignAmountAndCheckAssertionB . postingStripPrices) ps
+  ps' <- fmap concat $ mapM (addOrAssignAmountAndCheckAssertionB . postingStripPrices) ps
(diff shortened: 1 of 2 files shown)


wgslr:
Hi! As a hledger user, I wanted to add my input and note that I'm strongly against this change.
1. This is a breaking change, modifying the history of existing journals. If we want the functionality showed in the first post, it should be new operator. I particularly dislike making a specific number (`0`) a magic value.
2. I use subaccounts for budgeting, which raises instances where the behavior described in #1689 is not what I want. Consider the following balances:
```
assets:checking           10foo
assets:checking:clothes   200foo
assets:checking:food     -200foo
```
`hledger balance --tree checking` will show a total balance of 10foo.
Let's say I look at my bank statement and discover that the checking account balance is in fact 0foo. I will write a balance assignment to correct the value in hledger:
```
2021-10-16
   assets:checking  ==* 0foo
   equity:reconciliation
```
In my opinion, the expected behavior is to generate posting in the top-level account to adjust the balance and **not touch the subaccounts**:
```
2021-10-16
   assets:checking  -10foo ==* 0foo
   equity:reconciliation
```

I do not want to loose the sub-account balances, which show that I overspent on food instead of buying clothes.

simonmichael:
Hi @la-wu, are you still interested in this hledger PR: Set subaccount balances to zero on inclusive assignments of zero ?

Hi @wgslr, thanks for your feedback in 2021. I feel that since we have both subaccount-inclusive (`*`) and subaccount-variants of the assertion/assignment operators, there need be no loss of functionality.

However I'm hazy on the exact proposal and ramifications right now, and I don't remember feeling the need for this feature again. So perhaps we should drop it ?

PR #1730, opened 2021-10-08
Fix PR template comment syntax

toonn:
The # syntax for comments results in all the lines being rendered as
separate headers because it's markdown syntax for titles. I believe the
markdown custom is to use `<!-- HTML comments -->`.

Using the comment syntax customarily used in many other languages here meant I accidentally included a bunch of strange section headers in my last PR : )
I figured I might as well save other contributors those few seconds for every PR.

The URL in the PR template redirects to a general contributing guide,
which is useful, but it's hard to see the trees for the forest when just
looking for PR and commit guidelines.

I updated it with the URL for the PR page in the developer docs, as was the original intention AFAICT. Maybe we want to include the URL to the general contributor guide as well?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,11 @@
-# Thanks for your pull request! We appreciate it. 
-# If you're a new developer, FOSS contributor, or hledger contributor, welcome.
-# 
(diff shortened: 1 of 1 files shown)


simonmichael:
Fantastic! This has been mysteriously borked for a long time. :)

PR #1732, opened 2021-10-11
fix: roi: use posting dates when available

adept:
This will not change the results computed for the typical use-case,
just makes "roi" more thorough/consistent by supporting posting dates
when they are provided.

As far as I am aware, there are no open tickets for this


simonmichael:
Drat, how did those commit messages get in. This PR looks good, thanks for the improvement!

PR #1734, opened 2021-10-14
cln: Move posting rendering functions into Hledger.Data.Posting.

Xitian9:
Replace showPosting with a wrapper around postingAsLines.

The functions textConcat(Top|Bottom)Padded are no longer used anywhere
in the code base, and can be removed if desired.

This produces slightly different output for showPosting, in particular
it no longer displays the transaction date. However, this has been
marked as ‘for debugging only’ for a while, and is only used in
hledger-check-fancy assertions. The output there is still acceptable.


simonmichael:
Sounds reasonable. Thanks!

We can keep textConcat* around for now, general utilities like that can come in handy later or for scripters.

PR #1737, opened 2021-10-22
; imp: use megaparsec's prettyprinter for assertion parse errors

ShrykeWindgrace:
The `Show` instance for the error bundle is not human-readable, `megaparsec` provides a prettyprinter.


simonmichael:
Thank you!

PR #1743, opened 2021-10-31
Fix the link to the developer workdflows

jazcarate:
The "Developer workflows" was moved to the `WORKFLOWS.md` file


simonmichael:
Thanks!

PR #1747, opened 2021-11-06
Add Bifunctor and improve Semigroup instances for PeriodicReport and PeriodicReportRow

Xitian9:


Xitian9:
One thing to note is that this change in the Semigroup instance might cause unexpected migration issues. Previously when using `<>` on `PeriodicReportRows`, it would take the second name and combine all the amounts as usual. For example:

```
PeriodicReportRow "first" [Sum 1] (Sum 1) (Sum 1) <> PeriodicReportRow "second" [Sum 2] (Sum 2) (Sum 2)
  -> PeriodicReportRow "second" [Sum 3] (Sum 3) (Sum 3)
```

The new instance requires a `Semigroup` instance for the type of prrName. If it already has one, like `Text`, it would indeed have a Semigroup instance, and so the result would probably not be what you expect.

```
PeriodicReportRow "first" [Sum 1] (Sum 1) (Sum 1) <> PeriodicReportRow "second" [Sum 2] (Sum 2) (Sum 2)
  -> PeriodicReportRow "firstsecond" [Sum 3] (Sum 3) (Sum 3)
```

To work around this, you can use `first Data.Semigroup.Last` to get the previous behaviour.

I'm not sure how widely used the Semigroup instance is. It's not used in our code at all, was relatively recently introduced, and is not widely advertised. This might not be a problem in practice.

Xitian9:
I'm no longer certain about the wisdom of the Semigroup instance change. But what I definitely need is a separate way to the first name, rather than the second. It's a bit of an open problem which is the best behaviour, or even if one should be preferred.

Xitian9:
I've changed this so the Semigroup instance of PeriodicReportRow now preserves the first name, rather than the second. This seems a more sensible default.

PR #1751, opened 2021-11-12
Introduce multiline balance report display.

### LINKED ISSUE #1739: cashflow, incomestatement: "more..." in new version of hledger
I attach simple journal for a game, where player uses several commodities, he pays membership in USD and he earns GOLD, GEMS and STONES in the game. It is just a game to demonstrate some issue in the new hledger.

hledger 1.2, I like that I can see all details in "is" and "cf" reports:

```
$ hledger --version
1.2

$ hledger is
...
Total:
--------------------
          -1000 GEMS
           -643 GOLD
         -100 STONES
              10 USD
```

```
$ hledger --version
hledger 47b5b6b, linux-x86_64

$ hledger is
...
==========================++=================================
 Net:                     ||   1000 GEMS, 643 GOLD, 2 more.. 
```

The report with "more.." trick could be useful but I would like to have a switch to expand the report to see all details in similar way like old hledger reported. The same problem with "is", "bs", "bse" and "cf" report.
[demo4.txt](https://github.com/simonmichael/hledger/files/7405071/demo4.txt)

Xitian9:
We also stop eliding long amounts in balance reports, which has caused a lot of confusion (Fixes #1739).

### DIFF
--- a/hledger/Hledger/Cli/Commands/Balance.hs
+++ b/hledger/Hledger/Cli/Commands/Balance.hs
@@ -318,12 +318,12 @@ balancemode = hledgerCommandMode
     ,flagNone ["percent", "%"] (setboolopt "percent") "express values in percentage of each column's total"
     ,flagNone ["invert"] (setboolopt "invert") "display all amounts with reversed sign"
     ,flagNone ["transpose"] (setboolopt "transpose") "transpose rows and columns"
-    ,flagReq  ["commodity-layout"] (\s opts -> Right $ setopt "commodity-layout" s opts) "ARG"
+    ,flagReq  ["layout"] (\s opts -> Right $ setopt "layout" s opts) "ARG"
       (unlines
-        ["show multicommodity amounts in the given ARG. ARG can be:"
-        ,"'oneline':   show all commodities on a single line"
-        ,"'multiline': show each commodity on a new line"
-        ,"'column':    show commodity symbols in a separate column and amounts as bare numbers"
+        ["how to show multi-commodity amounts:"
+        ,"'wide[,WIDTH]': all commodities on one line [elided at WIDTH]"
+        ,"'tall'        : each commodity on a new line"
+        ,"'bare'        : bare numbers, symbols in a column"
         ])
     ,outputFormatFlag ["txt","html","csv","json"]
     ,outputFileFlag
@@ -407,13 +407,13 @@ balance opts@CliOpts{reportspec_=rspec} j = case balancecalc_ of
 -- | Render a single-column balance report as CSV.
 balanceReportAsCsv :: ReportOpts -> BalanceReport -> CSV
 balanceReportAsCsv opts (items, total) =
-    ("account" : ((if commodity_layout_ opts == CommodityColumn then (:) "commodity" else id) $ ["balance"]))
+    ("account" : ((if commodity_layout_ opts == CommodityBare then (:) "commodity" else id) $ ["balance"]))
   :  (concatMap (\(a, _, _, b) -> rows a b) items)
   ++ if no_total_ opts then [] else rows "total" total
   where
     rows :: AccountName -> MixedAmount -> [[T.Text]]
     rows name ma = case commodity_layout_ opts of
-      CommodityColumn ->
+      CommodityBare ->
           fmap (\(k, a) -> [showName name, k, renderAmount . mixedAmount . amountStripPrices $ a])
           . M.toList . foldl' sumAmounts mempty . amounts $ ma
       _ -> [[showName name, renderAmount ma]]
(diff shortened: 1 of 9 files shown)


simonmichael:
Good stuff! First notes:
```
     --commodity-layout=STYLE  show multicommodity amounts in the given
                               STYLE. STYLE can be:
                               'singleline': display all commodities on a
                               single line
                               'multiline':  display each commodity on a
                               separate line
                               'column':     display commodity symbols in a
                               separate column and amounts as bare numbers, one
                               row per commodity
```
- currently supported on the balance commands, I wonder if it makes sense anywhere else
- --commodity-column still supported as hidden flag, good
- let's support any unique value prefix like we do else where (single, multi, m..)
- with no argument, or any unrecognised argument, let's have it list the valid arguments
- singleline -> oneline for easier typing/saying ?
- 'elided' not supported right now, ok
- 'multiline''s help could also describe 'column', not sure if there's a better wording. "display one commodity per line" ?
- a small example shown in all possible modes might be worth putting in docs or at least in tests. Eg I haven't seen much difference between multiline and column yet

Xitian9:
Updated. I'm not completely sure how you intended to combine the multiline help and the column help. Let me know what you think.

Xitian9:
It was raised before that we may need a shorter abbreviation for ‘Commodity’ in the commodity column. Maybe let's bring that up here again. If I recall, options were ‘C’, ‘Com’, or ‘Comm’

Xitian9:
Currently with the implementation, if there are more commodities in one column than another, blank spaces will be displayed.

```
   ||  2020   2021 
===++==============
 a || 1.00A  1.00D 
 a || 1.00B  1.00E 
 a || 1.00C  1.00F 
 a || 1.00D        
---++--------------
   || 1.00A  1.00D 
   || 1.00B  1.00E 
   || 1.00C  1.00F 
   || 1.00D        
```

It is arguably preferable to just pad those with zeros with no commodity, like `--empty` does:
```
   ||  2020   2021 
===++==============
 a || 1.00A  1.00D 
 a || 1.00B  1.00E 
 a || 1.00C  1.00F 
 a || 1.00D      0
---++--------------
   || 1.00A  1.00D 
   || 1.00B  1.00E 
   || 1.00C  1.00F 
   || 1.00D      0
```

Let me know if you have a preference.

simonmichael:
Looking good. More nitpicks:

- it doesn't yet list the valid arguments when none is provided. But perhaps that's more difficult, not a blocker.
- ideally it would accept any unique prefix, not just a few hard coded ones.  (Arguable. We avoid this with command names. But generally we do it with option values, where it seems harmless enough, and it might feel more consistent.)
- probably out of scope for this PR, but we should think about how much we want to support this in the other output formats like csv.

simonmichael:
I'm not sure about space or 0. I haven't felt any pain from the space yet.

With `--commodity-layout=multiline` and a multi-column report, it's easy to see the same commodities on different lines in the totals row. It looks slightly sloppy and harder to compare them across periods. I wonder how hard it would be to insert empty lines to keep the layout consistent across periods.

Xitian9:
I think putting effort into aligning commodities is probably not the best use of effort here. It's a sticky problem, and if you really need alignment then --commodity-layout=column is the best choice.

That said, something of independent value would be a --commodity-order option, which controls the order of commodities in any report. I'm imagining something like --commodity-order="USD,GBP,CAD", the all displays would put USD first (if present), GBP second (if present), CAD third (if present), and any remaining quantities in alphabetical order.

simonmichael:
Fair enough. It's easy to spend too much time on low impact things.

Xitian9:
I've addressed many of your points.

- Making it display a specific message on `--commodity-layout` is possible, but would mean working around using `flagReq`. I haven't done this, and unless you feel strongly don't intend to.
- Yes, I think what good defaults are for other formats requires more thinking than just ‘enable everything everywhere’. In particular, I'm of the opinion that perhaps `--commodity-column` should be the default for `csv` output.
- I've switch to putting a `0` when there aren't enough commodities, because if it's just a blank space it looks odd in big reports. However, if you feel otherwise, I'm happy to change it back.

hledger/Hledger/Cli/Commands/Balance.hs
```diff
@@ -309,20 +312,28 @@ balancemode = hledgerCommandMode
...
     ,flagReq  ["format"] (\s opts -> Right $ setopt "format" s opts) "FORMATSTR" "use this custom line format (in simple reports)"
     ,flagNone ["sort-amount","S"] (setboolopt "sort-amount") "sort by amount instead of account code/name (in flat mode). With multiple columns, sorts by the row total, or by row average if that is displayed."
     ,flagNone ["percent", "%"] (setboolopt "percent") "express values in percentage of each column's total"
     ,flagNone ["invert"] (setboolopt "invert") "display all amounts with reversed sign"
     ,flagNone ["transpose"] (setboolopt "transpose") "transpose rows and columns"
-    ,flagNone ["commodity-column"] (setboolopt "commodity-column")
-      "show commodity symbols in a separate column, amounts as bare numbers, one row per commodity"
+    ,flagReq  ["commodity-layout"] (\s opts -> Right $ setopt "commodity-layout" s opts) "ARG"
+      (unlines
+        ["show multicommodity amounts in the given ARG. ARG can be:"
```
simonmichael, line 323: "how to show multicommodity amounts:"  ?

hledger/test/balance/multicommodity.test
```diff
@@ -1,31 +1,60 @@
...
+# 2. Display multiline if requested
+$ hledger -f- bal -Y --commodity-layout=multiline
+Balance changes in 2020-01-01..2021-12-31:
+
+   ||  2020   2021 
+===++==============
+ a || 1.00A  1.00D 
+ a || 1.00B  1.00E 
+ a || 1.00C  1.00F 
+ a || 1.00D      0
```
simonmichael, line 32: I have doubts about this 0. Usually a number in a report means something quantitative, but unlike `--commodity-layout=column`, here this cell has no real meaning, and the 0 is simply cosmetic to avoid empty spaces. I still need to try it out in practice.

simonmichael (commented):
Let's also mention this option's existence in Balance.md.

hledger/test/balance/multicommodity.test
Xitian9: Yes, happy to either have a space or 0. Your call on this one.

simonmichael: I'm inclined to show a space right now. It helps differentiate `multiline` and `column` modes. It would be harder for machine parsing but `multiline` isn't intended for that.

simonmichael:
> we may need a shorter abbreviation for ‘Commodity’ in the commodity column

I have some long commodity names, so the width commodity column didn't seem a problem. If the report includes only single-character commodity symbols, it looks like a bit of wasted space. I guess you could make it fit the column width to the widest commodity symbol, truncating the heading accordingly. Not sure if it's worthwhile.

simonmichael:
Nice one!

PR #1757, opened 2021-11-16
imp: ui: Display an error message on invalid regexp, rather than

### LINKED ISSUE #1394: ui: show the error when a malformed query is entered
Since 1.19, hledger-ui has become more robust: entering malformed regular expressions like `?` or `foo??` at the `/` prompt  no longer causes the program to exit. But the problem is just ignored; instead, it should display the error message - somehow or other.

Xitian9:
silently ignoring. (Fixes #1394)


simonmichael:
I think this can be improved:
- enter a first query, `/` `a`
- enter a second, malformed query, `/`, `DEL`, `?`
- now it shows an error message and the prompt, but also has restored the first query value, `a`. I think it should show no value after entering a malformed one.

Xitian9:
Do you think it would make more sense to display the unsuccessful query? On the theory that it might have been a long one with a single typo that somebody wants to correct?

simonmichael:
Looks good!

PR #1760, opened 2021-11-18
ui: add mouse support

simonmichael:
This PR adds mouse support to hledger-ui. These should work:

- left click on an account should enter its register
- left click on a register transaction should enter the transaction screen
- mouse wheel/trackpad swipe should scroll the accounts and register screens
- left click on the left-most column or top-most row should return to the previous screen (like the LEFT key)
- if mouse support doesn't work on some terminal/platform, it should at least be harmless (unnoticeable)

Known issues:
- ~There seems to be a bounce issue - sometimes one click has the effect of two.~
- ~I would like to left click on the top or left border to navigate back, but these areas seem unclickable.~


simonmichael:
Clicking is now debounced and robust. 
Clicking left-most column or top-most row exits to previous screen. 
Also, a small fix for an old accounts screen movement bug.

simonmichael:
Plenty more to do, but I think this is a good start.

PR #1764, opened 2021-11-21
imp: stats: also show run time and throughput

simonmichael:
CliOpts has a new field, progstarttime_.

Currently the new stats are always printed on stdout, ignoring --output-file/--output-format.


simonmichael:
[Docs](https://github.com/simonmichael/hledger/blob/c709a22c648c3667e7d5242fb5825eed74794dca/hledger/Hledger/Cli/Commands/Stats.md).

PR #1765, opened 2021-11-22
feat: bal: with --declared, include all declared accounts

simonmichael:
Together with -E, this allows showing a balance for all accounts, both
used and declared. I mainly want this for hledger-ui, but there's no
harm in exposing it in the balance command as well. This is somewhat
consistent with the accounts and payees commands.

### DIFF
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE NamedFieldPuns   #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-|
@@ -26,7 +27,6 @@ module Hledger.Reports.MultiBalanceReport (
   getPostingsByColumn,
   getPostings,
   startingPostings,
-  startingBalancesFromPostings,
   generateMultiBalanceReport,
   balanceReportTableAsText,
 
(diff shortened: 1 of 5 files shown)


simonmichael:
I reverted this merged PR after finding issues. I have just pushed a fixed version  d337f77e50749679581ca330679d40db8fd17e4d. This was quite tricky; I think it's good now, 
but I'll wait at least a day, and any feedback is welcome.

Commit message:
```
feat: bal: with --declared, include declared accounts (#1765)

Together with -E, this allows showing a balance for all accounts, both
used and declared. I mainly want this for hledger-ui, but there's no
harm in exposing it in the balance command as well. This is somewhat
consistent with the accounts and payees commands.
```

Doc:

> ### Showing declared accounts
> 
> With `--declared`, 
> accounts which have been declared with an [account directive](#declaring-accounts)
> will be included in the balance report, even if they have no transactions.
> (Since they will have a zero balance, you will also need `-E/--empty` to see them.)
> 
> More precisely, *leaf* declared accounts (with no subaccounts) will be included,
> since those are usually the more useful in reports.
> 
> The idea of this is to be able to see a useful "complete" balance report,
> even when you don't have transactions in all of your declared accounts yet.

simonmichael:
I reverted this merged PR after finding issues. I have just pushed a fixed version 6319d61. This was quite tricky; I think it's good now, but I'll wait at least a day, and any feedback is welcome.

Commit message:
```
feat: bal: with --declared, include declared leaf accounts (#1765)

Together with -E, this shows a balance for both used and declared
accounts (excluding empty parent accounts, which are usually not
wanted in list-mode reports).

This is somewhat consistent with --declared in the accounts and payees
commands, except for the leaf account restriction.

The idea of this is to be able to see a useful "complete" balance
report, even when you don't have transactions in all of your declared
accounts yet. I mainly want this for hledger-ui, but there's no harm
in exposing it in the balance CLI as well.
```

Doc:

> ### Showing declared accounts
> 
> With `--declared`, 
> accounts which have been declared with an [account directive](#declaring-accounts)
> will be included in the balance report, even if they have no transactions.
> (Since they will have a zero balance, you will also need `-E/--empty` to see them.)
> 
> More precisely, *leaf* declared accounts (with no subaccounts) will be included,
> since those are usually the more useful in reports.
> 
> The idea of this is to be able to see a useful "complete" balance report,
> even when you don't have transactions in all of your declared accounts yet.

simonmichael:
Repushed with fixes for bs/cf/is, showing the correct declared accounts in each subreport. No known open issues.

PR #1772, opened 2021-12-04
imp: check: Do not check commodities for zero amounts with empty

### LINKED ISSUE #1767: print -x and check -s are incompatible for zero-amounts
test.journal:

```
<directives>

2021-12-02 Test
    assets:test               $0 = $0
```

hledger print -x -f test.journal:

```
2021-12-02 Test
    assets:test               0 = $0
```

which when ran through hledger check -s:

```
hledger: undeclared commodity ""
in transaction at: "/home/nixos/finance/test.journal" (lines 4-5)

  2021-12-02 Test
      assets:test               0 = $0
```

I am fine with the commodity being dropped for 0 in `print -x`, but in that case `check -s` should not check commodities for 0. It would also be fine to fix it the opposite way and not drop commodities for 0.

This seems closely related to (maybe a more general form of) https://github.com/simonmichael/hledger/issues/1356?

Setting a default commodity makes `check -s` work.

Tested with hledger 1.21

Xitian9:
commodity. (Fixes #1767)

### DIFF
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -403,9 +403,14 @@ journalCheckCommoditiesDeclared j =
                           (linesPrepend "  " . (<>"\n") . textChomp $ showTransaction t)
       where
         mfirstundeclaredcomm =
-          find (`M.notMember` jcommodities j) . map acommodity $
-          (maybe id ((:) . baamount) pbalanceassertion) . filter (/= missingamt) $ amountsRaw pamount
+          find (`M.notMember` jcommodities j)
+          . map acommodity
+          . (maybe id ((:) . baamount) pbalanceassertion)
+          . filter (not . isIgnorable)
+          $ amountsRaw pamount
 
+    -- Ignore missing amounts and zero amounts without commodity (#1767)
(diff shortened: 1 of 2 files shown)


simonmichael:
> commodityless zero amounts will not fail

Doesn't this weaken the commodities check, such that you can no longer be certain there are no typos in your commodities  ?

Can we say for sure that accidentally omitting the commodity symbol from a zero amount is always fine and therefore we needn't allow it to be checked ?

Xitian9:
If you put an undeclared commodity on a zero amount that will still check as normal. The only case which changes is a zero amount with an empty commodity.

I think we're fairly safe doing this, because mixed amounts will (almost) automatically prune zero amounts in any case. Almost every usage of `MixedAmount` involves passing through `amounts :: MixedAmount -> [Amount]`, which will remove all zero amounts (unless there are only zero amounts, in which case there are some rules for pruning all but one of them).

simonmichael:
I imagine you're right that it probably won't cause problems in practice. 

I'm a little sad that we would give up the ability to check "all of my amounts have a declared commodity symbols", for the obsessive neat freaks among us. 

But on the whole this change sounds like an improvement, I can merge soon.

PR #1773, opened 2021-12-06
imp: csv: Handle --layout=tidy with the legacy balance report.

Xitian9:
Note that this gives output compatible with the multibalancereport,
which means it is different from the csv from the legacy balance report.
This means tools only need to be able to handle one tidy csv format.


Xitian9:
Based on discussion in the chat, this also makes the following changes to tidy csv.

1. Both start and end dates are reported for all intervals, rather than a string representation of the interval itself.
2. Total and average rows are not included in tidy csv, as they complicate the data representation and can be easily calculated.

PR #1788, opened 2021-12-15
fix: Ensure head and tail are not called on empty account names.

Xitian9:


simonmichael:
Here's a repro for the issue reported by sol:
```
alias /a:.*/=
2021-01-01
  a:aa     1
  b
```
```
$ hledger print
hledger: Data.Text.Internal.Fusion.Common.head: Empty stream
  error, called at libraries/text/src/Data/Text/Internal/Fusion/Common.hs:943:24 in text-1.2.4.1:Data.Text.Internal.Fusion.Common
```

This PR does prevent the error, but I think it needs a little more thought:
```
$ hledger print
2021-01-01                                                              
                    1                                                   
    b                                                                   
                                                                        
```

Xitian9:
Indeed, the problem occurs when using an alias to completely remove an account name, replacing it with an empty string. In this PR I opted obey the command, but I agree it's usually not what people want. We could replace an empty account name with `...`, as we do when using `--depth=0`, or we could throw a more helpful error.

simonmichael:
Tough one..

simonmichael:
Phrasing it more generally, should user be allowed to create malformed account names (and invalid-journal output from print) with regexp account aliases ? Another example, not involving empty account names:
```
2020-01-01
  old    1
  other
```
```
$ hledger -f c.j --alias old="new  USD" print
2021-01-01
    new  USD               1
    other

$ hledger -f c.j --alias old="new  USD" print | hledger -f- print
2021-01-01
    new             USD 1
    other
```

Easiest resolution: add a warning in alias and print docs.
Harder (at least): after applying aliases, validate the result somehow..

hledger-lib/Hledger/Utils/Text.hs
```diff
@@ -169,15 +169,17 @@ stripquotes s = if isSingleQuoted s || isDoubleQuoted s then T.init $ T.tail s e
 
 isSingleQuoted :: Text -> Bool
 isSingleQuoted s =
-  T.length (T.take 2 s) == 2 && T.head s == '\'' && T.last s == '\''
+  T.length s >= 2 && T.head s == '\'' && T.last s == '\''
 
 isDoubleQuoted :: Text -> Bool
 isDoubleQuoted s =
-  T.length (T.take 2 s) == 2 && T.head s == '"' && T.last s == '"'
+  T.length s >= 2 && T.head s == '"' && T.last s == '"'
```
simonmichael, line 176: I'm a little confused.. was one or both of these causing the reported bug ? I'm not quite seeing why the old code was wrong.

hledger-lib/Hledger/Data/Posting.hs
```diff
@@ -404,8 +404,8 @@ accountNamePostingType a
 
 accountNameWithoutPostingType :: AccountName -> AccountName
 accountNameWithoutPostingType a = case accountNamePostingType a of
-                                    BalancedVirtualPosting -> T.init $ T.tail a
-                                    VirtualPosting -> T.init $ T.tail a
+                                    BalancedVirtualPosting -> maybe "" fst . T.unsnoc $ T.drop 1 a
+                                    VirtualPosting -> maybe "" fst . T.unsnoc $ T.drop 1 a
```
simonmichael, line 408: Can't this give a different result from before, with a one char account name ? Perhaps not observable in practice, but.. perhaps better to use textUnbracket here ?

simonmichael (changes requested):

hledger-lib/Hledger/Utils/Text.hs
Xitian9: No, not wrong. They are completely equivalent, but the old code was much harder for me to parse. Saying that the length is 2 or longer seems much clearer to me than saying that if you take the first two characters, you have exactly two characters.

hledger-lib/Hledger/Data/Posting.hs
Xitian9: Seems reasonable.

hledger-lib/Hledger/Utils/Text.hs
Xitian9: I changed it because I thought that may have been the source, and spent 30 seconds trying to parse it to see if it was. It wasn't, but I thought I should simplify it to make it easier to parse next time.

Xitian9:
I vote for the easy resolution. Account aliases are a high-power feature which most users will not use. The consequences of failure are also low (your report looks funny). Unless you're rewriting your journal based on the output of print, in which case you know you're playing with fire anyway.

Besides, maybe somebody really does want an empty account name at some point, or they use account aliases to get around our restrictions on account names. Weird perhaps, but I don't see the need to go to extra effort to stop them from doing it.

hledger-lib/Hledger/Utils/Text.hs
simonmichael: What was the cause of the bug, actually ?

Xitian9: `accountNameWithoutPostingType`

simonmichael:
https://hledger.org/dev/hledger.html#aliases-can-generate-bad-account-names added.

PR #1790, opened 2021-12-20
imp: smartdates: Simplify and generalise the SmartDate constructor and

Xitian9:
parsers to allow for arbitrary numbers of periods in relative dates.

We now accept smart dates like “in 5 days, in -6 months, 2 quarters ago”.

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -12,7 +12,8 @@ Date parsing and utilities for hledger.
 For date and time values, we use the standard Day and UTCTime types.
 
 A 'SmartDate' is a date which may be partially-specified or relative.
-Eg 2008\/12\/31, but also 2008\/12, 12\/31, tomorrow, last week, next year.
+Eg 2008\/12\/31, but also 2008\/12, 12\/31, tomorrow, last week, next year,
+in 5 days, in -3 quarters.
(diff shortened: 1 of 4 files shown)


simonmichael:
```
in n days/weeks/months/quarters/years         | n periods from the current period
n days/weeks/months/quarters/years ago        | -n periods from the current period
```
Nice.
Asymmetry - anything we can do to make it more regular and memorable ? Eg add a `... ahead` form ?
Plurals - should it support both singular and plural forms ?

Xitian9:
IIRC, as implemented it accepts both plural and singular forms, and does not require agreement with the numbers.

Regarding symmetry: I really like your proposed `3 days hence` format, but agree that it's probably too stuffy. It's actually currently somewhat symmetric, in that both forms accept negative numbers, so the following pairs are equivalent:
- `in -3 days` and `3 days ago`
- `in 3 days` and `-3 days ago`

simonmichael:
It's good that the negative numbers are handled, though maybe we don't advertise that as it reads quite unnaturally..

Re symmetry, I was kind of hoping for a postfix variant of "in", like "ago", so you can remember a single phrase with just a word changing. "in N Ms" and "N Ms ago" are just different enough to confuse with other forms and slip the memory I feel. 

But it's not a blocker. Have you thought of any limitations/conflicts we might run to when combining this with other hledger features ?

PR #1798, opened 2022-01-04
test: Run all tests with --fast compilation, not just functests.

Xitian9:
Previously doctests did not do fast compilation, so a common workflow of `make functest; make doctest` would recompile everything and take along time.


simonmichael:
I had some reason for leaving tests optimised..  something like: in my workflow I'm typically doing an optimised build/install right after, building unoptimised causes the next optimised build to be redone in full, and it didn't seem to be saving time.. but let's try this.

PR #1808, opened 2022-01-15
fix: cost: Add inferred equity postings during journal finalisation, so

Xitian9:
they can be matched by queries.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -24,6 +24,7 @@ module Hledger.Data.Journal (
   commodityStylesFromAmounts,
   journalCommodityStyles,
   journalToCost,
+  journalAddInferredEquityPostings,
   journalReverse,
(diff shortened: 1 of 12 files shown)


Xitian9:
The `ConversionOp` data type has been simplified, due to the decoupling of equity posting generation and costing. This means it is now the same as it was before the introduction of inferred equity postings. The question is, do we restore things to how they were, or opt for minimal change from the status quo?

- Master: `data ConversionOp = NoConversionOp | InferEquity | ToCost`, with `conversionop_` a record of `ReportOpts`
- Old: `data Costing = NoCost | Cost`, with `cost_` a record of `ReportOpts`.
- Minimal change: `data ConversionOp = NoConversionOp | ToCost`, with `conversionop_` a record of `ReportOpts`

simonmichael:
This looks great to me. ConversionOp does look a little odd now, but who knows perhaps it'll change again later. I'm going to merge this and maybe add a little note to that haddock. Thanks!

PR #1812, opened 2022-01-26
Manual: clarify reference from --strict to check

pkel:
Me on the mailing list, after skimming the manual section on `--strict`:

> `--strict` does too much for me, ...

Simon kindly informs me that the check command can run individual checks.

https://groups.google.com/g/hledger/c/l3wPnljbLco/m/BvGYgyBTAwAJ

I went back to the manual and found that `--strict` refers to `check`. I'm not sure how I could miss it. I hope that my small change improves the situation a bit.


simonmichael:
Thanks!

PR #1814, opened 2022-01-27
imp: csv: Give an error if unable to substitute csv templates. (#1803)

### LINKED ISSUE #1803: csv import: 4 rule file accepts a record of only 3 fields
_If you're not sure this is a bug, or if some discussion would help_,
**
**Brief chat on [hledger.irc](https://www.irccloud.com/irc/libera.chat/channel/hledger) Jan 7th 16:08-16:56**
**

_Have you looked at (the right version of) the user manual?_
**Yes**

_A small reproducible example  (--anon can help anonymise data)_

**I created my CSV data file and inadvertently missed a ',' after the payee name in record 2.**
**CSV data:**

```
2021-12-22,garo_bike ;  ebici,expenses:trasporto:ebici,-50.0
2021-12-23,caffe_siciliaexpenses:cibo:dolce,-10.5
2021-12-27,simone ; pane:simone,expenses:cibo:pane,-4.8
```

**RULES file:**
```
account1 assets:bank:checking
fields date, description, account2, amount
```

**CMD and result:**

**I then did a `--dry-run`:**
**An noticed the strange output in the 2nd posting**

```
$ hledger import --dry-run --rules-file=./inc_exp2018_ods.csv.rules csv_missing_delim.csv
; would import 3 new transactions from csv_missing_delim.csv:

2021-12-22 garo_bike ;  ebici
    assets:bank:checking               -50.0
    expenses:trasporto:ebici            50.0

2021-12-23 caffe_siciliaexpenses:cibo:dolce
    assets:bank:checking              %4
    -10.5                            %-4

2021-12-27 simone ; pane:simone
    assets:bank:checking            -4.8
    expenses:cibo:pane               4.8
```

_The output of hledger --version_
**1.24, linux-x86_64**

_What platform you're on_
**Linux**

Xitian9:
Fixes #1803.

### DIFF
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -1249,10 +1249,10 @@ renderTemplate rules record t = maybe t mconcat $ parseMaybe
 
 -- | Replace something that looks like a reference to a csv field ("%date" or "%1)
 -- with that field's value. If it doesn't look like a field reference, or if we
--- can't find such a field, leave it unchanged.
+-- can't find such a field, replace it with the empty string.
 replaceCsvFieldReference :: CsvRules -> CsvRecord -> CsvFieldReference -> Text
 replaceCsvFieldReference rules record s = case T.uncons s of
-    Just ('%', fieldname) -> fromMaybe s $ csvFieldValue rules record fieldname
+    Just ('%', fieldname) -> fromMaybe "" $ csvFieldValue rules record fieldname
     _                     -> s
 
 -- | Get the (whitespace-stripped) value of a CSV field, identified by its name or
--- a/hledger/test/csv.test
+++ b/hledger/test/csv.test
@@ -1032,6 +1032,19 @@ $  ./csvtest.sh
 
 >=0
 
+# 51. Throw an error when unable to substitute csv templates
+<
+"2021-12-23","caffe_siciliaexpenses:cibo:dolce","-10.5"
+
+RULES file
+
+account1 assets:bank:checking
+fields date, description, account2, amount
+
+$  ./csvtest.sh
+>2 /could not balance this transaction/
+>=1
+
 ## . 
 #<
 #$  ./csvtest.sh


Xitian9:
This is a big change to how csv templates are handled. Previously, if the field name could not be substituted, it would be left as is (For example %4 would just be left as %4). Now, it will throw an error if it can't be substituted.

Maybe some people were relying on the previous behaviour, for example they wanted a literal %99 in their description, and new there would not be a 99th entry in any of the csv records. This would break things for them. Maybe we need to warn of this.

simonmichael:
Normally we do like to throw errors rather than handle them silently. CSV data is inherently messy though, so I wonder if it's the best call here. I was imagining a `%nosuchfield` reference just interpolating nothing.

I don't quite see how the provided test tests that. Maybe we could have a couple more: one testing %4 (with < 4 fields) and another testing %nosuchfieldname ?

Xitian9:
> Normally we do like to throw errors rather than handle them silently. CSV data is inherently messy though, so I wonder if it's the best call here. I was imagining a `%nosuchfield` reference just interpolating nothing.

Yes, it's a bit unclear. The main issue this was addressing is that when something like for an amount %4 is left uninterpolated, it was interpreted as the weird commodity %4, which was then negated as %-4. I worry that leaving it blank could result in situations where that gets interpreted as a missing amount, which might then actually balance in certain circumstances.

Leaving a description blank might be fine, but whatever our solution it would need to throw a sufficiently stern warning whenever it can't read (1) an amount (2) an account name (3) a date. Could we substitute something which could not possibly be interpreted as any of those?

> I don't quite see how the provided test tests that. Maybe we could have a couple more: one testing %4 (with < 4 fields) and another testing %nosuchfieldname ?

The existing test does test your case 1. I'm always happy to add more tests.

Xitian9:
I've changed it to substitute `%nosuchfield` if a field is missing. This still makes it throw an error in the relevant test, because it can't intrepret `%nosuchfield` as an amount, but the error message is actually much nicer.

simonmichael:
Here are some options:

1. insert "%fieldnamewithtypo" or "%11" (current hledger)
2. insert a standard symbol: "%nosuchfield" (current 1803 PR)
3. insert nothing ("")
4. raise an error
5. print a warning (the first of its kind)

2 is probably quote practical, but I'm not too fond of cluttering data with an anglocentric magic symbol. If someone mentions `nosuchfield` in the `fields` list, will they expect that field value to be interpolated ? Currently it's not.

CSVs with fewer values in certain records are fairly common. IIRC we used to enforce that all records had the same number of values, but dropped that to be more forgiving. I believe those cases usually mean "this field has no value", so interpolating "" would be correct. This would be my preference, if we could convince ourselves that quietly inserting "", eg for account name, commodity symbol, amount, or balance, does not cause major problems.

simonmichael:
I'm still inclined to leave it blank, until we find a good example of bad behaviour. 

If 3 is reasonable most of the time, but not always, we could switch to 4 in -s/--strict mode ?

Xitian9:
Oops, forgot about this. Updated.

Xitian9:
Anything more needed on this?

Issue #1817, opened 2022-01-28
query by account tags


simonmichael:
The next step in #217, #1073 etc.: the `account` directive supports [tags](https://hledger.org/hledger.html#tags-1), but you can't match on them in queries. This would allow external tools, like charting scripts, to reliably select accounts by their [type](https://hledger.org/hledger.html#account-types). 

Eg to select asset accounts, something like this:

    hledger bal tag=type:A

Possible problems ?

- Standard tags behaviour is that posting-matching reports will also see tags from the parent transaction, and transaction-matching commands will also see tags from all postings. I'm not sure if this will be a problem.
- `tag=type:VAL` is a little clunky and VAL is an infix match so you'd need `tag=type:'^A'` to be (somewhat) robust
- you might want to also allow writing the long form `Asset` for readability, as the account type tag does.

If necessary, a specialised `atype:` query might be one way to solve these.

simonmichael:
I have a prototype working. More issues to consider:
- When matching account tags, should subaccounts inherit tags from their parents (always, never, sometimes) ? The `type:` tag is inherited.
- When matching account types, should it be possible to match inferred account types, not just declared ones ? I think this would be out of scope for `tag:` and would require a specialised query type (`atype:`).

simonmichael:
Merged; see also #1820.

PR #1819, opened 2022-01-29
feat: queries can now match account tags (#1817)

simonmichael:
For #1817:

Accounts, postings, and transactions can now all be filtered by the
tags in an account's declaration. In particular it's now possible to
more reliably select accounts by type, using their type: tag rather
than their name:

```
account myasset       ; type:Asset
account myliability   ; type:Liability

$ hledger accounts tag:type=^a
myasset
```

API changes:
A finalised Journal has a new jdeclaredaccounttags field
for easy lookup of account tags.
Query.matchesTaggedAccount is a tag-aware version of matchesAccount.

The current tests, to show how it behaves:
```shell
# Querying accounts by tag.
<
account a   ; type:A
account l   ; type:Liability
account r   ; type:R
account o   ; othertag:
account u

2022-01-01
  (r)        1

2022-01-02
  (a)        1
  (l)        1

# 7. We can match declared accounts by having a tag,
$ hledger -f- accounts --declared tag:.
a
l
r
o

# 8. not having a tag,
$ hledger -f- accounts --declared not:tag:.
u

# 9. or a tag and it's value. Tag values are matched infix.
$ hledger -f- accounts --declared tag:type=a
a
l

# 10. So we must anchor the regex to match single-letter account types.
$ hledger -f- accounts --declared tag:type=^a$
a

# 11. But if account type was declared in the long form, matching just one letter fails
$ hledger -f- accounts --declared tag:type=^l$

# 12. so we need to match more loosely
$ hledger -f- accounts --declared tag:type=^l
l

# 13. In the same way, we can match used accounts by tag.
$ hledger -f- accounts --used tag:type=r
r

# 14. We can match postings by their account's tags.
$ hledger -f- register -w80 tag:type=^a
2022-01-02                      (a)                              1             1

# 15. We can match transactions by their accounts' tags.
$ hledger -f- print tag:type=^a
2022-01-02
    (a)               1
    (l)               1

>=

# 16. And negatively match them by tag.
$ hledger -f- print tag:type=^a not:tag:type=^l

# 17. We can filter balance reports by account tags.
$ hledger -f- bal tag:type=^a
                   1  a
--------------------
                   1  

# 18. Postingless declared accounts in balance reports are also filtered.
$ hledger -f- bal -N --declared -E o u tag:othertag
                   0  o

```

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -48,6 +48,8 @@ module Hledger.Data.Journal (
   journalAccountNamesDeclaredOrUsed,
   journalAccountNamesDeclaredOrImplied,
   journalAccountNames,
+  journalAccountTags,
+  journalInheritedAccountTags,
   -- journalAmountAndPriceCommodities,
   -- journalAmountStyles,
   -- overJournalAmounts,
@@ -103,7 +105,7 @@ import Control.Monad.State.Strict (StateT)
 import Data.Char (toUpper, isDigit)
 import Data.Default (Default(..))
 import Data.Foldable (toList)
-import Data.List ((\\), find, foldl', sortBy)
+import Data.List ((\\), find, foldl', sortBy, union)
 import Data.List.Extra (nubSort)
 import qualified Data.Map.Strict as M
 import Data.Maybe (catMaybes, fromMaybe, mapMaybe, maybeToList)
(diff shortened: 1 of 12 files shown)


simonmichael:
This PR page is strangely not updating right now. In case it doesn't fix itself, the latest commits can be seen at showing the latest https://github.com/simonmichael/hledger/commits/1817-query-acct-tags commits . Accounts now inherit tags from their parents, and this is ready for review/merge.

PR #1820, opened 2022-01-30
new type: query for easy matching by account type

simonmichael:
Follow-on from #1817/#1819: because matching account types with `tag:` is inconvenient, this adds the `type:TYPES` query, where `TYPES` is any of the (case insensitive) letters `ALERXCV` (see https://hledger.org/hledger.html#account-types.) This is expected to work appropriately with all commands. It handles types inherited from parents, overridden by children, and inferred from account names, so should be a pretty reliable way to select accounts by type if 1. you have declared account types or 2. you use common english top-level account names. Eg:

    $ hledger -f examples/sample.journal bal type:al
                      $1  assets:bank:saving
                     $-2  assets:cash
                      $1  liabilities:debts
    --------------------
                       0

API changes:

Journal has a new jaccounttypes map.
The journalAccountType lookup function makes it easy to check an account's type.
The journalTags and journalInheritedTags functions look up an account's tags.
Functions like journalFilterPostings and journalFilterTransactions,
and new matching functions matchesAccountExtra, matchesPostingExtra
and matchesTransactionExtra, use these to allow more powerful matching
that is aware of account types and tags.

### DIFF
--- a/hledger-lib/Hledger/Data/AccountName.hs
+++ b/hledger-lib/Hledger/Data/AccountName.hs
@@ -20,6 +20,13 @@ module Hledger.Data.AccountName (
   ,accountNameToAccountRegexCI
   ,accountNameTreeFrom
   ,accountSummarisedName
+  ,accountNameInferType
+  ,assetAccountRegex
+  ,cashAccountRegex
+  ,liabilityAccountRegex
+  ,equityAccountRegex
+  ,revenueAccountRegex
+  ,expenseAccountRegex
   ,acctsep
   ,acctsepchar
   ,clipAccountName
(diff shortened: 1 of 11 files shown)


simonmichael:
I asked #haskell about changing from fail to error' (for the better output).
> error is a bottom, but fail is different.  That can affect the way GHC treats the error condition.
> If you wanted control over the error type as well as its timing, you should use throwIO along with creating an exception of your desired type
> throwIO creates an IO action that throws the exception when it's executed. error creates an expression that throws the exception when it's evaluated. The former is less up to the whims of the optimizer, for instance. IO actions are executed serially

I think for us they are practically equivalent, and we already use error' extensively, but just noting it here.

simonmichael:
Assumed perfect. Account types for all! \o/

PR #1826, opened 2022-02-04
imp: cost: Allow inferring cost from conversion equity postings.

Xitian9:
When given --infer-equity, hledger will change transaction prices to
balancing equity postings. This introduces the inverse operation,
--match-equity, which will match balancing equity postings and transferm
them into a transaction price, allowing --cost to work properly with
them. This is only a partial inverse as it needs to use some heuristics
to match the postings which will not work in complicated cases.

Specifically, when hledger finds exactly two conversion postings in a
transaction (by default, subaccounts of equity:conversion or
equity:trad(e|ing)), it will find the first posting in the transaction
whose amount is the inverse of one of the conversion postings, and
insert the corresponding transaction price.


Xitian9:
Some open questions that need to be answered before this is ready for merge.

- What about the naming of the `matched:` tag. Would you prefer something else?
- With --infer-costs, when should the various metadata be displayed in the `print` command?
  - The inferred transaction price should probably always be displayed: that's why you called it with --infer-costs, right?
  - What about the auto-generated `matched:` tags? Should those always be displayed, or only when called with `--show-costs`?
- Currently, any posting which has a `matched:` tag will be ignored for balancing transactions. **This is definitely too lax**, but I can't think of a way to get around this without doing all the conversion posting matching in all cases, even when not called with `--infer-costs`. Feedback on the desired level of strictness here would be good.
- Currently it is not possible to _both_ provide a transaction price and manual conversion postings __unless__ you call it with `--infer-costs`. It will not balance the transaction.

simonmichael:
Some thoughts after playing with this, continuing from chat:

- With the latest changes, @ is allowed to coexist with equity postings without unbalancing the transaction, which opens up new possibilities.
- @ notation carries more information than equity postings: it indicates 1. which postings are/are not involved in a conversion, and 2. which is the priced commodity. So it's a useful notation and we needn't try to avoid it.
- Every use of @ notation implies a matching pair of equity:conversion postings. If these are not recorded explicitly, we can always mechanically infer them.
- Conversely, pairs of equity:conversion postings imply a conversion price. If that is not recorded explicitly with @ notation, we can't always infer it. (Only for transactions with a constrained, regular structure.) 
- So perhaps things could work like this ?
  - When we see @, we look for the corresponding pair of equity postings, and if not present we always infer them.
  - When we see a pair of equity:conversion postings and can't find a matching @ notation, we can do any of these, with increasing amounts of work: 
    1. Give an error; explicit @ notation is always required.
    2. Give an error but try to suggest some @ notation to add.
    3. Try to infer it. If it's too hard, we give an error advising to either simplify the transaction or to add explicit @ notation. 
  - No need for the --infer-equity/--infer-costs flags.

Xitian9:
Just to add to the above, the use of @ notation along with equity:conversion postings introduces a compatibility issue. Namely, such a transaction will not balance with `ledger`, or a version of `hledger` older than the change. This is because existing behaviour is that transactions balance with their @ price, but conversion postings would disrupt that, making them balance with their primary amount.

Sorting this issue would require detecting whether a posting with an @ price has corresponding equity:conversion postings, and only using the @ price if it does not. Ways in which this could be done include a manual tag (as in the original version of this PR), or by automatically matching up conversion postings (as in you proposal above). But whichever method we use, a program would need to know to look for them.

Xitian9:
I've significantly rewritten the logic, so it's worth a summary of where we are now. When called with `--infer-costs`, hledger will now do the following things.

At journal finalisation time, before inferring missing amounts:
1. It will make sure that any transaction which has conversion equity postings (determined by their account names and any account declarations) has them in adjacent pairs of two. It also must be the case that there is a bijection between conversion postings and non-conversion postings, where each conversion posting have exactly (negative) the amount of a non-conversion posting. If any of these don't hold, it will fail finalisation with an error message, similarly to how it will if postings are unbalanced.
2. For each pair of conversion equity postings, it will search through the postings in that transaction which have transaction prices. If exactly one such posting matches the conversion equity postings (i.e. one of the conversion amounts matches negative the primary amount, and the other matches the cost amount), then it will internally mark the conversion postings and the matched posting as paired, meaning that the primary amount of the priced posting will be used for balancing, and the conversion postings will be deleted when called with --cost. The matched posting will then not be available for matching future conversion posting pairs, nor will the corresponding unpriced posting.
3. If more than one priced posting matches the pair of conversion postings, it will fail finalisation with an error.
4. If no priced postings match, then it will search the unpriced postings. If exactly one matches (the negative of) each conversion posting, it will add an appropriate transaction price to the *first* of the matched postings, and mark it as matched, as above. Neither matched posting will be available for matching future conversion posting pairs.

Xitian9:
Here's a walkthrough of a somewhat complicated example:
```
2009-01-01
    assets:euros               €100 @ $1.35  ; 100 euros bought
    assets:dollars                    $-135  ; for $135
    equity:conversion                 €-100
    equity:conversion                  $135
    assets:pounds                       £80  ; 80 pounds bought
    assets:dollars                    $-100  ; for $100
    equity:conversion                  £-80
    equity:conversion                  $100
```

First step: the equity conversion postings definitely come in pairs of two. They could even be adjacent to each other, but here they're separated so it's clear.

```
2009-01-01
    assets:euros               €100 @ $1.35  ; 100 euros bought
    assets:dollars                    $-135  ; for $135
    equity:conversion                 €-100  <--
    equity:conversion                  $135  <--
    assets:pounds                       £80  ; 80 pounds bought
    assets:dollars                    $-100  ; for $100
    equity:conversion                  £-80  <--
    equity:conversion                  $100  <--
```

Each conversion posting matches exactly one non-conversion posting. Here they are easy to see from the order, but this need not be the case.
```
2009-01-01
    assets:euros               €100 @ $1.35  ; 100 euros bought  <-1-
    assets:dollars                    $-135  ; for $135          <-2-
    equity:conversion                 €-100  <-1-
    equity:conversion                  $135  <-2-
    assets:pounds                       £80  ; 80 pounds bought  <-3-
    assets:dollars                    $-100  ; for $100          <-4-
    equity:conversion                  £-80  <-3-
    equity:conversion                  $100  <-4-
```

We now match the first pair, and search through the transaction price postings for a match. Fortunately we find one.
```
2009-01-01
    assets:euros               €100 @ $1.35  ; 100 euros bought  <--
    assets:dollars                    $-135  ; for $135
    equity:conversion                 €-100  <--
    equity:conversion                  $135  <--
    assets:pounds                       £80  ; 80 pounds bought
    assets:dollars                    $-100  ; for $100
    equity:conversion                  £-80
    equity:conversion                  $100
```

These postings are internally marked and are no longer available for matching further conversion postings. We also find the first matching non-priced posting and make it unavailable.

```
2009-01-01
    assets:euros               €100 @ $1.35  ; 100 euros bought  <--
    assets:dollars                    $-135  ; for $135  <--
    equity:conversion                 €-100  <--
    equity:conversion                  $135  <--
    assets:pounds                       £80  ; 80 pounds bought
    assets:dollars                    $-100  ; for $100
    equity:conversion                  £-80
    equity:conversion                  $100
```

Next, we move to the second pair of conversion postings. Since there are no more priced transaction postings that they can match, we move on to non-priced postings, and find two which match. Note that the previously matched postings are unavailable, so even if their amounts matched (which they don't in this case), they would not be selected.

```
2009-01-01
    assets:euros               €100 @ $1.35  ; 100 euros bought
    assets:dollars                    $-135  ; for $135
    equity:conversion                 €-100
    equity:conversion                  $135
    assets:pounds                       £80  ; 80 pounds bought  <--
    assets:dollars                    $-100  ; for $100      <--
    equity:conversion                  £-80        <--
    equity:conversion                  $100      <--
```
We assign the transaction price to the *first* of these, and eliminate it and its other match from future consideration.

```
2009-01-01
    assets:euros               €100 @ $1.35  ; 100 euros bought
    assets:dollars                    $-135  ; for $135
    equity:conversion                 €-100
    equity:conversion                  $135
    assets:pounds                £80 @@ $100  ; 80 pounds bought (internally)
    assets:dollars                    $-100  ; for $100
    equity:conversion                  £-80
    equity:conversion                  $100
```

the-solipsist:
One request, please make `@@ $20.00` and `  ; cost: @@ $20.00` equivalent.  (`--infer-equity` already auto-generates `  ; cost: @@ $20.00` as a comment.)

Such that these two would be treated as equivalent:
```
2009-01-01
    assets:euros               €100 @ $1.35
    equity:conversion                 €-100
    equity:conversion                  $135
    assets:dollars                    $-135
```
and
```
2009-01-01
    assets:euros                       €100  ; cost: @@ $135
    equity:conversion                 €-100
    equity:conversion                  $135
    assets:dollars                    $-135
```

Xitian9:
As implemented this removes the `cost: @@ $135` notation. This was needed as a way to maintain the cost information without generating invalid journals which have both `@` prices and equity, but that is now allowed. `--generate-equity` would now produce the following:
```
2009-01-01
    assets:euros               €100 @@ $135
    equity:conversion                 €-100
    equity:conversion                  $135
    assets:dollars                    $-135
```

simonmichael:
After a little recent testing and discussion - I'm happy to merge this, and watch out for more UX/doc improvements we can make later. Thanks for the work!

simonmichael:
@Xitian9 just on the off chance you remember - in this PR adding --infer-cost, any idea why you also made --infer-cost affect `close` the same way `--explicit` does ? Ie --infer-cost makes the amount(s) in `close`'s equity posting(s) explicit.

simonmichael:
```haskell
    -- should we show the amount(s) on the equity posting(s) ?
    explicit = boolopt "explicit" rawopts || copts ^. infer_costs
```

Xitian9:
I can't remember exactly. Probably the reasoning was that if you called
close with --infer-costs you would obviously want it displayed, but that
may have been misguided. I don't see harm in removing that line, as you can
always just add --explicit.

simonmichael:
Thanks! Will do. It puzzled me when doing some end-of-year journal rewrites. Merry Christmas / Happy Holidays.

Issue #1033, opened 2019-05-31
glibc-using-nix-installed or guix-installed hledger won't handle non-ascii data, even with LANG set


simonmichael:
Reported by @Vifon: hledger installed via the nix command on the download page fails if the journal contains non-ascii characters. Reproduced on Gentoo and Ubuntu GNU/Linux; does not happen on mac. 

Eg:
```
$ nix-env -i -f https://github.com/NixOS/nixpkgs/archive/fe41fd.tar.gz -A hledger; curl -s https://raw.githubusercontent.com/simonmichael/hledger/master/examples/unicode.journal | ~/.nix-profile/bin/hledger -f- print
...
hledger: <stdin>: hGetContents: invalid argument (invalid byte sequence)
$ echo $LANG
en_US.utf8
$ locale -a | grep US
en_US.iso885915
en_US.utf8
$ curl -s https://raw.githubusercontent.com/simonmichael/hledger/master/examples/unicode.journal | ~/.local/bin/hledger -f- print
(works)
```

simonmichael:
@peti, any idea what might be the cause ?

peti:
The Nix `glibc` does not know any locales (other than `C`). Locale data resides in a separate package, `glibcLocales`, and glibc can only find that package through an environment variable

    LOCALE_ARCHIVE="$(nix-build --no-out-link "<nixpkgs>" -A glibcLocales)/lib/locale/locale-archive"

that needs to be set at run-time. Alternatively, that variable can also point to the locale archive of the native host OS:

    LOCALE_ARCHIVE="/usr/lib/locale/locale-archive"

This separation exists so that Nix users can choose a custom set of locales they'd like to install without affecting `glibc`. If `glibc` would have a reference to the locale data, then any change to the locate data would trigger a complete re-build of the entire system.

simonmichael:
Wow, good info, thanks. What do you think is the best approach for reliable install commands that take this into account ? This worked on Linux:
```
$ export LOCALE_ARCHIVE="$(nix-build --no-out-link '<nixpkgs>' -A glibcLocales)/lib/locale/locale-archive"
$ nix-env -i -f https://github.com/NixOS/nixpkgs/archive/fe41fd.tar.gz -A hledger hledger-web hledger-ui
```
but not on Mac:
```
$ nix-build --no-out-link '<nixpkgs>' -A glibcLocales
error: expression does not evaluate to a derivation (or a set or list of those)
```

simonmichael:
[...]

Issue #708, opened 2018-02-12
hledger-web on Windows can show non-ascii characters as mojibake (garbled)


hymerman:
Anywhere the pounds sterling symbol is meant to be shown in the web interface, "┬ú" is shown instead. It's this way in the page source. Hledger on the console is fine.

This has been happening for a long time (years), on all machines I've tried it on (Windows 7 and Windows 10). I'm a bad person for not reporting it until now!

E.g.
![image](https://user-images.githubusercontent.com/154886/36105040-8a1fb424-100b-11e8-882d-4eb4dec58827.png)
----

Likely workaround [below](https://github.com/simonmichael/hledger/issues/708#issuecomment-1773836682)

simonmichael:
Thanks for the report. I don't see this on mac or linux. Can any other windows users confirm ? How are you starting hledger-web ? How is your journal file encoded ? (Not sure how to check this on windows, on unix it's `file FILE`).

hymerman:
No problem - thanks for the quick response!

I'm starting like this (using mingw64 provided with Git for Windows, but the same bug happens starting through Windows standard command prompt too):

    hledger-web.exe --file=/path/to/my.ledger

The journal file is encoded in UTF-8, with Windows (CRLF) line endings.

I'm using the binary from AppVeyor artifacts.

simonmichael:
Is it the same if you double click hledger-web.exe ? (You might need to create a journal file in default location eg with hledger add).  

Does the data entry mechanism matter ? Eg compare entry through hledger-web, hledger add, and a text editor.

hymerman:
I just tried creating a new journal with hledger add, and double clicking hledger-web. Same result - shows "┬ú". I can't figure out how to add transactions in the web interface, it doesn't do anything or give any error after clicking "Add". But now I've tried through hledger add, and just by reading a text file.

Entries in the text file look like this:

1999/06/30 Gross interest
    assets:savings:Nationwide  £1.17
    income

simonmichael:
Thanks for testing. Hmm, odd that you have no add form in hledger-web. I would expect at least a js error visible in web inspector. I suspect the appveyor binaries are not up to date.

hymerman:
[...]

Issue #536, opened 2017-04-08
`hledger bal Расходы` doesn't work under Cygwin


artkpv:
Hi, 
Thank you for the great tool!

How can I fix this to make hledger filter bal report using cyrillic charaters? 

1. Given Cygwin, mintty 2.6.2 (x86_64), hledger 1.1
2. 
```
$ locale
LANG=ru_RU.UTF-8
LC_CTYPE="ru_RU.UTF-8"
LC_NUMERIC="ru_RU.UTF-8"
LC_TIME="ru_RU.UTF-8"
LC_COLLATE="ru_RU.UTF-8"
LC_MONETARY="ru_RU.UTF-8"
LC_MESSAGES="ru_RU.UTF-8"
LC_ALL=
```
3. This: `$ hledger.exe -f main.ledger bal Расходы`
Outputs:
```
 --------------------
                    0
```

simonmichael:
You're welcome!

I don't have a windows machine accessible at the moment, but what is the output with --debug=9 added ?

w1ld:
@simonmichael 
With `--debug=9` it is has a lot of output. It reads the chars as '\1056\1072\1089\1093\1086\1076\1099'

Here is for `--debug=2`:

```
$ hledger.exe -f main.ledger bal Расходы --debug=2
running:    "hledger 1.1"
raw args:
 [ "-f"
 , "main.ledger"
 , "bal"
 , "\1056\1072\1089\1093\1086\1076\1099"
 , "--debug=2"
 ]
raw args rearranged for cmdargs:
 [ "bal"
 , "\1056\1072\1089\1093\1086\1076\1099"
 , "--debug=2"
 , "-f"
 , "main.ledger"
 ]
raw command is probably: "bal"
raw args before command: [ "-f" , "main.ledger" ]
raw args after command: [ "\1056\1072\1089\1093\1086\1076\1099" , "--debug=2" ]
processed opts:
 CliOpts
   { rawopts_ =
       [ ( "command" , "balance" )
       , ( "args" , "\1056\1072\1089\1093\1086\1076\1099" )
       , ( "debug" , "2" )
       , ( "file" , "main.ledger" )
       ]
   , command_ = "balance"
   , file_ = [ "main.ledger" ]
   , rules_file_ = Nothing
   , output_file_ = Nothing
   , output_format_ = Nothing
   , alias_ = []
   , ignore_assertions_ = False
   , debug_ = 2
   , no_new_accounts_ = False
[...]

Issue #73, opened 2013-04-08
handle misconfigured locale more gracefully (invalid argument / invalid character / illegal byte sequence)


simonmichael:
_Original author: mecca...@gmail.com (November 02, 2011 22:48:32)_

on hledger-0.16.1 I get &quot;mkTextEncoding: invalid argument (Invalid argument)&quot;
when invoking hledger

(archlinux 64bit, ghc 7.0.3, en_US.utf8 locale)

_Original issue: http://code.google.com/p/hledger/issues/detail?id=73_

### Related:

Error messages:
- `mkTextEncoding: invalid argument`
- `commitAndReleaseBuffer: invalid argument (invalid character)`
- `Invalid or incomplete multibyte or wide character`
- `Illegal byte sequence`

Docs:
- https://hledger.org/download.html#utf-8-locale
- https://hledger.org/hledger.html#unicode-characters
- https://hledger.org/hledger.html#troubleshooting

Issues: 
- https://gitlab.haskell.org/ghc/ghc/-/issues/17755 
- https://github.com/serokell/haskell-with-utf8/issues/10
- #420
- #1619

Other:
- https://hackage.haskell.org/package/encoding
- https://hackage.haskell.org/package/with-utf8
- https://serokell.io/blog/haskell-with-utf8
- https://en.wikipedia.org/wiki/Byte_order_mark

simonmichael:
We have https://hledger.org/download.html#utf-8-locale discussing this; I've added the error message reported above. 

Re catching those errors: https://github.com/serokell/haskell-with-utf8/issues/10

simonmichael:
Slight caveats for with-utf8: it seems not too actively maintained, and I hear on IRC that it is too forceful:

> The problem with "always UTF-8" is that it intentionally breaks for everyone who has a properly configured system with encoding different than utf-8
> So if someone in Japan is using a UTF-16 (or UTF-32) configuration system wide or something, then you are now unable to open their files, because they're not UTF-8, but you're overriding the environment

simonmichael:
https://gitlab.haskell.org/ghc/ghc/-/issues/17755 is the relevant GHC issue.

simonmichael:
I sure would love for us to handle this in a user friendly way, since it is a very frequent puzzle for new users of hledger. If we do, perhaps it might encourage other haskell apps to do similar.

simonmichael:
I'll offer a bounty of $50 for decent progress on this, or $100 for a complete satisfying solution.

[...]

PR #1834, opened 2022-03-10
Use with-utf8 and don't use Data.Text.IO.

### LINKED ISSUE #1154: Windows-built hledger executable fails in WSL
Hi!

I've encountered the following problem with character encoding.

My setup: 
- Win 10 with an Ubuntu18 in WSL
- hledger is built from source on the windows side:
```bash
bash $ hledger.exe --version
hledger 1.16.1
```
- locale in Ubuntu:
```bash
bash $ locale
LANG=en_US.UTF-8
LANGUAGE=
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL
```
- UTF-8 encoded journal

- run `hledger.exe -f journal register` inside WSL (launched via bash/conemu/ubuntu, no matter) to get
```bash
bash $ hledger.exe -f journal register
hledger.exe: E:\nosave\bookkeeping\journal: hGetContents: invalid argument (invalid byte sequence)
```
Yet on the windows side in a unicode-aware (via chcp 65001) powershell I get the correct output.

As suggested in the "troubleshooting" section in the manual, I tried running `LANG=something hledger -f journal register`; it does not help either for reasonable choice of "something".

What else can I try in this situation? Try to use an ubuntu-in-wsl-built version of hledger, maybe?

### LINKED ISSUE #1619: Data.Text.IO should not be used
[It has been documented](https://www.snoyman.com/blog/2020/10/haskell-bad-parts-1/) that `Data.Text.IO` does not play well with locales, and can cause problems. We should reduce/eliminate its use in our code, explicitly encoding to utf8 before writing the resulting ByteString.

[More context:
<https://www.snoyman.com/blog/2016/12/beware-of-readfile>
<https://hledger.org/install.html#check-your-locale>
<https://hledger.org/1.25/hledger.html#troubleshooting>
Basically by addressing this with [#1834](https://github.com/simonmichael/hledger/pull/1834) we aim to once and for all solve or at least reduce "hledger crashed because $LANG was unset / wrong for this file / the locale was not installed / [Nix](https://github.com/simonmichael/hledger/issues/1033)"-type reports (see also issues linked on #1834). --SM]

Xitian9:
fix: utf-8: Use with-utf8 to ensure all files are read and written with utf8 encoding. (Fixes #1619)

May also fix #1154, #1033, #708, #536, and #73. Testing is needed.

### DIFF
--- a/hledger-lib/Hledger/Utils.hs
+++ b/hledger-lib/Hledger/Utils.hs
@@ -36,7 +36,7 @@ import Data.FileEmbed (makeRelativeToProject, embedStringFile)
 import Data.List.Extra (foldl', foldl1', uncons, unsnoc)
 import qualified Data.Set as Set
 import Data.Text (Text)
-import qualified Data.Text.IO as T
+import qualified Data.Text.IO as TIO (hGetContents)  -- Only putStr and friends are safe
 import qualified Data.Text.Lazy.Builder as TB
 import Data.Time.Clock (getCurrentTime)
 import Data.Time.LocalTime (LocalTime, ZonedTime, getCurrentTimeZone,
@@ -48,10 +48,10 @@ import Lens.Micro ((&), (.~))
 import Lens.Micro.TH (DefName(TopName), lensClass, lensField, makeLensesWith, classyRules)
 import System.Console.ANSI (Color,ColorIntensity,ConsoleLayer(..), SGR(..), setSGRCode)
 import System.Directory (getHomeDirectory)
-import System.FilePath (isRelative, (</>))
-import System.IO
-  (Handle, IOMode (..), hGetEncoding, hSetEncoding, hSetNewlineMode,
-   openFile, stdin, universalNewlineMode, utf8_bom)
+import System.FilePath ((</>), isRelative)
+import System.IO (Handle, IOMode (..), hGetEncoding, hSetEncoding,
+                  hSetNewlineMode, stdin, universalNewlineMode, utf8_bom)
+import qualified System.IO.Utf8 as Utf8
 
 import Hledger.Utils.Debug
 import Hledger.Utils.Parse
@@ -175,7 +175,7 @@ expandHomePath = \case
 -- using the system locale's text encoding,
 -- ignoring any utf8 BOM prefix (as seen in paypal's 2018 CSV, eg) if that encoding is utf8.
 readFilePortably :: FilePath -> IO Text
-readFilePortably f =  openFile f ReadMode >>= readHandlePortably
+readFilePortably f =  Utf8.openFile f ReadMode >>= readHandlePortably
 
 -- | Like readFilePortably, but read from standard input if the path is "-".
 readFileOrStdinPortably :: String -> IO Text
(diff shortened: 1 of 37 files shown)


hledger-lib/Hledger/Utils.hs
```diff
@@ -186,12 +184,7 @@ readFileOrStdinPortably f = openFileOrStdin f ReadMode >>= readHandlePortably
     openFileOrStdin f m   = openFile f m
 
 readHandlePortably :: Handle -> IO Text
-readHandlePortably h = do
-  hSetNewlineMode h universalNewlineMode
-  menc <- hGetEncoding h
-  when (fmap show menc == Just "UTF-8") $  -- XXX no Eq instance, rely on Show
-    hSetEncoding h utf8_bom
-  T.hGetContents h
+readHandlePortably h = hSetNewlineMode h universalNewlineMode *> TIO.hGetContents h
```
simonmichael, line 187: IIRC, setting utf8_bom encoding made reading utf8 files more robust, by also accepting the utf8 bom prefix that some of them have. Why isn't this needed any more ?

simonmichael, line 187: I think >> might be a little more idiomatic and obvious in this codebase

simonmichael, line 187: Doesn't this read with the system locale's encoding as before ? Shouldn't we be enforcing utf8 now ?

simonmichael:
PS, I'm a little confused, github says this has no conflicts with master but merging locally fails with either a bunch of conflicts or "refusing to merge unrelated branch". Is it PEBKAC or is there something odd about this branch ?

simonmichael:
I remember we discussed this but I forgot where. This seems to have potential to be unexpectedly breaking and in that case quite annoying for users, so we should give it a detailed commit message (=> changelog/migration note), and ideally a preview release and a loud call for testing. I have missed apr 1 and may 1 but might still rush out a preview release when this lands.

hledger-lib/Hledger/Utils.hs
Xitian9: The reasoning was that I read something which recommended against it, which I thought was in the `with-utf8` documentation, but I cannot find it anymore. I can leave it in if you'd prefer.

Xitian9: Sure thing.

Xitian9: I believe it does. `openFile` comes from `with-utf8` (I'll qualify the import so that's more obvious), and `stdin` should be taken care of by the `withUtf8` wrapper around `main`.

Xitian9: I think the issue is that `with-utf8` already sets the encoding, and so if we do it ourselves it will just be overridden. If setting the BOM is indeed desirable (which I have little knowledge of), it should probably be done upstream.

Xitian9:
I've rebased onto master. I updated the commit message with a bit more detail, but my brain is too tired to have much of a go of it at the moment. Let me know if there are specific points you'd like addressed in the message.

simonmichael:
For the record, we asked about adding https://en.wikipedia.org/wiki/Byte_order_mark support at  https://github.com/serokell/haskell-with-utf8/issues/14 and meanwhile @Xitian9 has kept our existing BOM support in place.  It seems to still be reasonably common to encounter it in the wild.

I have proposed we add a backward compatibility flag to restore old file reading behaviour anticipating a long tail of bug reports from users somehow relying on the old behaviour as they upgrade over the coming years. Let's revisit this after some testing.

simonmichael:
Merged, release docs to follow. We may or may not add a flag to restore old reading behaviour  later if it seems worthwhile.

Otherwise, now that this has landed in master, any and all testing by folks with real-world non-ascii data is most welcome and helpful.

simonmichael:
We decided to revert this for 1.26 (#1864) and contemplate it a bit more, following discussion eg on the mail list:
https://groups.google.com/g/hledger/c/0gFC7xtsteI/m/lX3Fn2veBQAJ
Thanks @Xitian9.

PR #1839, opened 2022-03-11
dev: reg: areg: Be more clever about register and aregister alignment.

Xitian9:
When rendering register or aregister reports, calculate the amount /
balance width based on the first 100 items, and start rendering in that
way. If you encounter a longer one, update and continue rendering. This
will result in adjustment of column width for long reports, but allows
us to save a lot more performant/efficient.

We also only render each amount once, rather than twice as before, by
storing the rendered amount in a tuple.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Aregister.hs
+++ b/hledger/Hledger/Cli/Commands/Aregister.hs
@@ -56,6 +56,7 @@ aregistermode = hledgerCommandMode
 #endif
       ++ " or $COLUMNS). -wN,M sets description width as well."
      )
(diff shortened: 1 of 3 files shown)


simonmichael:
For the record: [a]register report is slow because it renders all lines in memory at once to calculate the perfect column widths (as well as because it renders a lot of amounts). This change chooses widths based on the first 100 lines. It means that after the 100th line there can be one or more discontinuities where column widths change, whenever a much-wider-than-previous value (description, account, posting amount, or balance amount) is encountered. This is ugly, but
- it saves about 75% time and 30% max residency
- we think that for most real world data, output will be the same as before
- the old behaviour (slower, perfect alignment) can be restored with --align-all.

PR #1840, opened 2022-03-11
imp: reg: areg: Align to first 1000 postings by default, rather than 100.

Xitian9:
This has the same performance characteristics for a big decrease in alignment
issues. 1000 postings seems to still put it under the memory bottleneck.


simonmichael:
Cool. Could you update "first 100 lines" in [A]Register.md also.

simonmichael:
Cool, thanks.

PR #1841, opened 2022-03-12
ref: Use ExceptT String IO a instead of IO (Either String a).

Xitian9:
This increases composability and avoids some ugly case handling.

### DIFF
--- a/bin/hledger-check-fancyassertions.hs
+++ b/bin/hledger-check-fancyassertions.hs
@@ -98,6 +98,7 @@ module Main where
 
 import Control.Arrow (first)
 import Control.Monad (mplus, mzero, unless, void)
+import Control.Monad.Except (runExceptT)
 import Control.Monad.Trans.Class (lift)
 import Control.Monad.Trans.State.Strict (runStateT)
 import Data.String (fromString)
@@ -167,7 +168,7 @@ main :: IO ()
 main = do
     opts <- execParser args
     journalFile <- maybe H.defaultJournalPath pure (file opts)
-    ejournal    <- H.readJournalFile (set H.ignore_assertions (ignoreAssertions opts) H.definputopts) journalFile
+    ejournal    <- runExceptT $ H.readJournalFile (set H.ignore_assertions (ignoreAssertions opts) H.definputopts) journalFile
     case ejournal of
       Right j -> do
         (journal, starting) <- fixupJournal opts j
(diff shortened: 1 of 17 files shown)


simonmichael:
When you have a chance, would you mind adding a list of API changes to the commit message(s).

simonmichael:
And let's move the second and third commits to a separate PR.

Xitian9:
Done.

simonmichael:
I feel like it's the API's job to make calling hledger as easy as possible, and runExceptT is one more strange thing for scripters to deal with, not to mention a potential breaking change for people's personal scripts. How would it be if we kept the old API, and added a new one with different names, such as the below (less like "readFile" etc., but more consistent with hledger's "typeOperation" naming scheme):

```haskell
-- old easy IO API
readJournal      :: InputOpts -> Maybe FilePath -> Text -> IO (Either String Journal)
readJournalFile  :: InputOpts -> PrefixedFilePath       -> IO (Either String Journal)
readJournalFiles :: InputOpts -> [PrefixedFilePath]     -> IO (Either String Journal)
-- new, more composable API
journalRead      :: InputOpts -> Maybe FilePath -> Text -> ExceptT String IO Journal
journalReadFile  :: InputOpts -> PrefixedFilePath       -> ExceptT String IO Journal
journalReadFiles :: InputOpts -> [PrefixedFilePath]     -> ExceptT String IO Journal
```
We could still deprecate and even remove the IO functions at some later point if needed.

simonmichael:
We went a different route, as discussed in chat. I'll . Thanks @Xitian9.

simonmichael:
Some build issues were fixed by #1847. Mini post-mortem:

> I did ask myself several times, "what could possibly go wrong just re-exporting runExceptT" 
> for a start, re-exporting is generally bad because it confuses people, they don't expect import Hledger to provide runExceptT which usually comes from Control.Monad.Except
> second, it was bound to introduce some redundant import warnings, and we treat those as build errors, and it's hard/slow to test for build errors across all supported GHC versions including all packages and components (make buildtest-stackX.Y.yaml or make buildtest-all is a good way to do it locally)
> third, I have not configured the main repo to require passing CI before merge (because it would mean I can no longer make quick pushes directly to master)

simonmichael:
Addressing the third point, I have now configured master to require a passing status check - currently, the `ghc: "86"...` job in either push.yml or pull.yml - before commit(s) can be pushed or merged there. This should somewhat help keep mistakes out.

PR #1848, opened 2022-03-26
Some cleanups related to use of ExceptT

Xitian9:


hledger-lib/Hledger/Read/Common.hs
```diff
@@ -249,42 +249,29 @@ commodityStyleFromRawOpts rawOpts =
...
+-- Timeclock and Timedot files.
+initialiseAndParseJournal :: ErroringJournalParser IO ParsedJournal -> InputOpts
+                          -> FilePath -> Text -> ExceptT String IO Journal
+initialiseAndParseJournal parser iopts f txt =
+    prettyParseErrors $ runParserT (evalStateT parser initJournal) f txt
+  where
+    y = first3 . toGregorian $ _ioDay iopts
+    initJournal = nulljournal{jparsedefaultyear = Just y, jincludefilestack = [f]}
+    prettyParseErrors = withExceptT customErrorBundlePretty . liftEither
+                    <=< withExceptT (finalErrorBundlePretty . attachSource f txt)
```
simonmichael, line 274: Woah, prettyParseErrors.. what is going on there, perhaps a short comment for future devs ?

Xitian9: What do you mean? It's completely self-explanatory. (end sarcasm). Okay, comment and type signature added.

simonmichael:
Thanks!

PR #1850, opened 2022-03-29
Switch to table-layout for tabular layout

### LINKED ISSUE #1679: Replacing/Breaking off Text.Tabular.AsciiWide.
There have been calls at various points to break off AsciiWide and WideString into its own library (see e.g. #1136).
The functionality is completely independent of hledger, and making it available for re-use/decoupling the versions would be useful. AsciiWide also happens to add a surprising amount to build time (a little under 5s on my system), so splitting this off would help build times.

That said, AsciiWide may not be (is probably not) the best solution to the problem, and instead of reinventing the wheel maybe we want to adopt another solution. [table-layout](https://hackage.haskell.org/package/table-layout) is a solution that people here have looked at before, and seems well-aligned to the job to be done. It also seems that it would make aligning on decimal points (a long-requested feature) easy. However, it does not seem to support wide characters (https://github.com/muesli4/table-layout/issues/8). @ony and @simonmichael put in some effort to see if that could be added about a year and a half ago; maybe I could put out feelers to see if they are open to more work on that.

Xitian9:
Fixes #1679. This gets rid of our home-customised Text.Tabular.AsciiWide and Text.WideString modules in favour of the external library table-layout.

At the moment this relies on a few PRs awaiting merge over at table-layout so it's probably not ready to be merged quite yet, but I'm putting it here for comment. I've broken it into several commits so that the changes are more bite-sized, but these changes are not atomic and tests will fail in the middle (mostly due to colouring negative amounts not being reintroduced until showMixedAmountB being updated), though it compiles at each step.


Xitian9:
This PR would open the door to a number of potential improvements/customisations.

- It would fairly easily allow aligning amounts on decimal points: https://github.com/muesli4/table-layout/issues/32
- table-layout has a fairly robust table styling interface, and we could easily make our tables fancier, if that's the sort of thing we felt like doing. Here's an example of one of the fancier tables it can create (ignore github-induced gaps between lines)
```
╭───╥───────────────────┬───────────────────┬────────────┰────────────┬────────────┬────────────┬────────────┰────────────┬────────────╮
│   ║     Some text     │   Some numbers    │     X      ┃     Z      │     W      │     A      │     B      ┃    Text    │     Y      │
╞═══╬═══════════════════╪═══════════════════╪════════════╋════════════╧════════════╪════════════╧════════════╋════════════╧════════════╡
│ 1 ║ This is long text ┆    4.20000000     ┆    foo     ┃    blah         bloo    ┆    blop         blog    ┃   Short         baz     │
│   ║       Short       ┆ 200300400500600.2 ┆    bar     ┃   yadda         yoda    ┆   yeeda         york    ┃   Short        wibble   │
├───╫┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┼┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┼┄┄┄┄┄┄┄┄┄┄┄┄╂┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┼┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╂┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┤
│ 2 ║ This is long text ┆    4.20000000     ┆    foo     ┃  bibbidy      babbidy   ┆    boo          blue    ┃   Short        wobble   │
╰───╨───────────────────┴───────────────────┴────────────┸─────────────────────────┴─────────────────────────┸─────────────────────────╯
```

simonmichael:
I know you already updated this a bunch of times, but when you get time could you do it once more and I'll review and merge if possible. 

The hledger-lib/Text/WideString.hs utils are might be useful down the road, could we to keep them around a bit longer ?

As you saw in chat I would like to tweak table borders, adding consistent separators for totals. Maybe also using the same style for all borders, making life simpler (rather than adding more complex styling - but no harm to have that ability in the toolbox).

How could using table-layout help with decimal aligning ?

simonmichael:
> How could using table-layout help with decimal aligning ?

"Align content in a column at a character occurence.", I see, nifty. Might be a bit tricky with our variable decimal marks..

Xitian9:
> "Align content in a column at a character occurence.", I see, nifty. Might be a bit tricky with our variable decimal marks..

It's possible with a custom implementation of `Cell`. See here: https://github.com/muesli4/table-layout/issues/32

simonmichael:
https://github.com/muesli4/table-layout/issues/43

Xitian9:
As much as I would love to see something like this integrated at some point – and I think `table-layout` provides an excellent technical solution – the project seems to have stagnated, with the last change necessary for passing all our tests being stalled for more than two years now.

I'll reluctantly close this pull request, but I continue to think that we should ideally use an external library for table formatting and layout, rather than a custom in-house solution.

simonmichael:
Thanks for following up @Xitian9.

Issue #1436, opened 2020-12-30
Parseable, consistent error messages


DamienCassou:
Right now, `hledger check` error messages are hard to parse which makes writing tools around hledger hard. Here is an example of an error output when an account is undeclared:

```
$ hledger check --strict
hledger: 
strict mode: undeclared account "debt"
in transaction at: "/home/cassou/Documents/configuration/ledger/2021.ledger" (lines 11-13)
```

and here is an error output when a transaction doesn't balance:

```
$ hledger check --strict
hledger: "/home/cassou/Documents/configuration/ledger/2021.ledger" (lines 16-18)
could not balance this transaction:
real postings' sum should be 0 but is: -1.00€
2021-01-01 Misc
    asset:account         -11.00€
    expense:food           10.00€
```

As you can see, the structures of the error messages are different, requiring different error parsers.

simonmichael:
I also have the feeling that, for maximum compatibility with tools, the first line should contain the word "error" and the error's position. Does that sound correct ?

We also want error messages to be readable on a standard 80-char terminal, so we usually have a challenge with long lines.

Perhaps it's not necessary to mention the program name.

And of course eventually we'd like to show a pretty, coloured human-friendly rendering of the error region.

Some mockups:
```
$ hledger check payees
Error at "/Users/simon/src/hledger/examples/sample.journal" (line 27):
Undeclared payee "income"

> 2018-01-01 income
      assets:bank:checking              $1
      income:salary                    $-1

$ hledger check --strict
Error at "/home/cassou/Documents/configuration/ledger/2021.ledger" (line 12):
Undeclared account "debt"

  2020-01-01
>   debt    1
    assets

$ hledger check
Error in transaction at "/home/cassou/Documents/configuration/ledger/2021.ledger" (lines 16-18):
Could not balance this transaction:
real postings' sum should be 0 but is: -1.00€

> 2021-01-01 Misc
>     asset:account         -11.00€
>     expense:food           10.00€
```

We also have an cosmetic issue where some errors include text like "*** Exception:" or "ExitFailure 1" and others don't, depending on where they are handled.

simonmichael:
More notes:
- best to include a brief error description on the first line as well, even if it makes lines wide ?
- and/or include a unique searchable short error code ? Maybe our errors are simple enough this is overkill.
[...]

PR #1861, opened 2022-04-25
more consistent error messages, per #1436

simonmichael:
This updates a number of journal-related error messages per #1436, and adds tests for all of them. There are still a few to do, but I won't be able to get back to this for a few weeks, and want to avoid a long-lived conflicful branch, so I thought I'd show this now and maybe merge it soon.

https://github.com/simonmichael/hledger/blob/d7360e3/hledger/test/errors/README.md#current-status


simonmichael:
Thanks! In that case I'll go ahead and merge, to lock down some progress. Sorry about the conflicts.

PR #1868, opened 2022-06-05
CREDITS.html -> CREDITS.md

the-solipsist:
Corrected the extension for the CREDITS file.


simonmichael:
I delayed because I thought I'd have to add a redirect to keep the link working at both https://hledger.org/dev-README.html#funding and https://github.com/simonmichael/hledger#funding, but no, .md works in both places. Thanks!

Issue #1229, opened 2020-04-20
hledger-web: adding a transaction always add it to the default journal file


legrostdg:
Given this `ferme.journal`:

```
include 2019.journal
include 2020.journal
```

I tried to add a transaction to `2020.journal` with hledger-web. I selected `2020.journal` in the `to:` field, but the transaction got added to `ferme.journal`

simonmichael:
Thanks for the report, it sounds like a regression.

erictapen:
Just stumbled over this issue, I proposed a fix in https://github.com/simonmichael/hledger/pull/1873.

simonmichael:
It looks like this regressed with [cc1241fa2](https://github.com/simonmichael/hledger/commit/cc1241fa205beee103fe39d1c1c9e0523c74edfb#diff-002b5322a2a570277540532bc117176af530456ac9a24fc7629a76abfece5009) in 2018.

I'd like to restore the old behaviour where the file field is part of normal form processing (and add your extra validation, which I think we never had). Yesod form code is tricky so it'll take a bit longer, or if you'd like to have a go let me know and I'll post my diff.

erictapen:
I understand but I think I won't put more effort into it, as indeed the form code deterred me a bit.

simonmichael:
Understandable. Thanks for pushing it forward!

simonmichael:
This is currently blocked on figuring out Yesod types; it's not clear how to use selectField here: 
https://github.com/simonmichael/hledger/commit/cfc05540daad75a358a47741c16e7611fe42f54f#diff-4f785631673350333a037eca53769f9320f8b8a2856371597ef0687f52f86849R72-R124.

simonmichael:
Fixed in master, at last. Thanks again @erictapen for moving this forward.

And also for pointing out the lack of server-side validation of the journal file path. This might deserve further action, eg test if existing hledger-web versions can be tricked into append transactions to arbitrary files the hledger-web process has write permission on, and if so do some announcing and deprecating. Help on this is welcome from anyone. I'll leave this open for now.

erictapen:
Thats awesome news! I'm glad there is now a fix in master.

simonmichael:
Happily, we did always validate the file path when adding to a different file (last seen in hledger-web 1.9). So, no security issue after all

PR #1873, opened 2022-06-16
hledger-web: actually add transaction to the selected journal file

erictapen:
[Fix for #1229]

The hledger-web interface currently allows for selecting a file on which to append a transaction, but currently that data isn't used anywhere. It just writes to the default file, regardless of the selection.

tmp CpdrFTokbK

My solution is somewhat of a hack; I use the `tsourcepos` metadata in the `Transaction` type to carry the preffered journal file around.

Also this currently has the security implication, that it allows users to append transactions to arbitrary files. It probably would make sense to only restrict this to the files that are discovered by hledger, but I'm not familiar with the logic.


simonmichael:
>  it allows users to append transactions to arbitrary files

Thanks for the PR and for calling out the above, which does sound like a blocker.

erictapen:
> Thanks for the PR and for calling out the above, which does sound like a blocker.

Do you have any pointer to where I could get a list of allowed files from? I'm a bit lost in the code base currently.

simonmichael:
Sure, hledger-lib:Hledger.Journal.Data.journalFilePaths will list all the journal's files.

simonmichael:
PS, this worked once - possible some spelunking in code history is useful ?

erictapen:
I implemented the check for the journal file. Didn't look at the code history though, so I have no idea what broke this.

simonmichael:
We went with a different fix, see #1229.

Issue #1754, opened 2021-11-14
timedot: add payee/description and trailing comment/tags


linuxcaffe:
The most awesome timedot would be even awesomer with a description (payee field). 
Currently a timedot entry carries no description, but can handle a trailing comment. 

`<acct:sub:sub><spaces><qtty><spaces><;><comment>`

This enhancement would use the comment as description.

linuxcaffe:
Closed in favour of #1220 https://github.com/simonmichael/hledger/issues/1220#issuecomment-968392952

simonmichael:
Related to #1220 (comments in timeclock format). Let's keep this one open for timedot specifically.

simonmichael:
Following up on djp's [timedot comment](https://github.com/simonmichael/hledger/issues/1220#issuecomment-968392952) there: 

- [timedot format](https://hledger.org/hledger.html#timedot-format) supports descriptions, written after the date
- You can add a `; comment` and tags there too, but it's treated as part of the description. Instead we should parse text after semicolon as a comment and tags, as in journal format
- We should do the same for `; comment` on a time dots line, where currently it is ignored.

linuxcaffe:
I'm confused. "timedot supports descriptions".. how? I get an error putting anything after the date. I often put a comment like `;Friday` on the line following the date, but.. 
Does that description relate to the day?  Not to each timedot event, right? When I review my ledger combining the timedot and timeclock, using hledger-ui, for example, entries from the timeclock have descriptions, those from timedot do not.
I DO agree that the trailing `<spaces>;comment :tags:` would be dandy..

simonmichael:
Eg:
```
* 2021-11-14 Sun
fos.hledger              .... .... .
```
```
$ hledger -f a.timedot print desc:sun
2021-11-14 * Sun
    (fos.hledger)            2.25
```

simonmichael:
(It works for me at least back to hledger 1.17..)

linuxcaffe:
Oh that's cool on the date (emacsism?) but I'm after a description on the entry
```
      (fos.hledger)        2.25       more dumb suggestions in irc
```

simonmichael:
Yes, in that position I propose we should support comments  (`; comment maybe with a tag:`). There's no room for a posting description in our data model.

simonmichael:
[...]

PR #1878, opened 2022-07-08
timedot: support ranges

alhirzel:
This is a follow-up to a thread I created here: https://groups.google.com/g/hledger/c/FaEBpC8JYus

This change allows the following to parse correctly in a timedot file:

```
2022-06-28 description
Time:ABC:Task1  8.2-8.9 ; comment 1
Time:ABC:Task1  9-10.3 ; comment 2
```

Using this syntax, the duration associated with each line is computed by subtracting the two numbers (8.9 - 8.2 = 0.7 or 10.3 - 9 = 1.3). The default time unit is applied just as if the hour count itself (0.7 or 1.3) had been written instead. The primary advantage is that I can use a 24-hour timestamp for start and end times, and I can leave a "hanging entry" while I'm working. It also makes a 6-minute rule very easy to observe by using zero or one decimal place everywhere, and is also less verbose than the timeclock format for this use case.

### DIFF
--- a/examples/sample.timedot
+++ b/examples/sample.timedot
@@ -19,4 +19,5 @@ fos:hledger   .... .... ....
 fos:ledger    0.25
 fos:haskell   .5
 inc:client1   2
+inc:client2   14-16
 
--- a/hledger-lib/Hledger/Read/TimedotReader.hs
+++ b/hledger-lib/Hledger/Read/TimedotReader.hs
@@ -175,7 +175,7 @@ entryp = do
   lift skipNonNewlineSpaces
   hrs <-
     try (lift followingcommentp >> return 0)
-    <|> (lift durationp <*
+    <|> (lift (try timerangep <|> durationp) <*
          (try (lift followingcommentp) <|> (newline >> return "")))
   mcs <- getDefaultCommodityAndStyle
   let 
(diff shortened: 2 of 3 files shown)


simonmichael:
Thanks! We might end up doing this, perhaps adding `..` for consistency with ranges elsewhere. As written it seems a little non-self-explanatory. How do you (would you) choose those numbers in practice ? Do they represent times of day ? 

I am concerned about timedot format becoming rather complicated. The more we add to it, the less likely it is to spread beyond hledger. I'd vote we do #1754 first and maybe experiment with this via preprocessor script.

alhirzel:
Good plan, will work #1754 first.

I indeed use this with times of day as the range numbers so that I can more easily relate a timedot sheet to the blocks of time on my calendar (e.g. meeting for project X from 9am-10am). Perhaps I can improve the documentation to make it more clear.

I think the `..` range syntax is very sensible as well; were you thinking of supporting both the `..` and `-`? You mention `..` is also consistent with other ranges, where are examples of this? (I quickly searched the repo and didn't see any examples of parsing the `".."` string.)

simonmichael:
https://hledger.org/1.26/hledger.html#period-expressions

I think if people are writing times they will find the decimal notation unexpected, and may try to write variants like HH.MM[.SS[.DEC]] or   HH:MM[:SS[.DEC]], etc..

alhirzel:
Ah, I see. You raise a good point about ambiguity wrt `hh.mm` syntax (or more likely, `h.m` syntax). I think it may indeed be best to table this syntax or have my own preprocessor script or a separate format to avoid the conflict. Thank you so much for your thoughts on this Simon! What are your thoughts on closing this PR?

simonmichael:
Sounds good for now!

PR #1881, opened 2022-07-09
examples/csv: add example for capital one credit cards [ci skip]

maxthomas:
hey there! hledger newbie here and just wanted to share some rules for .csv import for a financial institution that's popular in the US.

it also serves as an example of how to set posting dates given secondary dates from .csvs.

let me know if i can change anything, or if it doesn't belong :)

thank you all so much for this project!!


simonmichael:
Thank you very much!

Issue #35, opened 2013-04-08
hledger-web startup failure


simonmichael:
_Original author: simon@joyful.com (December 18, 2010 21:34:26)_

W. Ludwick is having a startup failure with the hledger-web 0.13 mac binary that I can't reproduce:

$ ./hledger-web-0.13-mac-i386
No journal file &quot;/Users/walt/.journal&quot;, creating it.
Edit this file or use &quot;hledger add&quot; or &quot;hledger web&quot; to add transactions.
(no further output. But shows $ prompt ? Not sure)

With the same binary I get (from a clean start):

$ ./hledger-web-0.13-mac-i386
No journal file &quot;/Users/simon/.journal&quot;, creating it.
Edit this file or use &quot;hledger add&quot; or &quot;hledger web&quot; to add transactions.
Installing support files in ./.hledger/web/ - done, please run again.
$ ./hledger-web-0.13-mac-i386
Using support files in ./.hledger/web/
Starting http server on port 5000 with base url http://localhost:5000
Attempting to start a web browser

_Original issue: http://code.google.com/p/hledger/issues/detail?id=35_

PR #1884, opened 2022-07-12
Update achievements listing

alhirzel:
noticed these when digging thru the docs!

### DIFF
--- a/ACHIEVEMENTS.md
+++ b/ACHIEVEMENTS.md
@@ -19,7 +19,7 @@ Some project achievements to unlock!
 - ~~pretty good tutorial docs~~
 - ~~pretty good cookbook docs~~
 - ~~100 IRC chatters~~
-- **2k github stars**
+- ~~2k github stars~~
 - **in top 30 github starred haskell projects**
 - **match ledger speed**
 - **match ledger committers**


simonmichael:
Thanks! Got a link for the "in top 30 github starred haskell projects" ?

simonmichael:
(Reckoning by http://stars.hledger.org we are still at #35.)

alhirzel:
I just manually counted down this list, looks like number 23: https://github.com/topics/haskell?o=desc&s=stars

Do you prefer to use the link you posted? If so I'll modify the PR. Thanks for the detailed check, I'd hate to get this wrong!

alhirzel:
And note that it's number 17 on this list, which filters the last link I posted to projects that indicate their language is Haskell AND are tagged as #haskell. https://github.com/topics/haskell?l=haskell&o=desc&s=stars

simonmichael:
It's not a big deal but yes I think the http://stars.hledger.org list is best (I've now trimmed it to >2k stars, making it shorter). Because the others find only the projects who have somehow configured (?) their topic; ours seems more accurate, eg it includes https://github.com/github/semantic, and excludes https://github.com/hmemcpy/milewski-ctfp-pdf.

alhirzel:
@simonmichael - understood and updated PR to only take credit for the 2000 stars

PR #1890, opened 2022-07-18
imp: Add ' as possible digit separator

ThreeFx:
It is used for Swiss Francs: CHF 1'000


simonmichael:
@ThreeFx, thanks for this. Some ideas to move it forward:

- update the docs in hledger.m4.md > Decimal marks, digit group marks
- add a test or two, I think numbers.test would be the right place
- at least manually sanity-check that it works as expected in the `commodity` directive and `--commodity` option

simonmichael:
Closing for now.

PR #1917, opened 2022-08-14
fix multi-file account display order (#1909)

### LINKED ISSUE #1909: account display order can be wrong when accounts are declared in multiple files
The order of `account` directives is tracked during parsing, but restarts from 1 in each file, instead of being a total order across all files.

It means that if you have account directives in multiple files, those accounts tend to be displayed interleaved, instead of in the order you would expect (first the accounts declared in first file, then the accounts declared in second file).

simonmichael:
This PR fixes #1909 (by renumbering account directives after merging or finalising Journals), and adds some useful debug logging (at levels 2, 5 and 6) of file reading/including and account declaration order. 

- lib: save account directive positions, for troubleshooting (1909)
- imp: at --debug 5, log account declarations info while parsing (1909)
- imp: accounts: at --debug 2, show account declaration positions (1909)
- dev: tests for multi-file account display order (1909)
- fix: fix multi-file account display order; improve file read logging (1909)
- dev: journalFilePath, include: show (unknown) instead of nothing
- dev: lib: hlint improvements
- imp: add a missing space after colon in some debug output

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -308,7 +308,7 @@ journalFilePaths :: Journal -> [FilePath]
 journalFilePaths = map fst . jfiles
 
 mainfile :: Journal -> (FilePath, Text)
-mainfile = headDef ("", "") . jfiles
+mainfile = headDef ("(unknown)", "") . jfiles
 
 addTransaction :: Transaction -> Journal -> Journal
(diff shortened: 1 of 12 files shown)


simonmichael:
The --debug=2 addition is slightly unorthodox: it affects the account command's standard output. Perhaps it should be a new flag like --types (--declarations ?), instead.

simonmichael:
Added `account --declarations` flag.

Issue #1921, opened 2022-08-20
cf/bs/bse/is don't show child accounts if there is an intervening account of another type


simonmichael:
Reported by scara. Eg 
```
account a         ; type:L
account a:aa      ; type:X
account a:aa:aaa  ; type:L

2021-01-01
    (a)                                            1
    (a:aa)                                         1
    (a:aa:aaa)                                     1
```
```
$ hledger bs
Balance Sheet 2021-01-01

             || 2021-01-01 
=============++============
 Assets      ||            
-------------++------------
-------------++------------
             ||            
=============++============
 Liabilities ||            
-------------++------------
 a           ||         -1 
-------------++------------
             ||         -1 
=============++============
 Net:        ||          1 
```
It should show `a:aa:aaa` as well (compare `bal type:L` and `accounts --types`. These commands need to be updated to use the newer account type queries.

https://github.com/simonmichael/hledger/blob/b32b1300459736968cc10fdce10f7ae9b328a9d6/hledger-lib/Hledger/Data/Journal.hs#L481-L510

simonmichael:
Fixed by #1923.

PR #1923, opened 2022-08-20
fix!: cbr: Remove old account type query code. (#1921)

Xitian9:
This replaces the old journal*AccountQuery with the new Type query. This
enables uniform treatment of account type, and fixes a subtle bug
(#1921).

Note that cbcsubreportquery no longer takes Journal as an argument.


simonmichael:
Lovely, thanks!

PR #1929, opened 2022-09-07
ui: simpler, more debuggable Screen type

simonmichael:
- dev: lib: improve ReportSpec doc
- dev: ui: refactor: simpler, more debuggable Screen type


simonmichael:
I believe this is better - easier to understand, debug, and change without introducing bugs - though it may not be the last word (it would be interesting to see a solution using typeclasses, eg). Any review or comments welcome. https://github.com/simonmichael/hledger/blob/a5f4d2fd6e33f11d1dada30b0a173e1aa1f5fbf6/hledger-ui/Hledger/UI/UITypes.hs#L150 is a good starting point. This passes the ui tests (uitest.org).

Issue #1942, opened 2022-10-28
hledger-git: git: 'record' is not a git command


pfiaux:
I installed `hledger-git` and while looking through the script I noticed record calls `git record ...`.

https://github.com/simonmichael/hledger/blob/25c9083d0dcbed80498b227b46a3e3a12b17315d/bin/hledger-git#L72-L79

Should it call `git commit` instead on line 78?

To reproduce (with hledger-git installed):
```
⟩ hledger git record "test"
git: 'record' is not a git command. See 'git --help'.

The most similar command is
        restore
```
(same with `hledger-git record "test"`)

I'm on macOS, and I think hledger version isn't relevant here since the issue seems to be in the add on script.

simonmichael:
Thanks for the report, you're quite right - I had this alias defined in .gitconfig:
```
  record = ! sh -c '(git add -p -- $@ && git commit) || git reset' --
```

It should be commit (in some form). If you are able to try it and test that it performs as required (reliable, no prompts) that would be very helpful.

pfiaux:
I think calling `git add -p` isn't needed since the files are added in the for loop (L76). Also the patch option could result in prompts if some files were not added:
```
-p, --patch
   Interactively choose hunks of patch between the index and the work tree and add them to the
   index. This gives the user a chance to review the difference before adding modified contents to
   the index.

   This effectively runs add --interactive, but bypasses the initial command menu and directly jumps
   to the patch subcommand. See “Interactive mode” for details.
```

I think the following will do:
```
$GIT commit -m "$MSG" "$@" -- "$FILES" || $GIT reset
```

- Calls commit with a message and any optional arguments
- If commit fails unstage all the changes as your alias did (I don't think arguments are needed here)

I'll update my hledger-git and run some tests.

simonmichael:
Fixed by #1949.

PR #1949, opened 2022-11-13
fix: cli: updates hledger-git record command to use plain git

pfiaux:
This fixes the issue in https://github.com/simonmichael/hledger/issues/1942.
Instead of relying on a `git record` alias being present `hledger git record`
uses only vanilla git.

The expected alias for record is:
```
record = ! sh -c '(git add -p -- $@ && git commit) || git reset' --
```

The removed `git record` is in the script is implemented using:
- `git commit` with the given messages (and additional arguments if given)
- `git reset` if the commit fails

`git add -p` isn't needed because the files are added already above. Also
this would be interactive which isn't the goal for this call.

The message still defaults to the date if not given, however I had to add
a check to only shift if arguments were passed in, otherwise shift fails.

### DIFF
--- a/bin/hledger-git
+++ b/bin/hledger-git
@@ -8,7 +8,7 @@ hledger-git - easyish version control for your hledger journal, using git.
 An experimental prototype, currently works for the default journal only.
 A git repo in the main file's directory will be autocreated if needed.
 
-Subcommands: 
+Subcommands:
 
 hledger git record [MSG] - record the journal's files (as listed by 'files')
 hledger git status       - show unrecorded changes (after first record)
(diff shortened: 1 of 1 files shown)


bin/hledger-git
```diff
@@ -74,8 +74,11 @@ record() {
     ensure_git_repo
     cd "$DIR"
     for F in $FILES; do $GIT add -f "$F"; done
-    MSG=${1:-$(date +'%Y-%m-%d %H:%M:%S %Z')}; shift
-    $GIT record -m "$MSG" "$@" -- "$FILES"
+    MSG=${1:-$(date +'%Y-%m-%d %H:%M:%S %Z')}
+    if [ $# -ge 1 ]; then
+        shift
+    fi
+    $GIT commit -m "$MSG" "$@" || $GIT reset
```
pfiaux, line 81: minor: `"$@"` is only possible if a custom message is given. However to change this would change the previous/existing behavior since it was like that.

simonmichael:
Merged as b3de7e59a. Thanks!

PR #1955, opened 2022-12-05
Change inheriting valued tags to override

### LINKED ISSUE #1950: Inherited Tag Values are not Overwritten

Consider this `book.hledger`:

```hledger
2022-11-17 Aldi
    ; concerns: me
    Assets               -30 €
    Costs:Food            20 €
```

```bash
❯ hledger --version
hledger 1.27.1, linux-x86_64

               -10 €  me
                10 €  you
--------------------
                   0  
# ☝️ This is expected.


# ☝️ This is expected.

# ☝️ Why does this appear? 

# ☝️ This is expected.
```

## Conclusion

## Proposal

chrislemaire:
Resolves #1950 

Old behaviour of inheriting tags with values was that tags added to, for instance, a posting, would be added to the tags, possibly overriding the tags of the transaction. Or, in other words, the transaction tags were added to the posting tags, as there was no sense of overriding a tag.

The new behaviour is that tags are now overridden when a lower level re-uses that tag name. For instance, when defining a transaction with tag t:v and posting with tag t:v2, only t:v2 remains on the posting, overriding the transaction tag.

Perhaps it would be desirable to add an option for additive tags over overriding tags, as it may well be useful at times to add to tags in the parent.


hledger/test/query-tag.test
```diff
@@ -214,3 +214,23 @@ $ hledger -f- bal -N tag:type=a
...
+2022-11-17 Aldi                 me                           -30 €         -30 €
+                                me                            20 €         -10 €
+
+# 23. Accounts can override the tags of their parents
+<
+account a    ; atag:A
+account a:b  ; atag:B
+
+$ hledger -f- accounts tag:atag=a
+a
```
lukasbestle, line 236: There could be more tests for the combination of transaction/posting and account tags:

```suggestion
a

# 24. Transactions can override the tags of used accounts
<
account a    ; atag:A
account a:b  ; atag:B

2023-01-01 Test
    ; atag:C
    (a:b)  10 €

$ hledger -f- bal --pivot=atag
                10 €  C
--------------------
                10 €

# 25. Postings can override the tags of their accounts
<
account a    ; atag:A
account a:b  ; atag:B

2023-01-01 Test
    ; atag:C
    (a:b)  10 €  ; atag:D

$ hledger -f- bal --pivot=atag
                10 €  D
--------------------
                10 €
```

lukasbestle (commented):
I think it makes sense to override in hierarchical order, meaning that the closer a tag is to the posting, the higher the precedence. This means:

1. Tag defined for the posting
2. Tag defined for the transaction
3. Tag defined for the account
4. Tag defined for the nextmost parent of the account up to the top of the account hierarchy

Another way to look at it: This order ensures that the "most specific" tag is used. Accounts are defined for the whole accounting setup and are therefore really general. Transactions are more specific as they are one instance where the account is used. The posting is the most specific.

And a third reason for this order: The closer a tag is to the posting, the higher the likelihood that the author intended the tag to be used. Tags that are defined further away from the posting (maybe even in another file) can be forgotten, so it would be weird if they would have a higher precedence.

hledger/test/query-tag.test
chrislemaire: I've added your tests as suggested and I'm thinking of more tests that should be relevant to this ordering with multi-valued tags. Thank you very much for taking the time to look into some tests!

simonmichael:
Here's my related comment on the parent issue #1950, suggesting things to help move this forward: https://github.com/simonmichael/hledger/issues/1950#issuecomment-1382899511

simonmichael:
Hi @chrislemaire, are you still interested in this hledger PR: Change inheriting valued tags to override ?

PR #1969, opened 2022-12-28
Update hledger.m4.md

RobertNielsen1:
Updated section on pivoting. Used synonyms for "member" in cases where there could be confusion with the tag named "member."

### DIFF
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -4710,8 +4710,8 @@ Some examples:
 
 ```journal
 2016/02/16 Yearly Dues Payment
-    assets:bank account                    2 EUR
-    income:dues                    -2 EUR  ; member: John Doe
+    assets:bank account                 2 EUR
+    income:dues                        -2 EUR  ; member: John Doe
 ```
 Normal balance report showing account names:
 ```shell
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -4709,15 +4709,15 @@ Values containing `colon:separated:parts` will form a hierarchy, as account name
 Some examples:
 
 ```journal
-2016/02/16 Member Fee Payment
+2016/02/16 Yearly Dues Payment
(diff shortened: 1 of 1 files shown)


hledger/hledger.m4.md
```diff
@@ -4709,15 +4709,15 @@ Values containing `colon:separated:parts` will form a hierarchy, as account name
 Some examples:
 
 ```journal
-2016/02/16 Member Fee Payment
+2016/02/16 Yearly Dues Payment
     assets:bank account                    2 EUR
-    income:member fees                    -2 EUR  ; member: John Doe
+    income:dues                    -2 EUR  ; member: John Doe
```
simonmichael, line 4714: Should we keep the 2's aligned, just for clarity ?

(Procedure: change it locally and "amend" (git commit --amend) the old commit, then force push (git push -f) to this PR branch again.)

simonmichael: Sorry for the nit pick, I figure it's good practice! :)

RobertNielsen1:
I think it worked… Also, I marked the conversation as resolved, but now am not sure if I was supposed to do that.

Rob

simonmichael:
You added another commit rather than amending the old one. That's fine, I'll just squash them into one as I merge. I'm waiting for CI to pass before I can do that. 

Another tip, for low-impact commits like this you can begin the commit message with ; to bypass the slow/costly parts of CI testing.

And another, CI usually requires the commit message to begin with a prefix and colon (more about this at https://hledger.org/COMMITS.html <https://hledger.org/COMMITS.html>, but the doc is over zealous, don't pay too much attention to the details). In short, for this kind of change I'll usually use a message like 

";doc: journal: change example"

I have mucked things up by pressing the wrong button in Github, happens all the time, so I'm going to pull your commits and tweak the message. You'll see my reply on the PR when done.

simonmichael:
Merged as 4d8337513. Thank you!

PR #1973, opened 2023-01-06
imp: close: add clopen command / provide more useful defaults

simonmichael:
- cli: fix some Commands file permissions
- cli: simplify command doc inclusion, give them normal heading structure
- lib: Hledger.Data.RawOptions: add unsetboolopt
- feat: clopen command (documentation)
- feat: clopen command (implementation)
- ;cabal: update cabal files
- ;doc: update CLI help texts
- ;doc: update manuals
- ;dev: shake: cmdhelp: tweak commit message


simonmichael:
This was born out of the perpetual confusion between
- RX-closing: zeroing out revenues/expenses to reset them and focus on recent activity (and balance the accounting equation)
- AL(E)-closing: zeroing out assets/liabilities/equity at the end of a PTA file (an unzeroing them at the start of the next file) to allow flexible file combining

Even though these are mathematically similar they are distinct use cases and I figured it was worth giving separate commands with more appropriate defaults for each, for ease of use and ease of learning and discussion. `close` prints one closing transaction, closing RX by default. `clopen` prints a closing and an opposite opening transaction, AL by default. But both of them can be customised to do other things.

I'm not sure if this arrangement makes sense and is worthwhile, given very few of us close RX balances and it seems a little bit obsolete in computer-based accounting (we are used to using date filters, no need to "reset"). So I'm going to let this sit for a bit, I'd welcome any comment.

the-solipsist:
I don't have strong feelings either way. It might be a good idea given that they are conceptually different.

However, I _do feel strongly_ that clopen should by default include Equity as well, since when you shift to a new journal, all your Equity balances (e.g., Retained Earnings, Conversions, and so on) need to shift as well.

the-solipsist:
Though, I'd hasten to add that one can get rid of previous FY's expenses and revenue by simply shifting to a new file and not migrating RX accounts, and migrating only ALE accounts.

Additionally, I'm not sure how one could say, "so few of us close RX accounts", given that it's a standard requirement if one uses hledger for business purposes. Thus it may or may not be true.

simonmichael:
"so few of us close RX accounts" - This is entirely my guess, but an informed one; I would put a lot of money on it! Other RX-closers out there, let us know you're there.

simonmichael:
An alternate plan: Keep as one command, with different modes/defaults and better docs:

- `close [--close]` - print a closing transaction, retaining earnings by default
- `close --clopen` - print closing and opening transactions, migrating ALE balances by default
- `close --open` - print an opening transaction, opening ALE balances by default

simonmichael:
Or `close [--close] [--open]`, as we have now, but printing just a closing transaction by default instead of both.

simonmichael:
Or `close [--retain | --move | --open] [QUERY]`:

- By default it prints a closing transaction for all accounts
- --retain prints only a RX-closing transaction and sets default desc/acct for retaining earnings
- --move prints ALE-closing and -opening transactions and sets defaults for moving to a new file
- --open prints an ALE-opening transaction and sets defaults for initialising a new file
- A QUERY can be added to further restrict the accounts above

the-solipsist:
Perhaps "--migrate" is better than "--move"? I think it indicates migrating
to a new file more intuitively / unambiguously.

simonmichael:
Well definitely nothing here is final... I moved away from "migrate" because that word could be needed for someting else, like upgrading to a newer file format; and you might not be "migrating" to a new file, but "moving" balances around for some other reason...  (there is also bin/hledger-move, which does something not unlike this, though with extra features)..

simonmichael:
I dropped the separate clopen command and have implemented the `close [--retain | --migrate | --open] [QUERY]` proposal. Testers welcome.

simonmichael:
Merged as 0db44f11b.  (Unrelated commits appearing on this PR were me trying to optimise CI.)

PR #1974, opened 2023-01-06
doc: add pandas to recommended tools; fix manual budgeting-and-foreca…

maxthomas:
hey there! noticed that the link to `budgeting-and-forecasting` in the manual seems [broken](https://hledger.org/1.28/budgeting-and-forecasting.html) and instead should point to the versionless [link](https://hledger.org/budgeting-and-forecasting.html).

i *believe* this fixes the issue; seems to work locally after running `pandoc`, which i added a brief blurb to in the dev docs.

i wasn't super sure if i should commit the generated pages, since it seems to have trailing whitespace trimmed, which would have created a lot of files to review/etc. 

:heart: for all you do!

### DIFF
--- a/doc/DOCS.md
+++ b/doc/DOCS.md
@@ -80,7 +80,7 @@ $ make Shake
 
 Then render the per-package manuals from markdown-m4 source files (*.m4.md) 
 to text, man, info, and markdown formats. This requires some unix tools 
-such as m4:
+such as m4, as well as pandoc:
 
 ```
 $ ./Shake manuals
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -4834,7 +4834,7 @@ each periodic transaction rule generates recurring budget goals in specified acc
 and goals and actual performance can be compared.
 See the balance command's doc below.
  
-See also: [Budgeting and Forecasting](/budgeting-and-forecasting.html).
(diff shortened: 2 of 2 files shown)


simonmichael:
Thanks @maxthomas!

Hmm, `/../budgeting-and-forecasting.html` doesn't look right.

You might be seeing a broken link depending how you are previewing, but on the rendered site the current link should be correct, leading to https://hledger.org/budgeting-and-forecasting.html.

maxthomas:
thanks @simonmichael ! apologies, i'm probably being unclear - the current manual is almost certainly broken (or, i'm not doing this right, also possible :) )

here's the flow i'm using:

- hit [this versionless link to the manual](https://hledger.org/hledger.html#budgeting-with-periodic-transactions)
- get redirected to [1.28](https://hledger.org/1.28/hledger.html#budgeting-with-periodic-transactions)
- try to click [Budgeting and Forecasting](https://hledger.org/1.28/budgeting-and-forecasting.html) link, which contains `1.28`
- get a `404`

would an absolute, and not relative link, work? i'll push that up right now if so; if not - let me know!

simonmichael:
I see what you mean now, thank you. I had fixed this in the dev manual but not in the older manuals. I have fixed it in all manuals now (by making the path absolute, which is the convention throughout the docs). Does this resolve it ?

maxthomas:
right on! sure thing - i'll close this. thank you again for all you do!

Issue #1962, opened 2022-12-16
Solve Ledger file reading


simonmichael:
Capturing this from chat:

Despite all the work on supporting Ledger syntax, the vast majority of Ledger users who try to read their file with hledger, fail and move on. Including the folks who don't use value expressions, and the folks who have been told the ledger print | hledger -f- ...  trick.
you'd think that last would work - ledger print is always valid h/ledger syntax, right ? I think now there are two main snags: 
needing to set LANG, because everyone has non-ascii and haskell programs throw up their hands if they see that without proper LANG
needing to add commodity directives or -c options, because ledger print adds decimal zeros forcing hledger to check transaction-balancedness more precisely than ledger does
definitely time we had a better strategy here

simonmichael:
Documented the second issue for users at https://hledger.org/ledger.html#incompatible-balancing

simonmichael:
https://github.com/simonmichael/hledger/tree/master/hledger/test/ledger-compat is the start of a test suite for Ledger file compatibility. It uses Ledger's functional tests as a source of diverse sample Ledger files, and others collected manually can be added over time. Let me know if you can think of another good source.

https://gist.github.com/simonmichael/052703b1641669bfe067c68b81f707cc is the categorised results of a test run.. easier to read in Emacs, but to summarise, we currently read about 80% of Ledger's tests' sample data files. The most frequent causes of read failure were amount expressions and lot notation. There was ~20 other distinct causes of failure as well.

simonmichael:
https://hledger.org/ledger.html#ledger-file-format-support-status is a new status table.

alensiljak:
> design and implement some kind of Ledger file detection

What do you think of https://github.com/ledger-rs/incubator/discussions/2? The main suggestion is the data format specified in the header, similar to shebangs. The program handling becomes easier with semantic versioning.

simonmichael:
@alensiljak seems a good idea. hledger uses file extension as a hint for input/output format also - .csv/.tsv/.ssv/.timeclock/.timedot/.journal/.hledger (/.ledger/.beancount/...)

PR #1982, opened 2023-01-13
imp: journal: periodic txns need not start on an interval boundary

simonmichael:
Eg,`~ monthly from 1/15` now works, instead of giving an error message or having to rewrite it as `~ every 15th day of month`.

The motivation here was #1962; such periodic transaction rules are one of the things that can prevent hledger reading a Ledger file. Also more generally this limitation just looks a bit arbitrary to users.

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -165,20 +165,21 @@ spanYears (DateSpan ma mb) = mapMaybe (fmap (first3 . toGregorian)) [ma,mb]
 spansSpan :: [DateSpan] -> DateSpan
 spansSpan spans = DateSpan (spanStart =<< headMay spans) (spanEnd =<< lastMay spans)
 
--- | Split a DateSpan into consecutive whole spans of the specified interval
--- which fully encompass the original span (and a little more when necessary).
+-- | Split a DateSpan into consecutive spans of the specified Interval.
+-- If the first argument is true and the interval is Weeks, Months, Quarters or Years,
+-- the start date will be adjusted backward if needed to nearest natural interval boundary
+-- (a monday, first of month, first of quarter or first of year).
 -- If no interval is specified, the original span is returned.
 -- If the original span is the null date span, ie unbounded, the null date span is returned.
 -- If the original span is empty, eg if the end date is <= the start date, no spans are returned.
 --
---
 -- ==== Examples:
--- >>> let t i y1 m1 d1 y2 m2 d2 = splitSpan i $ DateSpan (Just $ fromGregorian y1 m1 d1) (Just $ fromGregorian y2 m2 d2)
+-- >>> let t i y1 m1 d1 y2 m2 d2 = splitSpan True i $ DateSpan (Just $ fromGregorian y1 m1 d1) (Just $ fromGregorian y2 m2 d2)
 -- >>> t NoInterval 2008 01 01 2009 01 01
 -- [DateSpan 2008]
 -- >>> t (Quarters 1) 2008 01 01 2009 01 01
 -- [DateSpan 2008Q1,DateSpan 2008Q2,DateSpan 2008Q3,DateSpan 2008Q4]
--- >>> splitSpan (Quarters 1) nulldatespan
+-- >>> splitSpan True (Quarters 1) nulldatespan
 -- [DateSpan ..]
 -- >>> t (Days 1) 2008 01 01 2008 01 01  -- an empty datespan
 -- []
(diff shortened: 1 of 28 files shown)


simonmichael:
I have updated this to also allow all report periods to start on flexible dates. Note the commit message:

Eg, -p 'monthly from 1/15' or -M -b 1/15 now works,
instead of giving an error message
or requiring the special -p 'every 15th day of month from 1/15' form.
Report start date is now adjusted to an interval boundary only if
it has not been specified explicitly.
This keeps period expressions consistent between report periods (-p)
and periodic transactions (~).

`in` period expressions, like 'in 2023-01', are a grey area; they
actually specify the start date explicitly, although they may look a
bit implicit. Now, `weekly in 2023-01` will start weeks exactly on
2023-01-01 (a sunday, also causing ugly verbose column headings),
whereas previously it would adjust the start date to the previous
monday (2022-12-26).

simonmichael:
The last item mentioned above is blocking this. In short: 

- removing start dates' "snapping" behaviour means you have to specify them more carefully to get the same reports as before
- and for weekly reports in particular, you might have specify a monday date, which is inconvenient (it's not easy to mentally adjust a date to the nearest monday).

Questions:

- How to reduce the inconvenience/disruption ? 
- What's the most convenient and intuitive way to re-enable date snapping when it's wanted ?

simonmichael:
I lean towards: "All things considered, the most intuitive UX is that partial/implicit dates will be snapped when appropriate, fully-specified dates will not. Eg, these will snap to a natural start date:

weekly from YYYY-MM (snaps to the monday on or before YYYY-MM-01)
weekly from YYYY
weekly in YYYY-MM
weekly in YYYY
every 7 days from YYYY-MM
-W -b YYYY
quarterly from 2023-02 (snaps to 2023-01-01)

but these will not:

weekly from YYYY-MM-DD
every 7 days from YYYY-MM-DD
-W -b YYYY-MM-DD
quarterly from 2023-02-01

simonmichael:
There's a danger of overcomplicating. So here's a draft showing how the user doc could look:

## Date snapping
To help produce clean reports, when a weekly, monthly, quarterly or yearly report interval is used, report start dates which are partially specified (like YYYY-MM) or inferred (eg from the journal) will "snap" to the previous natural interval boundary. And report end dates will snap to make the last interval the same length as the others. But when you don't want this snapping behaviour, specify start/end dates exactly, including the day. 

Examples:

| | starts | ends |
|-|-|-|
| `weekly from YYYY-MM` | the monday on or before YYYY-MM-01 | the sunday on or after journal end date |
| `-W -b YYYY-MM` | as above | as above |
| `weekly` | the monday on or before journal start date | as above
| `weekly in YYYY` | the monday on or before YYYY-01-01 | the sunday on or after YYYY-12-31
| `quarterly from 2022-02`  | 2022-01-01, 2022-04-01, ... | a whole number of 3-month periods since 2022-01-01, on or after journal end date
| `-W -b YYYY-MM-DD` | YYYY-MM-DD (no snapping) | a whole number of weeks since YYYY-MM-DD, on or after journal end date |
| `quarterly from 2022-02-15 to today` | 2022-02-15, 2022-05-15, ... (no snapping) | today's date (no snapping; the last period will probably be shorter than the rest)

Periodic transaction rules with these report intervals behave similarly: their start/end dates will be snapped to week/month/quarter/year boundaries unless you specify an exact day.

simonmichael:
This PR is ready for review/testing. It was a big yak shave to allow periodic transactions to start on any date, but can be summed up by: 

- partial or inferred dates are flexible, and in periodic reports or periodic transactions will be adjusted automatically to period boundaries (traditional hledger behaviour)
- fully-specified dates are exact and not adjusted (new in hledger 1.29)

Here's the new doc:

## Date adjustment

With a report interval (other than daily), report start / end dates which
have not been specified explicitly and in full (eg not `-b 2023-01-01`, 
but `-b 2023-01` or `-b 2023` or unspecified) are considered flexible:

- A flexible start date will be automatically adjusted earlier if needed to
  fall on a natural interval boundary.
- Similarly, a flexible end date will be adjusted later if needed
  to make the last period a whole interval (the same length as the others).

This is convenient for producing clean periodic reports (this is traditional hledger behaviour).
By contrast, fully-specified exact dates will not be adjusted (this is new in hledger 1.29).

An example: with a journal whose first date is 2023-01-10 and last date is 2023-03-20:

- `hledger bal -M -b 2023/1/15 -e 2023/3/10`\
  The report periods will begin on the 15th day of each month, starting from 2023-01-15,
  and the last period's last day will be 2023-03-09.
  (Exact start and end dates, neither is adjusted.)

- `hledger bal -M -b 2023-01 -e 2023-04` or  `hledger bal -M`\
  The report periods will begin on the 1st of each month, starting from 2023-01-01,
  and the last period's last day will be 2023-03-31.
  (Flexible start and end dates, both are adjusted.)

simonmichael:
User impact of this change is still TBD. Eg, users upgrading past 1.29 might notice:

- starting periodic transactions on arbitrary dates is now easy; no more "Unable to generate transactions ... because not a first day of ..." errors.
- periodic reports might now start/end on different dates (more precisely what you wrote in the command line) than before. To restore the old behaviour, it might be necessary to write less precise dates (like changing `-b 2023-01-05` to `-b 2023-01`).

simonmichael:
Merging optimistically to reduce backlog, feedback and changes still welcome.

PR #1989, opened 2023-01-23
Boolean Queries

chrislemaire:
This PR attempts to resolve issue #203.

This enables queries like:
```shell
hledger -f - print "(NOT tag:'transactiontag=B' OR desc:4) AND tag:cooltag"
```

## Compatibility
I tried to implement the solution conservatively, hopefully breaking as little workflows as possible.
The changes needed for fixing the tests seemed minimal enough that this mostly worked.
The only change to old queries that needed to be made was that queries like `acct:'a a'` would be parsed as two queries. The workaround in my PR would be to write `"acct:'a a'"` instead.

In other words, all old queries without spaces in the regex should work exactly the same as they did before.
Queries with spaces might need additional escaping.

## TODO
I noticed the `words''` function is used in a few other places than just in `Query.hs`.
I didn't work out the exact need for the `words''` function, but it should probably be checked what it is used for, as the way queries can be written is quite different now.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -8,6 +8,7 @@ transactions..)  by various criteria, and a SimpleTextParser for query expressio
 {-# LANGUAGE FlexibleContexts   #-}
 {-# LANGUAGE OverloadedStrings  #-}
 {-# LANGUAGE ViewPatterns       #-}
+{-# LANGUAGE TupleSections      #-}
 
 module Hledger.Query (
   -- * Query and QueryOpt
@@ -19,7 +20,7 @@ module Hledger.Query (
   generatedTransactionTag,
   -- * parsing
   parseQuery,
-  parseQueryList,
+  parseQueries,
   parseQueryTerm,
   parseAccountType,
   -- * modifying
@@ -82,8 +83,9 @@ import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Time.Calendar (Day, fromGregorian )
 import Safe (readDef, readMay, maximumByMay, maximumMay, minimumMay)
-import Text.Megaparsec (between, noneOf, sepBy)
+import Text.Megaparsec (between, noneOf, sepBy, try, (<?>), notFollowedBy)
 import Text.Megaparsec.Char (char, string)
+
 
 import Hledger.Utils hiding (words')
 import Hledger.Data.Types
@@ -154,23 +156,13 @@ data QueryOpt = QueryOptInAcctOnly AccountName  -- ^ show an account register fo
 -- showAccountMatcher _ = Nothing
 
 
--- | A version of parseQueryList which acts on a single Text of
--- space-separated terms.
+-- | Parses a query from the string containing a query expression.
+-- Parts of the query expression are either (sub-)queries or query options.
 --
 -- The usual shell quoting rules are assumed. When a pattern contains
 -- whitespace, it (or the whole term including prefix) should be enclosed
--- in single or double quotes.
---
--- >>> parseQuery nulldate "expenses:dining out"
--- Right (Or [Acct (RegexpCI "expenses:dining"),Acct (RegexpCI "out")],[])
---
--- >>> parseQuery nulldate "\"expenses:dining out\""
--- Right (Acct (RegexpCI "expenses:dining out"),[])
-parseQuery :: Day -> T.Text -> Either String (Query,[QueryOpt])
-parseQuery d = parseQueryList d . words'' queryprefixes
-
--- | Convert a list of query expression containing to a query and zero
--- or more query options; or return an error message if query parsing fails.
+-- in ESCAPED single or double quotes or the whole term should be between
+-- parentheses to denotate a subquery.
 --
 -- A query term is either:
 --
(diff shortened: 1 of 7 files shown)


chrislemaire:
I updated a part of the documentation talking about combining queries, but it seems like the body of text on this subject is a little small. Could it be I'm missing some sources of documentation that I should be adjusting accordingly? Is there information on queries in the man pages I should change? If so, which file(s) should still be changed for that?

simonmichael:
@chrislemaire it looks like you updated the right source file for the hledger manual (hledger.m4.md). That plus some m4 macros and a pandoc template produces the man pages, via `Shake manuals`. Perhaps related: https://hledger.org/DOCS.html . Does that help ?

hledger/hledger.m4.md
```diff
@@ -4615,6 +4615,15 @@ These are most often [account name](#account-names) substrings:
...
 
+When quotes are used to escape spaces and/or special characters, the entire query should be enclosed with quotes as well:
+
+   `"'^expenses\b' 'accounts (payable|receivable)\""`
+
+Note that we use double quotes to escape the query and single quotes to escape the individual parts of the query.
+Escaping parts of the query with double quotes would also be possible, but those quotes would need to be escaped to be part of the query:
+
+   `"'^expenses\b' \"accounts (payable|receivable)\""`
+
```
simonmichael, line 4626: I appreciate the attention to easy migration.

Am I understanding rightly this would break some existing reports involving multi-word queries ? Eg a report like `hledger bal 'my bank'` would now instead be equivalent to our current `hledger bal my bank`, possibly showing different results ?

simonmichael: I am frankly scared of complicating quoting, escaping and parsing of the command line any further. I feel we are already pushing the limit, as you can maybe tell from `words'` and `words''` and https://hledger.org/dev/hledger.html#special-characters. If we must break current command line semantics anyway - not saying we will, but if we do - perhaps we should just accept the boolean query terms as arguments without the extra level of quoting ?

nobodyinperson: Awesome @chrislemaire that you're working ln this, thank you! 🎉 

I also find it problematic that queries with accounts that contain sapaces will be broken with this.

How about introducing a new directive for boolean queries (e.g. `query:`) that is parsed normally but then adds to the filtering? With the new proposed expressions one has to use quotes anyway, so just putting a `query:` in front of the potentially elaborate expressions doesn't hurt, but maintains 100% backwards compatibility.

simonmichael:
Nice idea I think, also for your consideration `expr:'foo and bar..'`

We could do something compatible like this to start, and plan a hledger 2.0 with more disruptive changes ?

hledger/test/cli/query-args.test
```diff
@@ -4,6 +4,10 @@
   a a  1
   b
 
+2010/3/1 y
+  a    1
+  b
+
```
chrislemaire, line 10: Added some parts input to this as I noticed this test would not fail on some faulty implementations where it should have. The faulty implementation would interpret this query as `Or [Acct $ toRegexCI' "a", Acct $ toRegexCI' "a"]`, where it should have been `Acct $ toRegexCI' "a a"`. The test did not catch this as 'a a' is indeed matched by a regex 'a', so I added this extra transaction.

hledger-lib/Hledger/Query.hs
```diff
@@ -270,41 +278,109 @@ defaultprefix = "acct"
 
 -- | Parse a single query term as either a query or a query option,
 -- or return an error message if parsing fails.
-parseQueryTerm :: Day -> T.Text -> Either String (Either Query QueryOpt)
-parseQueryTerm _ (T.stripPrefix "inacctonly:" -> Just s) = Right $ Right $ QueryOptInAcctOnly s
-parseQueryTerm _ (T.stripPrefix "inacct:" -> Just s) = Right $ Right $ QueryOptInAcct s
+parseQueryTerm :: Day -> T.Text -> Either String (Query, [QueryOpt])
+parseQueryTerm _ (T.stripPrefix "inacctonly:" -> Just s) = Right (Any, [QueryOptInAcctOnly s])
+parseQueryTerm _ (T.stripPrefix "inacct:" -> Just s) = Right (Any, [QueryOptInAcct s])
```
chrislemaire, line 283: The diff looks a bit messy here so I will add a short explanation of what's changed in most of these lines.

I changed the type of `parseQueryTerm` to `(Query, [QueryOpt])` as "expr:" prefixed queries can contain multiple query options and queries that are combined to a single query. Hence it was no longer possible to return a single query or a single query option. I believe this plays reasonably nice with other components, as there were very little direct invocations of this function.

The only 'functionality' that is lost by this change as far as I saw is the very specific behaviour for `not:option`, where 'option' is some query option. This no longer respects that the "not:" prefix ignores query option queries. Instead, it always negates a query. But I believe it would be a bit silly to write a query that is just `not:option`, right?

hledger/hledger.m4.md
chrislemaire: I see. I do agree the many quotes would be a bit overwhelming. Some extra quoting will still be required with `expr:`, but at least the rules for that will be consistent with the existing quoting rules!

I changed the PR so that boolean queries need to be prefixed with `expr:` as Simon suggested. The test now passes without having to change it, so that resolves our quoting issue.

chrislemaire:
> Another idea: `q:'QUERY EXPR'` ?

Whichever you like most! :)
Or maybe both? It's not too hard to swap prefix names later in the reviewing process.

simonmichael:
I tested briefly:

- AND/OR/NOT must be capitalised currently; I think case insensitive would be good
- People often want to match on two accounts, eg to see food expenses paid with cash, but they want to do it with `register`. This PR won't help with that, because register is a posting report, and report items only involve a single account.
- `aregister` is a transaction report, like `print`, so it should allow that. You can't really use this PR's syntax with areg (or you don't need to), because areg has a special first argument. But I thought this should have worked: `areg cash food`, and it doesn't, which needs investigation (unrelated to this PR).

simonmichael:
I did confirm that `hledger print expr:'cash AND food'` works equivalently to (or perhaps better than) our standard workarounds `hledger print cash not:not:food` and `hledger print cash | hledger -f- -I print food`. 

I haven't thought of any more complex queries to show this feature's real power. I don't seem to often need more than what we've had till now. It could be useful for docs if we can show some useful complex queries from real usage, or some reproducible complex queries that work with examples/bcexample.hledger.

simonmichael:
I was unclear; those are just general musings about experiences that users can have when trying this new feature, not requiring any action on your part. But I'm trying to clarify for myself some use cases/examples that do exercise this feature , so I can test it more, and also understand better its impact for users.

simonmichael:
I am confused by the merge conflict, which seems to show your code based on parseQueryTerm returning a single Either (rather than Either .. (Either ..)), which was changed several years ago. Perhaps you could rebase it when you get a chance @chrislemaire

chrislemaire:
> I am confused by the merge conflict, which seems to show your code based on parseQueryTerm returning a single Either (rather than Either .. (Either ..)), which was changed several years ago. Perhaps you could rebase it when you get a chance @chrislemaire

Looks like the merge conflict was just about [these lines](https://github.com/simonmichael/hledger/pull/1989/files#diff-b9c50261875d42d10c15d4b2f3327b50f59da86ca778c01565ee27bdfbe29072R974-R976), git was unable to merge the added `Flex` and `Exact` constructors in.

PR #1992, opened 2023-01-25
infer costs from equity with missing amounts; ignore redundant costs

simonmichael:
- lib: add journalMarkRedundantCosts to help with balancing; slight refactoring
- imp: journal: ignore redundant costs; --infer-costs more robustly
- ;doc: journal: redundant costs
- dev: transactionInferCostsFromEquity: clarify partitionPs

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -27,6 +27,7 @@ module Hledger.Data.Journal (
   journalToCost,
   journalAddInferredEquityPostings,
   journalInferCostsFromEquity,
+  journalMarkRedundantCosts,
   journalReverse,
   journalSetLastReadTime,
   journalRenumberAccountDeclarations,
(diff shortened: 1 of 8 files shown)


simonmichael:
Here are improvements arising from a recent mail list thread:

- --infer-costs now works in transactions with an implicit amount
(inferring costs from equity now happens after transaction balancing,
not before).

- As an indirect consequence, uou can now write both @/@@ costs and corresponding equity conversion postings
in a transaction at any time, not just when using --infer-costs or --infer-equity.
hledger will recognise the redundancy and ignore it.

- One thing has become more strict: hledger now requires conversion postings
to occur in adjacent pairs; an odd number of them is not allowed.
(Conversion postings are postings to accounts of type `V`/`Conversion`,
or named `equity:conversion`, `equity:trade`, `equity:trading`,
or subaccounts of these.)

simonmichael:
A modification suggested by NobbZ: we will enforce conversion postings occurring in adjacent pairs only in strict mode.

PR #1997, opened 2023-02-01
Removed a redundant paragraph in the documentation.

jbrains:
I found two nearly-identical copies of a paragraph in the documentation, so I propose to remove one of them. I chose the one to remove somewhat arbitrarily.

The copy of the removed lines can be found a few lines under the removed lines---two paragraphs after it.

### DIFF
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -4780,11 +4780,6 @@ Some queries can also be expressed as command-line options:
 When you mix command options and query arguments, 
 generally the resulting query is their intersection.
 
-## Queries and account aliases
-
-When account names are [rewritten](#account-aliases) with `--alias` or `alias`,
-`acct:` will match either the old or the new account name.
-
 ## Queries and valuation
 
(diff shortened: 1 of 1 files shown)


simonmichael:
Thank you @jbrains!

If you have time, might you force-push this again with a `;doc: ` prefix added to the message ? If not don't worry about it, I'll amend it in a bit.

(Background: we use commit message prefixes described at https://hledger.org/COMMITS.html ; the PR template probably needs to state this more clearly; our CI is supposed to warn about this; but it won't for your PR because it's unreliable (not always able to detect the commit range, https://github.com/simonmichael/hledger/actions/runs/4068803589/jobs/7009066690 -> Check commit messages); instead it will cause me trouble in future pushes.)

simonmichael:
And the semicolon is just a hint to do a briefer CI, which saves time (and energy).

simonmichael:
Never mind, I did the git twiddling. Merged as c7eb9aa94, thank you!

PR #2000, opened 2023-02-08
Aregister: implement HTML output formatter

jmtd:
This is a first attempt at adding a HTML output formatter for `areg`.

A checklist of things that need to happen (please edit+add if you want):

 - [ ] @jmtd read https://hledger.org/PULLREQUESTS.html again
 - [ ] @jmtd read https://hledger.org/COMMITS.html again
 - [ ] @jmtd write docs.
 - [ ] @jmtd write tests.
 - [ ] @jmtd prefix-colon the commit appropriately

I think, also, amounts with more than one unit type should be on separate rows, and we should rowspan appropriately since we can do that in HTML. e.g. for `examples/bitcoin-split-coinbase.journal`,the first entry would generate one row, but the second (which has two commodities, BTC and BCH) should result in two rows, with the date, description, otheramounts columns row-spanning the two.


simonmichael:
Great! I support getting HTML support merged first and exploring layout changes in the next PR.

jmtd:
I tentatively suggest this is ready for review… I have not written any tests, but none seem to exist for `aregister` already, at least inside the .hs. I ran the whole package test suite locally and it passed, but  I don't know what coverage (if any) that has over `areg`. If you suggest I should look further at tests, I will.

simonmichael:
Good enough! Thank you very much @jmtd.

PR #2027, opened 2023-04-28
imp: generate auto postings on forecast transactions by default

simonmichael:
I was thinking that auto posting rules should be applied to --forecast transactions (only) by default. In my usage, they are best used for generating new entries, not for modifying existing entries which ideally are complete and finalised. But that is still accessible via --auto as before.

I suppose this is a breaking change: if you were generating forecast txns, and you have applicable auto posting rules, the auto postings will now be added to your forecast txns and there's no way to turn that off except by changing/removing those rules, perhaps moving them to a separate file selected with -f.

This also moves Auto Postings out of the Other Syntax sin bin, making it a first class feature again.


simonmichael:
This was merged as 2b5194238.

simonmichael:
And reverted again. I found at least one user for whom this would be a breaking change, and it doesn't seem worth it. Auto postings remains a first class feature though.

PR #2030, opened 2023-05-03
timedot: make one multi-posting transaction per date line (#1754)

simonmichael:
An implementation of alternative 2 at https://github.com/simonmichael/hledger/issues/1754#issuecomment-1532424480, ie generating one transaction per date line instead of one per account/amount line.

### DIFF
--- a/examples/sample.timedot
+++ b/examples/sample.timedot
@@ -1,22 +1,16 @@
-2016/2/1
-fos:haskell   ....
-biz:research  .
-inc:client1   .... .... .... .... .... ....
+# sample.timedot
+# This is a comment line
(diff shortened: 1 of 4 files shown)


simonmichael:
Value proposition:
- creates more compact and human-readable journal entries
- resolves complexity with supporting transaction description and transaction and posting comments
- makes timedot format more isomorphic to journal format

Impact:
- a breaking change; print and register reports will change. Balance reports will not.
- description, payee, note, transaction and posting comments and tags will now work as in journal format

simonmichael:
Merged as bbecd611f

PR #2050, opened 2023-06-09
feat: cli: Support colon-delimited --pivot

glguy:


simonmichael:
Nice! A test in pivot.test and explanation in hledger.m4.md # Pivoting will be good.

PR #2070, opened 2023-08-01
;doc:roi: mention annualization differences, clean up TWR section

adept:
Clean up docs a bit after PR #2069


simonmichael:
I missed this. Thanks!

PR #2075, opened 2023-08-24
imp:print: zero posting amounts are now shown with commodity & style

simonmichael:
I noticed that `print` strips the commodity symbol from zero posting amounts, transforming eg `$0` to `0`. This can be awkward when trying to clean up/regenerate a journal, eg. This PR makes print (and print-like) reports preserve the commodity of zeroes more carefully. Other reports like balance and register are expected to work as before, but any additional testing is welcome.

```
print now shows zero posting amounts with their original commodity
symbol and the corresponding style (instead of stripping the symbol).

If an inferred amount has multiple zeroes in different commodities,
a posting is displayed for each of these.

Possible breaking changes:

showMixedAmountLinesB, showAmountB, showAmountPrice now preserve
commodityful zeroes when rendering. This is intended to improve print output,
but it seems possible it might also affect balance and register reports,
though our tests show no change in those.
```

Example:
```
# 1. print preserves the commodity symbol of zero amounts.
2023-01-01
  (a)    0 A @ 0 B = 0 A @ 0 B

# 2. The inferred balancing amount for zeros in multiple commodities
# is preserved and shown accurately, with a posting for each commodity.
2023-01-01
  a   0 A
  b   0 B
  z
```
```
$ hledger print -x
2023-01-01
    (a)       0 A @ 0 B = 0 A @ 0 B

2023-01-01
    a             0 A
    b             0 B
    z             0 A
    z             0 B

```

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -911,11 +911,18 @@ showMixedAmountOneLineB opts@AmountDisplayOpts{displayMaxWidth=mmax,displayMinWi
     -- Add the elision strings (if any) to each amount
     withElided = zipWith (\n2 amt -> (amt, elisionDisplay Nothing (wbWidth sep) n2 amt)) [n-1,n-2..0]
 
+-- Get a mixed amount's component amounts with a bit of cleanup (like @amounts@),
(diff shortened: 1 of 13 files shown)


simonmichael:
Merging this to get more testing.

PR #2076, opened 2023-08-25
dev: make groupByDateSpan return data when first DateSpan's begin is Nothing

jneubrand:
There doesn't seem to be any report code that would hit this bug—just noticed it while working on a script :)

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -317,7 +317,7 @@ groupByDateSpan showempty date colspans =
     groupByCols (c:cs) ps = (c, map snd matches) : groupByCols cs later
       where (matches, later) = span ((spanEnd c >) . Just . fst) ps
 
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks!

Is it a bug ? I can't quite tell. If colspans is empty, there's nothing to dropWhile anyway, so the true or false fallback has no effect. It seems like this code could be simpler.

jneubrand:
Well, it's a bug in that the code does something unintuitive _but only in a path that isn't hit by any code that can be called via the CLI_ (because you can't provide custom spans to any function that uses `groupByDateSpan`.)

The fallback doesn't apply when colspans is an empty list, but rather when the first span has no begin date. So when passing colspans `[(DateSpan ..2022-12-31,[]),(DateSpan 2023,[]))]` to `groupByDateSpan`, *all* postings are discarded instead of *none*.

simonmichael:
I believe you. Thank you!

Issue #1956, opened 2022-12-09
--median flag as alternative to --average


jeremy-j-ackso:
The benefit of having `--median` available is that it will lessen or eliminate the impact to averages of rare but extreme values. Suppose I'm using my monthly historical data to come up with typical/baseline monthly budgets for the coming year, if I have an extreme value in a category due to a single very large purchase among my typical purchases then my averages are going to be skewed higher.

For instance, suppose I average $400/month in groceries, however once per year I also purchase a CSA Farm Share that I pay for completely in one month, and costs $400 on its own. This is actually happening for me this month as I'm purchasing my farm share for the 2023 season, and the actual numbers are not far off from the example. That means I have one month where I have double the grocery expense, which skews the average up to $433.33. However the median is still $400.

```
mean(400, 399, 401, 398, 402, 397, 403, 396, 404, 395, 405, 800) = 433.33
median(400, 399, 401, 398, 402, 397, 403, 396, 404, 395, 405, 800) = 400.50
```

The knowledge that there's such a difference between the mean average and the median is useful to know, and after some investigating I'd likely come to the conclusion that $400 is my typical grocery bill I should be budgeting for monthly, not $433. `--median` would help me get to that decision faster.

PR #2086, opened 2023-09-13
feat: balance: Add only-summary flag (#1012)

### LINKED ISSUE #1012: A wish for an "average-only" report
There are a (admittedly few) times when I want to see only the average balance of an account over some interval. For example when estimating a monthly budget amount for expenses, or bringing printed data to show a financial advisor, or just when the width of a report is unwieldy.

One possible implementation could be to add an `--average-only` option. This option would only make sense when using one of the interval options -D, -W, -M, -Q, or -Y

Output might look something like this:

```
$ hledger balance expenses:utilities -M --average-only --flat --drop 2

Average Monthly Balance yyyy/mm/dd-yyyy/mm/dd

            ||     Average
============++=============
 electric   ||     $128.63
 naturalgas ||      $55.88
 phone      ||      $78.59
 trash      ||      $11.72
 water      ||      $63.33
------------++-------------
            ||     $338.15

```

Xitian9:
Fixes #1012.

Add a flag --only-summary for multi-column balance reports, which only displays the summary columns (--row-total, --average). This is useful when there are many columns (a weekly summary over many years) where you're only interested in the average (or some other summary).

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -156,6 +156,7 @@ data ReportOpts = ReportOpts {
     ,declared_         :: Bool  -- ^ Include accounts declared but not yet posted to ?
     ,row_total_        :: Bool
     ,no_total_         :: Bool
+    ,summary_only_     :: Bool
     ,show_costs_       :: Bool  -- ^ Show costs for reports which normally don't show them ?
     ,sort_amount_      :: Bool
     ,percent_          :: Bool
@@ -206,6 +207,7 @@ defreportopts = ReportOpts
     , declared_         = False
     , row_total_        = False
     , no_total_         = False
+    , summary_only_     = False
     , show_costs_       = False
     , sort_amount_      = False
     , percent_          = False
@@ -259,6 +261,7 @@ rawOptsToReportOpts d rawopts =
           ,declared_         = boolopt "declared" rawopts
           ,row_total_        = boolopt "row-total" rawopts
           ,no_total_         = boolopt "no-total" rawopts
+          ,summary_only_     = boolopt "summary-only" rawopts
           ,show_costs_       = boolopt "show-costs" rawopts
           ,sort_amount_      = boolopt "sort-amount" rawopts
           ,percent_          = boolopt "percent" rawopts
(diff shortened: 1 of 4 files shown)


Xitian9:
Alternate names for the flag: `--only-summaries`, `--summaries-only`, `--summary-only`.

simonmichael:
Yay, @Xitian9 !

This looks great, thank you. `--only-summary` without `-T` or `-A` shows an empty report. I'm ok with this; other choices could be to show the normal report or to give an error saying it should be used with -T or -A.

simonmichael:
I think it's fine, and less work if other summary columns are added later. Thanks!

Xitian9:
Alternatively, `--only-summary` could imply `-T` and `-A`. I think this would be reasonable at this stage, but if more summary columns (e.g. #1956) are added this could get unwieldy.

simonmichael:
Alright, and in that case, should we call it just --summary ? It's a little bit of a stretch but I think this will be a useful flag and typing two words will feel tedious.

Xitian9:
Maybe a useful exercise is to determine which additional summaries might be useful. If there are too many, we don't want to be turning all of them on. We've already had a request for median, but others we might want to consider are here:  https://hackage.haskell.org/package/foldl-1.4.15/docs/Control-Foldl.html#g:3

Here are some we might consider:
- [X] `sum` (already implemented as `-T`)
- [X] `mean` (already implemented as `-A`)
- [] `median` (already requested in #1956)
- [] other useful percentiles?
- [] `variance`
- [] `std` (population standard deviation)
- [] `maximum`
- [] `minimum`

My feeling is that `median`, `std`, `maximum`, `minimum`, and possibly the 25th and 75th percentiles are conceivably useful. If they are all implemented that would be 8 summary columns, which is probably excessive. A smaller collection would be just total, average, plus median, 25th and 75th percentiles, which is 5 columns.

I'm happy to get a steer on what is in-scope here. Once we know what ‘all summaries’ will look like, we can determine if it's reasonable to turn them all on.

simonmichael:
My inclination for now:

- call it --summary-only (users can abbreviate to --summary if they want)

- don't have it turn anything on. The current behaviour (independent of -T and -A) is easy to understand/explain, and showing just the account names is also a kind of summary.

Xitian9:
Updates made!

simonmichael:
Sorry for the delay. Thank you!

Issue #2009, opened 2023-03-05
CSV import: support interpolating regex match groups into field templates


jmtd:
In CSV import rules, I want to incorporate portions of a CSV field into the hledger field that I am populating with the rules. Here are two supporting use-cases (interestingly these are both genuinely use-cases for me, and not contrived)

1. When importing credit card statement data, I have one repayment transaction each month. The real date of this transactions tends to be somewhere towards the middle of the date range of the statement, but on the human-readable statement, it is moved to the beginning of the period, cancelling out the carried-over balance, such that this month's transactions are against a zero-balance. I wish the same for my hledger data. This is achievable by hand with a comment tag, e.g. `some:account  £ X ; date:YYYY-MM-01`, but in order to do this automatically at import-time, I need to extract a sub-set of the CSV record's date field in a field match directive,  and reference that in the field assignment directive.

2. For some of my CSV imports, I wish to derive the hledger account for a posting from a field in the CSV. But I don't want the whole field, I only want a portion of it. ([This is actually for importing data from a separate ledger. The full worked use-case is described here](https://jmtd.net/log/hledger/separate/)).

In both cases, the solution is for *match groups* from the regular expression in a field matcher directive to be made available to be interpolated into the field assignment directive. For example (addressing use-case 1): 

```
if %date (....-..)-..
    comment2 date:\1-01
```

And (addressing use-case 2):

```
if %account1 liabilities:jon:(.*)
    account1 \1
```

Implementation note: interestingly part of the work required is already there: the field matcher regular expression already treats bracket-characters as denoting match groups (and not as literal brackets. I'm not sure how you would match a literal bracket, I haven't tested whether backspace-escapes work for example). I.e., `if (....)-(..)-(..)` matches `2022-12-31`.

jmtd:
By coincidence, another problem I'm working on (providing different 'views' of a common set of transactions; handling when I owe or are owed from different pots e.g. personal family) would benefit from regex match tokens (method 2, above). I'm returning to my hacking

alerque:
Yo thanks for this @jmtd, this is going to help cleanup some of my imports *a lot*!

alerque:
Also it seems this covers at least of most of what I was requesting in #1061.

PR #2087, opened 2023-09-14
feat: import: interpolate regex matches in field templates (#2009) 

jmtd:
Extend CSV rules handling to support interpolating regular expression match groups into field assignments.

For example. Given a conditional block

    if %date (....-..)-..
        comment1 date:\1-01

The regular expression includes a region denoted by parenthesis which describe a match group. If the regular expression matches, the field assignment has '\1' substituted for the first match group value from applying the regular expression.

In this example, a date-warping comment is generated, setting the posting's date to the first of the month.

Another example

    if %account1 liabilities:jon:(.*)
        account1 \1

Here we use the match group to throw away a prefix on account names.

## Implementation notes

We extend `renderTemplate` to detect and process strings of the form '\N+' (isDigit N). Match groups are scoped to a conditional block that might be in terms of a `hledgerField`, therefore, `renderTemplate` grows a `hledgerField` parameter (and all callers are updated).

The new match group handling function `regexMatchValue` needs to determine the relevant Conditional Block scope using `isBlockActive`, formerly a sub-function of `getEffectiveAssignment`, now changed to a top-level function.

(The diff-view of the changes to `getEffectiveAssignment` aren't terribly legible. They might be clearer with `git diff --ignore-all-space`. Besides moving `isBlockActive` to a top-level, I also collapsed one of the nested levels of `where` clauses).

The match groups are calculated and collected using a new function `Hledger.Utils.Regex.regexMatchTextGroups`.

### Testing

I need to add testing to your particular test suite/approach. I did ad-hoc testing as I went but I haven't grokked your scheme yet.

### TODO

(feel free to edit and add to this)

 - [ ] settle on the delimiter (`%matchN`,`%\1`,`\1`?)
 - [ ] unit tests
 - [x] functional tests (shelltest)
 - [x] docs
 - [ ] `examples/`?
 - [ ] can `regexMatchTextGroups` throw `error`? Is that a problem?
 - [ ] Possibly validateCsvRules could check that backreferences refer to existing match groups ?

### DIFF
--- a/hledger-lib/Hledger/Read/RulesReader.hs
+++ b/hledger-lib/Hledger/Read/RulesReader.hs
@@ -372,9 +372,11 @@ JOURNAL-PSEUDO-FIELD: amount-in | amount-out | currency
 
 ASSIGNMENT-SEPARATOR: SPACE | ( : SPACE? )
 
-FIELD-VALUE: VALUE (possibly containing CSV-FIELD-REFERENCEs)
+FIELD-VALUE: VALUE (possibly containing CSV-FIELD-REFERENCEs and REGEX-MATCHGROUP-REFERENCEs)
 
 CSV-FIELD-REFERENCE: % CSV-FIELD
+
+REGEX-MATCHGROUP-REFERENCE: \ DIGIT+
 
 CSV-FIELD: ( FIELD-NAME | FIELD-NUMBER ) (corresponding to a CSV field)
 
@@ -707,7 +709,7 @@ hledgerField = getEffectiveAssignment
 -- | Look up the final value assigned to a hledger field, with csv field
 -- references interpolated.
 hledgerFieldValue :: CsvRules -> CsvRecord -> HledgerFieldName -> Maybe Text
-hledgerFieldValue rules record = fmap (renderTemplate rules record) . hledgerField rules record
+hledgerFieldValue rules record f = (fmap (renderTemplate rules record f) . hledgerField rules record) f
 
 maybeNegate :: MatcherPrefix -> Bool -> Bool
 maybeNegate Not origbool = not origbool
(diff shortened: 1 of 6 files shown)


simonmichael:
Great!

Yes I think it would be better to return a Maybe (which will cause the reader to throw an exception in the usual way).

Possibly validateCsvRules could check that backreferences refer to existing match groups ?

We'd have to document that `matchN` may not be used for field names I think. How would `%\N` or `\N` work ?

jmtd:
> Yes I think it would be better to return a Maybe (which will cause the reader to throw an exception in the usual way).

Is this comment in relation to TDFA's `=~` throwing `error`?

I've looked a bit more closely at that. Their docs say it can do so if there is an "error in processing". Since our new function `regexMatchTextGroups` takes a `Regexp`, we can rule out throwing `error` due to a malformed expression (it would be caught when the `Regexp` was constructed, earlier). I don't know whether this is the only error case that is meant by "error in processing". TDFA has an alternative function `=~~`, which is described as "If a single match fails, then fail will be called.". A single match is not a failure condition from our POV, but we could detect and ignore that; it's not clear to me that any of other the circumstances that would cause `error` with `=~` would instead use `MonadFail` with `=~~`. To answer this conclusively, I will have to spend more time studying TDFA's source.

> Possibly validateCsvRules could check that backreferences refer to existing match groups ?

That is a good idea.

> We'd have to document that matchN may not be used for field names I think.

Ack

>  How would %\N or \N work, instead ?

I think the choice is a decision weighing the legibility of the chosen token versus the risk of clashing with its prior use. I think the risk of `%matchN` being used by someone in some rules already is higher than `%\N` or `\N`, but still fairly low. In terms of legibility, I think `%\N` or `\N` are a little more obscure, and less self-documenting (given the existing use of `%N`). Still, I'll go with your preference. If we move to a token including a backslash, I will have to check whether TDFA eats backslashes (as it already eats parentheses)

simonmichael:
It's just that I have never seen any other syntax for regexp backreferences than \N (well, occasionally $N). So \N, or %\N if we need more uniformity, seemed intuitive to me (%N is already used of course). As well as being less verbose than %matchN, it's more visually distinct from field names, indicating that it's something different. More ideas, anyone ?

simonmichael:
https://www.regular-expressions.info/refcapture.html has some more variations.

jmtd:
> It's just that I have never seen any other syntax for regexp backreferences than \N (well, occasionally $N). So \N, or %\N if we need more uniformity, seemed intuitive to me (%N is already used of course).

That makes sense! You've convinced me, my preference is `\N`, which doesn't run the risk of confusion with `%N`. It looks like TDFA does not eat backslashes.

simonmichael:
Though, \N is quite likely to appear in text that someone is trying to match. %\N would be safer in this regard, wouldn't it.

jmtd:
(It seems my last attempt at a comment on this issue was lost, sorry!)

Sorry for the hiatus on this. I'm returning to it now. I'm side-stepping the issue of exactly what token to use, and exploring the other necessary elements (docs, tests) for a bit.

jmtd:
Hi @simonmichael, I think this is substantially ready and I'd appreciate it if you could take another look!

With respect to the delimiter: `\` is it is consistent with the existing use for back-references in account aliases. `\N` appearing in text someone is trying to match is not a concern, because we only use this token in the field assignment value, not in the match regex. I think the risk of someone wishing to assign text containing `\` to a value is vanishingly small. So, I still think that's the best delimiter choice; but ultimately I'm not wedded to any given choice, I just want the feature :)

simonmichael:
Excellent. It's looking great! Would it make sense to support \0, which usually refers to the whole regexp's match ? If that's out of scope, it's also fine.

jmtd:
> Would it make sense to support \0, which usually refers to the whole regexp's match ? If that's out of scope, it's also fine.

We could do that, if it would be useful. TDFA does not do that, so we'd be synthesizing it in either `Utils.Regex.regexMatchTextGroups` or `hledger-lib/Hledger/Read/RulesReader.hs`. I'd prefer to do that in a follow-on PR, would that work for you?

simonmichael:
There's more to this than meets the eye, and it will be good to expand the docs and tests for this.

What's intended when there are multiple matchers in an if, each with it's own capture groups ?
Eg in this, are the groups numbered from 1 to 4 ? 
```
if
a (.)
b (.)
c (.) (.)
  ...
```
And how are they numbererd in more complicated matchers combinations using AND/NOT ? Eg:
```
# (a and b) or (c and d) or (not e and f)
if
a (.)
& b (.)
c (.) 
& d (.)
! e (.)
&  f (.)
  ...
```

simonmichael:
Related: https://hledger.org/dev/hledger.html#how-csv-rules-are-evaluated 
> collect all field assignments at top level and in matched if blocks. When there are multiple assignments for a field, keep only the last one.

simonmichael:
Currently I think regexMatchValue is trying all of the regexps in all of the active if blocks and collecting all match groups from all of them. 

Whereas we need to do something clever inside getEffectiveAssignment - keep track of which if block it ends up picking the assignment from (if any), and collect match groups from just that.

simonmichael:
And perhaps ideally it would collect match groups just from the matcher within the if block that succeeded. But limiting them to the successful if block would be a start.

simonmichael:
For people trying to use this, I think the workaround for now is when you use capture groups, make those regexps specific enough that at most one of them will match any given CSV record.

simonmichael:
Sorry, these comments are in the context of bug #2158 which was opened against this PR. But I'll leave them here.

Issue #2054, opened 2023-06-28
csv rules: support easy negation of conditions


simonmichael:
https://hledger.org/dev/hledger.html#matchers : "There's not yet an easy syntax to negate a matcher." We should have this, for if blocks (and possibly if tables ?)

simonmichael:
Since 1.32 you can negate a matcher by prepending !, but it's unfinished, you can't yet combine this with &. See

> **[Combining matchers](https://hledger.org/dev/hledger.html#combining-matchers)**
> 
> When an if block has multiple matchers, they are combined as follows:
> 
> - By default they are OR'd (any one of them can match)
> - When a matcher is preceded by ampersand (&) it will be AND'ed with the previous matcher (both of them must match)
> - *Added in 1.32* When a matcher is preceded by an exclamation mark (!), the matcher is negated (it may not match).
> 
> [Currently](https://github.com/simonmichael/hledger/pull/2088#issuecomment-1844200398) there is a limitation: you can't use both & and ! on the same line (you can't AND a negated matcher).

PR #2088, opened 2023-09-17
feat: CSV rule negation

bobobo1618:
I'm interested in hledger but really need https://github.com/simonmichael/hledger/issues/2054 for my use-case (certain locations for the same store fit in one purchase category, other locations in another).

I don't know a whole lot of Haskell but hopefully enough to make myself dangerous. I _think_ this change works? The tests I added seem to pass?

### DIFF
--- a/hledger-lib/Hledger/Read/RulesReader.hs
+++ b/hledger-lib/Hledger/Read/RulesReader.hs
@@ -300,7 +300,7 @@ type FieldTemplate    = Text
 type DateFormat       = Text
 
 -- | A prefix for a matcher test, either & or none (implicit or).
-data MatcherPrefix = And | None
+data MatcherPrefix = And | Not | None
   deriving (Show, Eq)
 
 -- | A single test for matching a CSV record, in one way or another.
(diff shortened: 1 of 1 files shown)


simonmichael:
Sorry for the delay @bobobo1618 . A valuable feature! In my excitement I forgot to ask for documentation. Could you suggest some wording and I'll add it.

bobobo1618:
I had a look at [the existing matcher documentation](https://hledger.org/1.31/hledger.html#matchers) and would suggest "When a matcher is preceded by an exclamation mark (`!`), the matcher will be negated; only lines that _don't_ match the original matcher will be matched."

simonmichael:
Perfect. Thank you!

simonmichael:
Howdy @bobobo1618 . I tried to use `!` in CSV rules for the first time, in an AND-ed condition, like this:
```
# stripe transfers where amount is positive
if STRIPE TRANSFER
& ! %amount -
```
but couldn't make it work. Would you have time to check this case and show a working example, or add a test and fix ?

bobobo1618:
> Howdy @bobobo1618 . I tried to use `!` in CSV rules for the first time, in an AND-ed condition, like this:
> 
> ```
> # stripe transfers where amount is positive
> if STRIPE TRANSFER
> & ! %amount -
> ```
> 
> but couldn't make it work. Would you have time to check this case and show a working example, or add a test and fix ?

I'd love to but I'm afraid this is beyond my Haskell abilities. Right now the logic only supports a simple prefix but this feature would require some kind of recursive logical structure and I'm not familiar enough with how to make that work.

simonmichael:
Mentioned this limitation in the manual for now.

PR #2097, opened 2023-10-12
imp: web: include account declaration info in accounts JSON

s-zeid:
This adds the accounts' declaration info to the `/accounts` endpoint in hledger-web.  My use case is to be able to access account tags in some (unpublished as of now) client-side JS I'm working on.

This does add a new field to `accountsKV` in `Hledger.Data.Json`.  That field can have a value of `Nothing`/`null` if the declaration info hasn't been added.  (I'm doing this in the Web-side code.)

I have compared the output from before and after this commit ([script](https://gist.github.com/s-zeid/05db466eb3158522a0d5f99fdc812663)), and at least for my journal (and some random ones in `/examples/`), there were no changes other than the declaration info being added.

This is my first time working with Haskell, so I apologize if I did this the wrong way or made some other noob mistake.

### DIFF
--- a/hledger-lib/Hledger/Data/Json.hs
+++ b/hledger-lib/Hledger/Data/Json.hs
@@ -165,20 +165,21 @@ accountKV ::
 #endif
   => Account -> [kv]
 accountKV a =
-    [ "aname"        .= aname a
-    , "aebalance"    .= aebalance a
-    , "aibalance"    .= aibalance a
-    , "anumpostings" .= anumpostings a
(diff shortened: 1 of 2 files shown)


simonmichael:
Nice work! Do we need to update anything in hledger-web.m4.md ?

s-zeid:
I don't think so.  It doesn't specifically discuss the output of `/account`; it just refers readers to the Hledger.Data.Json haddock docs for details.

simonmichael:
Great, thanks!

Issue #2099, opened 2023-10-15
hledger-web cannot serve on 0.0.0.0


klockeph:
If I run `hledger-web --serve --host 0.0.0.0` on my Raspberry Pi to expose web-hledger on all interfaces (in my case WiFi and VPN), some things break:

Most notably, opening `192.168.1.<Pi>:5000` redirects to `0.0.0.0:5000/journal`.
Navigating to the journal directly does not help, since it still tries to load additional resources from `0.0.0.0` - the browser console is flooded with `Failed to load resource: net::ERR_CONNECTION_REFUSED 0.0.0.0:5000/<resource>`.

simonmichael:
Maybe also setting --base-url helps ?

klockeph:
Well yes, but then it would still only work for the particular interface that has the matching IP to the base url.

klockeph:
Coming to think of it, it probably just needs relative paths in the frontend. I'll try to see if I can create a PR.

simonmichael:
"just" you say. :-)  Great, thanks for looking at it.

PR #2100, opened 2023-10-15
fix: Use guessAppRootOr to enable relative root if desired.

klockeph:
This is useful when serving on 0.0.0.0, such that querying from any other device with <IP>:<PORT> does not fall back to 0.0.0.0:PORT, which fails. (#2099)

Tested: Manually


simonmichael:
Cool. What's a description of the user-visible difference this makes, suitable for the manual ?

On mac, with or without your fix, running hledger-web with --host 0.0.0.0 or --host localhost seems to break it for web browsers - Safari shows an empty page and Firefox refuses to browse it. (Also one has to use --port to avoid AirTunes which is using port 5000.)

klockeph:
You mean serving from Mac?
Unfortunately, I cannot test that sorry.


The user-visible change would be that if the web interface is reachable from multiple domains or IPs, they will not necessarily be redirected to the `base_url`; instead `hledger-web` will try to keep them at the root/host from the URL of the request.
Apart from this it fixes the bug where 0.0.0.0 would only work for access from localhost.

If this is not desired behavior we could maybe check if `base_url` is set and if it is, keep ApprootMaster?
I just wasn't sure how the arguments are passed and if `base_url` will always contain "IP:PORT" if unset.

simonmichael:
Ah I see - when hledger-web is accessible at multiple IPs, it will keep responding from the same IP address you visited ? That sounds good to me, I guess you have observed it fixing your issue on GNU/linux.

I don't know if this interacts with the other options, do you think there might be a problem there needing more testing ? Otherwise I'm happy to merge this. I'll call for testers on chat also.

klockeph:
Tested on Linux as well now; also works.

simonmichael:
I think we need to clarify how this interacts with --base-url. The doc for that option says

--base-url=URL : set the base url (default: http://IPADDR:PORT). Note: affects url generation but not route parsing. Can be useful if running behind a reverse web proxy that does path rewriting.

Note to self: translate this to userese.

klockeph:
We could just let `base-url` overwrite the `guessAppRootOr`. So if `base-url` is given, don't use the guess. Otherwise use guess and fall back to `IP:PORT`.
However I'm not proficient in Haskell, so if you want to go for this I'd be happy if someone else picks up the patch.

simonmichael:
I can't fully test this on mac, but I see that we print both the listening address and port and the base url and port, so it should be easy enough to troubleshoot any issues.  I will shortly merge this as 6312446cb, thank you !

PR #2101, opened 2023-10-17
fix: print: more consistent/flexible precisions

### LINKED ISSUE #2085: print needs more control over amount precisions
Since 1.31, `print` preserves original amount precisions, without standardising them. This is the right default behaviour, but I meant to allow overriding it with `-c`, and ran out of time. That is needed because otherwise there's no way to generate journal entries with consistent precisions, eg from a CSV with inconsistent decimal places.

Here's a proposal for supporting several kinds of precision cleanup with `print`:
[see #2101]

(Not a regression, since the change was documented.)

simonmichael:
1.31 removed `print`'s behaviour of adding/removing decimal zeros to match commodity precisions (#2085). This adds an option restoring that behaviour, and adds some other rounding behaviours which may be useful for journal cleanup, and fixes some inconsistencies, and makes `print` show commodity symbols and decimal digits for zeros.

This stuff may not affect everyone, but it turns out to be important (and confusing!) when you're trying to clean up or re-parse or export your data (the Plain Text part of PTA). The UX and internals went through a lot of evolutions over the past month. Here are the docs:

## Change note

- `print` now shows zeros with a commodity symbol and decimal digits when possible, preserving more information.

- `print` now formats balance assertion costs with standard commodity styles, like other amounts.

- `print` has a new option for rounding amounts to commodity display precisions:
  
  - `--round=none` - show amounts with original precisions (default; like 1.31; avoids implying less or more precision than was recorded)
  - `--round=soft` - add/remove decimal zeros in non-cost amounts (like 1.30 but also affects balance assertion amounts)
  - `--round=hard` - round non-cost amounts (can hide significant digits)
  - `--round=all`  - round all amounts and costs

  **Past behaviour**

  Comparing amount styling behaviour across hledger versions can be confusing;
  here's a summary (for my own sanity).
  
  print shows four kinds of amount: posting amounts, balance assertion amounts, and costs for each of those.

  Which amounts does print do basic styling (eg symbol placement) on ?
  
  | hledger   | amt | cost | bal | balcost |
  |-----------|-----|------|-----|---------|
  | 1.1-1.14  | Y   | N    | N   | N       |
  | 1.15-1.22 | Y   | N    | Y   | N       |
  | 1.23-1.30 | Y   | Y    | Y   | N       |
  |           |     |      |     |         |
  | 1.31-     | Y   | Y    | Y   | Y       |
  
  Which kind of rounding does print do on each amount ?
  
  | hledger             | amt  | cost | bal  | balcost |
  |---------------------|------|------|------|---------|
  | 1.0-1.20            | hard | none | none | none    |
  | 1.21-1.30           | soft | none | none | none    |
  | 1.31                | none | none | none | none    |
  |                     |      |      |      |         |
  | 1.31.1              | none | none | none | none    |
  | 1.31.1 --round=soft | soft | none | soft | none    |
  | 1.31.1 --round=hard | hard | none | hard | none    |
  | 1.31.1 --round=all  | hard | hard | hard | hard    |

## Code doc

A start at more internal codebase documentation, which I felt the need for:

=== Precision and rounding

Numeric amounts in hledger are represented by "Decimal",
plus additional information like a "CommoditySymbol"
and an "AmountStyle" for display.
Together these form hledger's "Amount" type.

In hledger docs, "precision" means the number of digits to the right of the decimal mark.
Amounts have several precisions we can talk about:

- __Journal precision__ - the number of decimal digits recorded in the journal file / input data.
  We accept up to 255 decimal digits there.

- __Internal precision__ - the number of decimal digits stored internally in each Decimal value (after calling normalizeDecimal).
  Decimal supports up to 255 decimal digits.
  In amounts parsed from the journal, this will be the same as their journal precision.
  In amounts generated from calculations, internal precision may increase, and will not decrease.

- __Display precision__ - the number of decimal digits to be displayed in output ("AmountPrecision");
  this is part of the AmountStyle stored within each Amount.
  Immediately after parsing, this is the same as the journal and internal precisions;
  later it gets normalised, for consistent appearance.
  When display precision is less than the internal precision, fewer, rounded decimal digits are displayed ("rounding").
  When display precision is greater than the internal precision, additional decimal zeros are displayed ("padding").

We use the term "display rounding" for applying a target display precision to an existing amount, 
This can be done more or less forcefully, according to a "display rounding strategy", represented by the "Rounding" type.
This is stored within each AmountStyle for convenience, though semantically speaking it is not part of the amount.
The rounding strategies are:

- none - leave the amount's display precision unchanged
- soft - add or remove trailing decimal zeros to approximate the target precision, but don't remove significant digits
- hard - use the exact target precision, possibly rounding and hiding significant digits
- all  - do hard rounding of both the main amount and its cost amount (costs are normally not display-rounded).

Here is when display rounding happens:

1. After reading a journal,
  the standard commodity styles are applied to all amounts, except for precision (no display rounding at this stage).
  (still needed ? seems so)

2. After reading a journal, when checking each transaction for balancedness,
  the transaction amounts are hard-rounded, temporarily, before calculating their sum.
  (We'd like to update this to use transaction-local precisions only.)

3. Just before rendering a report (since 1.31).
  Most reports do hard rounding; @print@ (and maybe other print-like commands ?)
  can do any of the rounding strategies.


simonmichael:
PS it's really more than a bugfix. But I was thinking it's best to put out as a 1.31.1 bugfix release so we don't have another hledger version with different print behaviour out there being packaged. I'd rather skip unnecessary release work, but maybe it's needed anyway for other fixes.

PR #2104, opened 2023-10-23
imp:web: access control UX cleanups (fix #834)

### LINKED ISSUE #834: web: Rename/alias --capabilities and --capabilities-header
Follow-up to #821. [and [issuecomment-401393071](https://github.com/simonmichael/hledger/pull/821#issuecomment-401393071)]

I don't like `capabilities` either, but it seemed like the best conceptual fit to what it does. There are two problems here - one is the naming, the other is UX.

`--capabilities` enables features of the web UI that are available always (unless overridden by `--capabilities-header`), `--capablities-header` specifies a header (`X-Sandstorm-Permissions` on Sandstorm), which contains the permissions that the current role can access. We could reuse the Sandstorm terminology and say that the web application started by `hledger web --permissions=view,add,manage` has the permission to edit the journal in its entirety. `--access-level` also seems to fit.

The UX problem is a different one. It is currently possible to start the webapp like `hledger web --capabilities=manage`, and the app would be broken (perhaps functional, but with a broken UI). Do we foresee a need for more permissions? (E.g. `manage-users`, `switch-journal`)

If yes, then there should be built-in roles for the static-permissions case (e.g. `--role admin == --permissions=view,add,manage`), so that starting the app with misconfigured permissions is not as easy. If these permissions are all that will ever be needed (if hledger-web's intent is to only be a simple web UI without too many features), then the solution could be same as above, or the permissions/access-levels could be gradual - `manage` will always imply both `add` and `view`, and `add` will imply `view` (a sort of implicit roles)

I'm unsure what the best solution here is. `--role`, `--permissions`, and `--permissions-header` uses commonly known terminology and is both accessible and extensible. Only `--permissions` and `--permissions-header` (or alternatives) without `--role` but with implicit roles is not as clearly extensible, but would be simpler.

(In writing this issue, I almost decided on `--permissions` (`-P`?), which seems to be a better alternative to `--capabilities` (which is a good conceptual fit but an uncommon term), `--access-level` (which works but is conceptually closer to *role* than to *permission*) or others.)

Any thoughts on this? I didn't expect my PR to get merged and immediately released, but I think that in any case, we need to keep `--capabilities{,-header}` at least until next release.

simonmichael:
Changes:

1. rename the sandstorm "manage" permission to "edit" (old permission names: view, add, manage;
 new permission names: view, add, edit).

Rationale: "edit" best describes this permission's current powers, to users and to operators. If we ever added more manager-type features we'd want that to be a new permission, not a rename of the existing one (which would change the powers of existing users).

2. rename the sandstorm roles for consistency with permissions (old role names: viewer, editor, manager;
 new role names: viewer, adder, editor)

Rationale: it's needed to avoid confusion.

3. add a new option: --allow=view|add|edit|sandstorm (default: add). 'sandstorm' sets permissions according to the X-Sandstorm-Permissions header. Drop the --capabilities and --capabilities-header options.

Rationale: it's simpler and more intuitive.

4. replace "capability" with "permission" in ui/docs/code.

Rationale: consistent with the above, more familiar.

[5. reverse the order of role definitions.

Rationale: "increasing power" order is consistent with the permissions and --allow definitions, hence less confusing. We thought it wouldn't change the roles' meaning, but it did, but no-one is using this yet anyway.]

### DIFF
--- a/.sandstorm/launcher.sh
+++ b/.sandstorm/launcher.sh
@@ -29,4 +29,4 @@ set -euo pipefail
 mkdir -p /var/lib/hledger
 touch /var/lib/hledger/Ledger
 cd /var
-hledger-web --capabilities-header=X-Sandstorm-Permissions --serve --base-url='' -f /var/lib/hledger/Ledger --port 8000
+hledger-web --allow=sandstorm --serve --base-url='' -f /var/lib/hledger/Ledger --port 8000
--- a/.sandstorm/sandstorm-pkgdef.capnp
+++ b/.sandstorm/sandstorm-pkgdef.capnp
@@ -54,7 +54,7 @@ const pkgdef :Spk.PackageDefinition = (
         #marketBig = (svg = embed "path/to/market-big-300x300.svg"),
       ),
 
-      website = "http://hledger.org",
+      website = "https://hledger.org",
       # This should be the app's main website url.
 
       codeUrl = "https://github.com/simonmichael/hledger",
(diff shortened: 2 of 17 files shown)


simonmichael:
Here's some cleanup for hledger-web access control and #834. I'll merge unless @zarybnicky or @ocdtrekkie see problems.

ocdtrekkie:
Is there a particular reason to re-order the roles? I don't *think* it will matter, but we should definitely test it. I will try to get an experimental build within the next day for you based on this branch.

simonmichael:
Well spotted, yes I reordered the roles to be consistent with the "increasing power" order as in permissions and --allow option; I found it easier to work on. As you said, they warn about this for permissions for not for roles, and nobody is using this yet, so I think it'll be fine.

ocdtrekkie:
It'll be nice to test the theory we can reorder them so I'm cool with it.

ocdtrekkie:
If you want to make this easily testable from your branch, you should edit .sandstorm/launcher.sh to use the new --allow option. And maybe increment the appVersion in the package definition file, though we only need to do that before we actually release the update.

ocdtrekkie:
Screenshot from 2023-10-23 23-04-51

It looks like re-ordering the roles is a destructive change. The view-only link on the old PR now grants the editor access.

simonmichael:
Thanks for testing. Stupid thing. I vote to leave it tidy, as is, since we have nobody using the old order.

ocdtrekkie:
That scares me in a "technically we have no way to know for sure" sense, but if that's how you want to proceed, can we increment the appVersion ASAP so I can push a new release as fast as possible?

simonmichael:
Pushed - thanks.

Issue #869, opened 2018-09-11
add ssv, tsv output formats


simonmichael:
Follow-on from #176. If we can read ssv and tsv, we should probably be able to output them too. I propose we add `ssv` and `tsv` output formats, similar to existing `csv`.

PR #2108, opened 2023-11-05
feat: cli: Add tsv output (#869)

psagers:
I've been writing some add-on scripts to analyze my hledger data and frequently wishing I had tsv output. csv is just complicated enough to require a bit of parsing--usually through some external dependency--whereas tsv is trivial enough to decode directly in essentially any environment. (It's also the native and strongly preferred format of [visidata](https://www.visidata.org/).)

I see there's an old issue requesting this feature (#869). There also seems to be an old pull request that circled around it a bit, but never turned into anything.

I believe the implementation is in fact pretty trivial. The data is identical to csv output, so it should just be a matter of writing `printTSV` based on `printCSV` and hooking it up.

A few notes:

- This creates a small amount of code duplication, which could be resolved if it's worthwhile.
- There was one csv test that I couldn't duplicate, as it has output lines that begin with `<`. Perhaps someone familiar with `shelltest` could suggest a workaround, if desired.
- To be safe, I'm replacing all tabs, carriage returns, and newlines in field data with spaces. I'm assuming these characters are rare to nonexistent in output data to begin with.

### DIFF
--- a/hledger-lib/Hledger/Read/CsvUtils.hs
+++ b/hledger-lib/Hledger/Read/CsvUtils.hs
@@ -13,6 +13,7 @@ CSV utilities.
 module Hledger.Read.CsvUtils (
   CSV, CsvRecord, CsvValue,
   printCSV,
+  printTSV,
   -- * Tests
   tests_CsvUtils,
 )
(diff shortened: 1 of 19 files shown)


hledger/hledger.txt
simonmichael, line 1: hledger.txt is a generated file - hledger.m4.md is the right place to edit.

simonmichael (changes requested):
Looks good !

simonmichael:
For test data lines beginning with <, maybe shelltestrunner's format 2 helps (triple
character delimiters)

simonmichael (approved):

simonmichael:
Looks good to me. I appreciate that you added the new output type to all commands. Thank you!

PR #2117, opened 2023-11-29
;doc: Update pathspec for running roi-unrealised.ledger

soccja:
roi-unrealised.ledger got moved from examples -> examples/investing. Path-spec for commands documented in roi-unrealised.ledger updated accordingly. 

Documented output from running roi-unrealised.ledger was obsolete. Refreshed.


simonmichael:
Thanks!

PR #2120, opened 2023-12-01
fix: do not show average column in cumulative balance

### LINKED ISSUE #2118: hledger balance shows average column even if cumulative/historical
```
hledger bal -Y -AT --cumulative
```

shows a yearly cumulative balance. The total and average column make no sense for a cumulative balance.

The total column is therefore omitted – but not the average column.

I would provide a PR to fix this.

There might be similar bugs in register or other reports.

schoettl:
Fixes #2118

TODO:

- [ ] document in `--cumulative`/`--historical` option description that it disables total and average column


simonmichael:
Issue was closed.

PR #2143, opened 2023-12-19
imp: Support brick 2.2

Vekhir:
`brick` 2.2 is a minor improvement over 2.1.1, so no API breakages expected. See also https://github.com/simonmichael/hledger/pull/2128#discussion_r1431765057
Builds and tests fine with hledger-ui `1.32.1`.


simonmichael:
Thank you, let's update package.yaml as well, it is the master file.

Vekhir:
`brick` is updated in `package.yaml` and `hledger-ui.cabal`

Issue #2113, opened 2023-11-16
import can't check account names / account declarations not affecting sibling files


simonmichael:
Reported in chat: `hledger import -s` does not seem to see account declarations in the main journal, and will wrongly complain that account names are not declared. Current workaround: don't use -s with import, eg you could do this instead: `hledger import ... && hledger check -s`

simonmichael:
I believe this is related:
```
commodity ""
account a
account b

2023-01-01
  (a)   1
```
```
2023-01-02
  (b)   2
```
```
$ hledger -f a.j -f b.j print
2023-01-01
    (a)               1

2023-01-02
    (b)               2

$ hledger -f a.j  -f b.j print -s
hledger: Error: /Users/simon/src/hledger/b.j:2:
  | 2023-01-02
2 |     (b)               2
  |      ^

Strict account checking is enabled, and
account "b" has not been declared.
Consider adding an account directive. Examples:

account b
account b    ; type:A  ; (L,E,R,X,C,V)

```

Ie, account directives are not affecting a sibling file. This is odd because https://hledger.org/dev/hledger.html#directive-effects says `account`'s effect does not end at file end and affects all files.

simonmichael:
And, `hledger -f a.b -f b.j check accounts` passes, but `hledger -f a.b -f b.j check accounts -s` does not.

simonmichael:
We'd like to be able to do strict account name checking on the generated transactions before proceeding with an import. Some reasons:

- you want to know when categorisation rules are incomplete and transactions involving "expenses:unknown" or "income:unknown" are generated
- you want to know when you accidentally changed the rules so as to generate undeclared account names
- you want to know when you forgot to update your csv rules after changing account names and account declarations

Currently it requires a clunky workflow like
```
$ hledger -f $LEDGER_FILE -f <(hledger import --dry-run foo.csv) check -s && hledger import foo.csv
```

PR #2155, opened 2024-01-13
;doc: add version annotations for new features

hamzashezad:
This draft PR aims to document the version a feature was introduced. As discussed in the Matrix room, `*Added in <version>*` will be prepended to;

1. heading, or if it doesn't exist
2. the description.

For v1.32,

- [x] timedot supports letters syntax (#2116)
- [x] print has `beancount` output
- [x] match groups in csv rules (#2009)
- [x] negating matchers (#2088)
- [x] tsv output support where csv is supported (#869)
- [x] print has a new option for controlling amount rounding (#2085)
- [ ] --summary-only, multi-column balance reports (#1012) (no docs present)

Dropped;
- [ ] import now works with -s/--strict (#2113) 
- [ ] https://hledger.org/1.32/hledger.html#part-3-reporting-concepts

### DIFF
--- a/hledger/Hledger/Cli/Commands/Aregister.md
+++ b/hledger/Hledger/Cli/Commands/Aregister.md
@@ -62,7 +62,7 @@ at the cost of more time and memory, use the `--align-all` flag.
 This command also supports the
 [output destination](hledger.html#output-destination) and
 [output format](hledger.html#output-format) options.
-The output formats supported are `txt`, `csv`, `tsv`, and `json`.
+The output formats supported are `txt`, `csv`, `tsv` (*Added in 1.32*), and `json`.
 
 ### aregister and posting dates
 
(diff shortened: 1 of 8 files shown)


hamzashezad:
Once a version is considered old enough, it's accompanying annotations can be removed. "Old enough" could be judged by when OS packages have caught up, e.g. nixpkgs, debian, etc. Looking at https://hledger.org/install.html#binary-packages, 1.30 is common, so maybe keep them for a fixed number of versions, say for the next two versions.

Personally, I'd keep unless they _need_ to be removed because why bother.

simonmichael:
@hamzashezad if you have time, maybe add these to the 1.32 manuals in hledger_site repo also, for https://hledger.org/1.32/hledger.html

simonmichael:
(I wonder about replacing "added in" with "since" for brevity)

hamzashezad:
@simonmichael sure, I can add to the hledger_site repo as well. Can you guide me how the files are generated? 

"Since" does sound better.

simonmichael:
1.32/* were generated by me at release time with make snapshot-1.32 but are just normal files now, you can commit to them. Thanks.

PR #2159, opened 2024-01-22
fix: Run shellcheck on hledger-bar

colindean:
I [encountered][1] an unexpected error when testing hledger-bar for
proposed inclusion in Homebrew's default installation of hledger:

    /opt/homebrew/Cellar/hledger/1.32.2_1/bin/hledger-bar: line 81:
    conditional binary operator expected

I'm doubtful that this is fixed by running shellcheck, but checking
a script against shellcheck is usually one of the first things I check
before debugging shell!

This patch is ~generated by shellcheck with

    shellcheck --shell=bash --enable=all --format=diff bin/hledger-bar | \
    git apply

[1]: https://github.com/Homebrew/homebrew-core/actions/runs/7606843601/job/20713321881?pr=160590

### DIFF
--- a/bin/hledger-bar
+++ b/bin/hledger-bar
@@ -78,7 +78,7 @@ poschar="+"
 
 # https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit and 24-bit colors.
 # Disable if stdout is not a terminal or NO_COLOR is defined.
-if [[ ! -t 1 || $NO_COLOR ]]; then
+if [[ ! -t 1 || -n ${NO_COLOR} ]]; then
     red=''
     green=''
     nocol=''
@@ -93,7 +93,7 @@ unquote() { s="${1%\"}"; echo "${s#\"}"; }
 # process command line
 shownum=0
 verbose=0
-scale=$defscale
+scale=${defscale}
 if [[ $# -eq 0 || $1 == --help || $1 == -h ]]; then usage; exit; fi
 if [[ $1 == -v ]]; then shownum=1; shift; elif [[ $1 == -vv ]]; then shownum=1; verbose=1; shift; fi
 if [[ $1 =~ ^[0-9]+$ ]]; then scale=$1; shift; fi
@@ -107,7 +107,7 @@ cmd="hledger balance -Ocsv --transpose -N -0 --layout=bare -M $*"
 
 printcmd() {
     echo "From command (might need added quotes):"  # don't know how to print all the slashes
-    echo "$cmd"
+    echo "${cmd}"
 }
 
 printamterr() {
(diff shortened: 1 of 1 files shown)


bin/hledger-bar
```diff
@@ -78,7 +78,7 @@ poschar="+"
 
 # https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit and 24-bit colors.
 # Disable if stdout is not a terminal or NO_COLOR is defined.
-if [[ ! -t 1 || -v $NO_COLOR ]]; then
+if [[ ! -t 1 || -v ${NO_COLOR} ]]; then
```
colindean, line 81: **suggestion:** I think this might actually fix the error in that build:
```suggestion
if [[ ! -t 1 || -v "${NO_COLOR}" ]]; then
```

simonmichael:
Thanks @colindean ! I never knew shellcheck could do that.

When I check it in master there's one shellcheck warning:
```
In bin/hledger-bar line 106:                                                                                                                                                                
cmd="hledger balance -Ocsv --transpose -N -0 --layout=bare -M $@"                                                                                                                           
    ^-- SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.                                                                            
```
which I've just pushed a fix for.

The homebrew error seems like something caused by a different shell version, but I'm not sure what it could be. Testing here with `shellcheck -s bash` or `ksh` reports no warnings, `sh` or `dash` report a ton of warnings, more than what you're seeing. My bash here is 5.2.21.

simonmichael:
Found it! Now fixed in master. hledger-bar was broken if $NO_COLOR was not defined. Thanks for the bug report.

colindean:
> Thanks @colindean ! I never knew shellcheck could do that.

It's a great tool, opinionated in that it never itself can modify code under inspection, unlike many if not most other linters. There are some idiosyncracies of doing `shellcheck --format=diff | git apply` but it works most of the time.

I've got something like this in many of my projects including shell files:

```makefile
.PHONY: format-sh
format-sh: $(SHELL_FILES)
    $(SHELLCHECK) --shell=bash --enable=all --format=diff $(SHELL_FILES) | git apply
```
 
> When I check it in master there's one shellcheck warning:

You'll see more if you pass `--enable=all`. This PR addressed all of them, which were all autocorrectable.

> 
> The homebrew error seems like something caused by a different shell version, but I'm not sure what it could be. Testing here with `shellcheck -s bash` or `ksh` reports no warnings, `sh` or `dash` report a ton of warnings, more than what you're seeing. My bash here is 5.2.21.

I'll have to dig into it a little more if the next hledger release to include bd8bd393f24cbef158763fbc902e46fc6584477a doesn't fix the problem.

simonmichael:
> You'll see more if you pass --enable=all. This PR addressed all of them, which were all autocorrectable.

Aha. If you care to repeat this for latest master and re-push, I'll merge it.

colindean:
Done!

Shellcheck doesn't like the bare `${NO_COLOR}` check, preferring `-n`.

simonmichael:
Would you mind rebasing it to one clean commit. (git rebase locally, then git push -f)

colindean:
Rebased!

PR #2160, opened 2024-01-23
dev: Runs shellcheck on hledger-simplebal

colindean:
Shellcheck handled the variable fixups. The pipefail flag avoids another Shellcheck warning w.r.t. handling pipe failures. This script is so minimal that it's safe to simply fail the pipe entirely if the hledger command fails.

_(Last shellcheck eval unless you're open to more and I find time. Doing these as I prep a PR to include them to be installed by default in the [Homebrew package][1].)_

[1]: https://github.com/Homebrew/homebrew-core/pull/160590

### DIFF
--- a/bin/hledger-simplebal
+++ b/bin/hledger-simplebal
@@ -3,8 +3,10 @@
 # total of `hledger balance`, optionally with arguments, as a single
 # machine-readable number.  Requires hledger 1.24 or newer.
 
+set -o pipefail
+
 VALUATION_COMMODITY="$"
-hledger bal -0 -N -X "$VALUATION_COMMODITY" --infer-market-prices -c "$VALUATION_COMMODITY 1000" --layout=bare "$@" | awk '{print $1}'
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks, and thanks for getting more scripts packaged. I never thought of them as packageable, but you show that they are, at least on unix.

Often they are prototypes for things that might end up in the main tool. If they are worth packaging, perhaps that means it's getting to be time to do that..

colindean:
That'd be awesome, esp. because I don't think the shell scripts are tested. -simplebal doesn't really need to be tested other than a "yep it works" but -bar has a lot of options.

simonmichael:
Yes, really I created bin/ for "things which I don't have time to test and support as much as the core tools". Packaging them implies more support than that in my mind, so thank you for volunteering to test and support these scripts for brew users ! 😂 Note brew is on GNU/Linux too. 

Let's see about RIIH (Rewrite It In Haskell) these two before very long.

PR #2163, opened 2024-02-04
fix: css to preserve linebreaks in messages

Zegnat:
Super minimal change. Just preserving line breaks in hledger-web. See the following before and after screenshots:

Before:
the error message about an unbalanced transaction is all on one line.

After:
the error message about an unbalanced transaction is split over multiple lines.

I was not sure if this would be nice to have in the changelog or not. Let me know if you would prefer me to prefix the commit message with `;`.

### DIFF
--- a/hledger-web/static/hledger.css
+++ b/hledger-web/static/hledger.css
@@ -86,6 +86,11 @@ ul {
     padding-left: 30px;
 }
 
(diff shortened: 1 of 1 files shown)


simonmichael:
Very nice and changelog worthy, thanks!

Issue #2164, opened 2024-02-05
support tsv: and ssv: file path prefixes, as well as csv:


simonmichael:
As reported in chat, `-f tsv:-` and `-f ssv:-` don't work like `-f csv:-`. That seems natural to expect, so we'd like to support it. 

Currently hledger's file path prefixes must be a canonical format (internally: reader) name, and the name of the reader that handles CSV, SSV and TSV is `csv`. 

Here is a patch that partially adds support for `ssv:` and `tsv:` prefixes:

```diff
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -151,2 +151,4 @@ readerNames = map rFormat (readers'::[Reader IO])
 -- Find the reader named by @mformat@, if provided.
+-- ("ssv" and "tsv" are recognised as alternate names for the csv reader,
+-- which also handles those formats.)
 -- Or, if a file path is provided, find the first reader that handles
@@ -155,3 +157,3 @@ findReader :: MonadIO m => Maybe StorageFormat -> Maybe FilePath -> Maybe (Reade
 findReader Nothing Nothing     = Nothing
-findReader (Just fmt) _        = headMay [r | r <- readers', rFormat r == fmt]
+findReader (Just fmt) _        = headMay [r | r <- readers', let rname = rFormat r, rname == fmt || (rname=="csv" && fmt `elem` ["ssv","tsv"])]
 findReader Nothing (Just path) =
@@ -170,2 +172,7 @@ type PrefixedFilePath = FilePath
 -- split that off. Eg "csv:-" -> (Just "csv", "-").
+-- These reader prefixes can be used to force a specific reader,
+-- overriding the file extension. They are the readers' canonical names,
+-- not all the file extensions they recognise. But as a special case,
+-- "csv", "ssv" and "tsv" are all recognised as prefixes selecting the csv reader,
+-- since it handles all three of those formats.
 splitReaderPrefix :: PrefixedFilePath -> (Maybe String, FilePath)
@@ -173,3 +180,3 @@ splitReaderPrefix f =
   headDef (Nothing, f) $
-  [(Just r, drop (length r + 1) f) | r <- readerNames, (r++":") `isPrefixOf` f]
+  [(Just r, drop (length r + 1) f) | r <- readerNames ++ ["ssv","tsv"], (r++":") `isPrefixOf` f]
```

[...]

PR #2165, opened 2024-02-08
imp: Support tsv and ssv prefixes (#2164)

reesmichael1:
With these changes, `hledger` now supports using the `ssv:` and `tsv:` prefixes to determine the file separator. For example:

```
$ cat basic_s
Date;Description;Id;Amount
12/11/2019;Foo;123;10.23
$ cat basic_s.rules
skip         1
fields       date, description, , amount
date-format  %d/%m/%Y
$ hledger print -f ssv:basic_s # stock hledger cannot parse the file
The hledger journal file "/home/michael/builds/hledger/ssv:basic_s" was not found.
Please create it first, eg with "hledger add" or a text editor.
Or, specify an existing journal file with -f or LEDGER_FILE.
$ stack exec -- hledger print -f ssv:basic_s # But this new build can!
2019-11-12 Foo
    expenses:unknown           10.23
    income:unknown            -10.23
```

Some notes from my end:
* Since I was already poking around this area, I went ahead and created a new type for all of the storage formats supported by `hledger`, but if this is too wide-reaching a change, it would be straightforward to use something like `Raw String`, which would in turn be `Raw "journal"`, `Raw "timeclock"`, etc., for everything except the CSV/etc. files.
* The comments in `CsvReader.hs` suggest that it should be possible to pass the _rules_ file as an argument and have the _data_ file be inferred. However, I wasn't able to figure out how to do this, either with stock `hledger` or with my build. This means I very well might have broken this feature. Could you please share how this is supposed to work so I can verify if it still functions?
* I don't love adding `ssvtest.sh`, which is mostly a copy of `csvtest.sh`, but I wasn't immediately sure how else to avoid the `csv:.*.csv` pattern used in `csvtest.sh`.

### DIFF
--- a/hledger-lib/Hledger/Data/Types.hs
+++ b/hledger-lib/Hledger/Data/Types.hs
@@ -622,9 +622,38 @@ data Journal = Journal {
 -- The data is partial, and list fields are in reverse order.
 type ParsedJournal = Journal
 
+-- | One of the standard *-separated value file types known by hledger,
+data SepFormat 
+  = Csv  -- comma-separated
+  | Tsv  -- tab-separated
+  | Ssv  -- semicolon-separated
+  deriving Eq
+
 -- | The id of a data format understood by hledger, eg @journal@ or @csv@.
 -- The --output-format option selects one of these for output.
-type StorageFormat = String
+data StorageFormat 
+  = Rules 
+  | Journal' 
+  | Ledger' 
+  | Timeclock 
+  | Timedot 
+  | Sep SepFormat 
+  deriving Eq
+
+instance Show SepFormat where
+  show Csv = "csv"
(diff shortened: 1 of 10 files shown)


simonmichael:
Great, thank you!

I had a few thoughts but they're not too important; I'm inclined to merge this and maybe (or not) do a followup PR or tweaks at some point.

simonmichael:
Our csv reader is really a Character Separated Values reader, it can handle any separator character. I guess these changes don't disrupt with that; for other characters the `.csv` extension or `csv:` prefix should be used, as before.

reesmichael1:
Cool, thanks! I was also thinking about clarifying those docs in the shower this morning, but looks like you beat me to it :grinning:

Issue #2158, opened 2024-01-21
CSV: match group references get mixed up when multiple ifs match a record


amoralesch:
When reading CSV and _rules_ files and using REGEX expressions, if more than one `if` block matches the same record, the REGEX group index gets mixup.

## Example:

* CSV file:

```csv
date,desc
2019-01-01,SUFFIX First
2019-02-01,SUFFIX Text 1 - Text 2
```

* Rules file:

```
skip          1

fields        date,desc
separator     ,
date-format   %Y-%m-%d
description   Test

if %desc SUFFIX (.*)
  account1      account:\1
  amount1       1
  account2      equity

if %desc SUFFIX (.*) - (.*)
  account1      account:\1:\2
  amount1       1
  account2      equity
```

## Expected Result

```
2019-01-01 Test
    account:First               1
    equity

2019-02-01 Test
    account:Text 1:Text 2               1
    equity
```

## Current Result

```
2019-01-01 Test
    account:First               1
    equity

2019-02-01 Test
    account:Text 1 - Text 2:Text 1               1
    equity
```

## Note

If the rules file is modified with the following:

```
if %desc SUFFIX (.*) - (.*)
  account1      account:\2:\3
  amount1       1
  account2      equity
```

(note the index for matching groups (`\2` and `\3`), then the result works as expected.

## More info

HLedger version: `hledger 1.32.2, mac-aarch64`

[Background: wish #2009, pr #2087. 
Workaround: don't allow multiple ifs with capture groups to match the same CSV record. So if you use capture groups, make those regexps specific enough that at most one of them will match any given CSV record.]

amoralesch:
Related? https://github.com/simonmichael/hledger/issues/2009

amoralesch:
FYI: @jmtd

jmtd:
Interesting: I'll take a look, thank you.

simonmichael:
I'd like to get this one in today's release, and have time to take a look if needed.

simonmichael:
I left some comments on the original PR #2158.

jmtd:
> I'd like to get this one in today's release, and have time to take a look if needed.

Sorry -- I've just seen your comment -- but I'm not able to take a look at the moment, I've got too much work on. I might have a chance to start looking at the end of this weekend (FOSDEM/travel back from FOSDEM), but I don't know how long it will take to fully address.

simonmichael:
No worries @jmtd; I went ahead with the release and noted a workaround on the issue.

[...]

PR #2167, opened 2024-02-09
feat: ui: Add a dark theme

jmtd:
I wanted a theme with roughly the same degree of colour as the default but which was easier on my eyes late at night.

(Also; I wanted to build hledger on my new machine to get back to #2158, and this was an itch to scratch)

### DIFF
--- a/hledger-ui/Hledger/UI/Theme.hs
+++ b/hledger-ui/Hledger/UI/Theme.hs
@@ -103,6 +103,29 @@ themesList = [
     (attrName "list" <> attrName "selected"                             , defAttr & reverseVideo)
   ])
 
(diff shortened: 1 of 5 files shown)


simonmichael:
Nice!  I've also opened #2168, in case you have any thoughts.

Dark theme seems very dark and low contrast here (dark grey text on black background). Is it a bit too extreme for the prominent "dark" name ?

simonmichael:
Added as is for now. Thanks!

jmtd:
> Added as is for now. Thanks!

Great, thanks! I've got some more work to raise that renames default->light and uses the fixed colour palette, and I may make some further change suggestions to the dark theme too.

PR #2174, opened 2024-02-18
fix: Hide ambiguous instance for (!?)

Vekhir:
base 4.19.0.0 introduces Data.List.!? which clashes with Data.Vector.!?, the latter of which is needed for vector operations.

This PR hides Data.List.!? in the relevant locations to prevent the name clash.


simonmichael:
Thanks!

PR #2175, opened 2024-02-22
WIP: rename default theme to light, rework light+dark themes with consistent colors

jmtd:
_This is Work In Progress, it will take me some time to write up properly what's done and why, and what the caveats are -- but I wanted to raise this now before the branch gets too stale._

Rename the default theme to "light" and rework both it and the dark theme in terms of colours picked from the "fixed" 240-color palette. This should result in consistent colours between terminal emulators, with some caveats

 * the new colours here will likely be different to what people experienced prior to this patch
 * it's not impossible for folks to redefine these palette indexes, but it's much less common than redefining the 16 color indexes used prior, as the former are specified as an RGB triplet, and the latter merely as a name e.g. "yellow"

A drawback of doing this is that the colours will not display at all in terminals which don't support the 240 palette. I think that it's extremely uncommon for modern users to have a Terminal that doesn't support these; but more common would be having a misconfigured terminal (or some intermediary software: ssh, tmux, screen, etc. introducing a problem). This can be simulated by overriding `TERM`, e.g. with `TERM=xterm-color` (where xterm-color is a terminfo definition which doesn't support the 240 color palette).

To make this failure mode less catastrophic, many of the "selected" lines have `reverseVideo` set, and also the fg/bg colors swapped, which should look like a no-op when the color is supported, but is visible as reversevideo when the color is not supported. This means it can still be used. Example:
hledger-256color-nosupport-rv

 However, using this reverseVideo hack appears to introduce an unfortunate artifact in the right-hand border of selected lines when color is working:
hledger-256color-line-artifact
I'm not sure yet whether this is a bug in this patch/hledger-ui, brick or vty. (The dark theme isn't affected)

Some TODOs

 - [x] demo screenshot of artifact bug
 - [x] document rework of dark theme in fixed colors
 - [ ] rename "terminal" theme to "inherit"?
 - [ ] rename "greenterm" to "greenonblack"
 - [X] close GH issue commands into one or more commit msgs
 - [ ] add data about color blind / other visual disability experiences


simonmichael:
Hi @jmtd, do you have time to revisit this hledger-ui PR: rename default theme to light, etc.. ?

simonmichael:
As it happens I have used hledger-ui very little in recent months. Right now I'm mainly interested in resolving open PRs. Generally, I'm interested in

1. any cleanup/improvement of what we have now
2. more visual appeal that delights and attracts users - especially if it's effective and practical
3. more/easier customisability of colours and styles - for tinkerers and devs, and supporting 2.

I don't think the miscoloured margin character need be a blocker.

jmtd:
I've rebased, cleaned up and tested the branch.

simonmichael:
Let's keep supporting --theme=default as a (undocumented) alias for --theme=light, to avoid breakage.

You suggested renaming greenterm to greenonblack - I'd be ok with it (with a compatibility alias).

I'd also be ok with renaming terminal. I have a hard time finding a name that communicates intuitively what it will look like. I guess "inherit" is another reasonable choice, though will be a little unclear to some. Looking at the code, it just sets the borders to white on black, and sets the selected item to reverse video. I'm not sure how useful that is. The goal was a theme that disturbs your usual terminal colours as little as possible. Perhaps your monochrome-with-reverse-video mode should be used here. Would "mono" or "monochrome" also be reasonable names ?

simonmichael:
Hi @jmtd, any thoughts on this ?

jmtd:
Hi @simonmichael , sorry I started a new job this week so things (including in the run up) have been hectic. I'm travelling for some days next week so I should have some time to look at it again then.

simonmichael:

hledger-ui 1.52 had some similar changes; the current theme names are light, dark, greenterm, terminal.

PR #2179, opened 2024-02-27
import: fix scoping problems with match groups

### LINKED ISSUE #2158: CSV: match group references get mixed up when multiple ifs match a record

jmtd:
This is a beginning of an attempt to fix #2158. First a test to capture the problem:

```
    issue-2158:                   FAIL
      /home/jon/git/hledger/hledger-lib/Hledger/Read/RulesReader.hs:1547:
      expected: Just "account:Text 1:Text 2"
       but got: Just "account:Text 1 - Text 2:Text 1"
```

I'm not sure why the order of `rconditionalblocks` appears to be reversed from the order in rules files. That's where I'll look when I next pick this up.


jmtd:
>  I'm not sure why the order of rconditionalblocks appears to be reversed from the order in rules files. 

Probably relevant: https://github.com/simonmichael/hledger/blob/master/hledger-lib/Hledger/Read/RulesReader.hs#L457

jmtd:
I think this is nearly there.

 - [ ] I've had to (temporarily?) adjust `regexMatchValue` to use `rconditionalblocks` rather than `rblocksassigning`, as I'm not  sure yet how to update that field to reflect an altered set of ConditionalBlocks
 - [ ] `Maybe (Either FieldTemplate ConditionalBlock)` is a bit of a mouthful. We could introduce a type alias for the `Either` bit. On the other hand, the function is now only used in two places, so perhaps that's overkill.
 - [ ] I don't know how legible `hledgerField` and `hledgerFieldValue` are. I've been staring at them too long to judge right now.
 - [ ] The new unit tests need cleaning up; I've left dev experiments and comments and suchlike there

hledger-lib/Hledger/Read/RulesReader.hs
```diff
@@ -728,15 +737,22 @@ maybeNegate _ origbool = origbool
 -- Note conditional blocks' patterns are matched against an approximation of the
 -- CSV record: all the field values, without enclosing quotes, comma-separated.
 --
-getEffectiveAssignment :: CsvRules -> CsvRecord -> HledgerFieldName -> Maybe FieldTemplate
-getEffectiveAssignment rules record f = lastMay $ map snd $ assignments
+getEffectiveAssignment
```
simonmichael, line 740: I think a couple of comments & haddocks might need updating, eg lines 734 and 751.

simonmichael (changes requested):
Just a minor thing. I haven't understood every detail but this is looking and sounding great.

hledger-lib/Hledger/Read/RulesReader.hs
```diff
@@ -730,9 +731,10 @@ maybeNegate :: MatcherPrefix -> Bool -> Bool
 maybeNegate Not origbool = not origbool
 maybeNegate _ origbool = origbool
 
--- | Given the conversion rules, a CSV record and a hledger field name, find
--- the value template ultimately assigned to this field, if any, by a field
--- assignment at top level or in a conditional block matching this record.
+-- | Given the conversion rules, a CSV record and a hledger field name, collect
+-- the value templates (and their parent Conditional Blocks where applicable)
+-- assigned to this field, if any, by top-level field assignments and
+-- conditional blocks matching this record.
```
simonmichael, line 737: Excuse quibbling, it's because I expect I'll be staring confusedly at this file in future:

"find the value template" changed to "collect the value templates (and their parent Conditional Blocks where applicable)".

but the type signature looks more like "find a single template or its parent conditional block" ?

jmtd: You're absolutely right; I forgot that `getEffectiveAssignment` applied `lastMay` to the list, I thought it was done by its callers (it was in an earlier revision of mine). I'll address it.

simonmichael:
Thanks for the fix @jmtd!

(I tried to merge keeping history, probably made a mess as usual..)

alerque:
Thanks for the work on this! I'm looking forward to being able to use it.

Which reminds me I need to get that Nix Flake working so I can run arbitrary versions in production...

Issue #2181, opened 2024-03-05
Documentation: if|table, priority of search patterns


PSLLSP:
This is just a request to update documentation for CSV import, command if|table

IF table can have several patterns that match record. Important point is that `hledger` uses THE LATEST pattern that matches the expression. Information describing this behavior should be added to the documentation.

Example, records in CSV file don't have full date, only year and month. We want to translate that information in the way that payment is done in the last day of the month, like 2020-04 -> 2024-04-30. This is not straight forward task, months have different number of days, between 28 and 31 and February can have 28 or 29 days.

NOTE: The payment is done on the first day of the next month but I do not see any "easy" way to code such translation without coding rule for each month, like  2020-04 -> 2020-05-01 or 2020-12 -> 2021-01-01. That is why I use the last day of the month, it is easier to code...

The translation of date from Y-M format to Y-M-D is coded with if|table; note that rules matching any February are before rules those defines years where February has 29 days. Rule with higher priority has to be added to the end of the "if table".

IF|table cannot have "else" rule but such rule can be defined with default value that could be overwritten with a rule in the table.

```
$ cat demo8.csv
2021-12,1749424,EUR,3.61
2021-11,1722985,EUR,3.48
2021-10,1693457,EUR,2.70
2021-09,1675066,EUR,3.33
2021-08,1634982,EUR,5.41
2021-07,1626884,EUR,5.41
2021-06,1586496,EUR,5.23
2021-05,1558019,EUR,5.41
2021-04,1535096,EUR,5.23
2021-03,1513987,EUR,5.41
2021-02,1473282,EUR,4.90
2021-01,1435894,EUR,5.41
2020-12,1405040,EUR,5.41
2020-11,1381326,EUR,5.23
2020-10,1363741,EUR,4.83
2020-09,1316067,EUR,4.83
2020-08,1289920,EUR,7.05
2020-07,1264568,EUR,10.15
2020-06,1235759,EUR,10.15
2020-05,1200729,EUR,10.15
2020-04,1165835,EUR,10.15
2020-03,1145395,EUR,10.15
2020-02,1122715,EUR,10.58
2020-01,1091401,EUR,10.14
```

```
$ cat demo8.csv.rules
# hledger import rules

fields        ym, code, curr, amou

description   DEMO|Payment %ym
status        *

amount1       %amou %curr

account1      expenses:demo
account2      assets:bank:card

[...]

PR #2183, opened 2024-03-08
Allow comment lines in the "if" table body

adept:
As discussed in #2181 , it seems sensible to allow comment lines in the "if" table body.

This small patch does just this (and adds a test that demonstrates behaviour).


simonmichael:
Thanks! As usual I clicked too soon, please mention it in the docs when you get a chance.

adept:
Did this in #2184

Best regards, Dmytro

simonmichael:
Thanks!

Issue #588, opened 2017-07-22
empty total row in balance sheet ?


simonmichael:
At https://hub.darcs.net/simon/darcshub-docs#finances, the balance sheet looks a bit strange; I think the grand total row should not be empty.
```
~/src/darcshub-docs$ hledger -f darcshub.journal bs -M 
Balance Sheet

               ||  2017/01/31  2017/02/28  2017/03/31  2017/04/30  2017/05/31  2017/06/30  2017/07/31 
===============++=====================================================================================
 Assets        ||                                                                                     
---------------++-------------------------------------------------------------------------------------
 assets:cash   ||    $-155.00    $-560.00   $-2790.00   $-3720.00   $-3950.00   $-3950.00   $-3950.00 
 assets:paypal ||           0           0     $117.30     $168.12     $287.75     $347.97     $382.93 
---------------++-------------------------------------------------------------------------------------
               ||    $-155.00    $-560.00   $-2672.70   $-3551.88   $-3662.25   $-3602.03   $-3567.07 
===============++=====================================================================================
 Liabilities   ||                                                                                     
---------------++-------------------------------------------------------------------------------------
---------------++-------------------------------------------------------------------------------------
               ||                                                                                     
===============++=====================================================================================
 Total         ||                                                                                     
```

mstksg:
Hm, I'll try to look at this this week.

simonmichael:
Maybe hold off for a day or two, I got into some refactoring for another task which might lead to a fix soon.

simonmichael:
Fixed now I believe. I went ahead and pushed some related refactoring of BalanceView that makes it a little clearer for me, pardon my renaming.

PR #2186, opened 2024-03-17
fix: Fill totals with correct amount of zeroes for empty reports

aragaer:
In current implementation balancesheet and balancesheetequity result in incorrect CSV output if one of the subreports is empty -- the "total" row for that subreport contains only "total" field and nothing else.

The solution is to know in advance the number of columns and fill total with zeroes if subreport is empty.

This might make #588 unneeded since now subreport's totals is never empty.


simonmichael:
Looks good, thank you!

PR #2204, opened 2024-06-14
Abbreviate week naming for weekly reports

victormihalache:
This PR changes how the week name is displayed in the column header by shortening it for all reports, and omitting the year if all column are within the same year.

We rename `showPeriodMonthAbbrev` to `showPeriodAbbrev` as it's now used for weeks too, so not strictly just months.

It adds the following method
```haskell
showPeriodAbbrev (WeekPeriod b) = T.pack $ formatTime defaultTimeLocale "W%V" b
```

and changes the multi-year naming:
```diff
- showPeriod (WeekPeriod b) = T.pack $ formatTime defaultTimeLocale "%FW%V" b
+ showPeriod (WeekPeriod b) = T.pack $ formatTime defaultTimeLocale "%0Y-W%V" b
```

### DIFF
--- a/hledger-lib/Hledger/Data/Dates.hs
+++ b/hledger-lib/Hledger/Data/Dates.hs
@@ -43,7 +43,7 @@ module Hledger.Data.Dates (
   showEFDate,
   showDateSpan,
   showDateSpanDebug,
-  showDateSpanMonthAbbrev,
+  showDateSpanAbbrev,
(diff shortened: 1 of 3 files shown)


simonmichael:
Just for the record: I noticed that single-year balance reports omit the year from week names, but register reports (where saving space is less urgent) don't omit the year. I don't think it's a problem.

Issue #2180, opened 2024-03-01
Support --sort for register command


igxlin:
Look like currently `hledger` only print register in the order of date. The `--sort` option of `ledger register` is very helpful when users want to get the expenses of most amount. Hope to have that option in `hledger`.

```bash
--sort value-expression (-S)
                                 Sort  postings  by evaluating the given value-expression.  Note that a
                                 comma-separated list of expressions is allowed,  in  which  case  each
                                 sorting  term  is  used in order to determine the final ordering.  For
                                 example, to search by date and then amount, one would use:
                                       ledger reg --sort 'date, amount'
                                 The sort order may be controlled with the '-' sign.  For  example,  to
                                 sort in reverse chronological order:
                                       ledger reg --sort '-date'
```

reesmichael1:
I've started working on this, and I'm curious how many `value-expressions` we'd want to support. I think it might be easiest to start a draft PR with support for, say, `date`, `account`, and `amount`, and then add more from there as needed.

Are there any other commonly used ones that anyone thinks should be included to start?

samahri:
Hi, is this issue still open? I notice a PR for it is merged

igxlin:
Closed now. Thanks for reminder.

PR #2211, opened 2024-07-16
feat: register: add --sort as in ledger

reesmichael1:
This begins work towards implementing the `--sort` argument for `register`, as requested in #2180. This still needs some polishing (e.g., documentation), but I wanted to get this up to discuss the implementation. The biggest question to me is if `SortSpec` is better placed in `ReportOptions`, which would let it be reused across the other reports along with just `register` (at the very least, it seems like this should be reused across `aregister`).

Moving it would also let us calculate it internally in `postingsReport` instead of awkwardly passing in a `Maybe SortSpec` wherever `postingsReport` is called.

### DIFF
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -129,11 +129,14 @@ comparePostings ropts (ex:es) (a, pa) (b, pb) =
             description = fmap (\t -> tdescription t) tx
         -- If there's no transaction attached, then use empty text for the description
         in fromMaybe "" description
+    absamount = mapMixedAmount (\amt -> amt { aquantity = abs (aquantity amt)})
     comparison = case ex of
+          AbsAmount' False -> compare (absamount (pamount a)) (absamount (pamount b))
           Amount' False -> compare (pamount a) (pamount b)
           Account' False -> compare (paccount a) (paccount b)
           Date' False -> compare (postingDateOrDate2 (whichDate ropts) a) (postingDateOrDate2 (whichDate ropts) b)
           Description' False -> compare (getDescription a) (getDescription b)
+          AbsAmount' True -> compare (Down (absamount (pamount a))) (Down (absamount (pamount b)))
           Amount' True -> compare (Down (pamount a)) (Down (pamount b))
           Account' True -> compare (Down (paccount a)) (Down (paccount b))
           Date' True -> compare (Down (postingDateOrDate2 (whichDate ropts) a)) (Down (postingDateOrDate2 (whichDate ropts) b))
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -25,7 +25,7 @@ where
 
 import Data.List (nub, sortBy, sortOn)
 import Data.List.Extra (nubSort)
-import Data.Maybe (isJust, isNothing)
+import Data.Maybe (isJust, isNothing, fromMaybe)
 import Data.Ord
 import Data.Text (Text)
 import Data.Time.Calendar (Day)
@@ -123,13 +123,21 @@ registerRunningCalculationFn ropts
 comparePostings :: ReportOpts -> SortSpec -> (Posting, Maybe Period) -> (Posting, Maybe Period) -> Ordering
 comparePostings _ [] _ _ = EQ
 comparePostings ropts (ex:es) (a, pa) (b, pb) = 
-    let comparison = case ex of
+    let 
+    getDescription p = 
+        let tx = ptransaction p 
+            description = fmap (\t -> tdescription t) tx
+        -- If there's no transaction attached, then use empty text for the description
+        in fromMaybe "" description
+    comparison = case ex of
           Amount' False -> compare (pamount a) (pamount b)
           Account' False -> compare (paccount a) (paccount b)
           Date' False -> compare (postingDateOrDate2 (whichDate ropts) a) (postingDateOrDate2 (whichDate ropts) b)
+          Description' False -> compare (getDescription a) (getDescription b)
           Amount' True -> compare (Down (pamount a)) (Down (pamount b))
           Account' True -> compare (Down (paccount a)) (Down (paccount b))
           Date' True -> compare (Down (postingDateOrDate2 (whichDate ropts) a)) (Down (postingDateOrDate2 (whichDate ropts) b))
+          Description' True -> compare (Down (getDescription a)) (Down (getDescription b))
     in 
     if comparison == EQ then comparePostings ropts es (a, pa) (b, pb) else comparison
 
--- a/hledger-lib/Hledger/Reports/PostingsReport.hs
+++ b/hledger-lib/Hledger/Reports/PostingsReport.hs
@@ -129,14 +129,13 @@ comparePostings ropts (ex:es) (a, pa) (b, pb) =
             description = fmap (\t -> tdescription t) tx
         -- If there's no transaction attached, then use empty text for the description
         in fromMaybe "" description
-    absamount = mapMixedAmount (\amt -> amt { aquantity = abs (aquantity amt)})
     comparison = case ex of
-          AbsAmount' False -> compare (absamount (pamount a)) (absamount (pamount b))
+          AbsAmount' False -> compare (abs (pamount a)) (abs (pamount b))
           Amount' False -> compare (pamount a) (pamount b)
           Account' False -> compare (paccount a) (paccount b)
           Date' False -> compare (postingDateOrDate2 (whichDate ropts) a) (postingDateOrDate2 (whichDate ropts) b)
           Description' False -> compare (getDescription a) (getDescription b)
-          AbsAmount' True -> compare (Down (absamount (pamount a))) (Down (absamount (pamount b)))
+          AbsAmount' True -> compare (Down (abs (pamount a))) (Down (abs (pamount b)))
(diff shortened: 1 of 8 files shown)


hledger-lib/Hledger/Reports/PostingsReport.hs
```diff
@@ -61,15 +65,44 @@ instance HasAmounts PostingsReportItem where
...
+data ValueExp
+    = Date' Bool
+    | Account' Bool
+    | Amount' Bool
+    deriving (Show)
+type SortSpec = [ValueExp]
+
+-- Load a SortSpec from the argument given to --sort
+-- If there is no spec given, then sort by [Date' False] by default
+getSortSpec :: RawOpts -> SortSpec
```
reesmichael1, line 80: As I said in the main body, this feels weird in `PostingsReport`, but I wasn't sure if it would make sense in the main `ReportOptions` either, since there might be many commands that won't use it.

hledger-lib/Hledger/Reports/PostingsReport.hs
```diff
@@ -61,15 +65,44 @@ instance HasAmounts PostingsReportItem where
...
+                where isNegated = isPrefixOf "-" t
+                      trimmed = fromMaybe t (stripPrefix "-" t)
+          in map termParser terms
+    in maybe [Date' False] optParser opt 
+
 -- | Select postings from the journal and add running balance and other
 -- information to make a postings report. Used by eg hledger's register command.
-postingsReport :: ReportSpec -> Journal -> PostingsReport
-postingsReport rspec@ReportSpec{_rsReportOpts=ropts@ReportOpts{..}} j = items
+postingsReport :: ReportSpec -> Maybe SortSpec -> Journal -> PostingsReport
```
reesmichael1, line 97: Similarly, I think passing `Nothing` or `Just sspec` to this is awkward. If the `SortSpec` was part of `ReportOpts` instead, then we could calculate the correct `sspec` inside of this function and not have to pollute other call sites.

hledger-lib/Hledger/Reports/PostingsReport.hs
```diff
@@ -61,15 +65,44 @@ instance HasAmounts PostingsReportItem where
 -- with a tuple.
 type SummaryPosting = (Posting, Period)
 
+-- Possible value expressions taken by the --sort command
+-- Each of these takes a bool, which shows if it has been inverted
+-- (True -> has been inverted, reverse the order)
+data ValueExp
```
reesmichael1, line 71: So far, I've only added support for `date`, `account`, and `amount` as `value-expressions`, although [there are many more](https://manpages.org/ledger) in `ledger`. I felt like this was a reasonable subset to start with, and it's straightforward to add more as needed/requested, either in this PR or down the road.

simonmichael: I've seen people ask to sort by amount, but not by account name. Is that (or any other sort field) actually useful ? And multi-parameter sorting, and reversing ? These are cool and might indeed be useful, but I for one am wary of overengineering ahead of real world usefulness.

If not, there's an existing balance --sort-amount flag/sort_amount_ field we could reuse. (And even if so, perhaps we can still adapt and reuse it.)

When sorting by amount, what will happen with the running total/running balance column ?

hledger-lib/Hledger/Reports/PostingsReport.hs
```diff
@@ -61,15 +65,44 @@ instance HasAmounts PostingsReportItem where
...
+    = Date' Bool
+    | Account' Bool
+    | Amount' Bool
+    deriving (Show)
+type SortSpec = [ValueExp]
```
simonmichael, line 76: Naming: consider not using "value expressions" here. That specific phrase is used in Ledger's manual for Ledger's built in programming language, and in a number of docs I say that Ledger value expressions are something hledger doesn't support; and we'll never support value expressions exactly like Ledger's.

hledger/Hledger/Cli/Commands/Register.hs
```diff
@@ -49,6 +49,7 @@ registermode = hledgerCommandMode
     ("fuzzy search for one recent posting with description closest to "++arg)
   ,flagNone ["related","r"] (setboolopt "related") "show postings' siblings instead"
   ,flagNone ["invert"] (setboolopt "invert") "display all amounts with reversed sign"
+  ,flagReq  ["sort"] (\s opts -> Right $ setopt "sort" s opts) "FlagHelp" "Help"
```
simonmichael, line 52: More help to come I expect. Also once things settle we'll want to mention it in Register.md.

simonmichael (commented):
Thank you @reesmichael1 ! Some comments added.

hledger-lib/hledger-lib.cabal
```diff
@@ -143,6 +143,7 @@ library
     , pretty-simple >4 && <5
     , regex-tdfa
     , safe >=0.3.20
+    , split >=0.1
```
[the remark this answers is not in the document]
simonmichael: Or better, I think you could use Hledger.Utils.splitAtElement

hledger/Hledger/Cli/Commands/Register.hs
reesmichael1: Whoops, can't believe I forgot to update that...will fix along with adding documentation!

hledger-lib/Hledger/Reports/PostingsReport.hs
reesmichael1: Sure, there's plenty of room for discussion here. I started out pretty much porting the exact behavior of `--sort` from `ledger` (the test cases are direct copies from `ledger`'s output), and I tried to choose what seemed like, at a quick glance, a reasonable subset of sortable fields from the ~60 it supports.

Personally, I do think sorting by account name is useful. For testing while developing, I often ran `hledger register --sort account tag:trip=RecentTrip`, which let me see all of my individual transactions over the trip in `Expenses:Dining`, `Expenses:Travel`, etc. grouped together. This is more granular than just `hledger balance tag:trip=RecentTrip`, and I enjoyed having access to it. However, I'm very open to switching out to any set of fields that are deemed useful by consensus.

The running balance column shows the total from each individual transaction in whatever order they're in.

@igxlin As the original requester in #2180, any thoughts on what fields are most useful?

hledger-lib/Hledger/Reports/PostingsReport.hs
reesmichael1: Makes sense! How about `SortField` or something more directly descriptive then?

hledger-lib/Hledger/Reports/PostingsReport.hs
igxlin: Hi @reesmichael1 , personally, I only use `amount` a lot, as I want to know why I spent that much money from time to time. :see_no_evil: `date`, `amount`, `account` are good enough from my perspective.

simonmichael: Fair enough, date/amount/account then. 

Running total just summing items in display order as usual, makes sense. In -H mode, where register shows historical balances after each posting, I'm not sure there's anything we can show that wouldn't be misleading ? Let's see.

reesmichael1: Thanks everyone! I think this is mostly ready to go now, although I'll flag that I added an error message when someone tries to use both `--sort` and `--historical`. If you think it's better to allow that and live with the misleading results, I'm happy to remove it.

simonmichael:
Looking good!

Now that I've played with it a bit more - I think people will want `description` sorting as well. Could we accept that, and also `desc`, as that spelling is accepted elsewhere.

More ideas, though not blockers:
- somebody will probably want `absamount` sooner or later ?
- it would be nice to have this work the same way with `aregister`.

reesmichael1:
Thanks! I actually started adding it to `aregister` but stopped for some reason I can't remember, so I'll try to wire it in and see what it's actually like. I think it was probably because I thought the balance total column would be weird, but on more reflection, I think having it for transactions there would be useful.

I'll add in the `description`/`desc` as well! I'll see about `absamount` while I'm working on it.

I'll verify again, but I'm pretty sure the `date` and `-date` sorting does indeed preserve order within the same date as it should.

reesmichael1:
Ok, I think I've hacked together a way to get `abs` working for `MixedAmount` in 2d45393, along with adding `desc`/`description`. I'll look at `aregister` next, although it might be a day or two before I have time to wrap that up.

reesmichael1:
I'd hoped to get `aregister` working before leaving on vacation, but getting ready ended up taking all of my time (I'm typing this from a layover!). It'll be a few weeks before I have a chance to get back to `aregister`, so if you're happy with this as is, feel free to merge it now, or I can continue when I get back!

simonmichael:
Thanks @reesmichael1. I'd be ok with merging this now but there's one niggling change needed I think: also mention the allowed --sort keywords in 1. the option's help and 2. the error message when a bad value is given (and ideally also when no value is given, but I suspect that's too hard). No hurry on this, vacation++ 🏝️

simonmichael:
PS: I rebased.

simonmichael:
I have merged this to catch the next release. Thanks a lot @reesmichael1! If/when you work on it again (for aregister), this is still worth considering:

> mention the allowed --sort keywords in the option's help, and in the error message when a bad value is given, (and ideally also when no value is given, but I suspect that's too hard)

reesmichael1:
Thanks for taking over for me! Sorry for disappearing, I have been swamped with life over the last several weeks. I'll try to get back to continuing this soon!

PR #2220, opened 2024-09-03
Budget transpose

thielema:
bugfix for a duplicate transposition in Commands.Balance.budgetReport


simonmichael:
Thanks! I don't know the user-visible impact of this, but I don't need to document it so I'll merge shortly.

PR #2222, opened 2024-09-06
Add border and class attributes to Spreadsheet data type

thielema:
Here are some more commits for unifying existing multibalance HTML export with my Spreadsheet data type.
One visible effect is that now FODS tables also get the double lines like the multibalance HTML
and that the formats CSV, HTML, FODS should be transposable for all three kinds of balance exports.

I was also able to remove the dissection of already formated rows in
Commands.Balance.multiBalanceReportHtmlHeadRow, multiBalanceReportHtmlBodyRow, multiBalanceReportHtmlFootRow.

However, there was a comment "TODO pad totals ..." that I do not understand.
It belonged to multiBalanceReportHtmlHeadRow, which is now gone.
Thus the comment is probably no longer at a sensible position if it is still relevant, at all.

### DIFF
--- a/hledger-lib/Hledger/Write/Html.hs
+++ b/hledger-lib/Hledger/Write/Html.hs
@@ -6,6 +6,8 @@ This is derived from <https://hackage.haskell.org/package/classify-frog-0.2.4.3/
 -}
 module Hledger.Write.Html (
     printHtml,
+    formatRow,
+    formatCell,
     ) where
 
 import qualified Hledger.Write.Spreadsheet as Spr
@@ -15,7 +17,7 @@ import qualified Data.Text as Text
 import qualified Lucid.Base as LucidBase
 import qualified Lucid
 import Data.Text (Text)
-import Data.Foldable (for_)
+import Data.Foldable (traverse_)
 
 
 printHtml :: (Lines border) => [[Cell border (Lucid.Html ())]] -> Lucid.Html ()
@@ -26,9 +28,10 @@ printHtml table = do
         "th, td {padding-left:1em}" :
         "th.account, td.account {padding-left:0;}" :
         []
-    Lucid.table_ $ for_ table $ \row ->
-        Lucid.tr_ $ for_ row $ \cell ->
-        formatCell cell
+    Lucid.table_ $ traverse_ formatRow table
+
+formatRow:: (Lines border) => [Cell border (Lucid.Html ())] -> Lucid.Html ()
+formatRow = Lucid.tr_ . traverse_ formatCell
 
 formatCell :: (Lines border) => Cell border (Lucid.Html ()) -> Lucid.Html ()
 formatCell cell =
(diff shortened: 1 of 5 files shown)


simonmichael:
Thanks! It's release time, so much as I'd like to have these enhancements I think I'll save this for next release, rather than do a last-minute merge. Let me know if you disagree.

thielema:
On Fri, 6 Sep 2024, Simon Michael wrote:

> Thanks! It's release time, so much as I'd like to have these 
> enhancements I think I'll save this for next release, rather than do a 
> last-minute merge. Let me know if you disagree.

I don't care. I am working on more stuff regarding table export. I want to 
introduce a new extended tabular data type and I want to add anchors to 
cells in HTML and FODS.

simonmichael:
Thanks @thielema. I'll merge this if/when you are ready.

I suspect you use haddock comments less than me - I tend to add a line of english description to most things. They can be a big help to someone else debugging/improving code later.

simonmichael:
I hear you! This is of course a recurring debate among programmers. 

I agree that obviously redundant haddocks that add no value are to be avoided.

In the hledger codebase, there's no hard requirement for haddocks.
And sometimes the bar can be lower, eg in single-purpose less-frequently-developed modules like FODS.hs.

But I believe it's always worth at least considering them when adding code, and it's something I consider when reviewing code.
I'll give some reasons why I think so, for the record (and maybe they'll persuade you just a little).

Haddocks help guide and anchor the developer while writing or changing code. 
They are the cheapest kind of doc, spec and test suite. 
They also create a place to add actual doctests, now or later.

Haddocks can be helpful to contributors who are not expert haskellers, which happens quite often in the hledger project.
The human language descriptions can complement the code, reducing cognitive effort and helping with mental chunking.
I think people find the hledger's code, with its pervasive haddocks, above average in readability.

Haddocks can also help experienced developers move faster.
Code which seems quite clear and obvious when you are writing it is often less obvious to oneself a few weeks or years later.
And certainly it can be less obvious to others than we might think.
When debugging or coding we are often mentally stretched and we would prefer to conserve brainpower for the main task.

hledger is a documentation-driven project, with docs in general being a top priority.
I think this is one reason it has held together and kept improving over a long period.

simonmichael:
I had few more notes in mind, adding them (possibly via github edits) as I remember them:

In my experience debugging/writing/changing hledger code,

- Number of times I've regretted seeing a haddock comment attached to some code: almost zero.
- Effort to remove an excessive haddock: almost zero.
- Number of times I've been trying to understand some contributed code and haddocks would have saved my time: quite a few.
- Number of times I've appreciated code haddocks and found them helpful: many.
- Number of times I've found errors or staleness in haddocks: not often.
- Effort to fix wrong haddocks, or to write new ones: usually very low. And if it's high, it's usually very worthwhile because it alerts me to a confusion in the code or clarifies my thinking.

simonmichael:
And I've gone a bit off topic for this PR, perhaps I should have moved this to the mail list. If folks would like to discuss it more, I'd be happy to do that.)

thielema:
On Wed, 11 Sep 2024, Simon Michael wrote:

> And I've gone a bit off topic for this PR, perhaps I should have moved 
> this to the mail list. If folks would like to discuss it more, I'd be 
> happy to do that.)

I hear your arguments. Is there specific information you want to know 
about my added functions that I could put into Haddock comments?

simonmichael:
No need right now, thank you. Most likely you’ll be the one touching this and I’m sure it’s bug free… I just wanted to mention this generally. I’ll merge shortly.

simonmichael:
Merged, thank you!

I will update the todo comment you mentioned.

Issue #986, opened 2019-03-09
better packaging & easier setup of hledger shell completions


simonmichael:
[2019-10-19: this is about packaging any available shell completions for hledger (bash, zsh etc.) within the hledger executable. Creating such shell completions is discussed at: #136 ]

[2024-09-13: broadened this issue to better packaging & easier setup by any means]

[Here's](https://hackage.haskell.org/package/chronos-bench) an example where the command line (bash) completions are built in, for convenience. Something like this would make hledger's command line completions available to more folks.
```
You can as well use autocomplete of your shell with following command:

source <(chronos --bash-completion-script `which chronos`)
```

schoettl:
I think we only need to package the existing script into the distro packages.

People who install hledger using the source code can as well add `source ~/software/hledger/shell-completion/hledger-completion.bash` to their `.bashrc`.

Other people install it with their package manager and in future hopefully have the completion for free without doing anything.

How is the release cycle with hledeger? Do you know when package maintainers create a new version for hledger?

Last time, I asked at some distros, they closed/deferred the issue because the completion script was not yet in the sources they used for the package...

Status:
- Ubuntu told me to submit this issue to debian repos: https://answers.launchpad.net/ubuntu/+source/haskell-hledger/+question/678127
- Debian told me that the completion script must first be in the sources they use for the release.
- AUR told me nothing - I don't know who is the package maintainer: https://aur.archlinux.org/packages/hledger-git/#news
- Arch deferred the task, I guess for the same reason Debian did: https://bugs.archlinux.org/task/61546

simonmichael:
> On Mar 10, 2019, at 6:20 AM, Jakob Schöttl <notifications@github.com> wrote:
> I think we only need to package the existing script into the distro packages.
> 
> People who install hledger using the source code can as well add source ~/software/hledger/shell-completion/hledger-completion.bash to their .bashrc.
> 
> Other people install it with their package manager and in future hopefully have the completion for free without doing anything.
> 

That would be ideal, but consider what a small proportion of hledger users that applies to, at least right now. Currently no package does this. Many people build from source since packages are too old, or use hledger-install because they want something easy, or download binaries for windows.

> How is the release cycle with hledeger? Do you know when package maintainers create a new version for hledger?
> 
hledger's usual release window is the the first day or two of each month. We don't yet have many active packagers; some updates come quickly, some come when you prod them, others come very rarely. I've been working to make the situtation clearer on the download page.
[...]

PR #2223, opened 2024-09-10
Add bash completion to bindist

hseg:
From a cursory check, this appears to be what's needed, but I may be mistaken.
Would appreciate a review

Closes: #986

### DIFF
--- a/.github/workflows/binaries-linux-arm32v7.yml
+++ b/.github/workflows/binaries-linux-arm32v7.yml
@@ -34,7 +34,7 @@ jobs:
           docker cp $container_id:/root/hledger tmp
           docker rm -v $container_id
           cd tmp
-          tar cvf hledger-linux-arm32v7.tar hledger hledger-ui hledger-web
+          tar cvf hledger-linux-arm32v7.tar hledger hledger-ui hledger-web hledger/shell-completion/hledger-completion.bash
 
       # upload-artifact loses execute permissions, so we tar the binaries to preserve them.
       # github UI always zips artifacts when they are downloaded, so we don't bother compressing the tar.
--- a/.github/workflows/binaries-linux-x64-stack.yml
+++ b/.github/workflows/binaries-linux-x64-stack.yml
@@ -118,6 +118,7 @@ jobs:
     - name: Gather binaries
       run: |
         mkdir tmp
+        cp hledger/shell-completion/hledger-completion.bash tmp
         cd tmp
         cp ~/.local/bin/hledger .
         cp ~/.local/bin/hledger-ui .
@@ -125,7 +126,7 @@ jobs:
         strip hledger
         strip hledger-ui
         strip hledger-web
-        tar cvf hledger-mac-x64.tar hledger hledger-ui hledger-web
+        tar cvf hledger-mac-x64.tar hledger hledger-ui hledger-web hledger-completion.bash
 
     # upload-artifact loses execute permissions, so we tar the binaries to preserve them.
     # github UI always zips artifacts when they are downloaded, so we don't bother compressing the tar.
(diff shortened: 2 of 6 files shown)


hseg:
Oops, misread the workflow -- one still needs to copy completion to `tmp` first.
Let me quickly fix that...

.github/workflows/binaries-linux-arm32v7.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           docker cp $container_id:/root/hledger tmp
           docker rm -v $container_id
           cd tmp
-          tar cvf hledger-linux-arm32v7.tar hledger hledger-ui hledger-web
+          tar cvf hledger-linux-arm32v7.tar hledger hledger-ui hledger-web hledger/shell-completion/hledger-completion.bash
```
simonmichael, line 37: This looks wrong - is `hledger` the binary file or a directory here ?

hseg: Was meant to be the directory

simonmichael:
And thanks for making a start on this. CI scripts are hard to test and debug. I guess the alternative is to embed shell completions in the hledger binaries and add a flag that writes them to stdout ? Is that the standard/preferred idiom, or just unnecessary complication ?

hseg:
I know projects using `optparse-applicative` do this, cf eg https://gitlab.archlinux.org/archlinux/packaging/packages/stack/-/blame/main/PKGBUILD#L84
But that appeared imo to be more work, especially after I'd tracked down your releng.

hseg:
Hrm. Perhaps by adding the completion to `extra-source-files`, it'll be copied somewhere useful by `stack`?

hseg:
further oops -- just noticed I've opened this PR from my master branch, without bothering to create a new branch. merging this might end up being messy, sorry for that.

hseg:
Yeah, I'm not seeing anything there.
However... before the move to `tmp` to construct the tarball, aren't we at the repository root? Couldn't we just copy the completion in there?

hseg:
I think this might do the trick -- I note that `binaries-linux-x64.yml` does something similar.

hseg:
I've got to leave in a minute, I'll pick this up again tomorrow.

hseg:
I tried testing locally using https://github.com/nektos/act, but I don't have the github tokens. Would you mind testing this?

simonmichael:
I've broadened the scope of #986 a little and posted some problem reports there. I'll get back to this issue once I can see completions working on my machine.

.github/workflows/binaries-linux-arm32v7.yml
simonmichael: @hseg, all the other changes look right, but am I right in thinking this one is still wrong ? We rarely use this workflow, but as written it seems to treat `hledger` as both the binary and the directory at the same time.

hseg:
Yeah, it doesn't follow the same flow as the others, which is a smell. You want to add a `docker cp` line to copy out the shell completions into `tmp` as well.
In theory, this means you could write the `tar` line as just
```bash
tar cvf hledger-linux-arm32v7.tar *
```
Don't know how I missed it.

BTW, after my prodding, the AUR maintainer also added the manpages to the package on their own, but the same logic goes for them as well - they're user-facing content, they should be in the bindist.

simonmichael:
Thanks for helping with packaging and distribution @hseg. I wasn't really aware of packagers using the release binaries from github; now I am. I have merged this and we'll keep improving things. Perhaps the "release bindists" should have all the same extra source files that the hackage tarballs have.

hseg:
No problem. Though I notice you didn't change the linux-arm32v7 workflow you were asking about - won't this create a different directory structure in the tarball from the others? I'm afk, so I can't test this right now.

simonmichael:
Yup, I’ll follow up on that.

Issue #2196, opened 2024-04-25
1.25: balance reports' HTML output stopped showing digit group marks


simonmichael:
A regression reported by @the-solipsist: since hledger 1.25 (2022), the HTML output of balance reports (eg `bal -Y -o html`) has not been showing digit group marks, regardless of commodity display styles. (It was using some CSV output code, which suppresses them.)

simonmichael:
Fixed in master. Thanks for the report! Please claim your https://hledger.org/REGRESSIONS.html#regression-bounty when convenient.

PR #2224, opened 2024-09-11
fix: amount formatting of the Net row in the balancesheet HTML output

basvandijk:
The "Net" row in the HTML output of `balancesheet` formatted amounts using `machineFmt` which is inconsistent with all the other amounts in the balance sheet. This fixes that by formatting Net amounts using `oneLineNoCostFmt`.

### DIFF
--- a/hledger/Hledger/Cli/CompoundBalanceCommand.hs
+++ b/hledger/Hledger/Cli/CompoundBalanceCommand.hs
@@ -364,7 +364,7 @@ compoundBalanceReportAsHtml ropts cbr =
     totalrows =
       if no_total_ ropts || length subreports == 1 then []
       else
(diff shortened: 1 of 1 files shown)


simonmichael:
Thank you for the fix!

This was a regression in 1.25, reported and fixed in 1.33.1 by #2196.. but not completely.
Claim your $50 regression fixer's bounty: https://hledger.org/REGRESSIONS.html#regression-bounty

basvandijk:
Thanks for the merge.

> Claim your $50 regression fixer's bounty ...

I would happily donate to hledger instead since it's a vital tool in handling my finances.

simonmichael:
> I would happily donate to hledger instead since it's a vital tool in handling my finances.

Thanks! So noted.

PR #2226, opened 2024-09-12
Add hyperlinks to accounts and dates in HTML and FODS export

thielema:
This adds the option `--anchor` to `hledger-cli`. If the user passes this option, then in HTML and FODS output the account and date header cells become hyperlinks to `hledger-web`.
If you like this feature, I add some documentation.
I could also rename the option to `--base-url` for consistency with `hledger-web`.
There is a problem though. Some special characters like `+` or umlauts in account names are not correctly escaped in percent URL syntax by Lucid. I do not know how to solve this correctly with Lucid and whether it is solvable with Lucid, at all.

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -140,6 +140,7 @@ data ReportOpts = ReportOpts {
     ,no_elide_         :: Bool
     ,real_             :: Bool
     ,format_           :: StringFormat
+    ,balance_base_url_ :: Maybe T.Text
     ,pretty_           :: Bool
     ,querystring_      :: [T.Text]
     --
@@ -199,6 +200,7 @@ defreportopts = ReportOpts
     , no_elide_         = False
     , real_             = False
     , format_           = def
+    , balance_base_url_ = Nothing
     , pretty_           = False
     , querystring_      = []
     , average_          = False
@@ -255,6 +257,7 @@ rawOptsToReportOpts d rawopts =
           ,no_elide_         = boolopt "no-elide" rawopts
           ,real_             = boolopt "real" rawopts
           ,format_           = format
+          ,balance_base_url_ = T.pack <$> maybestringopt "base-url" rawopts
           ,querystring_      = querystring
           ,average_          = boolopt "average" rawopts
           ,related_          = boolopt "related" rawopts
--- a/hledger-lib/Hledger/Write/Html.hs
+++ b/hledger-lib/Hledger/Write/Html.hs
@@ -37,6 +37,10 @@ formatRow = Lucid.tr_ . traverse_ formatCell
 formatCell :: (Lines border) => Cell border (Lucid.Html ()) -> Lucid.Html ()
 formatCell cell =
     let str = cellContent cell in
+    let content =
+            if Text.null $ cellAnchor cell
+                then str
+                else Lucid.a_ [Lucid.href_ $ cellAnchor cell] str in
     let border field access =
             map (field<>) $ borderLines $ access $ cellBorder cell in
     let leftBorder   = border "border-left:"   Spr.borderLeft   in
@@ -51,7 +55,7 @@ formatCell cell =
             map Lucid.class_ $
             filter (not . Text.null) [Spr.textFromClass $ cellClass cell] in
     case cellStyle cell of
-        Head -> Lucid.th_ (style++class_) str
+        Head -> Lucid.th_ (style++class_) content
         Body emph ->
             let align =
                     case cellType cell of
@@ -62,7 +66,7 @@ formatCell cell =
                     case emph of
                         Item -> id
                         Total -> Lucid.b_
-            in  Lucid.td_ (style++align++class_) $ withEmph str
+            in  Lucid.td_ (style++align++class_) $ withEmph content
 
 
 class (Spr.Lines border) => Lines border where
--- a/hledger-lib/Hledger/Write/Ods.hs
+++ b/hledger-lib/Hledger/Write/Ods.hs
@@ -314,9 +314,16 @@ formatCell cell =
                         (cellContent cell)
                 _ -> "office:value-type='string'"
 
+        anchor text =
+            if T.null $ Spr.cellAnchor cell
+                then text
+                else printf "<text:a xlink:href='%s'>%s</text:a>"
+                        (escape $ T.unpack $ Spr.cellAnchor cell) text
+
     in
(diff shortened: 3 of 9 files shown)


thielema:
The feature could also be used in `hledger-web`. That is, in `hledger-web` there could be hyperlinks to auto-generated multi-balance HTML's, that link back to `hledger-web`.

simonmichael:
Interesting feature!

`--base-url` sounds good. I assume its value will be the same as what hledger-web prints at startup ? Eg 
```
...
with base url http://127.0.0.1:5000
...
```

I don't know the url escaping function offhand but it sounds resolvable. Maybe there's ideas in hledger-web.

The reverse feature (linking to static HTML reports) could be a separate PR. I'd like to have multi-period tabular reports as a normal view in hledger-web, but linking to static html reports saved in temporary files, or better, serving the static html generated in memory, might be easier.

thielema:
On Thu, 12 Sep 2024, Simon Michael wrote:

> Interesting feature!
> 
> --base-url sounds good. I assume its value will be the same as what hledger-web prints at startup ? Eg
> 
> ...
> with base url http://127.0.0.1:5000
> ...

I expect the trailing slash to be part of the base URL.

> I don't know the url escaping function offhand but it sounds resolvable. 
> Maybe there's ideas in hledger-web.

hledger-web generates HTML via `blaze` instead of `lucid`.


> The reverse feature (linking to static HTML reports) could be a separate PR.

of course

> I'd like to have multi-period tabular reports as a normal view in 
> hledger-web, but linking to static html reports saved in temporary 
> files, or better, serving the static html generated in memory, might be 
> easier.

Hledger.Write.Html.printHTML returns a Lucid.Html () which can be served 
from memory or written to a file.

thielema:
We could also add the query from the command-line to the hyperlinks. That is, if someone restricts transactions in a certain way for the `balance` command, the `hledger-web` register page would respect these restrictions, too.

simonmichael:
Sorry it has taken a while to get back to this, I have been backlogged.

I have no objection to this. Do you use it much ?

Is it implemented only for the `balance` command right now ?

thielema:
On Tue, 24 Sep 2024, Simon Michael wrote:

> I have no objection to this. Do you use it much ?

Not yet. I guess, it would see more use if integrated in hledger-web.

> Is it implemented only for the balance command right now ?

Yes. What other commands could benefit from the feature?

thielema:
On Wed, 25 Sep 2024, Simon Michael wrote:

> Things added to the balance command can often also apply to the 
> balancesheet, balancesheetequity, cashflow, and incomestatement 
> commands; they share many options and features. Sometimes I call them 
> collectively "balcmds".

I guess supporting these commands happens in CompoundBalanceCommand.

>       it would see more use if integrated in hledger-web.
> 
> Oh, if the HTML balance reports were directly served within hledger-web ?

exactly

simonmichael:
> I guess supporting these commands happens in CompoundBalanceCommand.

That's right.

thielema:
In `CompoundBalanceCommand` only the `--base-url` option was missing. Hyperlinking accounts in HTML output now works.

thielema:
CI build fails because of missing module `Hledger.Write.Html.Attribute`. However, I can locally checkout the repository and the module is present and building with `cabal` and `stack` succeeds.

thielema:
The balance query is now passed to the cell anchor URLs.
For URL percentage encoding I assume I have to call an escaping function myself. Do you already import such a function from somewhere? Otherwise I think network-uri will do it.

thielema:
I think I have now resolved the URL escaping issue using the package `modern-uri`.
`network-uri` only supports `String`.

simonmichael:
Thank you very much!

Let's try this and keep it as long as it's useful. I'll mark it experimental to start with.

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> Let's try this and keep it as long as it's useful. I'll mark it 
> experimental to start with.

I have just proposed an extension to hledger-web that uses those links. 
With this extension, the hyperlinks in balance reports become very handy.

simonmichael:
Nice.

> The export function will not add any slash in order to support relative hyperreferences.

Could you explain this to me a bit more ?

simonmichael:
Ie, can we make it not care whether they write a slash or not ? Where would relative links come in.

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> Ie, can we make it not care whether they write a slash or not ? Where would relative links come in.

I use relative links in my Balance Report extension in hledger-web. The 
links just read "register?q=date:2024". No slash involved. It is 
equivalent to the command line option --base-url="".

simonmichael:
I see the PR but I haven't run it locally yet and I don't quite understand. Can we eg add the trailing/separating slash automatically when needed ?

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> Ie, can we make it not care whether they write a slash or not ?

I could add a slash whenever base-url is non-empty and not terminated with 
a slash. This way, --base-url="" would enable relative links and 
--base-url=/ would enable server-absolute links. Additionally, 
--base-url=simple would not result in links like 
"simpleregister?q=date:2024".

simonmichael:
That sounds great. I'd just like to not make users work harder than needed.

simonmichael:
If possible I'd like to clean up this slash thing before the next wave of PR merges. I have a few small followups to this one. If you prefer, I can help with the slash. No urgency.

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> If possible I'd like to clean up this slash thing before the next wave 
> of PR merges. I have a few small followups to this one. If you prefer, I 
> can help with the slash. No urgency.

The slash issue is already fixed in commit 
1024ad90135db74a02374d9e23cdb4875a23346e.

simonmichael:
Great! Uh, in what pr or branch do I find that ?

thielema:
That commit was not yet pushed. Take this one instead:
https://github.com/simonmichael/hledger/commit/37705f3434506d384178bd0850d0470252ee37dd

simonmichael:
Will do, when I figure out how. :)

simonmichael:
Ah, I think the PR was made against your own master by mistake: https://github.com/thielema/hledger/pull/3 . I can get it from there.

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> Will do, when I figure out how. :)

I opened a regular pull request:

https://github.com/simonmichael/hledger/pull/2245

simonmichael:
Thank you! This and followups pushed to master.

Issue #2225, opened 2024-09-11
Unnecessary <th></th> before the "Average" column in the HTML output of the incomestatement


basvandijk:
In hledger-1.40 the HTML output of the incomestatement has an extra unnecessary column before the "Average" column:
![is-html-hledger 1 40](https://github.com/user-attachments/assets/d9671930-bb4f-4d34-a822-a188909b3b64)
This has the consequence that the average amounts are not aligned with the "Average" column header.
This was not the case in hledger-1.34:
![is-html-hledger-1 34](https://github.com/user-attachments/assets/3b6e495a-056e-4488-ab5f-0f96a2ad98dd)

thielema:
This can be a problem caused by my recent changes to `Hledger.Cli.CompoundBalanceCommand`. With what command-line options do you create the HTML?

thielema:
I suspect the problem lies here:
https://github.com/simonmichael/hledger/blob/1.40/hledger/Hledger/Cli/CompoundBalanceCommand.hs#L346
There empty TH's are generated where there should be no TH's, at all.
It is simple to fix, but we must check whether other callers of CompoundBalanceCommand need the empty TH's.

basvandijk:
> With what command-line options do you create the HTML?

I think I used the following:
```
hledger is -O html -b thisyear -e thismonth -M -s -S -A -X CHF
```

thielema:
Bug seems to be introducted here:
de617ec91bc0dec485f97d3abb51d9aaec37d06f

thielema:
https://github.com/simonmichael/hledger/pull/2228

simonmichael:
Thanks for the report @basvandijk, and for investigating @thielema. I will add this to my queue, unless anyone figures it out before me.

simonmichael:
Fixed by #2228. Please claim your $50 finder and fixer bounties, @basvandijk and @thielema !

https://hledger.org/REGRESSIONS.html#regression-bounty

simonmichael:
@basvandijk I think you told me you are donating your $50 regression finder's bounty for 2225 to the project.

@thielema, I encourage you to claim your $50 regression fixer's bounty within the next 7 days. If not, it will be recorded as a project donation.

https://hledger.org/REGRESSIONS.html#regression-bounty

PR #2228, opened 2024-09-12
compoundBalanceReportAsHtml: create according TH cells only if row-total or average is requested

thielema:
fixes bug #2225


simonmichael:
Wow! I replied too soon. Thank you very much, I'll also test soon.

Issue #2227, opened 2024-09-12
hledger-web register does not show From/To accounts


thielema:
The Register page of current hledger-web-40 with an inacct query does not show To/From accounts for many transactions, even if there are only two accounts involved in the transaction, i.e. the To/From accounts column needs to display only one account.

Last time it worked correctly in hledger-web-1.32, but fails in hledger-web-1.33.

thielema:
On Thu, 12 Sep 2024, thielema wrote:

> Last time it worked correctly in hledger-web-1.32, but fails in hledger-web-1.33.

With git-bisect I identified the commit 
2a99b3d45634f26cd62745ddce6136361001b3f8.

thielema:
Problem is that the rewritten code with `tailDef` always removes the last account, whereas the original intention was to replace all accounts by a comma except the last one.
I have a patch here using NonEmpty. I'll submit it shortly.

simonmichael:
Ouch, that's quite an obvious regression. It seems to suggest there aren't that many active hledger-web users..

Thanks for the report and fix! Please claim your $100 finder+fixer https://hledger.org/REGRESSIONS.html#regression-bounty when convenient.

thielema:
On Tue, 24 Sep 2024, Simon Michael wrote:

> Ouch, that's quite an obvious regression. It seems to suggest there aren't that many active hledger-web users..

That would be sad. I use it frequently.

However, I missed the bug for several months because I do not always 
update to latest hledger when I am not looking for new features. It might 
be that other users just use hledger-web from their GNU/Linux distribution 
and the bug might not have been shipped with Debian&Co.

simonmichael:
That's true, many users upgrade infrequently.

simonmichael:
Hi @thielema, I encourage you to claim your $100 regression finder and fixer's bounty within the next 7 days. If not, it will be recorded as a project donation.
https://hledger.org/REGRESSIONS.html#regression-bounty

wjdp:
![Image](https://github.com/user-attachments/assets/e2c80821-c5f4-42b8-8b41-6454468ea2cb)

Is this the reason the "To/From Account(s)" column is always empty?

PR #2230, opened 2024-09-13
web: RegisterR.getRegisterR.addCommas: do not drop last account

thielema:
Bug was introduced in commit 2a99b3d45634f26cd62745ddce6136361001b3f8 in an effort to get rid of partial List.tail,
because GHC-9.8 started to warn about it.

Problem is that the rewritten code with tailDef always removes the last account, whereas the original intention was to replace all accounts by a comma except the last one. addCommas should prepare a comma separated list like List.intercalate.

### DIFF
--- a/hledger-web/Hledger/Web/Handler/RegisterR.hs
+++ b/hledger-web/Hledger/Web/Handler/RegisterR.hs
@@ -9,9 +9,9 @@
 
 module Hledger.Web.Handler.RegisterR where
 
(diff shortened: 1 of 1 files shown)


thielema:
The patch fixes issue https://github.com/simonmichael/hledger/issues/2227.

thielema:
Would you mind reviewing the commit? It is a simple fix to a simple bug.

PR #2232, opened 2024-09-14
Vertically merge cells showing account names and Total

thielema:
I think this is the best compromise for multibalance:

* In HTML there is only one merged Total cell.
* In CSV there is a Total header in every totals row.
* In FODS the Total header is contained in every row, but only the first one is shown in a merged cell. Formulas can still access the covered Total text e.g. for VLOOKUP.

Same applies to account names in `tidy` layout.

### DIFF
--- a/hledger/Hledger/Cli/CompoundBalanceCommand.hs
+++ b/hledger/Hledger/Cli/CompoundBalanceCommand.hs
@@ -376,7 +376,7 @@ compoundBalanceReportAsHtml ropts cbr =
             Total simpleDateSpanCell totalrow
                              -- make a table of rendered lines of the report totals row
         & map (map (fmap wbToText))
-        & zipWith (:) (Spr.defaultCell "Net:" : repeat Spr.emptyCell)
+        & addRowSpanHeader (Spr.defaultCell "Net:")
                              -- insert a headings column, with Net: on the first line only
         & addTotalBorders    -- marking the first for special styling
         & map (Html.formatRow . map (fmap L.toHtml))
--- a/hledger-lib/Hledger/Write/Html/Blaze.hs
+++ b/hledger-lib/Hledger/Write/Html/Blaze.hs
@@ -46,18 +46,33 @@ formatCell cell =
     let class_ =
             map (HtmlAttr.class_ . Html.textValue) $
             filter (not . Text.null) [Spr.textFromClass $ cellClass cell] in
+    let span_ makeCell attrs =
+            case Spr.cellSpan cell of
+                Spr.NoSpan -> foldl (!) makeCell attrs
+                Spr.Covered -> pure ()
+                Spr.SpanHorizontal n ->
+                    foldl (!) makeCell
+                        (HtmlAttr.colspan (Html.stringValue $ show n) : attrs)
+                Spr.SpanVertical n ->
+                    foldl (!) makeCell
+                        (HtmlAttr.rowspan (Html.stringValue $ show n) : attrs)
+            in
     case cellStyle cell of
-        Head -> foldl (!) (Html.th content) (style++class_)
+        Head -> span_ (Html.th content) (style++class_)
         Body emph ->
             let align =
                     case cellType cell of
                         TypeString -> []
                         TypeDate -> []
                         _ -> [HtmlAttr.align "right"]
-                valign = [HtmlAttr.valign "top"]
+                valign =
+                    case Spr.cellSpan cell of
+                        Spr.SpanVertical n ->
+                            if n>1 then [HtmlAttr.valign "top"] else []
+                        _ -> []
                 withEmph =
                     case emph of
                         Item -> id
                         Total -> Html.b
-            in  foldl (!) (Html.td $ withEmph content)
-                    (style++align++valign++class_)
+            in  span_ (Html.td $ withEmph content) $
+                style++align++valign++class_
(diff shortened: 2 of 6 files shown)


simonmichael:
I see there's one new commit here; maybe it also depends on #2226 ?
When you get a chance to rebase both I'll look again.

thielema:
Yes, it depends on https://github.com/simonmichael/hledger/pull/2226. I have rebased just now.

simonmichael:
@thielema, what's the problem this is solving, and the visible effects ?

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> @thielema, what's the problem this is solving, and the visible effects ?

There is a tension between nice appearance and the opportunity for further 
processing. Showing a row header once looks nice, but for further 
processing it is better to repeat the header in every row. There was 
recently the commit 9788a062233c2d18373b8e26ec695919eb78c06e that replaced 
repeated headers by single ones. This indicates for me that HTML shall 
look nice but is not intended for further processing.

I first thought about adding an option for controlling repetition of row 
headers. But then I thought that formats like CSV and FODS are for further 
processing and HTML is for human inspection, thus decided to repeat 
headers in the first two formats but not in HTML. In FODS we can have both 
features at once: Row headers can be repeated but only the first one is 
shown in a merged cell. The HTML equivalent of this is a merged cell, but 
HTML does not store covered cells.


If you prefer explicit control I suggest adding two options:

   --replicate-row-header = auto|once|duplicate

   --use-row-span - boolean option
      can be combined with any setting for replicate-header
      and affects formulas accessing covered cells.

simonmichael:
I'm a bit lost/forgetful, so I have questions. 

I cleaned up some visual glitch with html output a few months back, with 9788a06. 

I think I also tweaked some row headings in csv output also, in another commit.

Perhaps we should talk about html to start with. 
You're right that I haven't been thinking of it as something to be reused in other ways.
What are the changes to html output you propose, compared to 1.40 ?

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> Perhaps we should talk about html to start with.
> You're right that I haven't been thinking of it as something to be reused in other ways.

Btw. HTML output can also be re-used, because LibreOffice-Calc recognizes 
HTML tables and import them as such. However, detection of numbers does 
not work well. Thus, I'd prefer importing FODS or CSV to LibreOffice for 
further processing.


> What are the changes to html output you propose, compared to 1.40 ?

We have

$ hledger-1.40 -f hledger-journal balance --output-file=multibalance.html --layout=bare --yearly

It shows:

Total:	Euro
<blank>	USD


Whereas hledger from thielema/header-row-span shows:

Total:	Euro
<covered>	USD


In 1.40, "Total" and "<blank>" are two separate cells.

New is, "Total" and "<covered>" are one merged cell. "<covered>" is not 
shown.

You would see the difference if there were borders. In 1.40 there would be 
a line between "Total" and "<blank>". In the merged cell there would be no 
line.

simonmichael:
Thanks, that got me closer. I'm with you now:

Screenshot 2024-09-29 at 08 27 55

becomes

Screenshot 2024-09-29 at 08 28 13

The first two columns lost their alignment, let's fix that.

simonmichael:
And you've left csv output as-is (with the header repeated in each row, for ease of machine processing). And you've done something fancier with fods. Sounds great.

thielema:
On Sun, 29 Sep 2024, Simon Michael wrote:

> Thanks, that got me closer. I'm with you now:
> 
> Screenshot.2024-09-29.at.08.27.55.png (view on web)
> 
> becomes
> 
> Screenshot.2024-09-29.at.08.28.13.png (view on web)
> 
> The first two columns lost their alignment, let's fix that.

Interesting. The only difference between old and new output is that the 
cells got `valign="top"` attributes. This is important for vertically 
merged cells, but I expected that it has no effect on atomic cells. 
Surprisingly, valign also affects horizontal alignment.

simonmichael:
Looks nice - thank you.

The headings are now center aligned - it might be ok, was it intended ?

Old:

Screenshot 2024-09-29 at 20 58 52


New:

Screenshot 2024-09-29 at 21 00 01

thielema:
On Mon, 30 Sep 2024, Simon Michael wrote:

> The headings are now center aligned - it might be ok, was it intended ?

Now I see. It was not valign=top that made the difference.


The change came with commit 
https://github.com/thielema/hledger/commit/ff397f79cc70a0ad061ec7b591dc9985a9ac16c6#diff-eaee0cecc4187da2292f3659e7656b09b43a00e14759d40732b03a89d094b76a.

I had thrown away multiBalanceReportHtmlHeadRow, 
multiBalanceReportHtmlBodyRow that contained some custom alignment styles. 
I instead relied on default alignments. If you prefer the original 
alignments I can add them, again.

simonmichael:
I've seen advice that center-aligned headings are often best, but I probably preferred the content-aligned headings at some point in the past. Let's leave it for a bit and see how it feels!

hledger/Hledger/Cli/CompoundBalanceCommand.hs
```diff
@@ -376,7 +376,7 @@ compoundBalanceReportAsHtml ropts cbr =
             Total simpleDateSpanCell totalrow
                              -- make a table of rendered lines of the report totals row
         & map (map (fmap wbToText))
-        & zipWith (:) (Spr.defaultCell "Net:" : repeat Spr.emptyCell)
+        & addRowSpanHeader (Spr.defaultCell "Net:")
```
simonmichael, line 379: @thielema: At some point, this cell lost its "account" class, causing "Net" to no longer be left-aligned. I didn't see how exactly to fix it myself.

simonmichael: By the way you may have seen this already but: adding `td, th {border: thin solid red;}` in a `hledger.css` in the current is quite handy for troubleshooting.

Issue #117, opened 2013-05-31
wish: hledger should support transactions with timestamps, following ISO8601


tphyahoo:
Currently hledger transactions look like 

2009/1/1
    assets:cow_account      COW1   ; one cow at 100USD
    equity:cow_account      -COW1

The proposal is to allow transactions that have a precise timestamp such as 

2007-04-05T14:30Z
    assets:cow_account      COW1   ; one cow at 100USD
    equity:cow_account      -COW1

, where the Z indicates UTC time zone.

If no time zone is specified, the time zone could be inferred using rules in a config file, or directive.

simonmichael:
Do we have some use cases ?

Needs discussion/developers, closing for now.

adius:
Modern payment providers like PayPal, Bitcoin, etc. provide millisecond precision for their transactions. It should be possible to reflect that in a journal. This is an aboslute "must have" feature. I was really surprised that it's not already integrated!

simonmichael:
Hi, why is it "must have" ? 

When someone truly needs it, I expect we'll have it. For now, you can model this with a transaction tag.

ony:
I miss that feature only for merging multiple journals. Example I have some balance assertions/assignments from bank account in one journal and automatic export of expenses on my shop account in other. Without precise time in transactions they are not intermixed for a single day.
Work around: keep only one journal and periodically merge in transactions from generated ones.

If someone will start implementing it, I would like that person to consider treatment of transactions that are within the same day but with mix of those that have time and that don't.
I can imagine that sequence of transactions without time specification will be assigned to a time range that spans between closest pair of transactions with time specification in that but don't cross day boundary. During merge of transactions from a different sources that belong to overlapping ranges we can follow current approach (order of read) effectively preserving current behaviour.

ellis:
@simonmichael I would love this for bookkeeping as an Amazon seller because Amazon pays out settlements over time periods that begin and end sometime in the middle of the day.  We want to validate the settlement reports, and ISO8601 timestamps would certainly be the easiest way to handle such cases.  Adding a `time` tag seems like quite a hack, but if I use that, can you say what the query format should be for select the entries between two timestamps?

simonmichael:
This sounds like a good use case, but can you explain it in more detail ? Why must you have a timestamp ?

ellis:
[...]

PR #2233, opened 2024-09-16
imp: allow timestamp syntax in transaction dates

Stealthmate:
This commit makes it possible to specify timestamps in hledger transactions. For example: 2024-01-01 12:43:56+09:00. All existing logic that uses the transaction's date information has been left unchanged - that is to say, the time of day is not used anywhere yet. The intention behind this is to modify such logic in future commits/pull requests.

Furthermore, backwards compatibility with existing ledgers using plain dates has been fully preserved (barring any bugs that I missed).

ref: #117

### DIFF
--- a/hledger-lib/Hledger/Data/Balancing.hs
+++ b/hledger-lib/Hledger/Data/Balancing.hs
@@ -793,6 +793,7 @@ tests_Balancing =
                   0
                   ""
                   nullsourcepos
+                  Nothing
                   (fromGregorian 2007 01 28)
                   Nothing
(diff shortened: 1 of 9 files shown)


simonmichael:
Thanks for this @Stealthmate. When you have a chance, rebase it and fix tests and I'll check it out.

simonmichael:
Hi @Stealthmate. Do you have time to revisit this hledger PR: allow timestamp syntax in transaction dates ? It needs rebasing and test passing.

PR #2237, opened 2024-09-22
web: fix #2215 autocomplete in dynamically added account fields

### LINKED ISSUE #2215: UI bug in hledger web - autocomplete not working on the 5th account field and beyond
### How to reproduce:

1. Start hledger web
2. Add a transaction
3. On the form you should fill the first 4 account fields and amounts
4. The form should now show a 5th account field and amount
5. The 5th account field does not have autocompletion working

### Ledger version

hledger 1.34-g7a83578ec-20240601, windows-x86_64

### How severe is this bug ?

severity2

### Who is likely to be impacted by this bug ?

Since it appeared to fly under the radar for sometime, I believe only a small portion of the user base would be impacted. Most people would be satisfied with the first 4 account fields. So **impact3**.

### Proposed solution

I'm not a Haskell developer, nor do I know which web framework is being used, but the issue seems to be on the javascript side. When adding the new field, we should be destroying the typeahead instance and reloading it to include the new field. Since we are not doing this, the current typeahead instance doesn't know about the new field.

Looking at the source code, I reckon these 2 files should be affected:
- [add-form.hamlet](https://github.com/simonmichael/hledger/blob/master/hledger-web/templates/add-form.hamlet): This is where typeahead is instantiated. The instantiation code should probably be put inside a global function for it to be called from the other file, when adding a new field.
- [hledger.js](https://github.com/simonmichael/hledger/blob/master/hledger-web/static/hledger.js): I believe this is where we are adding a new field to the form. More specifically on the `addformAddPosting` function. Here we should be re-instantiating typeahead.

I could do a PR, but I really wouldn't know how to start the server for testing.

simonmichael:
A fix for #2215, autocomplete in account fields 5 and beyond, that's having no effect - maybe someone else can see why ?

### DIFF
--- a/hledger-web/static/hledger.js
+++ b/hledger-web/static/hledger.js
@@ -163,8 +163,8 @@ function addformReset(showmsg) {
 // Set the add-new-row-on-keypress handler on the add form's current last amount field, only.
 // (NB: removes all other keypress handlers from all amount fields).
 function addformLastAmountBindKey() {
-  $('.amount-input').off('keypress');
-  $('.amount-input:last').keypress(addformAddPosting);
+  $('input[name=amount]').off('keypress');
+  $('input[name=amount]:last').keypress(addformAddPosting);
 }
 
 // Pre-fill today's date and focus the description field in the add form.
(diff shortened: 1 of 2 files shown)


simonmichael:
Found it. Cloning the last form row also cloned typeahead's helper elements, and initialising typehead on the new account field then created too many new helper elements, causing confusion.

PR #2242, opened 2024-09-29
Web Balance Report

thielema:
This is a preview for a new feature of hledger-web.
It adds serving balance reports from hledger-web.

So far there remain some issues:
* I generate HTML with Lucid and then convert to Blaze. Instead I will add a new spreadsheet export backend Write.Html.Blaze. We can later discuss whether we want to keep Lucid, at all.
* Balance reports are currently integrated in the `defaultLayout`, which means there is always the sidebar. Do we want that? Of course, the user can hide the sidebar with `s` key. However, I found the query field useful in conjunction with the balance reports.
* Links to the balance report are only integrated in the Journal view for now. I could also move them to the query field component. Or above the Journal link in the sidebar.
* In a further commit I want to add radio buttons for the Layout type, maybe also the format type (TXT, CSV, FODS, HTML), maybe turn the period selection into a radio button, as well.
* The sidebar link to `Journal` does not maintain the current query. In the past I did it this way in order to have quick way out of restrictions posed by queries. However, all other accounts in the sidebar in the query and the hyperlinks behind dates in transactions in the `Register` page also maintain the query. Since I currently have the links to balance reports on the Journal page it would be nice if the top Journal link in the sidebar would also contain the query.

### DIFF
--- a/hledger-web/hledger-web.cabal
+++ b/hledger-web/hledger-web.cabal
@@ -144,6 +144,7 @@ library
   other-modules:
       Hledger.Web.App
       Hledger.Web.Handler.AddR
+      Hledger.Web.Handler.BalanceR
       Hledger.Web.Handler.EditR
       Hledger.Web.Handler.JournalR
       Hledger.Web.Handler.MiscR
@@ -187,6 +188,7 @@ library
     , http-client
     , http-conduit
     , http-types
+    , lucid
     , megaparsec >=7.0.0 && <9.8
     , mtl >=2.2.1
     , network
--- a/hledger-web/package.yaml
+++ b/hledger-web/package.yaml
@@ -131,6 +131,7 @@ library:
   - http-conduit
   - http-client
   - http-types
+  - lucid
   - megaparsec >=7.0.0 && <9.8
   - mtl >=2.2.1
   - network
--- a/hledger-web/templates/default-layout.hamlet
+++ b/hledger-web/templates/default-layout.hamlet
@@ -19,6 +19,8 @@ $if elem ViewPermission perms
     <form#searchform.input-group method=GET>
       <input .form-control name=q value=#{qparam} placeholder="Search"
         title="Enter hledger search patterns to filter the data below">
+      $maybe period <- maybePeriod
+        <input hidden name=period value=#{period}>
       <div .input-group-btn>
         $if not (T.null qparam)
           <a href=@{here} .btn .btn-default title="Clear search terms">
--- a/hledger-web/templates/journal.hamlet
+++ b/hledger-web/templates/journal.hamlet
@@ -6,6 +6,20 @@ $if elem AddPermission perms
      data-toggle="modal" data-target="#addmodal" title="Add a new transaction to the journal">
     Add a transaction
 
+<p>
+  Report:
+  <a href=@?{ballink} title="Show balance report">Balance
+  <a href=@?{multiballink "yearly"}
+      title="Show daily multi-period balance report">Yearly
+  <a href=@?{multiballink "quarterly"}
+      title="Show daily multi-period balance report">Quarterly
+  <a href=@?{multiballink "monthly"}
+      title="Show daily multi-period balance report">Monthly
+  <a href=@?{multiballink "weekly"}
+      title="Show daily multi-period balance report">Weekly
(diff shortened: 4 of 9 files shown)


simonmichael:
I played around with this. Exciting to see changes in hledger-web, and especially multi-period reports!

This one seems to be a balance changes report, like the `balance` command's default mode.

With hledger-web I have always been aiming for an easy UI useful to non technical users, that's simple enough to maintain and keep robust. It doesn't look like much but it has had a lot of rough edges smoothed off (some remain..). 

It has also probably grown quite a bit of tech debt, and may be hard to extend in big ways without some reorg or even a fresh approach. I haven't focussed on it in some time though.

I am wondering about having an experimental mode, so that we could try out new stuff more freely without creating churn for regular users. Or even a separate binary, hledger-web-x, if needed.

simonmichael:
PS shall we change this PR's status to draft.

thielema:
Yes, I think draft is the right status. However, it seems I cannot change the status myself.

simonmichael:
Hi @thielema, any new thoughts on this PR ? Are you using it ?

simonmichael:

hledger-web/templates/default-layout.hamlet
```diff
@@ -19,6 +19,8 @@ $if elem ViewPermission perms
     <form#searchform.input-group method=GET>
       <input .form-control name=q value=#{qparam} placeholder="Search"
         title="Enter hledger search patterns to filter the data below">
+      $maybe period <- maybePeriod
```
acinader, line 22: should the input be conditional, so it is only set when the param exists?

hledger-web/templates/journal.hamlet
```diff
@@ -6,6 +6,20 @@ $if elem AddPermission perms
      data-toggle="modal" data-target="#addmodal" title="Add a new transaction to the journal">
     Add a transaction
 
+<p>
+  Report:
+  <a href=@?{ballink} title="Show balance report">Balance
```
acinader, line 11: The titles are wrong, right? They all say "show daily multi-period balance report",

hledger-web/hledger-web.cabal
```diff
@@ -181,12 +182,13 @@ library
...
     , hjsmin
-    , hledger >=1.50.99 && <1.51
-    , hledger-lib >=1.50.99 && <1.51
+    , hledger >=1.43.99 && <1.44
+    , hledger-lib >=1.43.99 && <1.44
     , hspec
     , http-client
     , http-conduit
     , http-types
+    , lucid
```
acinader, line 191: not used

acinader:
image

image


Column title alignment and column spacing need work

hledger-web/package.yaml
```diff
@@ -131,6 +131,7 @@ library:
   - http-conduit
   - http-client
   - http-types
+  - lucid
```
acinader, line 134: remove

acinader (commented):
Nice!!

1. We shouldn't show the account list side column in this context
2. This is one of many reports that we could expose

We should decouple coding up new views from figuring out how to work them into the existing hledger-web. It would be good to have a solid toolbox of reports and graphs, and then figure out how to present them as a whole.

If we take this as an example, we could:
- fix it up, comment it well, and cover  it with appropriate tests
- document its existence, document accepted query string parameters

With that as a model, we could add other reports like income statement and ???

Once we have worked through a few that we have deployed and used, we could figure out what to do with them: hledger-web-x or ???

My proposed todo list:
- ~~reopen this pr @simonmichael~~ <-- i'm gonna open a new draft pr instead.
- ~~rebase and test against main @acinader~~
- ~~test coverage @acinader~~
- ~~candidate view document/README @acinader~~
- ~~open new pr @acinader~~ 
- code review @thielema

acinader:
image

result of a quick google for good tabular presentation of accounting data by period.

acinader:
> image image
> Column title alignment and column spacing need work

ok. i now see that this is not how you intended them to look.

acinader:
image

I think this is what you intended, but the CSP header was blocking the inline styles....

hledger-web/templates/journal.hamlet
acinader: fixing

PR #2243, opened 2024-09-29
lib: Write.Html.Attribute.tableStyle: style definitions taken from Balance.multiBalanceReportAsHtml

thielema:
avoid duplication


simonmichael:
Superseded by #2244.

Issue #1846, opened 2022-03-23
tree-mode balance reports do not show account hierarchy properly in HTML output


simonmichael:
$ hledger bal --tree  # subaccounts are properly indented
$ hledger bal --tree -o a.html  # subaccounts are not indented

simonmichael:
FYI @thielema, in case it affects FODS output.

thielema:
I have also encountered the missing indentation when experimenting with a LaTeX output.

simonmichael:
The reports at https://github.com/simonmichael/hledger_finance have been updated, showing recent HTML improvements including a tree-mode report.

PR #2246, opened 2024-09-30
Spreadsheet tree mode

thielema:
A proposal to fix https://github.com/simonmichael/hledger/issues/1846

### DIFF
--- a/hledger/Hledger/Cli/Commands/Balance.hs
+++ b/hledger/Hledger/Cli/Commands/Balance.hs
@@ -1265,7 +1265,7 @@ budgetReportAsCsv ropts report
 budgetReportAsSpreadsheet ::
   ReportOpts -> BudgetReport -> [[Ods.Cell Ods.NumLines Text]]
 budgetReportAsSpreadsheet
-  ReportOpts{..}
+  ropts@ReportOpts{..}
   (PeriodicReport colspans items totrow)
   = (if transpose_ then Ods.transpose else id) $
 
@@ -1279,14 +1279,18 @@ budgetReportAsSpreadsheet
   ) :
 
   -- account rows
-  concatMap (rowAsTexts Value prrFullName) items
+  concatMap (\row -> rowAsTexts Value (accountCell row) row) items
 
   -- totals row
   ++ addTotalBorders
-        (concat [ rowAsTexts Total (const totalRowHeadingBudgetCsv) totrow | not no_total_ ])
+        (concat [ rowAsTexts Total (cell totalRowHeadingBudgetCsv) totrow | not no_total_ ])
 
   where
     cell = Ods.defaultCell
+    accountCell row =
+        let name = prrFullName row in
+        setAccountAnchor (balance_base_url_) querystring_ name $
+        cell $ renderPeriodicAcct ropts nbsp row
     {-
     ToDo: The chosen HTML cell class names are not put in stone.
     If you find you need more systematic names,
@@ -1298,17 +1302,18 @@ budgetReportAsSpreadsheet
         maybe Ods.emptyCell (fmap wbToText . curry (cellFromMixedAmount oneLineNoCostFmt) cls) mval
 
     rowAsTexts :: RowClass
-               -> (PeriodicReportRow a BudgetCell -> Text)
+               -> Ods.Cell Ods.NumLines Text
                -> PeriodicReportRow a BudgetCell
                -> [[Ods.Cell Ods.NumLines Text]]
-    rowAsTexts rc render row@(PeriodicReportRow _ as (rowtot,budgettot) (rowavg, budgetavg))
-      | layout_ /= LayoutBare = [accountCell : map showNorm vals]
-      | otherwise =
-            addRowSpanHeader accountCell -- add name
-          . zipWith (:) (map cell cs)   -- add symbols
+    rowAsTexts rc acctCell (PeriodicReportRow _ as (rowtot,budgettot) (rowavg, budgetavg)) =
+      addRowSpanHeader acctCell $
+      case layout_ of
+        LayoutBare ->
+            zipWith (:) (map cell cs)   -- add symbols
           . transpose                   -- each row becomes a list of Text quantities
           . map (map (fmap wbToText) . cellsFromMixedAmount dopts . second (fromMaybe nullmixedamt))
           $ vals
+        _ -> [map showNorm vals]
       where
         cs = S.toList . mconcat . map maCommodities $ mapMaybe snd vals
         dopts = oneLineNoCostFmt{displayCommodity=layout_ /= LayoutBare, displayCommodityOrder=Just cs, displayMinWidth=Nothing}
@@ -1319,11 +1324,6 @@ budgetReportAsSpreadsheet
             ++ concat [[(rowAverageClass rc, rowavg),
                         (budgetAverageClass rc, budgetavg)]
                             | average_]
-
-        accountCell =
-            let name = render row in
-            setAccountAnchor (guard (rc==Value) >> balance_base_url_)
-                querystring_ name (cell name)
 
 
 nbsp :: Text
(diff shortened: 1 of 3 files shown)


thielema:
This changes the text output from spaces to non-breakable spaces. The visual appearance is the same, but the character is different. Thus the tests fail.
Shall I alter the tests or shall I replace non-breakable spaces by regular ones for text output?

thielema:
I have maintained a FixMe comment on `renderacct` without knowing, whether it still applies.

I see, Tree mode in bugdet report is still missing.

simonmichael:
Fast work, thank you very much!

`balance` with no report interval, also is not yet implemented.

All the rendering code paths are unfortunate. I would have liked to drop the no-report-interval balance reports, but kept them for familiarity to Ledger users and because they are still sometimes useful.

simonmichael:
PS if you don't get to implementing all cases, I'll take whatever you've done.

simonmichael:
When you said non-breaking space, I assumed the html entity `&nbsp;`, used only in HTML output. But maybe you are using a unicode non-breaking space in all output formats. I do think we should avoid that.

thielema:
On Mon, 30 Sep 2024, Simon Michael wrote:

> balance with no report interval, also is not yet implemented.

What do you mean with "balance with no report interval"?

If I run `hledger balance` without a date query, I get a result. Can you 
give an example?

thielema:
On Mon, 30 Sep 2024, Simon Michael wrote:

> When you said non-breaking space, I assumed the html entity &nbsp;, used 
> only in HTML output. But maybe you are using a unicode non-breaking 
> space in all output formats.

I use them consistently in TXT, CSV and HTML&Co. output.

> If so I do think we should avoid that.

I am about to avoid them in TXT, but still struggle with budget reports.

In CSV non-breakable spaces should prevent spreadsheet processors from 
stripping away leading spaces on CSV import.

simonmichael:
> What do you mean with "balance with no report interval"?

I meant: `hledger balance --tree -o a.html` (no `--yearly`, `--monthly` etc.) also does not yet show indentation

thielema:
On Mon, 30 Sep 2024, Simon Michael wrote:

>       What do you mean with "balance with no report interval"?
> 
> I meant: hledger balance --tree -o a.html (no --yearly, --monthly etc.) also does not yet show indentation

This is already solved in my pull-request draft.

thielema:
Now use space for indentation in Text mode. Budget reports not yet ready.

thielema:
The tests
hledger/test/balance/balance.test:8
hledger/test/balance/drop.test:4  (q vs j:q)
fail. I think the expected output is not correct, because not in tree format.

thielema:
multiBalanceReportAsTable.renderacct always adds indentation, as required for tree mode. This is because multiBalanceReport already distinguishes between flat and tree mode. This is inconsistent with simple balance and budget reports, where tree mode is handled in the output functions.

thielema:
I am puzzled. Here is a test failure:
~~~~
:hledger/test/balance/balance.test:8: [Failed]
Command (at line 146):
hledger -f - balance -N --output-format=csv --tree
Expected stdout: 
"account","balance"
"Assets","$-1, ""AB.C"" 1"
"Assets:ABC","""AB.C"" 1"
"Assets:Cash","$-1"

Got stdout:      
"account","balance"
"Assets","$-1, ""AB.C"" 1"
"ABC","""AB.C"" 1"
"Cash","$-1"
~~~~

That is, "ABC" and "Cash" are not indented. I can reproduce this result when running `shelltest` locally.
However, when I run hledger manually to produce CSV output, "ABC" and "Cash" are indented.
It might have been a problem of the failure reporting, however, when I alter `hledger/test/balance/balance.test`, then indented "ABC" yields a test failure, whereas unindented "ABC" is accepted.

I do not understand, why hledger generates indentation in CSV in the test environment, but not when run manually.

simonmichael:
Are you sure your manual test is correct ? Here I'm seeing both shell test and manual tests fail:

```
~/src/hledger$ COLUMNS=80 stack exec -- shelltest --exec hledger/test/balance/balance.test -i8
:hledger/test/balance/balance.test:8: [Failed]
Command (at line 146):
hledger -f - balance -N --output-format=csv --tree
Expected stdout: 
"account","balance"
"Assets","$-1, ""AB.C"" 1"
"Assets:ABC","""AB.C"" 1"
"Assets:Cash","$-1"

Got stdout:      
"account","balance"
"Assets","$-1, ""AB.C"" 1"
"ABC","""AB.C"" 1"
"Cash","$-1"


         Test Cases  Total 
 Passed  0           0     
 Failed  1           1     
 Total   1           1     
~/src/hledger$ stack exec -- hledger -f b.j balance -N -O csv -t
"account","balance"
"Assets","0"
"ABC","$10000.00"
"Cash","$-10000.00"
~/src/hledger$ j ghci
GHCi, version 9.10.1: https://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /Users/simon/.ghci
Loaded GHCi configuration from /Users/simon/src/hledger/.ghci
[ 1 of 95] Compiling Hledger.Utils.Regex ( hledger-lib/Hledger/Utils/Regex.hs, interpreted )
...
Ok, 95 modules loaded.
ghci> :main -f b.j balance -N -O csv -t
"account","balance"
"Assets","0"
"ABC","$10000.00"
"Cash","$-10000.00"
```

thielema:
On Thu, 3 Oct 2024, Simon Michael wrote:

> Are you sure your manual test is correct ? Here I'm seeing both shell test and manual tests fail:

I think I now see the problem. It seems that the depth parameter in 
BalanceReportItem is counted from 0 and PeriodicReportRow/prrDepth is 
counted from 1. Is this right? If yes, this is very unexpected.

simonmichael:
I think that's right. BalanceReport's docstring says

> the number of indentation steps for rendering a Ledger-style account tree

--depth's doc says

> will show accounts only to the specified depth

They are not the same - eg when boring parent accounts are squashed onto one line, an account of depth 2 or 3 might be displayed with indentation 0.

simonmichael:
And that prrDepth helper says "display depth", by which I meant "indent". Perhaps it should be renamed.

thielema:
On Thu, 3 Oct 2024, Simon Michael wrote:

> I think that's right. BalanceReport's docstring says
> 
> > the number of indentation steps for rendering a Ledger-style account tree
> 
> --depth's doc says
> 
> > will show accounts only to the specified depth
> 
> They are not the same - eg when boring parent accounts are squashed onto 
> one line, an account of depth 2 or 3 might be displayed with indentation 
> 0.


These are numbers used in the user interface. I am talking about the 
internal values. In module BalanceReport I found the line:

      , prrDepth row - 1  -- BalanceReport uses 0-based account depths

The problem seems to be known. But not to me, so far.

simonmichael:
prrDepth is not actually an account depth, it is a display indent. I should rename it.

Account depth is derived from the account name. Display indent is about presentation; it is intended to help renderers generate output like Ledger's original tree format, where boring parents are elided, without having to understand that eliding logic. In fact it used to be included as a field in CSV output for that reason. 

Eg in the balance commands default text output here, depth 1 assets is displayed with 0 indents and depth 3 assets:bank:saving is displayed with 1 indent.

```
$ hledger -f examples/sample.journal bal -Y -t
Balance changes in 2008:

                   || 2008 
===================++======
assets            ||  $-1 
   bank:saving     ||   $1 
...
```

I hope it makes sense.

thielema:
On Thu, 3 Oct 2024, Simon Michael wrote:

> prrDepth is not actually an account depth, it is a display indent. I should rename it.

Sure, but it is still one-based, whereas 'dep' in BalanceReportItem is 
zero-based. Both values are indentations, right?

simonmichael:
Sorry, I am short of time now, I will look again with a fresh mind before answering.

simonmichael:
(PS: if the balance report is being inconsistent here, it might originate with MultiBalanceReport.displayedAccounts).

simonmichael:
I see you are right, and I have clarified and fixed this old off-by-one field, after merging this PR. Thank you!

PR #2249, opened 2024-10-01
budgetReportAsSpreadsheet: vertically merge duplicate account name cells

thielema:
It might appear less readable. In this case we might use the alternating grey bars as in `compoundBalanceReportAsHtml`.


simonmichael:
This seems like more vertical merging.. I trust it's better. Thanks!

thielema:
On Wed, 2 Oct 2024, Simon Michael wrote:

> This seems like more vertical merging.. I trust it's better. Thanks!

I am not sure myself, anymore. In my example with layout=bare I get for 
some accounts two lines, where one line consists entirely of zeros. This 
does not look right.

simonmichael:
Sorry, I lost context on this. If you have time, maybe you could show a small repro ?

thielema:
On Mon, 14 Oct 2024, Simon Michael wrote:

> Sorry, I lost context on this. If you have time, maybe you could show a 
> small repro ?

I think my observed problem is a separate issue. I opened a new ticket 
#2264.

PR #2253, opened 2024-10-05
new layout bare-wide

thielema:
This is a draft for a new type of layout. This is the original reason why I started to work on hledger export functions ...

I called it `bare-wide`, because it is like `bare` but the amounts are put in one column per commodity. Sometimes it requires two header rows and horizontally merged headers for dates. I implemented it for HTML and FODS output of plain balance, multi-period balance and compound balances. Not implemented for budget reports, but those reports support only bare and non-bare layouts, anyway.

I thought about alternative names like `sub` (one sub-column per commodity), `columns`, `separate`. Maybe you have better ideas.

I have also some draft code for a table type with labels for header groups. This could enable layout bare-wide for the text outputs. This draft code is not contained in this pull request.

I found a command in `multiBalanceReportHtmlRows` that reads:
~~~~
-- TODO: should the commodity_column be displayed as a subaccount in this case as well?
~~~~
I do not know whether it is still relevant. I have moved it to `compoundBalanceReportAsHtml` but it might be even more confusing there.

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -120,6 +120,7 @@ instance Default AccountListMode where def = ALFlat
 data Layout = LayoutWide (Maybe Int)
             | LayoutTall
             | LayoutBare
+            | LayoutBareWide
             | LayoutTidy
   deriving (Eq, Show)
 
@@ -374,6 +375,7 @@ layoutopt rawopts = fromMaybe (LayoutWide Nothing) $ layout <|> column
                      , ("tall", LayoutTall)
                      , ("bare", LayoutBare)
                      , ("tidy", LayoutTidy)
+                     , ("bare-wide", LayoutBareWide)
                      ]
         -- For `--layout=elided,n`, elide to the given width
         (s,n) = break (==',') $ map toLower opt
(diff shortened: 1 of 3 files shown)


simonmichael:
Interesting! I will try to use it, when it can also display the date headings again. (Perhaps this PR should be marked draft.)

simonmichael:
I don't remember what that TODO comment was about, we can drop it.

thielema:
On Wed, 9 Oct 2024, Simon Michael wrote:

> Interesting! I will try to use it, when it can also display the date 
> headings again. (Perhaps this PR should be marked draft.)

yes, currently it is a draft

thielema:
On Wed, 9 Oct 2024, Simon Michael wrote:

> Interesting! I will try to use it, when it can also display the date headings again.

hledger -f hledger-journal balance --yearly --layout=bare-wide --output-file=multibalance.html


This shows date headings. Or what do you mean?

thielema:
On Wed, 9 Oct 2024, Simon Michael wrote:

> Ah, I was testing terminal output. No date headings there.

Text output does not support the new layout for now, because this would 
require a Table type that allows to give labels to Header groups. (I have 
a draft for such an alternative Table type, though.)

simonmichael:
It is affected by --layout=bare-wide though, just FYI.

simonmichael:
Any new thoughts on this hledger PR: new layout bare-wide ? Is the feature still valuable ?

simonmichael:

thielema:
I have updated the pull request to the latest commits at the main branch.
I am using the new layout type for two years now and it allows for postprocessing in LibreOffice. Together with the option `--full-names` and `--declared`, LibreOffice' `VLOOKUP` function now works nicely. So I kindly ask to merge the feature now.

simonmichael:
I think you haven't pushed yet ?

simonmichael:
And, it seems I can't reopen this because master was renamed to main. Please use a new pull request, with updated docs in "Balance report layout".

thielema:
On Tue, 11 Aug 2026, Simon Michael wrote:

> [10144?s=20&v=4] simonmichael left a comment (simonmichael/hledger#2253)
> 
> I think you haven't pushed yet ?

I have now included five recent commits, and pushed to the pull-request, 
again.

PR #2255, opened 2024-10-05
dev: remove the dead link: hledger/embeddedfiles/hledger-journal.md

basvandijk:
`hledger/embeddedfiles/hledger-journal.md` was pointing to the non-existing `../../doc/tldr/hledger-journal.md` so this commit removes that dead link.


simonmichael:
Thanks!

Issue #2251, opened 2024-10-04
balancesheet --row-total with HTML output lacks data in the Total column


thielema:
While working on `compoundBalanceReportAsHtml` I noticed that
~~~~
$ hledger balancesheet --yearly --row-total --output-file=balancesheet.html
~~~~
shows a `Total` header but misses data in the `Total` column. In Text output the `Total` header is absent, so header and data is consistent.
Missing data also means, that `--row-total --average` together put average values into the total column.
The problem exists at least since hledger-1.28.

This is certainly related to: https://github.com/simonmichael/hledger/issues/2225

Is there a reason to omit the `Total` column in `balancesheet`? In `incomestatement` it is present.
I think the simplest thing would be to support total columns in all flavors of `CompoundBalance`.

simonmichael:
Thanks for the report. Yes, it was intended that -T/--row-total be ignored by balancesheet and balancesheetqeuity, since they show end balances and summing those is not meaningful. This was changed in hledger 1.23 apparently, though not mentioned in release notes. But the headings in html output were overlooked until now.

thielema:
I see there is the `totalscolumn` variable, that is `False` for Historical and Cumulative reports. I think I can fix the problem.

I am still uncertain, why are `balancesheet` and `incomestatement` different in this respect.

simonmichael:
Think of a report like `hledger is -M` (`hledger incomestatement --monthly`). Each column shows revenues/expenses in the period, and it makes sense to add those up and see a total across all periods. The same is not true for `bs`/`bse`, where each column shows an end balance at the end of each period, showing a total of those does not make sense. And of the many sources of confusion for new accountants, this is one we can prevent, by not showing it.

thielema:
On Wed, 9 Oct 2024, Simon Michael wrote:

> The same is not true for bs/bse, where each column shows an end balance 
> at the end of each period - showing a total of those does not make 
> sense. And of the many sources of confusion for new accountants, this is 
> one we can prevent, by not showing it.

I see. I would prefer if invalid option combinations would be rejected by 
the option parser. But the current policy seems to be to ignore 
conflicting flags. With optparse-applicative we could express such logic 
connections. However, this would be a major rewrite and it would also 
affect RecordOpts et.al. since it should comply with the flag logic.

simonmichael:
> I see. I would prefer if invalid option combinations would be rejected by the option parser. 

[...]

PR #2256, opened 2024-10-09
cli: CompoundBalanceCommand: row_total_ -> multiBalanceHasTotalsColumn

thielema:
this should fix https://github.com/simonmichael/hledger/issues/2251


simonmichael:
Thank you!

PR #2257, opened 2024-10-12
Fix all scripts in bin/

adept:
I've fixed the compilation errors in all the scripts in bin/

This necessitatied adding `budgetReportAs...` to the export list of Commands.Balance - I think there are no special reasons for them being omitted, as all the other reporting functions are exported.


simonmichael:
Many thanks!

Issue #2247, opened 2024-09-30
check accounts warns about --infer-equity's accounts, check tags warns about _price-matched


PainFile:
Hello everyone, I currently use hledger to track my finances, I normally do Strick checks to my files once every month to make sure that everything is ok. With the release of 1.40 I notice that **check accounts** and **check tags** send some errors with my information (not present on 1.34 or lower).  

This is the MRE I can come up with:

```
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
; Start of Minimal, Reproducible Example (MRE).

; Commodities
commodity "CD200130"
  format 1,000.00000000 "CD200130"

commodity MXN
  format 1,000.00 MXN

; Accounts
account Assets:Investing:Bonds:CetesDirecto:Invested:Cetes
account Assets:Investing:Bonds:CetesDirecto:Wallet

2024/01/01 * Test transaction
    Assets:Investing:Bonds:CetesDirecto:Invested:Cetes     100 "CD200130" @ 9.94392220 MXN
    Assets:Investing:Bonds:CetesDirecto:Wallet

; End of MRE.
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```

I run on **Windows 11**. Under **hledger 1.34-g7a83578ec-20240601, windows-x86_64** if I go with "_.\bin\hledger.exe -f test.journal check -s_" nothing comes out, i. e. everyting is OK.

If I change to **hledger 1.40-g0f31974c5-20240909, windows-x86_64** still under **Windows 11** and I run the same command I got the following error:

> hledger.exe: Error: C:\Temp\test.journal:17:
> 17 | 2024-01-01 * Test transaction
>    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>    |     Assets:Investing:Bonds:CetesDirecto:Invested:Cetes    100 "CD200130" @ 9.94392220 MXN
>    |     Assets:Investing:Bonds:CetesDirecto:Wallet                          -994.39222000 MXN
> 
> Strict account checking is enabled, and
> account "equity:conversion:CD200130-MXN:CD200130" has not been declared.
> Consider adding an account directive. Examples:
> 
[...]

PR #2262, opened 2024-10-15
simpler conversion account with --infer-equity

simonmichael:
This implements today's proposal on ~~mail list~~ & chat:
[Actually, it never reached the mail list; never mind.]

> The --infer-equity flag generates postings to accounts named something like `equity:conversion:A-B:A` and `equity:conversion:A-B:B`, where A and B are the commodity symbols being converted between. 
> 
> I feel these :A-B:* subaccounts 
> 
> - add noise 
> 
> - complicate exporting to beancount, since currency symbols in account names generally aren't allowed there. 
> 
> - complicate hledger check accounts and check -s, as seen in https://github.com/simonmichael/hledger/issues/2247 
> 
> - are not much needed; you can always filter commodities with cur: 
> 
> So I think I’d like to drop the subaccounts. —infer-equity would just use equity:conversion (or other conversion account you have designated). Do you see any problems with it ?

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -96,11 +96,13 @@ module Hledger.Data.Journal (
   journalAccountTypes,
   journalAddAccountTypes,
   journalPostingsAddAccountTags,
+  defaultConversionAccount,
   -- journalPrices,
   journalConversionAccount,
   journalConversionAccounts,
   -- * Misc
   canonicalStyleFrom,
+
   nulljournal,
   journalConcat,
   journalNumberTransactions,
--- a/hledger-lib/Hledger/Data/JournalChecks.hs
+++ b/hledger-lib/Hledger/Data/JournalChecks.hs
@@ -67,7 +67,6 @@ journalCheckAccounts j = mapM_ checkacct (journalPostings j)
           ,"Consider adding an account directive. Examples:"
           ,""
           ,"account %s"
-          ,"account %s    ; type:A  ; (L,E,R,X,C,V)"
           ]) f l ex (show a) a a
         where
           (f,l,_mcols,ex) = makePostingAccountErrorExcerpt p
--- a/hledger-lib/Hledger/Data/JournalChecks.hs
+++ b/hledger-lib/Hledger/Data/JournalChecks.hs
@@ -58,7 +58,7 @@ journalCheckAccounts :: Journal -> Either String ()
 journalCheckAccounts j = mapM_ checkacct (journalPostings j)
   where
     checkacct p@Posting{paccount=a}
-      | a `elem` journalAccountNamesDeclared j = Right ()
+      | a `elem` declaredaccounts = Right ()
       | otherwise = Left $ printf (unlines [
            "%s:%d:"
           ,"%s"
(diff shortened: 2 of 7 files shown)


Xitian9:
I don't have a strong personal opinion on whether these currency tags should be included, but originally added them because they were requested and the person who requested presented a sufficiently convincing case for why they were beneficial. I can't remember exactly who, and have looked around github for a record. I think it may have been @the-solipsist. Perhaps the conversation occurred in the #hledger chat.

It would be good to dig up the reasons for this so we can weigh them against the benefits of simplification.

the-solipsist:
The discussions happened in December 2021:

https://matrix.to/#/!GYlETOAdUBSCcvnEMY:matrix.org/$QOHDt3va8CsIC4X0gIcrzcpr6s2a96drptM6-hDC44I?via=matrix.org&via=mozilla.org&via=envs.net

I prefer `COMM1-COMM2:COMM[1/2]` because it allows me to use the `balance` command to find out the gains from a particular investment.  E.g., COMM1 might be an ISIN for a mutual fund, and COMM2 might be the INR value I've invested.  The balance of the two (when converted into INR with `-X₹` with the latest price) gives me the gains.

This technique works across all PTA software, regardless of whether they have functions like `--gains` built-in or not.  Given that, I think it is worth preserving.

I don't think "noise" is a good enough reason to get rid of this. Automated conversion to Beancount is a better reason.  Beancount requires all account names to start with a capital Latin letter:

> An account name is a colon-separated list of capitalized words which begin with a letter, and whose first word must be one of five account types

So if a commodity starts with a number or a symbol (which none of my commodities do), then a seamless conversion won't work.  I'd suggest flagging this to users and allowing them to just use `--alias` with `print` to fix the few instances where that might be the case.

Having said that, I won't personally be affected by this change.

simonmichael:
Thank you both! (And chatters.)

I see that the commodity pair subaccount that --infer-equity generates, at least, can be quite useful. So I'm going to withdraw this PR.

What then should change about --infer-equity's account names, if anything ? I had two followup questions in chat:

> I think most utility comes from the :TRADINGPAIR subaccount; the deeper :TRADINGPAIR:COMM1 and TRADINGPAIR:COMM2 subaccounts are less necessary. They make it easier to precisely report eg "what's the current net cost in USD of my AACO holdings", and see just USD without AACO. Without those leaf accounts, the same report would require cur:USD.

sol agreed with this, though he likes to keep one commodity per leaf account (as I do, usually).

> --infer-equity needs a single english-language parent account name to use by default. What is the best choice for this ? I picked "equity:conversion" thinking that was most generic and clear. Would "equity:trade" be more familiar and recognisable in practice ?
(trade chosen over trades or trading, for being most generic and spellable)

Currently,

- equity:conversion is the base account used by --infer-equity
- When detecting equity conversion postings during transaction balancing, all of the following are recognised:
  equity:conversion, equity:trade, equity:trades, equity:trading.. in fact any account matching `"^equity:(trad(e|ing)|conversion)s?(:|$)"`;
  also any subaccounts of these four.

This is complicated, eg for `check accounts`. I feel we should probably simplify this to a single default conversion account name.
Any other name the user prefers can simply be declared with account type V (Conversion).

sol uses "equity:trading" (plus a bunch of custom subaccounts), but didn't feel strongly.
I am currently re-reading the excellent [Tutorial on multiple currency accounting](https://www.mscs.dal.ca/~selinger/accounting/tutorial.html) (Selinger 2005); he calls them "currency trading accounts".

If starting today I would probably call it equity:trade.
But equity:conversion is also fine and it's probably not worth changing unless other related things are changing too.
If you care, let me know.

What to do for `check accounts` ? I suggest the default conversion account name, and any subaccounts of that, are allowlisted so it never warns about those ?

What to do for `print -O beancount` ? It already converts recognised currency symbols to ISO-4217 currency codes. I'd like it to automatically perform the same translation to any account names of the form CONVERSIONACCT:COMM1-COMM2:COMM. Hopefully those can be detected sufficiently robustly.

simonmichael:
Closing this in favour of #2267.

PR #2266, opened 2024-10-17
Register export to HTML and FODS

thielema:
Unified export to CSV, HTML, FODS via Write.Spreadsheet.

Supports --base-url and hyperlinks to hledger-web.
I intended to hyperlink from txnidx to hledger-web/journal page, however, this would require to move Web.Widget.Common.transactionFragment to `hledger` package. It's not super important for now.

### DIFF
--- a/hledger-lib/Hledger/Write/Ods.hs
+++ b/hledger-lib/Hledger/Write/Ods.hs
@@ -69,6 +69,9 @@ printFods encoding tables =
           "    <number:text>-</number:text>" :
           "    <number:day number:style='long'/>" :
           "  </number:date-style>" :
+          "  <number:number-style style:name='integer'>" :
+          "    <number:number number:min-integer-digits='1'/>" :
+          "  </number:number-style>" :
           customStyles ++
           "</office:styles>" :
           []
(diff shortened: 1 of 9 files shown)


simonmichael:
Great!

Is Hledger.Cli.Anchor not used yet ? Should we leave it out ?

simonmichael:
We can update the table at https://hledger.org/dev/hledger.html#output-format too.

thielema:
On Thu, 17 Oct 2024, Simon Michael wrote:

> Is Hledger.Cli.Anchor not used yet ? Should we leave it out ?

It is used by Commands.Balance and Commands.Register.

There might be a better place and name for the module, though. What do you 
think?

simonmichael:
I guess it's fine for now, unless you think it could just be added to Hledger.Cli.Utils.

I'll merge this when you're ready.

thielema:
On Fri, 18 Oct 2024, Simon Michael wrote:

> I guess it's fine for now, unless you think it could just be added to Hledger.Cli.Utils.

I think a separate module is justified.


> I'll merge this when you're ready.

yes please, I am ready

simonmichael:
Thank you!

PR #2270, opened 2024-10-19
Print export to HTML and FODS

thielema:
As in `balance` and `register` and friends I generate CSV, HTML, FODS from the same Spreadsheet structure. This means that HTML output deviates from the text output of `print`. It presents data in a table and not in the journal format.


simonmichael:
Rebased to work around some conflict and merged as 3306abfd4. Thank you!

simonmichael:
Great additions to flesh out the output formats.

PR #2271, opened 2024-10-20
Aregister export to FODS

thielema:
This completes the matrix of output formats with respect to HTML and FODS.


simonmichael:
Excellent, thank you!

Issue #2077, opened 2023-08-27
Option --cumulative for `hledger aregister`


thielema:
I like to have support for a cumulative running total additional to the standard historical running total in `aregister`.

See https://groups.google.com/d/msgid/hledger/0F2587EB-DF9A-4CE7-9A95-3C13B8042589%40joyful.com .

thielema:
I thought it would be an easy addition since `Aregister.hs` already contains out-commented `cumulative` and `historical` options and `Register.hs` already supports `cumulative` totals. However, after adding the options they are still ignored.
Additionally I do not understand how `Register.hs` sets `cumulative` as default, because it mentions `cumulative` only in the flag section of `Register.registermode` and nowhere else and `def :: BalanceAccumulation` is defined as `PerPeriod`.
I can change the line in `aregister.ropts'` to `balanceaccum_= Cumulative` and get cumulative totals. Thus it seems to work in principle, but somehow the option `cumulative` is not translated to `Cumulative`, although `balanceAccumulationOverride` says how to do it.

simonmichael:
hledger:Hledger.Cli.Commands.* (Register.hs, Aregister, etc) are just the "front end" of commands. Usually they call one or more generic reports implemented in hledger-lib:Hledger.Reports (PostingsReport, AccountTransactionsReport etc). So eg in PostingsReport.hs you can see postingsReport checking balanceaccum_, which comes from the ReportOpts argument, which was gathered from the raw CLI arguments in main. To navigate the code, I can recommend VS Code + HLS. Otherwise an editor with up to date tags file or good project-wide search. Hope it helps.

simonmichael:
I found this old PR of yours. Is it still a feature that's needed / worth the effort ?

simonmichael:
Excuse me, not a PR but rather a discussion issue.

thielema:
On Tue, 24 Sep 2024, Simon Michael wrote:

> Excuse me, not a PR but rather a discussion issue.

It is still an issue. I currently use a local patch because I did not find 
a way to properly implement the feature. Currently postponed in favor of 
the table export issues.

thielema:
Solved by b2aad5216809de9725125cb1d6079b8aa2cee328

PR #2273, opened 2024-10-21
Aregister: support option --cumulative

thielema:
This addresses #2077.

Now I translate `PerPeriod` to `Historical`, since `PerPeriod` is the default. This is a way to have different defaults for `aregister` (Historical) and `register` (Cumulative).


thielema:
On Tue, 22 Oct 2024, Simon Michael wrote:

> I think 2207 is not the right link..

I have corrected it to #2077.

simonmichael:
"Transactions before the report start date are always included in the running balance (--historical mode is always on)." in the doc will need updating. Or I can do it if you prefer.

simonmichael:
Looks good, thank you!

PR #2274, opened 2024-10-23
aregister: implement --invert option

thielema:
The option was already added but outcommented. I extended `accountTransactionsReport` accordingly.

### DIFF
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -58,8 +58,9 @@ module Hledger.Data.Posting (
   conversionPostingTagName,
   costPostingTagName,
 
--- * arithmetic
+  -- * arithmetic
   sumPostings,
+  negatePostingAmount,
   -- * rendering
   showPosting,
   showPostingLines,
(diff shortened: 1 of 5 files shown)


thielema:
I have added another commit that turns new option `--no-header` into `--header=no`. Does not actually belong to here, I hope you don't mind.

simonmichael:
No problem. Thanks!

simonmichael:
@thielema while updating changelogs I remembered that "header" and "heading" are often confusable. I think "heading" would actually be the most correct here, per https://ell.stackexchange.com/questions/191301/are-header-and-heading-interchangable-in-the-context-of-information-at-the . How would you feel if we changed it to --heading=YN before release ?

thielema:
On Wed, 30 Oct 2024, Simon Michael wrote:

> @thielema while updating changelogs I remembered that "header" and 
> "heading" are often confusable. I think "heading" would actually be the 
> most correct here, 
> perhttps://ell.stackexchange.com/questions/191301/are-header-and-heading-interchangable-in-the-context-of-information-at 
> -the . How would you feel if we changed it to --heading=YN before 
> release ?

I am fine with any naming.

simonmichael:
--title could also work. I'll ponder it.

thielema:
On Thu, 31 Oct 2024, Simon Michael wrote:

> --title could also work. I'll ponder it.

Problem is, that text output inserts a title and HTML/CSV/FODS output 
inserts column headings.

simonmichael:
Hi @thielema, sorry for responding slowly, I didn't totally understand the implications. I'm going to merge #2277 and will get clearer on it. Thanks!

PR #2283, opened 2024-11-06
lib: fix division by zero in roi when all assets are sold (fixes #2281)

### LINKED ISSUE #2281: roi "Ratio has zero denominator" when --pnl account balances to 0
I am trying to use hledger-roi and keep track of average cost using something similar to the method described in this [multicurrency accounting tutorial](https://www.mscs.dal.ca/~selinger/accounting/tutorial.html#4). That means that both the cash and stock amounts should always balance to zero.  In order to do that, all buy and sell commodity transactions go through an Income:Trading account (simplified below to just Income).

This crashes roi, I think, because it can't handle the Income account evaluating to 0

```
$ cat roi-example.journal 
P 2020/1/1 A GBP 100
2020/1/1 foo
    Assets                                  GBP -100
    Assets                                         1 A
    Income                                        -1 A
    Income                                   GBP 100

2020/2/1 doesn't matter what's in this transaction
    Equity                                  GBP -100
    Assets                                   GBP 100

$ hledger roi -f roi-example.journal --inv Assets --pnl Income -V
hledger: Ratio has zero denominator
$ hledger roi -f roi-example.journal --inv Assets --pnl Income -V --cashflow

IRR cash flow for 2020-01-01 - 2020-02-01
+------------++----------+
|            ||   Amount |
+============++==========+
| 2020-02-01 || GBP -100 |
| 2020-02-02 ||  GBP 100 |
+------------++----------+


TWR cash flow for 2020-01-01 - 2020-02-01
hledger: Ratio has zero denominator


$ hledger bal -f roi-example.journal -EV
             GBP 100  Assets
            GBP -100  Equity
                   0  Income
--------------------
                   0  
```

adept:
TWR computation works by computing the running total of "investment units" and their cost. If at some point we run out of units, we could divide by zero trying to compute new cost. Instead lets just reset the cost the the initial one as we have essentially went to the initial state (no units on hand, cost is an arbitrary chosen number 100).

This fixes #2281

### DIFF
--- a/hledger/Hledger/Cli/Commands/Roi.hs
+++ b/hledger/Hledger/Cli/Commands/Roi.hs
@@ -239,7 +239,9 @@ timeWeightedReturn styles showCashFlow prettyTables investmentsQuery trans mixed
                Left pnl' ->
                  -- PnL change
                  let valueAfterDate = valueOnDate + unMix pnl'
-                     unitCost' = valueAfterDate/unitBalance
+                     unitCost' =
+                       if unitBalance == 0 then initialUnitCost -- everything was sold, let's reset the cost to initial cost
+                       else valueAfterDate/unitBalance
                  in (valueOnDate, 0, unitCost', unitBalance))
           (0, 0, initialUnitCost, initialUnits)
           $ dbg3 "changes" changes


adept:
@simonmichael I dont understand the CI run failure here, but I think it is unrelated to my change

PR #2292, opened 2024-11-23
Regular-expression limited depth limits

Xitian9:
Previously depth-limiting was universal across all accounts, e.g. all
accounts are clipped to depth 2. However, sometimes you want certain
accounts clipped to a different depth than others, e.g. all expenses to
depth 3, while all assets to depth 2. This commit enables depth-limiting
to optionally include a regular expression, which limits the accounts it
applies to.

More than one depth limit can be passed, and they are applied to each
account name by the following rules:
- If one or more regular-expression depth limit applies, use the
  smallest one
- If no regular-expression depth limits apply, and a flat depth limit is
  supplied, use that
- Otherwise, do not do any depth limiting

For example, this will clip all accounts matching "assets" to depth 3,
all accounts matching "expenses" to depth 2, and all other accounts to
depth 1.
--depth assets=3 --depth expenses=2 --depth 1

### DIFF
--- a/hledger-lib/Hledger/Data/Account.hs
+++ b/hledger-lib/Hledger/Data/Account.hs
@@ -175,11 +175,11 @@ clipAccounts d a = a{asubs=subs}
 -- | Remove subaccounts below the specified depth, aggregating their balance at the depth limit
 -- (accounts at the depth limit will have any sub-balances merged into their exclusive balance).
 -- If the depth is Nothing, return the original accounts
-clipAccountsAndAggregate :: Maybe Int -> [Account] -> [Account]
-clipAccountsAndAggregate Nothing  as = as
-clipAccountsAndAggregate (Just d) as = combined
+clipAccountsAndAggregate :: DepthSpec -> [Account] -> [Account]
+clipAccountsAndAggregate (DepthSpec Nothing []) as = as
+clipAccountsAndAggregate d                      as = combined
     where
-      clipped  = [a{aname=clipOrEllipsifyAccountName (Just d) $ aname a} | a <- as]
+      clipped  = [a{aname=clipOrEllipsifyAccountName d $ aname a} | a <- as]
       combined = [a{aebalance=maSum $ map aebalance same}
                  | same@(a:_) <- groupOn aname clipped]
 {-
--- a/hledger-lib/Hledger/Data/AccountName.hs
+++ b/hledger-lib/Hledger/Data/AccountName.hs
@@ -35,6 +35,7 @@ module Hledger.Data.AccountName (
   ,acctsepchar
   ,clipAccountName
   ,clipOrEllipsifyAccountName
+  ,getAccountNameClippedDepth
   ,elideAccountName
   ,escapeName
   ,expandAccountName
@@ -61,9 +62,10 @@ where
 
 import Control.Applicative ((<|>))
 import Control.Monad (foldM)
-import Data.Foldable (asum, toList)
+import Data.Foldable (asum, find, toList)
 import qualified Data.List.NonEmpty as NE
 import qualified Data.Map as M
+import Data.Maybe (mapMaybe)
 import Data.MemoUgly (memo)
 import qualified Data.Set as S
 import Data.Text (Text)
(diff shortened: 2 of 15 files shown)


Xitian9:
There are still some parts of the design space to be decided on here. As implemented, then if more than one regular expression matches a given account name, then the most aggressive depth clipping will be done (i.e. the minimum of the provided depths).

A more advanced method would be to choose the most specific regular expression and use that depth. I didn't originally implement it this way because I didn't see a straightforward way to evaluate which regular expression was more specific, but I think the following algorithm may work:

The *specificity* of a match is the first place in the chain of parents of the account where the regular expression starts matching. So for `assets:bank:savings`, the *specificity* of the regular expression `assets` would be 1, as it starts matching right at the first level. On the other hand, the *specificity* of the regular expression `savings` (or `assets:bank:savings`) would be 3, as this doesn't start matching until you get three levels in.

The proposed algorithm would then be to apply the depth limit for the most specific match. Or take the minimum if two regexps have the same specificity: we can't get around tie breakers entirely.

Xitian9:
I think the modified proposal is an unambiguous win, so I've made the change. Let me know if you disagree.

simonmichael:
Thanks @Xitian9 ! I'm sorry for the slow response. I was a little resistant.

This adds complexity; hopefully only for people who use it.

It also means we can infer less about account structure when reading hledger report output.

But it's something people ask for occasionally, and I'm sure when we want it we'll be happy to find it available. (What did you need it for ?)

I find the explanation of regexp specificity confusing, I think we could improve it. 

The Query types -> depth: section in the manual could be updated also.

These aren't blockers; I'll merge soon.

Xitian9:
Not a problem at all. A week isn't a particularly long lag time for PR response.

My use case is that I run a cashflow report. Because of the structure of my accounts, the most relevant information occurs at different depths for different accounts. My expenses are mostly at depth 2 (expenses:food), but for some I'd like more specificity (expenses:taxes:income:Xitian). My revenues are mostly at depth 5 (revenues:<country code>:Xitian:salary:<employer>).

I've previously handled this in a number of ways:
1. I sometimes add dummy parents to accounts just to get everything to the same depth.
2. I use account aliases to cut certain accounts to a given depth.

The account aliases are the main workhorse here, but since account aliases happen first it precludes being able to filter on anything that you want to depth clip. This requires a separate alias stage to first get rid of unwanted subaccounts before using the account alias depth limiting. But this is all very awkward and not at all natural. This feature has allowed me to transform from this:
```
      --alias "/liabilities:.*subacct:.*payable/=ignore" \
      --alias "/revenues:.*subacct/=ignore" \
      --alias "/(revenues|liabilities):([A-Z][A-Z]):(.*)/=\1:\3" \
      --alias "/(liabilities:BANK1:subacct:[^:]*):.*/=\1" \
      --alias "/(liabilities:BANK2:subacct:[^:]*):.*/=\1" \
      --alias "/(liabilities:BANK3:subacct):.*/=\1" \
      --alias "/(revenues:[^:]*):.*/=\1" \
      --alias "/(equity:conversion:[^:]*).*/=\1" \
      --alias "/expenses:(beauty|childcare|classes|clothing|entertainment|financial|gifts|home|medical|miscellaneous|professional|transport|travel|utilities|unknown):.*/=expenses:\1" \
      --alias "/expenses:taxes:income:([^:]*).*/=expenses:taxes:income:\1" \
      --alias "/(expenses:food:(groceries|restaurants)):.*/=\1" \
```

to this

```
      --depth revenues=6 --depth liabilities=5 --depth expenses=2 \
      --depth expenses:taxes=4 --depth expenses:housing=5 --depth expenses:food=3 \
```

which is much clearer and easier to understand.

Xitian9:
It may be worth calling out that there is a change of behaviour for some calls. I believe this falls into the "making behaviour consistent" bin, but I should check with you.

Previously, if you used more than one `--depth` option, the last one would be taken. However, if depth clipping were supplied with `depth:` options, then the most aggressive depth-clipping would be performed. If a combination of the two were used, then the minimum of all the `depth:` options and the last `--depth` option would be taken.

This changes the behaviour so that the last of any of the `depth:` and `--depth` options is always taken, fully acknowledging that order is not well-defined when we have a mix of `depth:` and `--depth` options.

Xitian9:
Is this clearer as a description?

In place of a single number which limits the depth for all accounts, you can also provide separate depth limits for different accounts using regular expressions. For example, `--depth assets=2` (or, equivalently: `depth:assets=2`) will collapse accounts matching the regular expression "assets" to depth 2. So "assets:bank:savings" would be collapsed to "assets:bank", while "liabilities:bank:credit card" would not be affected. This can be combined with a flat depth to collapse other accounts not matching the regular expression, so `--depth assets=2 --depth 1` would collapse "assets:bank:savings" to "assets:bank" and "liabilities:bank:credit card" to "liabilities".

You can supply multiple depth arguments and they will all be applied, so `--depth assets=2 --depth liabilities=3 --depth 1` would collapse:
  - accounts matching "assets" to depth 2,
  - accounts matching "liabilities" to depth 3,
  - all other accounts to depth 1.
 
If an account is matched by more than one regular expression depth limit, then the more specific one is used. For example, if `--depth assets=1 --depth assets:bank:savings=2` is provided, then "assets:bank:savings" will be collapsed to depth 2 rather than depth 1. This is because "assets:bank:savings" matches at depth 3 in the account name, while "assets" matches at depth 1. The same would be true with the argument `--depth assets=1 --depth savings=2`. 

The full rules for depth limiting are:
1. If exactly one regular expression depth argument matches an account name, use the depth limit provided in that argument.
2. If more than one regular expression depth argument matches an account name, use the depth limit provided in the argument of the most specific regular expression. This is the regular expression that requires the most levels of the account name in order to match.
3. If more than one regular expression depth argument matches and they are equally specific, then use the depth limit provided by the last one.
4. If no regular expression depth arguments match, apply the simple flat depth argument.

simonmichael:
much clearer, thank you. I like having the full detail in the reference manual, but in this case I’d be tempted to omit the “full rules” section - your call.

simonmichael:
> This changes the behaviour so that the last of any of the depth: and --depth options is always taken

Thanks for the info. I'm not sure about this - it seems to me that "rightmost wins" is natural for command line options, but not for hledger query terms, which can be combined in complex expressions.

Xitian9:
I've updated the manual with the suggested text, and also the "query types" section of the manual.

I'm happy for you to make a call on the "rightmost wins" versus "take the minimum approach". We could very well maintain the old behaviour for flat depths. Or would you want the "minimum" approach to also apply for regexp-depths?

simonmichael:
I feel we should keep the old semantics if it's not too hard -

- for command line options, rightmost wins is best - it means you can always customise an alias or complex command line by adding another option at the end

- for query expressions, combining query terms according to the expression's structure is best, since there's not always a simple right or left

simonmichael:
(And for depth: specifically, I felt the natural combination of `depth:N depth:M` is `depth:MIN(N,M)`, ie AND those constraints. I'm not sure if we handle things like `(depth:N AND ...) OR (depth:M AND ...)` properly (don't let me send you down a rabbit hole if it's already imperfect)

Xitian9:
Okay, I've restored the previous behaviour and added a test for it.

simonmichael:
Lovely. Thanks!

PR #2311, opened 2025-01-02
;bin: git: make ensure_git more robust

larsks:
Instead of manually checking for a .git directory, ask git to check whether
or not we are in a work tree using 'git rev-parse --is-inside-work-tree'.
This takes care of things like being in a subdirectory of a work tree, or
someone having GIT_DIR pointing somewhere other than .git, etc.

### DIFF
--- a/bin/hledger-git
+++ b/bin/hledger-git
@@ -62,12 +62,11 @@ EOF
     fi
 }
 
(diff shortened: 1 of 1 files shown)


simonmichael:
Thank you!

PR #2319, opened 2025-01-21
Implement `encoding` in .rules

### LINKED ISSUE #1777: CSV import rules, encoding
Is it difficult to add new option for CSV import rules, option to define `encoding` of CSV file? I assume that hledger works with UTF-8 encoding by default. I have some CSV files in encoding CP1250 (Windows-1250), so I have to convert them before import with `recode cp1250..utf8 FILE.csv`.

jokesper:
An implementation for `encoding` in rules file:
The cases are:
- [X] `-f *.rules`
- [x] `-f *.csv`
- [x] `-f *.csv --rules *.rules`

The second and third case are effectively the same. The problem is that for them we call a `Reader`s `rReadFn` value taking the contents rather then the filepath / handle.
So we either would have to change `Reader` to take a `FilePath` instead of `Text`, which would break how we do Tests,
or we add an extra field  `rGetContents :: FilePath -> IO Text`.
I would rather get some feedback on this before I implement a potentially bigger change.

If implemented this PR should:
- fix #1777

### DIFF
--- a/hledger-lib/Hledger/Read.hs
+++ b/hledger-lib/Hledger/Read.hs
@@ -102,6 +102,7 @@ module Hledger.Read (
 
   -- * Easy journal parsing
   readJournal',
+  readJournal'',
   readJournalFile',
   readJournalFiles',
   orDieTrying,
@@ -125,7 +126,7 @@ module Hledger.Read (
 
 --- ** imports
 import qualified Control.Exception as C
-import Control.Monad (unless, when, forM)
+import Control.Monad (unless, when, forM, (<=<))
 import "mtl" Control.Monad.Except (ExceptT(..), runExceptT, liftEither)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Default (def)
@@ -145,7 +146,7 @@ import System.Environment (getEnv)
 import System.Exit (exitFailure)
 import System.FilePath ((<.>), (</>), splitDirectories, splitFileName, takeFileName)
 import System.Info (os)
-import System.IO (hPutStr, stderr)
+import System.IO (Handle, hPutStr, stderr)
 
 import Hledger.Data.Dates (getCurrentDay, parsedateM, showDate)
 import Hledger.Data.Types
@@ -205,7 +206,7 @@ type PrefixedFilePath = FilePath
 
 -- | @readJournal iopts mfile txt@
 --
--- Read a Journal from some text, with strict checks if enabled,
+-- Read a Journal from some handle, with strict checks if enabled,
 -- or return an error message.
 --
 -- The reader (data format) is chosen based on, in this order:
@@ -219,11 +220,11 @@ type PrefixedFilePath = FilePath
 -- If none of these is available, or if the reader name is unrecognised,
 -- we use the journal reader (for predictability).
 --
-readJournal :: InputOpts -> Maybe FilePath -> Text -> ExceptT String IO Journal
-readJournal iopts@InputOpts{strict_, _defer} mpath txt = do
+readJournal :: InputOpts -> Maybe FilePath -> Handle -> ExceptT String IO Journal
+readJournal iopts@InputOpts{strict_, _defer} mpath hdl = do
   let r :: Reader IO = fromMaybe JournalReader.reader $ findReader (mformat_ iopts) mpath
   dbg6IO "readJournal: trying reader" (rFormat r)
-  j <- rReadFn r iopts (fromMaybe "(string)" mpath) txt
+  j <- rReadFn r iopts (fromMaybe "(string)" mpath) hdl
   when (strict_ && not _defer) $ liftEither $ journalStrictChecks j
   return j
 
@@ -264,11 +265,11 @@ readJournalFileAndLatestDates iopts prefixedfile = do
     (mfmt, f) = splitReaderPrefix prefixedfile
     iopts' = iopts{mformat_=asum [mfmt, mformat_ iopts]}
   liftIO $ requireJournalFileExists f
-  t <-
+  h <-
     traceOrLogAt 6 ("readJournalFile: "++takeFileName f) $
-    liftIO $ readFileOrStdinPortably f
+    liftIO $ openFileOrStdin f
     -- <- T.readFile f  -- or without line ending translation, for testing
-  j <- readJournal iopts' (Just f) t
+  j <- readJournal iopts' (Just f) h
   if new_ iopts
     then do
       ds <- liftIO $ previousLatestDates f
@@ -313,8 +314,13 @@ readJournalFilesAndLatestDates iopts pfs = do
 
 -- | An easy version of 'readJournal' which assumes default options, and fails
 -- in the IO monad.
-readJournal' :: Text -> IO Journal
+readJournal' :: Handle -> IO Journal
 readJournal' = orDieTrying . readJournal definputopts Nothing
+
+-- | An even easier version of 'readJournal' which additionally to 'readJournal''
+-- also takes a 'Text' instead of a 'Handle'.
+readJournal'' :: Text -> IO Journal
+readJournal'' = readJournal' <=< inputToHandle
 
 -- | An easy version of 'readJournalFile' which assumes default options, and fails
 -- in the IO monad.
--- a/hledger-lib/Hledger/Read/Common.hs
+++ b/hledger-lib/Hledger/Read/Common.hs
@@ -34,6 +34,7 @@ module Hledger.Read.Common (
   HasInputOpts(..),
   definputopts,
   rawOptsToInputOpts,
+  handleReadFnToTextReadFn,
 
   -- * parsing utilities
   parseAndFinaliseJournal,
@@ -148,6 +149,7 @@ import Data.Time.Clock.POSIX (getPOSIXTime)
 import Data.Time.LocalTime (LocalTime(..), TimeOfDay(..))
 import Data.Word (Word8)
 import System.FilePath (takeFileName)
+import System.IO (Handle)
 import Text.Megaparsec
 import Text.Megaparsec.Char (char, char', digitChar, newline, string)
 import Text.Megaparsec.Char.Lexer (decimal)
@@ -179,9 +181,9 @@ data Reader m = Reader {
     ,rExtensions :: [String]
 
      -- The entry point for reading this format, accepting input options, file
-     -- path for error messages and file contents, producing an exception-raising IO
+     -- path for error messages and file contents via the handle, producing an exception-raising IO
      -- action that produces a journal or error message.
-    ,rReadFn   :: InputOpts -> FilePath -> Text -> ExceptT String IO Journal
+    ,rReadFn   :: InputOpts -> FilePath -> Handle -> ExceptT String IO Journal
 
      -- The actual megaparsec parser called by the above, in case
      -- another parser (includedirectivep) wants to use it directly.
@@ -230,6 +232,10 @@ rawOptsToInputOpts day usecoloronstdout postingaccttags rawopts =
       ,strict_            = boolopt "strict" rawopts
       ,_ioDay             = day
       }
+
+handleReadFnToTextReadFn :: (InputOpts -> FilePath -> Text -> ExceptT String IO Journal) -> InputOpts -> FilePath -> Handle -> ExceptT String IO Journal
+handleReadFnToTextReadFn p iopts fp =
+  p iopts fp <=< lift . readHandlePortably
 
 -- | Get the date span from --forecast's PERIODEXPR argument, if any.
 -- This will fail with a usage error if the period expression cannot be parsed,
--- a/hledger-lib/Hledger/Read/CsvReader.hs
+++ b/hledger-lib/Hledger/Read/CsvReader.hs
@@ -28,7 +28,7 @@ where
 import Prelude hiding (Applicative(..))
 import Control.Monad.Except       (ExceptT(..), liftEither)
 import Control.Monad.IO.Class     (MonadIO)
-import Data.Text (Text)
+import System.IO                  (Handle)
 
 import Hledger.Data
 import Hledger.Utils
@@ -54,10 +54,10 @@ reader sep = Reader
 -- This file path is normally the CSV(/SSV/TSV) data file, and a corresponding rules file is inferred.
 -- But it can also be the rules file, in which case the corresponding data file is inferred.
 -- This does not check balance assertions.
-parse :: SepFormat -> InputOpts -> FilePath -> Text -> ExceptT String IO Journal
-parse sep iopts f t = do
+parse :: SepFormat -> InputOpts -> FilePath -> Handle -> ExceptT String IO Journal
+parse sep iopts f h = do
   let mrulesfile = mrules_file_ iopts
-  readJournalFromCsv (Right <$> mrulesfile) f t (Just sep)
+  readJournalFromCsv (Right <$> mrulesfile) f h (Just sep)
   -- apply any command line account aliases. Can fail with a bad replacement pattern.
   >>= liftEither . journalApplyAliases (aliasesFromOpts iopts)
       -- journalFinalise assumes the journal's items are
@@ -65,7 +65,7 @@ parse sep iopts f t = do
       -- But here they are already properly ordered. So we'd
       -- better preemptively reverse them once more. XXX inefficient
       . journalReverse
-  >>= journalFinalise iopts{balancingopts_=(balancingopts_ iopts){ignore_assertions_=True}} f t
+  >>= journalFinalise iopts{balancingopts_=(balancingopts_ iopts){ignore_assertions_=True}} f ""
 
 --- ** tests
 
--- a/hledger-lib/Hledger/Read/JournalReader.hs
+++ b/hledger-lib/Hledger/Read/JournalReader.hs
@@ -194,7 +194,7 @@ reader :: MonadIO m => Reader m
 reader = Reader
   {rFormat     = Journal'
   ,rExtensions = ["journal", "j", "hledger", "ledger"]
-  ,rReadFn     = parse
+  ,rReadFn     = handleReadFnToTextReadFn parse
   ,rParser    = journalp  -- no need to add command line aliases like journalp'
                            -- when called as a subparser I think
   }
--- a/hledger-lib/Hledger/Read/RulesReader.hs
+++ b/hledger-lib/Hledger/Read/RulesReader.hs
@@ -52,6 +52,7 @@ import Control.Monad.State.Strict (StateT, get, modify', evalStateT)
 import Control.Monad.Trans.Class  (lift)
 import Data.Char                  (toLower, isDigit, isSpace, isAlphaNum, ord)
 import Data.Bifunctor             (first)
+import Data.Encoding              (encodingFromStringExplicit)
 import Data.Functor               ((<&>))
 import Data.List (elemIndex, mapAccumL, nub, sortOn)
 #if !MIN_VERSION_base(4,20,0)
@@ -69,6 +70,7 @@ import Data.Time ( Day, TimeZone, UTCTime, LocalTime, ZonedTime(ZonedTime),
   defaultTimeLocale, getCurrentTimeZone, localDay, parseTimeM, utcToLocalTime, localTimeToUTC, zonedTimeToUTC)
 import Safe (atMay, headMay, lastMay, readMay)
 import System.FilePath ((</>), takeDirectory, takeExtension, stripExtension, takeFileName)
+import System.IO       (Handle, hClose)
 import qualified Data.Csv as Cassava
 import qualified Data.Csv.Parser.Megaparsec as CassavaMegaparsec
 import qualified Data.ByteString as B
@@ -116,10 +118,11 @@ getDownloadDir = do
 -- file's directory. When a glob pattern matches multiple files, the alphabetically
 -- last is used. (Eg in case of multiple numbered downloads, the highest-numbered
 -- will be used.)
--- The provided text, or a --rules option, are ignored by this reader.
+-- The provided handle, or a --rules option, are ignored by this reader.
 -- Balance assertions are not checked.
-parse :: InputOpts -> FilePath -> Text -> ExceptT String IO Journal
-parse iopts f _ = do
+parse :: InputOpts -> FilePath -> Handle -> ExceptT String IO Journal
+parse iopts f h = do
+  lift $ hClose h -- We don't need it
   rules <- readRulesFile $ dbg4 "reading rules file" f
   -- XXX higher-than usual debug level for file reading to bypass excessive noise from elsewhere, normally 6 or 7
   mdatafile <- liftIO $ do
@@ -139,8 +142,8 @@ parse iopts f _ = do
       if not (dat=="-" || exists)
       then return nulljournal      -- data file inferred from rules file name was not found
       else do
-        t <- liftIO $ readFileOrStdinPortably dat
-        readJournalFromCsv (Just $ Left rules) dat t Nothing
+        dath <- liftIO $ openFileOrStdin dat
+        readJournalFromCsv (Just $ Left rules) dat dath Nothing
         -- apply any command line account aliases. Can fail with a bad replacement pattern.
         >>= liftEither . journalApplyAliases (aliasesFromOpts iopts)
             -- journalFinalise assumes the journal's items are
@@ -500,6 +503,7 @@ directivep = (do
 directives :: [Text]
 directives =
   ["source"
+  ,"encoding"
   ,"date-format"
   ,"decimal-mark"
   ,"separator"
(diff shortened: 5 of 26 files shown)


simonmichael:
Thanks for exploring this. It's probably a good idea. But what if rules files made it easier to run a preprocessor, eg

    source !iconv -f CP1250 FILE.csv

That requires a little more expertise from the user, I guess ?

jokesper:
> But what if rules files made it easier to run a preprocessor

I like the idea, though implementing it with the other cases is a bit wierd, since we ignore `source` with `-f`.

We could split it into two options with the other being `preprocessor iconv -f CP1250 {}`.
This has the drawback of either reserving a replacement or allowing one to specify it.

> The case of reading CSV from stdin might also be worth thinking about.

Is it even possible to read a csv from stdin?
- `-f - --rules *.rules` doesn't work, since we first figure out whether it's a csv based on the file ending.
- `source -` doesn't work, since it searches for `-` as a file and can't find it.
- `source /dev/stdin` works, but is a bit hacky.

Though if they would work, they would be handled the same as the cases mentioned above.
Opening the handle and reading from it are indpendent.

simonmichael:
> Is it even possible to read a csv from stdin?
> 

Like so: `-f csv:- --rules-file=foo.rules`

simonmichael:
But it can run x86_64 ghc. And yes, it's similar on Windows: text-icu isn't an easy thing to install.
```
C:\Users\Simon>stack install text-icu
[1 of 3] Compiling Main             ( C:\sr\setup-exe-src\setup-O_vy6YIf.hs, C:\sr\setup-exe-src\setup-O_vy6YIf.o )
[2 of 3] Compiling StackSetupShim   ( C:\sr\setup-exe-src\setup-shim-O_vy6YIf.hs, C:\sr\setup-exe-src\setup-shim-O_vy6YIf.o )
[3 of 3] Linking C:\sr\setup-exe-cache\x86_64-windows\tmp-Cabal-simple_O_vy6YIf_3.10.3.0_ghc-9.8.4.exe
text-icu> configure
text-icu> Configuring text-icu-0.8.0.5...
text-icu> Error: Cabal-simple_O_vy6YIf_3.10.3.0_ghc-9.8.4.exe: The program 'pkg-config'
text-icu> version >=0.9.0 is required but it could not be found.
text-icu>

Error: [S-7282]
       Stack failed to execute the build plan.

       While executing the build plan, Stack encountered the error:

       [S-7011]
       While building package text-icu-0.8.0.5 (scroll up to its section to see the error) using:
       C:\sr\setup-exe-cache\x86_64-windows\Cabal-simple_O_vy6YIf_3.10.3.0_ghc-9.8.4.exe --verbose=1 --builddir=.stack-work\dist\f24b2e15 configure --with-ghc=C:\Users\Simon\AppData\Local\Programs\stack\x86_64-windows\ghc-9.8.4\bin\ghc-9.8.4.exe --with-ghc-pkg=C:\Users\Simon\AppData\Local\Programs\stack\x86_64-windows\ghc-9.8.4\bin\ghc-pkg-9.8.4.exe --user --package-db=clear --package-db=global --package-db=C:\sr\snapshots\ce0f12f1\pkgdb --libdir=C:\sr\snapshots\ce0f12f1\lib --bindir=C:\sr\snapshots\ce0f12f1\bin --datadir=C:\sr\snapshots\ce0f12f1\share --libexecdir=C:\sr\snapshots\ce0f12f1\libexec --sysconfdir=C:\sr\snapshots\ce0f12f1\etc --docdir=C:\sr\snapshots\ce0f12f1\doc\text-icu-0.8.0.5 --htmldir=C:\sr\snapshots\ce0f12f1\doc\text-icu-0.8.0.5 --haddockdir=C:\sr\snapshots\ce0f12f1\doc\text-icu-0.8.0.5 --dependency=base=base-4.19.2.0-943b --dependency=bytestring=bytestring-0.12.1.0-290b --dependency=deepseq=deepseq-1.5.1.0-f918 --dependency=text=text-2.1.1-3967 --dependency=time=time-1.12.2-0f79 -fhomebrew --extra-include-dirs=C:\Users\Simon\AppData\Local\Programs\stack\x86_64-windows\msys2-20240727\mingw64\include --extra-lib-dirs=C:\Users\Simon\AppData\Local\Programs\stack\x86_64-windows\msys2-20240727\mingw64\lib --extra-lib-dirs=C:\Users\Simon\AppData\Local\Programs\stack\x86_64-windows\msys2-20240727\mingw64\bin --exact-configuration --ghc-option=-fhide-source-paths
       Process exited with code: ExitFailure 1
```

jokesper:
Should I instead switch to something like [`encoding`](https://hackage.haskell.org/package/encoding)?
It would be easier to install, as it's installed entirely by `cabal` / `stack` and would be statically build into `hledger`.
The drawback is: It works on `String` not `Text`. I.e. we would need a possibly costly `pack` (not tested).

simonmichael:
Yes that looks pretty good. Converting from Text to String won't affect anyone noticeably I predict, though if it can be avoided when unnecessary, all the better.

jokesper:
I just switched the part I already implemented to use `encoding` instead of `text-icu`.

I noticed `encoding` uses a custom setup which is currently affected by [this bug](https://github.com/haskell/cabal/issues/6505). For some developers (on Arch btw.) this leads to an extra option they should pass: `--ghc-options=-dynamic` when using cabal.

Should I document this anywhere, as it's already documented in the Arch wiki.

jokesper:
I implemented the remaining two cases. To do this I had to change the arguments to `rReadFn`. The details on why are in the commit message.

I think this is ready for review.

jokesper:
> Docs and tests will help motivate this, when the time comes.

Writing tests for this would require quite a bit of rework in how they are treated.
Currently the readers for `csv` and `rules` only look at the filepath and not at their passed contents.
Instead they get them themselves.

simonmichael:
Hi @jokesper. Sorry for the slow reply. If this would make hledger installers need to use special build flags, that would be a problem unfortunately.

simonmichael:
I have inquired at https://github.com/dmwit/encoding/issues/23

jokesper:
While trying to fix the CI, I noticed you already needed to supply `--ghc-options=-dynamic` to build `hledger-web`, since it depends on `xml-conduit` which has a custom setup.
Furthermore this is only a problem with cabal (stack should work fine) and on the first build (since then the dependencies are build).

Though I would still prefer if we didn't need a flag just to build the core `hledger-lib`, which would also affect every project depending on `hledger`.

jokesper:
Since `encoding` now dropped the custom build type, the associated problems have now been fixed

simonmichael:
Thanks @jokesper . Adding encoding-0.9 to stack*.yaml should fix the current CI failure.

simonmichael:
(Or in the case of stack.yaml, bumping it to the latest nightly-2025-02-11.)

jokesper:
Thanks for the info. It's fixed now

simonmichael:
Thanks @jokesper.
Now it's time for more review, and I'd welcome help from other reviewers.
Here's some relevant info:

26 files changed

encoding >0.9 dependency added

## Main commit description

    feat: Implement encoding in rules file

    Instead of `text-icu` as [recommended by `text`](https://hackage.haskell.org/package/text-2.1.2/docs/Data-Text-Encoding.html):
    
    > To gain access to a much larger family of encodings, use the `text-icu` package.
    
    we use `encoding`, since `text-icu` requires an external library.
    `encoding` does require a custom setup, which is currently
    affected by a [bug in Cabal](https://github.com/haskell/cabal/issues/6505)
    but it doesn't require an external runtime dependency, which makes it
    easier to distribute on macos and windows.
    
    I also changed the `rReadFn` to get a `Handle` rather than a `Text`.
    Such that we can use non standard functions for reading it.
    This is kind of sideways, since the one place we care about it,
    we don't use it at all. Though it still caused issues, since the input
    wasn't parseable as text.
    I think not using the handle has something to do with how `-` is treated
    as stdin but I didn't do any further tests on that, since this is out of
    the scope of this PR.


## Github AI summary

This pull request includes several changes to the `hledger-lib` library, primarily focusing on modifying the journal reading functions to use `Handle` instead of `Text` and adding support for handling different text encodings. The changes span multiple files and involve updates to function signatures, imports, and helper functions.

Updates to journal reading functions:
* Changed `readJournal` and related functions to accept a `Handle` instead of `Text` for reading journal data (`hledger-lib/Hledger/Read.hs`). [[1]](diffhunk://#diff-bdf99fb149929f11680d239ef2f7508759abea3b651393dc6806a31e1e71e511L222-R227) [[2]](diffhunk://#diff-bdf99fb149929f11680d239ef2f7508759abea3b651393dc6806a31e1e71e511L316-R324)
* Added a new function `readJournal''` that allows reading from `Text` by converting it to a `Handle` (`hledger-lib/Hledger/Read.hs`).

Import and helper function updates:
* Updated imports to include `Handle` and other necessary modules (`hledger-lib/Hledger/Read.hs`, `hledger-lib/Hledger/Read/Common.hs`, `hledger-lib/Hledger/Read/CsvReader.hs`, `hledger-lib/Hledger/Read/RulesReader.hs`). [[1]](diffhunk://#diff-bdf99fb149929f11680d239ef2f7508759abea3b651393dc6806a31e1e71e511L148-R149) [[2]](diffhunk://#diff-fcee3c006b1d2e35b8c279c9daa9963ecdb5910bdc90099f7b3a37b8b529a65dR152) [[3]](diffhunk://#diff-52f2974c62ce69a30e2bddf4d8b6008441715fde0a9d0a4c3ea8c072fa465223L31-R31) [[4]](diffhunk://#diff-95e0d5df4894ae4a99b5ad4a37811c64ec30d82af3b01035561976afc3c4293eR73)
* Added helper functions such as `handleReadFnToTextReadFn` to facilitate the transition from `Text` to `Handle` (`hledger-lib/Hledger/Read/Common.hs`).

Changes in specific readers:
* Updated `CsvReader`, `JournalReader`, `RulesReader`, `TimeclockReader`, and `TimedotReader` to use the new `Handle`-based functions (`hledger-lib/Hledger/Read/CsvReader.hs`, `hledger-lib/Hledger/Read/JournalReader.hs`, `hledger-lib/Hledger/Read/RulesReader.hs`, `hledger-lib/Hledger/Read/TimeclockReader.hs`, `hledger-lib/Hledger/Read/TimedotReader.hs`). [[1]](diffhunk://#diff-52f2974c62ce69a30e2bddf4d8b6008441715fde0a9d0a4c3ea8c072fa465223L57-R68) [[2]](diffhunk://#diff-7f7e2a1fff079623a40926984dc95d7de0fac5a9575881e8ec09ce309903a574L197-R197) [[3]](diffhunk://#diff-95e0d5df4894ae4a99b5ad4a37811c64ec30d82af3b01035561976afc3c4293eL119-R125) [[4]](diffhunk://#diff-d567696a32088b29da11cf620c9180ca5000b98120fca1f7ef01d66836109d02L82-R82) [[5]](diffhunk://#diff-2f4d6afbe4caeb1b1438089f2fe978bdfb460346753b5559df20560b6b68ef9dL71-R71)

Test updates:
* Modified tests in `PostingsReport` to use the new `readJournal''` function (`hledger-lib/Hledger/Reports/PostingsReport.hs`). [[1]](diffhunk://#diff-0d9c798b1335019b9475e284450700481b3ce00c8f416523a5f73e83badcc112L304-R304) [[2]](diffhunk://#diff-0d9c798b1335019b9475e284450700481b3ce00c8f416523a5f73e83badcc112L316-R316) [[3]](diffhunk://#diff-0d9c798b1335019b9475e284450700481b3ce00c8f416523a5f73e83badcc112L328-R328)

Miscellaneous:
* Added new language pragmas and updated the module exports to include new functions (`hledger-lib/Hledger/Utils/IO.hs`). [[1]](diffhunk://#diff-5692bda46ff6e1e83272e0e8d71eb87ac9389f0c3695151b6c8370f5d7f42e55R7-R9) [[2]](diffhunk://#diff-5692bda46ff6e1e83272e0e8d71eb87ac9389f0c3695151b6c8370f5d7f42e55R36-R44)

These changes improve the flexibility and robustness of the journal reading functions by allowing them to handle different input sources and text encodings.

## SM summary/doc draft

Previously, hledger could read CSV files containing non-ascii characters only if they are UTF8-encoded.
Now there is a new CSV rule, `encoding ENCODING`, which allows reading CSV files with other encodings.
The following encodings are supported:
...

## Actions

- Let's squash these commits into one when you get a chance

- Let's update the message. I usually don't bother with markdown hyperlinks in these - Github doesn't render them.
  It could be something like

    feat: Implement encoding in rules file

    This uses the encoding library, which supports fewer encodings than text-icu
    but is more portable, not requiring a C library.

    ...

- Let's clarify a bit the user-facing changes as well as the developer-facing ones.
  So far we have this, which is more temporary PR discussion than informative permanent commit message.
  Happy to suggest ideas once I understand this PR myself.

    I also changed the `rReadFn` to get a `Handle` rather than a `Text`.
    Such that we can use non standard functions for reading it.
    This is kind of sideways, since the one place we care about it,
    we don't use it at all. Though it still caused issues, since the input
    wasn't parseable as text.
    I think not using the handle has something to do with how `-` is treated
    as stdin but I didn't do any further tests on that, since this is out of
    the scope of this PR.

simonmichael:
Thanks @jokesper.
Now it's time for more review, and I'd welcome help from other reviewers.
Here's some relevant info:

26 files changed

encoding >0.9 dependency added

## Main commit description

    feat: Implement encoding in rules file

    Instead of `text-icu` as [recommended by `text`](https://hackage.haskell.org/package/text-2.1.2/docs/Data-Text-Encoding.html):
    
    > To gain access to a much larger family of encodings, use the `text-icu` package.
    
    we use `encoding`, since `text-icu` requires an external library.
    `encoding` does require a custom setup, which is currently
    affected by a [bug in Cabal](https://github.com/haskell/cabal/issues/6505)
    but it doesn't require an external runtime dependency, which makes it
    easier to distribute on macos and windows.
    
    I also changed the `rReadFn` to get a `Handle` rather than a `Text`.
    Such that we can use non standard functions for reading it.
    This is kind of sideways, since the one place we care about it,
    we don't use it at all. Though it still caused issues, since the input
    wasn't parseable as text.
    I think not using the handle has something to do with how `-` is treated
    as stdin but I didn't do any further tests on that, since this is out of
    the scope of this PR.


## Github AI summary

This pull request includes several changes to the `hledger-lib` library, primarily focusing on modifying the journal reading functions to use `Handle` instead of `Text` and adding support for handling different text encodings. The changes span multiple files and involve updates to function signatures, imports, and helper functions.

Updates to journal reading functions:
* Changed `readJournal` and related functions to accept a `Handle` instead of `Text` for reading journal data (`hledger-lib/Hledger/Read.hs`). [[1]](diffhunk://#diff-bdf99fb149929f11680d239ef2f7508759abea3b651393dc6806a31e1e71e511L222-R227) [[2]](diffhunk://#diff-bdf99fb149929f11680d239ef2f7508759abea3b651393dc6806a31e1e71e511L316-R324)
* Added a new function `readJournal''` that allows reading from `Text` by converting it to a `Handle` (`hledger-lib/Hledger/Read.hs`).

Import and helper function updates:
* Updated imports to include `Handle` and other necessary modules (`hledger-lib/Hledger/Read.hs`, `hledger-lib/Hledger/Read/Common.hs`, `hledger-lib/Hledger/Read/CsvReader.hs`, `hledger-lib/Hledger/Read/RulesReader.hs`). [[1]](diffhunk://#diff-bdf99fb149929f11680d239ef2f7508759abea3b651393dc6806a31e1e71e511L148-R149) [[2]](diffhunk://#diff-fcee3c006b1d2e35b8c279c9daa9963ecdb5910bdc90099f7b3a37b8b529a65dR152) [[3]](diffhunk://#diff-52f2974c62ce69a30e2bddf4d8b6008441715fde0a9d0a4c3ea8c072fa465223L31-R31) [[4]](diffhunk://#diff-95e0d5df4894ae4a99b5ad4a37811c64ec30d82af3b01035561976afc3c4293eR73)
* Added helper functions such as `handleReadFnToTextReadFn` to facilitate the transition from `Text` to `Handle` (`hledger-lib/Hledger/Read/Common.hs`).

Changes in specific readers:
* Updated `CsvReader`, `JournalReader`, `RulesReader`, `TimeclockReader`, and `TimedotReader` to use the new `Handle`-based functions (`hledger-lib/Hledger/Read/CsvReader.hs`, `hledger-lib/Hledger/Read/JournalReader.hs`, `hledger-lib/Hledger/Read/RulesReader.hs`, `hledger-lib/Hledger/Read/TimeclockReader.hs`, `hledger-lib/Hledger/Read/TimedotReader.hs`). [[1]](diffhunk://#diff-52f2974c62ce69a30e2bddf4d8b6008441715fde0a9d0a4c3ea8c072fa465223L57-R68) [[2]](diffhunk://#diff-7f7e2a1fff079623a40926984dc95d7de0fac5a9575881e8ec09ce309903a574L197-R197) [[3]](diffhunk://#diff-95e0d5df4894ae4a99b5ad4a37811c64ec30d82af3b01035561976afc3c4293eL119-R125) [[4]](diffhunk://#diff-d567696a32088b29da11cf620c9180ca5000b98120fca1f7ef01d66836109d02L82-R82) [[5]](diffhunk://#diff-2f4d6afbe4caeb1b1438089f2fe978bdfb460346753b5559df20560b6b68ef9dL71-R71)

Test updates:
* Modified tests in `PostingsReport` to use the new `readJournal''` function (`hledger-lib/Hledger/Reports/PostingsReport.hs`). [[1]](diffhunk://#diff-0d9c798b1335019b9475e284450700481b3ce00c8f416523a5f73e83badcc112L304-R304) [[2]](diffhunk://#diff-0d9c798b1335019b9475e284450700481b3ce00c8f416523a5f73e83badcc112L316-R316) [[3]](diffhunk://#diff-0d9c798b1335019b9475e284450700481b3ce00c8f416523a5f73e83badcc112L328-R328)

Miscellaneous:
* Added new language pragmas and updated the module exports to include new functions (`hledger-lib/Hledger/Utils/IO.hs`). [[1]](diffhunk://#diff-5692bda46ff6e1e83272e0e8d71eb87ac9389f0c3695151b6c8370f5d7f42e55R7-R9) [[2]](diffhunk://#diff-5692bda46ff6e1e83272e0e8d71eb87ac9389f0c3695151b6c8370f5d7f42e55R36-R44)

These changes improve the flexibility and robustness of the journal reading functions by allowing them to handle different input sources and text encodings.

## SM summary/doc draft

Previously, hledger could read CSV files containing non-ascii characters only if they are UTF8-encoded.
Now there is a new CSV rule, `encoding ENCODING`, which allows reading CSV files with other encodings.
The following encodings are supported:
...

## Actions

- Let's squash these commits into one when you get a chance

- Let's update the message. I usually don't bother with markdown hyperlinks in these - Github doesn't render them.
  It could be something like

    feat: Implement encoding in rules file

    This uses the encoding library, which supports fewer encodings than text-icu
    but is more portable, not requiring a C library.

    ...

- Let's clarify a bit the user-facing changes as well as the developer-facing ones.
  So far we have this, which is more temporary PR discussion than informative permanent commit message.
  Happy to suggest ideas once I understand this PR myself.

    I also changed the `rReadFn` to get a `Handle` rather than a `Text`.
    Such that we can use non standard functions for reading it.
    This is kind of sideways, since the one place we care about it,
    we don't use it at all. Though it still caused issues, since the input
    wasn't parseable as text.
    I think not using the handle has something to do with how `-` is treated
    as stdin but I didn't do any further tests on that, since this is out of
    the scope of this PR.

jokesper:
Yes, this is due to the interface of `hGetContents` from `encoding` using an implicit parameter.

simonmichael:
Edited my last comment a little (re commit message).

I see. I have used ImplicitParams in the past and don't like it much, it always confuses me. But it seems a requirement here.. maybe a benefit. Do you have any insight into what it does for us here ?

simonmichael:
I see:

https://hackage.haskell.org/package/encoding-0.9/docs/System-IO-Encoding.html
> By using implicit parameters, it can be used almost as a drop-in replacement

An encoding can be assigned to `?enc`, which works like a global for the System.IO.Encoding functions, instead of needing to provide it as an extra argument.

jokesper:
Alternatively we could inline the body of `hGetContents`. It's just reading a lazy bytestring and decoding it.

simonmichael:
I kicked off the multi-platform CI builds. encoding is not building on Windows unfortunately, because of:
```
encoding                     > Configuring encoding-0.9...
encoding                     > Error: [Cabal-4345]
encoding                     > Missing dependency on a foreign library:
encoding                     > * Missing (or bad) header file: system_encoding.h
encoding                     > If the header file does exist, it may contain errors that are caught by the C compiler at the preprocessing stage. In this case you can re-run configure with the verbosity flag -v3 to see the error messages.
```

simonmichael:
I opened https://github.com/dmwit/encoding/issues/26

jokesper:
I applied the workaround. Though I don't know whether this fixes the issue for cabal on windows. Is this (cabal+windows) even part of our target developer base?

simonmichael:
Yes, being easy to build with stack or cabal on any platform is a hledger requirement. Better if we can fix this upstream if possible. 

Here's another note for the doc:

The following encodings are supported (case-insensitive, can also be written with inner spaces or hyphens):

ASCII
UTF8
UTF16
UTF32
ISO88591
ISO88592
ISO88593
ISO88594
ISO88595
ISO88596
ISO88597
ISO88598
ISO88599
ISO885910
ISO885911
ISO885913
ISO885914
ISO885915
ISO885916
CP1250
CP1251
CP1252
CP1253
CP1254
CP1255
CP1256
CP1257
CP1258
KOI8R
KOI8U
GB18030
MacOSRoman
JISX0201
JISX0208
ISO2022JP
ShiftJIS
CP437
CP737
CP775
CP850
CP852
CP855
CP857
CP860
CP861
CP862
CP863
CP864
CP865
CP866
CP869
CP874
CP932

jokesper:
I added the list to the manual. With a link to the cases for the alternative spellings.
though I don't know whether the either is necessary, as the intention of this function is to allow any spelling for an encoding one may have choosen to be work.

simonmichael:
@jokesper, beyond the scope of this PR but do you have any thoughts about wider support for text encodings - would we ideally support this for all input files ? Output files ? Or is it best left to external tools, with CSV reading being a special case because we are more often dealing with CSV from diverse sources ?

jokesper:
I think having it just for CSVs is fine, since this is the common use case. It's unlikely to get a journal file in a different encoding, though what would be possible, since we changed from `Text` to `Handle` one could maybe implement parsing electronic invoicing. Though this is way out of the scope of HLedger.

Expanding it to other files wouldn't be that hard, except for the design choice of where to document the encoding (command line parameter?). Though in my opinion it's really not needed at all for any other cases.
It doesn't help anyone now, since we effectively already enforced an encoding, and the journal format isn't any common import / export format so no benefit there.

What could be useful is in an csv export, though I don't quite know if this is really relevent. If anyone wants to use it in an CSV export please open an issue.

simonmichael:
@jokesper, all: I welcome any review of https://github.com/dmwit/encoding/pull/27, which aims to make encoding build more smoothly on Windows. I'm not quite sure of the breakage risk and how much care is needed.

simonmichael:
I have squashed and merged this, plus some doc edits and some test files in examples/. (I didn't get as far as adding automated tests.) I verified that this new feature allows us to successfully convert CSV files with other encodings. Thank you @jokesper !

hseg:
Note that the bumping of `encoding` to `0.10` to make use of https://github.com/dmwit/encoding/pull/27 means we shouldn't be disabling the `systemEncoding` since `encoding-0.10` removes it.

i.e. the `.flags.encoding.systemEncoding = false` keys in `stack*.yaml` should
be removed

Stack indeed complains

```
Error: [S-8664]
       Invalid flag specification:
       * Package encoding does not define the following flags (specified in the project-level
         configuration (e.g. stack.yaml)): systemencoding.
         Flags defined by package encoding are: generate-encodings.
```

simonmichael:
Thanks! I missed those, now fixed.

simonmichael:
Unfortunately, [this change in Common.hs](https://github.com/simonmichael/hledger/pull/2319/files#diff-b1e8037cff372bc60fe965e6d0c52648a764bd12ca3b2fb82db3af832f52e4b8) is causing hledger-web to hang when saving changes to a largeish file (#2389).

simonmichael:
A separate thread was needed to avoid deadlock when writing more than a single buffer-ful to the pipe. I have pushed a fix

simonmichael:
Just a note that requiring a Handle (or use of inputToHandle) for reading a journal has made hledger-lib a bit harder to use or port, eg to WASM. Was it required as part of journal reading to give a chance to set the encoding ?

jokesper:
no, but it was easier to implement as both encoding and text have functions to read from a handle,
but reading into a bytestring and converting (which would be the alternative). Though the alternative shouldn't be to hard.

jokesper:
While trying to change the implementation from `Handle` to `ByteString`, I remembered why I choose `Handle`.

`readHandlePortably'`already checks the encoding on the Handle, so we do get some encoding support. changing it to `ByteString`means loosing this. Another option I considered was passing a `Maybe DynEncoding -> Text` instead of a `Handle` and it would work with WASM but it felt kind of hacky.

I think the `Maybe DynEncoding -> Text` is probably the better choice if we want to port to WASM and the `ByteString` option is better if we want to allow other readers in the future which don't work on text.

jokesper:
\* `Maybe DynEncoding -> IO Text`

jokesper:
It seems there is also `System.IO.mkTextEncoding` with `hSetEncoding` allowing the indirect use of iconv / windows conversions. Though `encoding` is probably the more portable choice.

PR #2321, opened 2025-01-23
refactor Hledger.Write.Html etc

simonmichael:
- **dev: rename printHtml -> styledTableHtml**
- **dev: refactor Hledger.Write.Html etc, reducing Lucid references**

### DIFF
--- a/hledger-lib/Hledger/Write/Html.hs
+++ b/hledger-lib/Hledger/Write/Html.hs
@@ -1,38 +1,41 @@
-{-# LANGUAGE OverloadedStrings #-}
 {- |
-Common definitions for Html.Blaze and Html.Lucid
+HTML writing helpers.
+This module would ideally hide the details of which HTML library is used, but it doesn't yet.
+
+Currently hledger-web uses blaze-html, but hledger CLI reports use lucid.
+lucid has a more usable API than blaze-html (https://chrisdone.com/posts/lucid).
+lucid2's is even better.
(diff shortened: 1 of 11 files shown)


simonmichael:
This is FYI @thielema:

So that I could troubleshoot hledger's HTML output or inspect diffs in reports, I attempted to "quickly" add newlines to it. Instead I did some refactoring and research in that direction. Here's the new doc for Hledger.Write.Html:

> HTML writing helpers.
> This module would ideally hide the details of which HTML library is used, but it doesn't yet.
> 
> Currently hledger-web uses blaze-html, but hledger CLI reports use lucid.
> lucid has a more usable API than blaze-html (https://chrisdone.com/posts/lucid).
> lucid2's is even better.
> Unfortunately lucid* can not render multi-line or indented text.
> We want this so that humans can read and troubleshoot our HTML output.
> So a transition to blaze-html may be coming.

It seems to me that moving hledger to blaze-html is forced here, even though its API is ugly. At least until/unless https://github.com/chrisdone/lucid/discussions/161 happens. Let me know if you agree.

thielema:
I do not see the actual advantages of the Lucid API and I am completely ok with Blaze.

simonmichael:
For the record: 
Currently hledger-web uses blaze-html and hledger uses lucid.
In Blaze.hs there's a parallel set of blaze-html-based helpers for hledger, not yet used.
Today I added newlines to the lucid output, for human-readable html output. That has removed one reason for migrating away from lucid.

PR #2323, opened 2025-02-11
New command: "run"

adept:
Adds a new command, "run", that allows one to run hledger commands in interactive REPL or from command/script files:

```sh
#!/usr/bin/env hledger run
echo "Depth 1:"
balance assets --depth 1

echo "Depth 2:"
balance assets --depth 2

echo "Depth 3:"
balance assets --depth 3

echo "Uncategorized expenses:"
print expenses:unknown
```

### DIFF
--- a/doc/FILES.md
+++ b/doc/FILES.md
@@ -450,6 +450,7 @@ src/hledger/
           Register.md
           Rewrite.md
           Roi.md
+          Run.md
           Stats.md
           Tags.md
           Test.md
--- a/hledger/Hledger/Cli.hs
+++ b/hledger/Hledger/Cli.hs
@@ -118,6 +118,7 @@ import Hledger
 import Hledger.Cli.CliOptions
 import Hledger.Cli.Conf
 import Hledger.Cli.Commands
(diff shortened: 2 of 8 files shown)


jokesper:
Why not use a standard `bash` script?

adept:
> Why not use a standard `bash` script?

to avoid re-parsing the same journal file (quite large, in my case) over and over and over again

simonmichael:
> Why not use a standard `bash` script?

We can also say, for the REPL - allowing fast interactive reporting sessions without rereading the data, which could be helpful if it's large. 

We had GHCI already as a REPL for the haskell API, but this one is much more like the familiar command line (though not identical).

simonmichael:
Thank you @adept !

V1 of the run command now merged, and also a couple of [starter scripts](https://github.com/simonmichael/hledger/commit/d68a832d1ca50f89e2b10bc6d254ce6b6863cc9a).

Xitian9:
This is amazing, and will be a big help for my workflow. I'm looking forward to trying it. Thank you very much @adept!

PR #2328, opened 2025-02-22
run: cache all journal files, allow commands to use -f, and other improvements

adept:
Further improvement of 'run'.

- Now commands given to `run` could have `-f` argument and all those journals would be cached and read only once.

- `run` supports numeric arguments (`-3`, this was actually already true, just needed a doc fix) and @ARGFILEs

- `run` can call `run`


hledger/Hledger/Cli/Commands/Run.md
```diff
@@ -33,7 +33,7 @@ You can use `#!/usr/bin/env hledger run` in the first line of the file to make i
 
 - If you meant to provide file name as an argument, but made a mistake and a gave file name that does not exist, "run" will attempt to interpret it as a command.
 
-- Numeric flags like `-3` do not work, use long form `--depth 3`
+- You can pass `-f` to the `run` itself, and also to any commands given after it (or in the command file, or via REPL). When specific command does not have `-f` in its flags, it will use the journal(s) specified in the arguments of `run`. If command does have `-f` flag, this journal would be read and its contents would be cache, so if several commands specify the same `-f` flag, they will read the journal only once.
```
simonmichael, line 36: Let's see if this gets confusing or not.

I am reminded of GHCI's `:l(oad) FILE` and `:m(odule) (+-*)MODULE` commands, which let you bring things in and out of memory and move things in an out of scope.

simonmichael (commented):
> run can call run 

😂 mind blown!

This sounds great, I look forward to trying, though it might take another day or two. Actually I have no objection to merging, if we can hear from another reviewer/tester.

simonmichael:
Actually, please enlighten us: what happens with run in run ?

adept:
Lets run (pun intended!) through some examples:

You do: `hledger run -f file1.journal -- cmd1 -- cmd2 -- cmd3` (here and below `cmd1` etc refers to some hledger command)
What happens: `run` loads `file1.journal` and runs three commands on that journal

You do: `hledger run -- cmd1 -- cmd2 -- cmd3`
What happens: `run` loads your default journal and runs three commands on that journal

You do: `hledger run -f file1.journal -- cmd1 -- cmd2 -f file2.journal -- cmd3 -- cmd4 -f file2.journal`
What happens: `run` loads `file1.journal` and runs `cmd1` on it. It then loads `flie2.journal` and runs `cmd2` on it. It then runs `cmd3` on the cached contents of `file1.journal` and finally runs `cmd4` on a cached contents of `file2.journal`

You do: `hledger run -- cmd1 -- run -- cmd2`, trying to nest `run` invocations on the command line
What happens: `run` loads your default journal, runs `cmd1` on it, and then drops you into REPL. 
We can't effectively "nest" `--`-separated command sections on the command line, so if you specify everything on the command line, nested `run` invocations are not really possible

You do: `hledger run -f file1.journal script1.txt`, where script1.txt contains:
```
cmd1
run -f file2.journal script2.txt
cmd2 -f file2.journal
```
What happens: `run` will load `file1.journal`, execute `cmd1` on it, then it will load (and cache) `file2.journal` and run commands from `script2.txt` on it, then it will use cached `file2.journal` to run `cmd2`

Does it clear things up a little bit, or makes them worse? :)

simonmichael:
That is very helpful, and it probably deserves to be in the manual, as long as things work this way. 

Though, there's a risk of scaring users with complexity which they hopefully won't encounter. (But they probably will.)

Some questions:

> You do: hledger run -f file1.journal -- cmd1 -- cmd2 -f file2.journal -- cmd3 -- cmd4 -f file2.journal
...
> It then runs cmd3 on the cached contents of file1.journal and finally runs cmd4 on a cached contents of file2.journal

I find that intuitive for a REPL but surprising on the command line, where (at least for hledger) right-most options usually take precedence. So this example will be helpful.

> You do: hledger run -- cmd1 -- run -- cmd2, trying to nest run invocations on the command line
> What happens: run loads your default journal, runs cmd1 on it, and then drops you into REPL.

And cmd2 would never be executed ? Or would it execute if you exit the REPL without error ? Let's spell that out for completeness.

simonmichael:
Related: 

I gave the claude code and aider REPLs a test drive today. They are very pretty, and give ideas for what could be added to a hledger REPL.
I almost wish the REPL front end could be written in a "softer" language like python, which could evolve faster and use python's polished TUI libs. But I guess that's out, because it would complicate packaging and it would lose the key benefit of having native hledger data in memory.

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -108,5 +115,46 @@ runREPL findBuiltinCommand j = do
...
+{-# NOINLINE journalCache #-}
+journalCache :: MVar (Map.Map [String] Journal)
+journalCache = unsafePerformIO $ newMVar Map.empty
+
+-- | Key used to cache the journal given in the arguments to 'run'.
+defaultJournalKey :: [String]
+defaultJournalKey = ["journal specified in args of run"]
+
+hasStdin :: [String] -> Bool
+hasStdin fps = "-" `elem` fps
```
simonmichael: splitReaderPrefix should be used here in case it's a prefixed path like `csv:-`.

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -108,5 +115,46 @@ runREPL findBuiltinCommand j = do
       Just "quit" -> return ()
       Just "exit" -> return ()
       Just input -> do
-        liftIO $ runCommand findBuiltinCommand j $ parseCommand input
+        liftIO $ runCommand findBuiltinCommand $ parseCommand input
         loop
+
+{-# NOINLINE journalCache #-}
+journalCache :: MVar (Map.Map [String] Journal)
+journalCache = unsafePerformIO $ newMVar Map.empty
```
simonmichael, line 142: I guess this works by running newMVar only the first time journalCache is called, relying on GHC's usual unsafePerformIO behaviour. 

Is this a common idiom ? Is it behaviour we could rely on with future GHC, or GHC compiling to wasm, or other haskell compilers like microhs ?

Are there any implications for thread safety of the Hledger.Cli library ?

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -108,5 +115,46 @@ runREPL findBuiltinCommand j = do
...
+
+addJournalToCache :: Journal -> [String] -> IO ()
+addJournalToCache j journalpaths = modifyMVar_ journalCache $ \cache ->
+  if hasStdin $ dbg1 "addJournalToCache" journalpaths
+  then return cache
+  else return $ Map.insert journalpaths j cache
+
+withJournalCached :: CliOpts -> (Journal -> IO ()) -> IO ()
+withJournalCached cliopts f = do
+  journalpaths <- journalFilePathFromOptsNoDefault cliopts
```
simonmichael: I am a little confused, doesn't this return a single file path ? Wouldn't "journalpath" or "p" be clearer ? Or following my usual convention, "mp" (maybe p).

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -108,5 +115,46 @@ runREPL findBuiltinCommand j = do
...
+  -- if command does not have -f flags, use the same journal that was given to the "run" invocation
+  let key = case journalpaths of
+              Nothing -> defaultJournalKey
+              Just paths -> NE.toList paths
```
simonmichael: It seems to be converted to a (singleton) list here - why is that ?

hledger/Hledger/Cli/Commands/Run.txt
```diff
@@ -44,6 +44,14 @@ Caveats:
...
 -   Numeric flags like -3 do not work, use long form --depth 3
 
+-   You can pass -f to the run itself, and also to any commands given
+    after it (or in the command file, or via REPL). When specific
+    command does not have -f in its flags, it will use the journal(s)
+    specified in the arguments of run. If command does have -f flag,
+    this journal would be read and its contents would be cache, so if
+    several commands specify the same -f flag, they will read the
+    journal only once.
+
```
simonmichael, line 52: Typo: "cache". 
Also the wording seems to suggest caching is a special thing that happens when command has -f. Perhaps it could be clarified that all input files are cached.
(I sometimes say "input files", since they're not always .journals).

hledger/Hledger/Cli/Commands/Run.hs
simonmichael: Would the uglymemo library be of any help here ? We use it for other things, eg regexps.

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -81,23 +99,30 @@ parseCommand line =
...
+          withJournalCached opts (cmdaction' opts)
     [] -> return ()
 
-runREPL :: (String -> Maybe (Mode RawOpts, CliOpts -> Journal -> IO ())) -> Journal -> IO ()
-runREPL findBuiltinCommand j = do
-  putStrLn "Enter hledger commands, or 'help' for help."
+-- | Run an interactive REPL.
+runREPL :: (String -> Maybe (Mode RawOpts, CliOpts -> Journal -> IO ())) -> IO ()
+runREPL findBuiltinCommand = do
+  putStrLn "Enter hledger commands, or 'quit' for help."
```
simonmichael, line 125: Typo: quit for help.

hledger/Hledger/Cli/Commands/Run.hs
adept: This is fixed in later commits. Do you want me to squash everything in a single commit, perhaps?

hledger/Hledger/Cli/Commands/Run.hs
adept: MVar is used specifically for thread safety. Uglymemo does exactly this under the hood, btw: https://github.com/postgres/postgres/blob/0e42d31b0b2273c376ce9de946b59d155fac589c/src/backend/optimizer/path/costsize.c#L1171

So I am not making the situation any worse, as uglymemo is already doing all of this :)

I am not using uglymemo directly because it can't memoize `(a -> IO b)` functions, only pure ones.

hledger/Hledger/Cli/Commands/Run.hs
adept: This returns a list, if there are several `-f` options. I've called it `mbjournalpaths` in later commit

adept:
@simonmichael I realize that there is quite a lot of commits here, and I am always unsure if this is a good thing or not.

I am hesitant to squish them, at the same time I feel that they maybe make review more complicated? (Also, I dont use git often, so I am always scared of history rewriting lest I mess something up :)

simonmichael:
@adept I'm fine with leaving the commits as is for now if it's easier, though also feel free to rewrite them into fewer cleaner ones and force-push if you want. I can view them both ways as you say. At merge time we can decide separately how best to merge.

adept:
todo:
- split run and repl
- fix small issue with nested "run"s
- reconsider "echo"

simonmichael:
Noting some things we've discussed in chat, for the record:

I'm liking the split; at the very least it provides intuitive command names for both the batch mode and the interactive mode, and I think users will find it easier to grok these one at a time. Having said that, the split is not perfect:

- `run` can't yet read commands from stdin; I think it should support that.

- `repl`, in addition to its interactive mode, can also read commands from stdin and process them in batch mode. This will be counterintuitive to some, though not to unix hackers.

But it's early days. At the moment run and repl are still so similar it's not yet clear if they are best as one command or as separate commands. I do know as one command it was a lot to explain, test, and learn as a new user (see also #2338). But then again these are power user commands.

We'll keep the echo command for now and remove the "display unrecognized commands as output" fallback for robustness.

A few "LEDGER_JOURNAL" and "agument" typos need fixing and in the commands list, "repl" needs a one space indent.

simonmichael:
> But then again these are power user commands.

Counterargument: 
- combining commands into custom report scripts will be a pretty common use case
- entering a repl to be able to run reports faster (and in future, benefit from more UI aids) will also be desirable pretty early on
-

adept:
More changes:
- run can again read from stdin
- repl (and run) hides prompt and start banner when used non-interactively
- repl does not stop on errors
- typos fixed
- run and repl complain on unrecognized input instead of printing it out

adept:
@simonmichael by my count, there is nothing else planned (so far) left to implement. (just wanted to explicitly say that I am not sitting on any in-progress or uncommitted changes here :)

Xitian9:
I've played around with this a bit to see how it works. Thanks for the work!

The main problem I see is that `run` does not work with `--forecast` in internal commands. I suspect this has something to do with when journal finalisation occurs, and how that differs among different subcommands.

```
$ cat test.journal
2025-01-01  Buy apple pie
    expenses:food              $20
    assets:cash               -$20

~ every 6 months from 2025-02-01  Buy soylent green
    expenses:food:not people   $10
    assets:cash               -$10

$ cat cmdtest.txt
register --forecast

# It works if you call register with --forecast
$ hledger -f test.journal register --forecast
2025-01-01 Buy apple pie                  expenses:food                            $20           $20
                                          assets:cash                             $-20             0
2025-02-01 Buy soylent green              expenses:food:not people                 $10           $10
                                          assets:cash                             $-10             0
2025-08-01 Buy soylent green              expenses:food:not people                 $10           $10
                                          assets:cash                             $-10             0

# It works if you supply --forecast on the command line
$ hledger -f test.journal --forecast run cmdtest.txt
2025-01-01 Buy apple pie                  expenses:food                            $20           $20
                                          assets:cash                             $-20             0
2025-02-01 Buy soylent green              expenses:food:not people                 $10           $10
                                          assets:cash                             $-10             0
2025-08-01 Buy soylent green              expenses:food:not people                 $10           $10
                                          assets:cash                             $-10             0

# It does not work if --forecast is only in the run command
$ hledger -f test.journal run cmdtest.txt
2025-01-01 Buy apple pie                  expenses:food                            $20           $20
                                          assets:cash                             $-20             0
```

I suspect this may affect other journal modification commands (auto postings, aliases), but haven't explored in depth.

Here are some other minor things that came up when I attempted to use it. None are blockers. 

- `hledger run` does not pass common arguments to its subcommands, for example `hledger run --pretty=yes` doesn't pass `--pretty=yes` to all subcommands. Whether it should is a good question, but I naively thought it might.
- `hledger run` does not pass to a pager the way other commands do. Again, whether it should is an open question.

adept:
Regarding forecast, I think this is a symptom of a somewhat bigger issue that plagues other api parts as well, not just file reading. I would probably formulate it as follows: we have functions that take CliOpts (or ReportOpts, etc), but ignore some values from them and expect callers to do some extra work based on those ignored values.

I suspect that it is just too convenient to pass CliOpts around, so it is not feasible to require any function that takes CliOpts to use it to fullest possible extent... 

However, it is definitely true that unless you are very familiar with API, it is hard to know whether you are calling a "proper" high level entry point, or something mid-level - they both could take CliOpts and a filename, for instance. 


Alternative may be to use return type to somehow indicate that the Journal is "incomplete" and might need more processing steps?

adept:
Ok, let's write down some words.

This is how I want `run` to behave:

1. `run`, like any other command, would either read a default journal, or a journal given by `-f` , This should take into account other command-line flags that influence journal reading/parsing.

2. This journal is then passed to all the commands that `run` runs, except:
   - If the command is explicitly given `-f`, it will operate on that input file (+ whatever other command line options were given)
   - If the command is explicitly given some options that are influencing input, but no `-f`, then it would read the same input files as `run` read (implicitly or explicitly), with those options applied

3. As a consequence of 1+2, if run was was given `-f non-default.journal`, then the only way to read a default journal for some command is to give it `-f default.journal`. Default journal would never be read implicitly in this case.

4. For nested run calls, we first determin which journal it should operate on using rules (1),(2),(3), and then we start it and all the rules above apply to its behavior

Can we somehow get rid of implicit behavior of (2)? 

We could require `-f` to be always given to any command. This is ugly and bad ui

We could introduce some special pseudo-name so that one can say  (semantically) `-f <the same>`. This is also ugly, and special syntax is bad.

We can introduce scopes (gah!), like:
```
with -f some.journal --forecast:
   do this
   and this
   and this
```
and commands inside `with` blocks are forbidden to have `-f`. This looks like "even advanced system contain 50% of the implementation of lisp", and if we go there, we might as well think about embedding lua or python :)

adept:
Now, lets figure out what influences ingestion of input files.

Commands (other than `run`) dont read input files, `Cli.hs` does this for them. It does this via `withJournalDo`, which takes `CliOpts`. This is how that call unfolds:

`withJournalDo cliopts` calls:
- `readJournalFiles inputopts`
    - lots of functions that eventually call `readJournal iopts`, which in turn:
       -  determines reader from `iopts`
       -  calls `reader.readFn iopts`, where `readFn` is somethng like `JournalReader.reader` ....
- `journalTransform cliopts`

Ok, so iopts are essentially "part of the journal" (as they are passed into parser).
I think i will modify the code to cache files by (iopts, filename), and to apply `journalTransform` to the cached value every time journal is requested .. I think this should definitely fix the `--forecast` bug @Xitian9 experiened, and maybe fix all other issues as well.

adept:
@Xitian9 I've pushed this change, your forecast example now works :)

simonmichael:
I appreciate the valuable additional review/discussion/updates here, hope to catch up soon.

Xitian9:
Thanks! I've run it and can confirm that the `--forecast` example now works. However, speed seems to have taken a big hit, and running my example (5 balance commands in a single run script) is no faster than running hledger 5 separate times.

adept:
Hmm. Do you mind running it with --debug 1 and looking at whether it caches files or reads them every time?

Xitian9:
It seems to be reading and caching 5 times, and using the cache once. With 5 subcommands, that would be 1 attempt for each subcommand, and 1 for the overall run command. There is only one successful cache hit, probably due to the fact that the reportspan and forecast options change for most of them.

adept:
Would you be able to share the options that you use for all these commands?

Xitian9:
```
balance      --pretty=yes --value=end --daily -b "13 days ago" -e tomorrow @expenses-report.args                                  --output-file=/tmp/tmp.V86rCPRgsp/expenses.out
balancesheet --pretty=yes --tree --layout=wide @netliquid.args                                                                    --output-file=/tmp/tmp.V86rCPRgsp/netliquid.out
balance      --pretty=yes --tree @pay.args --period "every 21st day of month from 2025-02-21 to 2025-03-21"                       --output-file=/tmp/tmp.V86rCPRgsp/paynow.out
balance      --pretty=yes --tree @pay.args --period "every 21st day of month from 2025-02-21 to 2025-03-21" --forecast            --output-file=/tmp/tmp.V86rCPRgsp/paymonth.out
balance      --pretty=yes --tree @pay.args --period "every 21st day of month from 2025-02-21 to 2025-05-21" --forecast receivable --output-file=/tmp/tmp.V86rCPRgsp/pay3month.out
```

The arg files contain regular expression account matching and depth arguments.

simonmichael:
I'm caught up now. I see how this is complicated. And yes, the overall processing pipeline may have grown a few more offshoots than it should have, perhaps some cleanup will help later. 

I know users will have "fun" figuring out the effects/scopes also. But I suppose that optimal [functional spec](https://github.com/simonmichael/hledger/pull/2328#issuecomment-2695425992) that adept wrote will be the most intuitive overall, once understood, since it "follows the rules". If we can achieve it. But if we must impose some restrictions / document some limitations for now, that's ok too.

simonmichael:
For the record I'm overdue for the ideal release date, which is my fault, but that is flexible so it's ok to spend more time tinkering with this if needed. In fact I think it's a good that we're doing that. This is a big feature and the most prominent in 1.42 probably.

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -38,41 +49,70 @@ runmode = hledgerCommandMode
...
+-- We expose this do-nothing function so that it could be included in the list of all commands inside
+-- Hledger.Cli.Commands and ensure that "run" is recognized as a valid command by the Hledger.Cli top-level
+-- command line parser. That parser, however, would not call run'. It has a special case for "run", and
+-- will call "run" (see below), passing it `findBuiltinCommand`, thus breaking circular dependency.
+runOrReplStub :: CliOpts -> Journal -> IO ()
+runOrReplStub _opts _j = return ()
+
+-- | Default input files + InputOpts that would be used by commands if
+--   there is no explicit alternative given
+newtype DefaultRunJournal = DefaultRunJournal (NE.NonEmpty String)
```
simonmichael, line 75: Not a release blocker, but I think we'll be able to make this clearer. 
It's not a journal, it's a list of file paths.
What does "+ InputOpts" mean here ? I thought maybe it meant "also includes options like --infer-equity" but it seems not.

simonmichael: PS I found it useful to derive Show here for debugging, might be worth adding.

adept: This is an obsolete comment, sorry :( I'll clean this up

adept:
@Xitian9 I figured out why, it is because of this guy sneakily living inside InputOpts, so that `journalFinalise` would be able to add forecast transactions:
```
    ,reportspan_        :: DateSpan             -- ^ a dirty hack keeping the query dates in InputOpts. This rightfully lives in ReportSpec, but is duplicated here.
```

I'll push a thing to work around this

adept:
> For the record I'm overdue for the ideal release date, which is my fault, but that is flexible so it's ok to spend more time tinkering with this if needed. In fact I think it's good that we're doing that. This is a big feature and the most prominent in 1.42 probably.

I feel that we are at the final stage of ironing the wrinkles out, it should not be long :)

adept:
@Xitian9 if you can give it another spin now it would be great. It seem to be behaving better in my tests

simonmichael (commented):
The --version, --man, --info, and --tldr flags, with a command, all seem to be ignored. Normally they would show that information instead of running the command's normal action.

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -81,32 +121,92 @@ parseCommand line =
...
+    minput <- getInputLine prompt
     case minput of
       Nothing -> return ()
       Just "quit" -> return ()
       Just "exit" -> return ()
       Just input -> do
-        liftIO $ runCommand findBuiltinCommand j $ parseCommand input
-        loop
+        liftIO $ (runCommand defaultJournalOverride findBuiltinCommand $ parseCommand input)
+                  `catch` (\(e::ErrorCall) -> putStr $ show e)
```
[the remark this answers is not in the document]
simonmichael: (Comment edited)

simonmichael:
`run`: when specifying commands on the command line, the requirement for a leading -- is not mentioned in the command line help and without it the results can be confusing, eg try 
```
hledger run files
hledger run files -- files
hledger run files -- files -- files
hledger run files -- files -- files -- files
```

simonmichael:
Actually I see some confusing results even with a leading --. Eg shouldn't this print - twice ? It prints - once and $LEDGER_FILE once.
```
echo | hledger -f- run -- files -- files
```

simonmichael:
Here's a mockup of some func tests:
```
# * run command

# ** 0. Run with empty command line, does nothing.
$ hledger -f- run

# ** 0. Run one command from command line, no leading --
$ hledger -f- run files
>
-

# ** 0. Run one command from command line, with leading --
$ hledger -f- run -- files
>
-

# ** 0. Run two commands from command line, no leading --
$ hledger -f- run files -- files
>
-
-

# ** 0. Run two commands from command line, with leading --
$ hledger -f- run -- files -- files
>
-
-
```

simonmichael:
Adding Repl.txt to package.yaml extra-source-files will make `just test` pass.

adept:
> Actually I see some confusing results even with a leading `--`. Eg shouldn't this print `-` twice ? It prints `-` once and $LEDGER_FILE once.
> 
> ```
> echo | hledger -f- run -- files -- files
> ```

Ugh. This is because `Cli.hs` rearranges arguments even before they get to `run`, and `run` receives this:

```
run -- files -- files -f-
```

Perhaps `moveFlagsAfterCommand` should stop at first `--`? I feel this might be important for add-on commands as well, maybe?

simonmichael:
I guess our command line processing isn't complicated enough yet. 😂

This sounds like a sensible sort of thing, if tests pass.

simonmichael:
To fix that failing commands test, I needed to add `commands` to the list at section 6.5.2 in main. These are the commands that don't (and shouldn't) read an input file.

`run` and `repl` currently require an input file, eg `LEDGER_FILE= hledger repl` fails, but they probably shouldn't.

simonmichael:
Cc'ing here for the record:

> I'm sorry for the complexity of command line processing. It's just trying to make everything work as a user would want/expect... while also working around an annoying limitation/mystery of cmdargs (can't ignore unrecognised things). And doing all the same stuff with dynamically discovered addon commands.

adept:
I've tried to add `run.test`, but ran into a weird error: when tests run, i get the below:
```
hledger: <stdin>: hGetBuffering: illegal operation (handle is closed)
```

I dont understand where hGetBuffering is coming from :( I suspect it is "is stdin a terminal?" check, but I am not sure :( 
A problem for tomorrow, I think

adept:
Ok, I figured the hGetBufferring thing - it was an attempt to read stdin twice ...

I fixed it and added `run.test`

hledger/Hledger/Cli/Commands/Run.hs
```diff
@@ -211,11 +214,25 @@ withJournalCached defaultJournalOverride cliopts cmd = do
...
+            case cache of
+              Just cached -> do
+                dbg1IO "readStdin using cached stdin" "-"
+                return (cache, cached)
+              Nothing -> do
+                dbg1IO "readStdin reading and caching stdin" "-"
+                stdinContent <- readFileOrStdinPortably "-"
+                return (Just stdinContent, stdinContent)
+          hndl <- liftIO $ inputToHandle stdinContent
+          readJournal iopts Nothing hndl
```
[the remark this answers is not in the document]
simonmichael: Those tests are failing because the options before `run` get moved to the end by moveFlagsAfterCommand.
```
-- - XXX Relative order of flags is mostly but not entirely preserved, currently:
--   pre-command flags get moved to the end, after post-command flags. 
```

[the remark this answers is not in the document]
simonmichael: (You said this yesterday, I understand it now.)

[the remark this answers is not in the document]
simonmichael: I have pushed a fix to master; now only two tests fail:

```
# ** 0. Run with empty command line, does nothing.
$ hledger -f- run
```
This one gives "illegal operation (handle is closed)".

```
# ** 0. Run two commands from command line, no leading --
$ hledger -f- run files -- files
>
-
-
```
Something (cmdargs ?) consumes the first --. If we can't avoid that, we could maybe have run raise an error if the first non-file argument is not -- ?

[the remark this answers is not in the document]
adept: It is either cmdargs or Cli.hs. Whatever this is, the end result is as follow:

- Commandline "hledger -f- run files -- files" leads to Run.hs getting "files files" as args
- Commandline "hledger -f- run -- files -- files" leads to Run.hs getting "files -- files" as args

So Run.hs does not have enough info to know if there was leading "--" or not :(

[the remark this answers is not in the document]
adept: https://hackage.haskell.org/package/cmdargs-0.10.22/docs/System-Console-CmdArgs-Explicit.html says "Parsing rules: Command lines are parsed as per most GNU programs. Short arguments single letter flags start with -, longer flags start with --, and everything else is considered an argument. Anything after -- alone is considered to be an argument. For example:"

So I am pretty sure that it is cmdargs eating first "--"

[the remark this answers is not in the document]
adept: Well... If we are seriously considering this, maybe getting the original command line or if getArgs and looking at it could get us somewhere, but this feels... horrible :)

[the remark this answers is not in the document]
simonmichael: FTR: we did that, excellent workaround.

simonmichael:
Time to merge this. Thank you!

Xitian9:
Just following up to confirm here that my example now has 6 calls, of which 4 successfully use the cache, consistent with what is expected if `--forecast` needs to regenerate. Thanks!

PR #2333, opened 2025-02-24
feat: csv: allow multiple matchers on the same line

thomie:
### Description
If blocks and If tables now allow multiple matchers on the same line. The only requirement is that the first (n-1) matchers end with a '$' (i.e. end of string marker).

Example `if block` with two matchers on the same line:

	if %date ^2025-02-22$ & %description amazon
	    account2 expenses:books

Example `if table` with two matchers on the same line:

	if|account2
	%date ^2025-02-22$ & %description amazon | expenses:books
	

### Justification
I make heavy use of `if tables`.  Example:

```
if|account2|comment2
%description ^CARTE ../../.. AUCHAN SUPERMAR 4 CB | expenses:food:supermarket|
%description ^CARTE ../../.. CARREFOUR.FR CB | expenses:food:supermarket|
...
```

But then sometimes I need to match 2 separate fields, usually `%date` and `%description`, and  I will have to fall back to using an `if block`:

```
if
%description ^VIR INST STICHTING MOLLIE PAYMENTS$
& %date ^2024-06-20$
  account2 expenses:food:supermarket
```

My rules files would look much nicer like this:
```
if|account2|comment2
%description ^CARTE ../../.. AUCHAN SUPERMAR 4 CB | expenses:food:supermarket|
%description ^CARTE ../../.. CARREFOUR.FR CB | expenses:food:supermarket|
%description ^VIR INST STICHTING MOLLIE PAYMENTS$ & %date ^2024-06-20$ | expenses:food:supermarket|
...
```

That's what this MR accomplishes.

### Alternative solution

Instead of using field matchers, I could use whole record matchers like so:

``` 
if|account2|comment2
%description ^CARTE ../../.. AUCHAN SUPERMAR 4 CB | expenses:food:supermarket|
%description ^CARTE ../../.. CARREFOUR.FR CB | expenses:food:supermarket|
2024-06-20,VIR INST STICHTING MOLLIE PAYMENTS | expenses:food:supermarket|
```

This works, but I suspect whole record matchers are slower to execute than field matchers. I have some csv files where dates are in `%d/%m/%Y` format, and matching whole records with dates seemed very slow. I also just like my matchers to be as restrictive as possible, so using `^` and `$` on field matchers, so that I'm sure I'm not accidentally matching a record I didn't mean to.

### Discussion

Hello @simonmichael, let me know if this is a feature you would like to include. This MR is more a prove of concept; the implementation is probably not very efficient, and the documentation needs updating. Thanks.

### DIFF
--- a/hledger-lib/Hledger/Read/RulesReader.hs
+++ b/hledger-lib/Hledger/Read/RulesReader.hs
@@ -637,19 +637,19 @@ conditionaltablep = do
                  ])
   when (null body) $
     customFailure $ parseErrorAt start $ "start of conditional table found, but no assignment rules afterward"
-  return $ flip map body $ \(m,vs) ->
-    CB{cbMatchers=[m], cbAssignments=zip fields vs}
+  return $ flip map body $ \(ms,vs) ->
+    CB{cbMatchers=ms, cbAssignments=zip fields vs}
   <?> "conditional table"
   where
-    bodylinep :: Char -> [Text] -> CsvRulesParser (Matcher,[FieldTemplate])
+    bodylinep :: Char -> [Text] -> CsvRulesParser ([Matcher],[FieldTemplate])
     bodylinep sep fields = do
       off <- getOffset
-      m <- matcherp' $ void $ char sep
+      ms <- matcherp' (lookAhead . void . char $ sep) `manyTill` char sep
       vs <- T.split (==sep) . T.pack <$> lift restofline
       if (length vs /= length fields)
         then customFailure $ parseErrorAt off $ ((printf "line of conditional table should have %d values, but this one has only %d" (length fields) (length vs)) :: String)
-        else return (m,vs)
-      
+        else return (ms,vs)
+
 
 -- A single matcher, on one line.
 -- This tries to parse first as a field matcher, then if that fails, as a whole-record matcher;
@@ -700,7 +700,7 @@ matcherprefixp :: CsvRulesParser MatcherPrefix
 matcherprefixp = do
   lift $ dbgparse 8 "trying matcherprefixp"
   (do
-    char '&' >> lift skipNonNewlineSpaces
+    char '&' >> optional (char '&') >> lift skipNonNewlineSpaces
     fromMaybe And <$> optional (char '!' >> lift skipNonNewlineSpaces >> return AndNot))
   <|> (char '!' >> lift skipNonNewlineSpaces >> return Not)
   <|> return Or
@@ -718,10 +718,12 @@ regexp end = do
   lift $ dbgparse 8 "trying regexp"
   -- notFollowedBy matchoperatorp
   c <- lift nonspace
-  cs <- anySingle `manyTill` end
+  cs <- anySingle `manyTill` (double_ampersand <|> end)
   case toRegexCI . T.strip . T.pack $ c:cs of
        Left x -> Fail.fail $ "CSV parser: " ++ x
        Right x -> return x
+  where
+    double_ampersand = lookAhead . void $ char '&' >> char '&'
 
 -- -- A match operator, indicating the type of match to perform.
 -- -- Currently just ~ meaning case insensitive infix regex match.
@@ -1559,7 +1561,7 @@ tests_RulesReader = testGroup "RulesReader" [
    ,testCase "quoted name" $ parseWithState' defrules csvfieldreferencep "%\"csv date\"" @?= (Right "%\"csv date\"")
    ]
 
-  ,testGroup "matcherp" [
+  ,testGroup "recordmatcherp" [
 
     testCase "recordmatcherp" $
       parseWithState' defrules matcherp "A A\n" @?= (Right $ RecordMatcher Or $ toRegexCI' "A A")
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks @thomie! I will give more feedback on this when I can. Other reviewers welcome.

Thanks for including the justification notes, that is very helpful.

I wouldn't expect any detectable speed difference between field matches and whole record matchers. If that's truly what's happening, it may be worth a separate bug report with a repro.

I would be concerned about this limiting what you can express in a search pattern (eg maybe you want to match on payee names like "Barnes & Noble"), and also about it being a breaking change that could break existing csv rules.

thomie:
> I wouldn't expect any detectable speed difference between field matches and whole record matchers. If that's truly what's happening, it may be worth a separate bug report with a repro.

I'll see if I can make a benchmark test for csv rules files.

> maybe you want to match on payee names like "Barnes & Noble"

That still works, no changes there. 

> a breaking change that could break existing csv rules

I think only faulty csv rules that didn't match anything before could now start matching something. I think it would be very unlikely that we would see this in the wild though. But here's an example:

```
if
%description The $ store
  account2 expenses:gadgets
```

The intention is probably to match a csv row containing "The $ store", but the user forgot/didn't know that `$` is the end of string marker.

So with current hledger, this matcher doesn't match anything at all:

```
hledger-lib> parseWithState' defrules matcherp "%description The $ store"
Right (FieldMatcher Or "%description" (RegexpCI "The $ store"))
hledger-lib> regexMatchText (toRegexCI' "The $ store") "The "
False
hledger-lib> regexMatchText (toRegexCI' "The $ store") "The $"
False
hledger-lib> regexMatchText (toRegexCI' "The $ store") "The $ store"
False
```

After the change from this MR, hledger will treat `%description The $ store` as two separate matchers, or-ed together:
```
hledger-lib> parseWithState' defrules (many matcherp) "%description The $ store"
Right [FieldMatcher Or "%description" (RegexpCI "The $"),RecordMatcher Or (RegexpCI "store")]
hledger-lib> regexMatchText (toRegexCI' "The $") "The "
True
hledger-lib> regexMatchText (toRegexCI' "store") "Some other store"
True
```
It would match csv rows containing a description field that ends with "The ", or csv rows containing the word "store" anywhere.

I argue the user would quickly realize their mistake either way (old hledger or new hledger), and fix their rule to be:

```
if
%description The \$ store
  account2 expenses:gadgets
```

This rule works the same with or without this commit, indeed matching  "The $ store".

simonmichael:
>> maybe you want to match on payee names like "Barnes & Noble"
> That still works, no changes there.

Oh I see, because the same-line-AND operator is effectively "$ &" (unlike the next-line-AND operator, which is just "&").
I see how that makes this change safer.

> I think only faulty csv rules that didn't match anything before could now start matching something. I think it would be very unlikely that we would see this in the wild though. But here's an example:
>
> if
> %description The $ store
> ...
> After the change from this MR, hledger will treat %description The $ store as two separate matchers, or-ed together:

And as you say, if there's an unquoted $ anywhere but at the end of a match pattern, that seems to be a user error, which would never match. So changed behaviour here could be justfiable.

> I argue the user would quickly realize their mistake either way (old hledger or new hledger), 

This I'm not so sure about. Expert users who've already hit this once, yes.

> if
> %description The \$ store
> ...
> This rule works the same with or without this commit, indeed matching "The $ store".

Nice that you detect that case.

This is a clever proposal. 
It's slightly unfortunate that 
- same-line-AND and next-line-AND operators would be different
- it's a unique convention which many users trip up on at first, as I did. 
- and when they do it may not be easy for them to pinpoint the problem, forcing them to come to the support fora or study the docs.

Does anyone have ideas for making this better ?

simonmichael:
I imagine same-line-NOT (!) is not enabled by this PR, correct ?

simonmichael:
So we could make the doc a little more specific, eg

*(Since 1.42:)*
You can write multiple matchers on one line, separated by `&`, 
if each preceding matcher ends with `$`, the end-of-string marker.
(In other words: you can write multiple matchers separated by `$ &`.)
Note you can't negate these same-line-ANDs with `!` (there is no same-line-NOT).

thomie:
At the moment, all of the following work (I had only tried the first up to now):

* pattern$ & pattern
* pattern$ & ! pattern
* pattern$ ! pattern
* pattern$ pattern

```
hledger-lib> parseWithState' defrules conditionaltablep "if,account2\n%description ^abc$ & %date ^2025-02-22$,foo"
Right [CB {cbMatchers = [FieldMatcher Or "%description" (RegexpCI "^abc$"),FieldMatcher And "%date" (RegexpCI "^2025-02-22$")], cbAssignments = [("account2","foo")]}]

hledger-lib> parseWithState' defrules conditionaltablep "if,account2\n%description ^abc$ & ! %date ^2025-02-22$,foo"
Right [CB {cbMatchers = [FieldMatcher Or "%description" (RegexpCI "^abc$"),FieldMatcher AndNot "%date" (RegexpCI "^2025-02-22$")], cbAssignments = [("account2","foo")]}]
 
hledger-lib> parseWithState' defrules conditionaltablep "if,account2\n%description ^abc$ ! %date ^2025-02-22$,foo"
Right [CB {cbMatchers = [FieldMatcher Or "%description" (RegexpCI "^abc$"),FieldMatcher Not "%date" (RegexpCI "^2025-02-22$")], cbAssignments = [("account2","foo")]}]
 
hledger-lib> parseWithState' defrules conditionaltablep "if,account2\n%description ^abc$ %date ^2025-02-22$,foo"
Right [CB {cbMatchers = [FieldMatcher Or "%description" (RegexpCI "^abc$"),FieldMatcher Or "%date" (RegexpCI "^2025-02-22$")], cbAssignments = [("account2","foo")]}]
```

I do understand your concern. Perhaps this is adding complexity for too little gain. About my example from earlier,  `The$store`, perhaps it's less surprising, and easier to debug, if it doesn't match anything than if it matches all records containing the word "store".  I could also write a hledger preprocessor for my own use if I'm the only one who cares about this feature.

However, me myself I found it surprising that hledger _didn't_ work like this out of the box. The end of string marker `$` is supported, so to me a regex pattern ends either at the `$` or at the end of the line. I expected to be able to combine patterns on the same line with the same operators as when they are on multiple lines. I had to dig into the source to figure out why it didn't work.

simonmichael:
On Thu, Feb 27, 2025, at 08:08, Thomas Miedema wrote:
> At the moment, all of the following work (I had only tried the first up to now):
> 
>  • pattern$ & pattern
>  • pattern$ & ! pattern
>  • pattern$ ! pattern
>  • pattern$ pattern

Cool. And interesting, I'm not sure what 3 and 4 would match, maybe nothing.

> About my example from earlier, The$store, perhaps it's less surprising, and easier to debug, if it doesn't match anything than if it matches all records containing the word "store".

Yes I think you're right there; this is intuitive, at least once you find it, it will make sense; "that's how $ works in a regexp".
And, we could reliably detect this and raise an error "$ should be at the end of a pattern; if you meant a literal $, use \$".

> However, me myself I found it surprising that hledger didn't work like this out of the box. The end of string marker $ is supported, so to me a regex pattern ends either at the $ or at the end of the line. I expected to be able to combine patterns on the same line with the same operators as when they are on multiple lines.

That's understandable, and I agree that would be ideal.

I think requiring $ at the end of all preceding patterns is going to be unintuitive for most people.
So far I think the most intuitive and least disruptive options are

- change the meaning of & (and accept some annoying breakage, and try to find other ways to minimise that)
- use && (less breakage; and possibly support/suggest && as an alternate spelling for next-line-AND, for consistency)

thomie:
> I think requiring $ at the end of all preceding patterns is going to be unintuitive (and restricting) for most people.

Ah, I didn't understand your `&&` proposal earlier.

I think you mean the use of `$` wouldn't be _required_ for any of the patterns.

So `%description MOLLIE && %date 2024-06-20` would combine the matchers `%description MOLLIE` and `%date 2024-06-20` on the same line with `&&`.

I like it. It would allow combining even more matchers on the same line. 

What about `&& !` (and not),  `||` (or) and possibly `|| !` (or not)?

simonmichael:
That's right, no requirement for $. 
`&& !` would be great (seems like that will just work, based on your test above).

`||` seems less needed. It would be intuitive, but then people would probably try grouping with parentheses as well.

`|| !` I'm having trouble with, is that ever useful ?

Also we should not forget that the original goal of CSV rules was to be a minimal, terse, simple-as-possible DSL usable by non-programmers (so called).

thomie:
I force pushed my commit. 

Matchers can now be combined with `&&` or `&& !` on the same line or on multiple lines.

I updated the documentation.

What do you think?

adept (approved):
Sorry for the delay, i took this change out for a spin, and I very much like it!

simonmichael:
Great! Thank you both!

PR #2337, opened 2025-02-26
;cln: unittest.hs: remove bothersome PackageImports

thomie:
When running `cd hledger-lib && ghci test/unittest.hs`, ghci complains with:

```
  test/unittest.hs:7:1: error:
      Could not find module ‘Hledger’
      It is not a module in the current program, or in any known package.
    |
  7 | import "hledger-lib" Hledger (tests_Hledger)
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Failed, no modules loaded.
```

I suspect the "hledger-lib" package-qualified import isn't really necessary anymore. After removing it, above ghci command works as expected.

This instance of PackageImports was introduced in commit c7574b8005a84c6116a2871313ce8417b492a823 (2019).

### DIFF
--- a/hledger-lib/hledger-lib.cabal
+++ b/hledger-lib/hledger-lib.cabal
@@ -176,7 +176,6 @@ test-suite doctest
   type: exitcode-stdio-1.0
   main-is: doctests.hs
   hs-source-dirs:
-      ./
       test
   ghc-options: -Wall -Wno-incomplete-uni-patterns -Wno-missing-signatures -Wno-orphans -Wno-type-defaults -Wno-unused-do-bind
   build-depends:
@@ -239,7 +238,6 @@ test-suite unittest
   type: exitcode-stdio-1.0
   main-is: unittest.hs
   hs-source-dirs:
-      ./
       test
   ghc-options: -Wall -Wno-incomplete-uni-patterns -Wno-missing-signatures -Wno-orphans -Wno-type-defaults -Wno-unused-do-bind
   build-depends:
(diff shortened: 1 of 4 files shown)


thomie:
The comments mention:

> -- package-qualified import to avoid cabal missing-home-modules warning (and double-building ?)

How can I reproduce this issue?

thomie:
> The comments mention:
>
>>    -- package-qualified import to avoid cabal missing-home-modules warning (and double-building ?)
>
> How can I reproduce this issue?

To answer my own question:
    
The missing-home-modules warning and the double building can indeed be reproduced by running (after removing the "hledger-lib" package-qualified import): `cd hledger-lib && cabal build unittest`. It will first build `hledger-lib`, then show a warning about missing-home-modules, and then build `hledger-lib` again.

After comparing the `unittest` sections of `hledger.cabal` and `hledger-lib.cabal`, the solution turned out to be to remove `./` from `hs-source-dirs` for `unittest`. Don't ask me why it works though!

## Conclusion
With this commit, the missing-home-modules warning does not reappear, the double building doesn't reappear, and `ghci test/unittest.hs` now works as expected. Overall it's a nice cleanup.

simonmichael:
I trust you. :) For the record, which version(s) of cabal-install did you test with ?

Thanks for the explanation and cleanup!

thomie:
> For the record, which version(s) of cabal-install did you test with ?

cabal-install version 3.14.1.0

Issue #2332, opened 2025-02-23
Terminal width lookup fails unnecessarily if terminfo unpopulated


hseg:
(Note: Very niche reproducer)

As noted in https://github.com/simonmichael/hledger/blob/master/hledger/Hledger/Cli/CliOptions.hs#L626, ifthe `terminfo` entry for `$TERM` can't be read, `setupTermFromEnv` throws an error. In particular, I'm encountering this when building `hledger` in a clean chroot that contains all its build dependencies, but doesn't contain the terminfo definitions for the terminal (kitty) I'm running the chrooted build in.

I'm a little loath to add kitty to the chroot unnecessarily, especially in view of the fact that `hledger` doesn't actually need an entire terminfo entry, just enough to tell the number of columns in the terminal. Given that the tests also pass with `TERM=dumb`, I'm guessing that its fallback settings are widely supported enough to work as an alternative.

Hence, I'd either recommend replacing the call to `setupTermFromEnv` with

```haskell
setupTermFromEnv `catch` \(_ :: SetupTermError) -> setupTerm "dumb"
```

or document this dependency on `TERM` being valid somewhere (perhaps this ticket suffices).

For my part, I'm adding `TERM="dumb"` to my packaging script until this is fixed (much like I'd set `LANG=C` to reset to a known-good situation).

simonmichael:
Thanks for the report and suggestion. Is it something you could try to implement, @hseg ?

simonmichael:
Now fixed thanks to #2340 I believe ?

hseg:
Yes - if I did things right merging it should've closed this

28 feb 2025 03:31:06 Simon Michael ***@***.***>:

PR #2340, opened 2025-02-27
Make terminal width lookup more robust

hseg:
On some systems, `TERM` is set to a value that doesn't have a valid terminfo entry. Rather than hackily fall back on a value for `TERM` that appears to work in most contexts (`TERM=dumb`) but which isn't guaranteed anywhere to be valid, use proper POSIX ioctls to get the tty width.
This has the added bonus of also working on Windows.
In fact, we already settled on computing the terminal size in this way in hledger-lib, so this commit centralizes the choice of the logic there.

Also added a note for alternative methods and their tradeoffs, in case this turns out to be fragile on some systems.

Additionally, `COLUMNS` is a bashism that isn't particularly usable for our purposes (for one thing, it isn't updated when the terminal is resized), so I removed its usages.

Finally, reducing our sources of truth for terminal width to 2 permitted removing the partial `NE.fromList` which I took advantage of.

Fixes: #2332

### DIFF
--- a/hledger-lib/Hledger/Utils/IO.hs
+++ b/hledger-lib/Hledger/Utils/IO.hs
@@ -413,8 +413,30 @@ hasOutputFile = do
 
 -- Terminal size
 
+-- [NOTE: Alternative methods of getting the terminal size]
+-- terminal-size uses the TIOCGWINSZ ioctl to get the window size on Unix
+-- systems, which may not be completely portable according to people in
+-- #linux@liberachat.
+--
+-- If this turns out to be the case, supplementary coverage can be given by
+-- using the terminfo package.
+--
+-- Conversely, terminfo on its own is not a full solution, firstly because it
+-- only works on Unix (not Windows), and secondly since in some scenarios (eg
+-- stripped-down build systems) the terminfo database may be limited and lack
+-- the correct entries. (A hack that sometimes works but which isn't robust
+-- enough to be relied upon is to set TERM=dumb -- while this advice does appear
+-- in some places, it's not guaranteed to work)
+--
+-- In any case, $LINES/$COLUMNS should not be used as a source for the terminal
+-- size -- that is a bashism, and in particular they aren't updated as the
+-- terminal is resized.
+--
+-- See #2332 for details
+
 -- | An alternative to ansi-terminal's getTerminalSize, based on
 -- the more robust-looking terminal-size package.
+--
 -- Tries to get stdout's terminal's current height and width.
 getTerminalHeightWidth :: IO (Maybe (Int,Int))
 getTerminalHeightWidth = fmap (fmap unwindow) size
--- a/hledger/hledger.cabal
+++ b/hledger/hledger.cabal
@@ -98,11 +98,6 @@ flag debug
   manual: True
   default: False
 
-flag terminfo
-  description: On POSIX systems, build with the terminfo lib for detecting terminal width
-  manual: False
-  default: True
-
 flag threaded
   description: Build with support for multithreaded execution
   manual: False
@@ -195,9 +190,6 @@ library
     , utility-ht >=0.0.13
     , wizards >=1.0
   default-language: Haskell2010
-  if (!(os(windows))) && (flag(terminfo))
-    build-depends:
-        terminfo
   if (flag(debug))
     cpp-options: -DDEBUG
 
@@ -247,9 +239,6 @@ executable hledger
     , utility-ht >=0.0.13
     , wizards >=1.0
   default-language: Haskell2010
-  if (!(os(windows))) && (flag(terminfo))
-    build-depends:
-        terminfo
   if (flag(debug))
     cpp-options: -DDEBUG
   if flag(threaded)
@@ -300,9 +289,6 @@ test-suite unittest
     , utility-ht >=0.0.13
     , wizards >=1.0
   default-language: Haskell2010
-  if (!(os(windows))) && (flag(terminfo))
-    build-depends:
-        terminfo
   if (flag(debug))
     cpp-options: -DDEBUG
 
@@ -353,8 +339,5 @@ benchmark bench
     , wizards >=1.0
   buildable: False
   default-language: Haskell2010
-  if (!(os(windows))) && (flag(terminfo))
-    build-depends:
-        terminfo
   if (flag(debug))
     cpp-options: -DDEBUG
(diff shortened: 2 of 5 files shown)


simonmichael:
>  it isn't updated when the terminal is resized

Slightly more detail:
Normally, in an ordinary bash window, it will update. 
With bash running in an emacs shell buffer, it will typically update; but sometimes it gets stuck and stops updating.
In osh (a bash replacement), it doesn't update, yet.
Not sure about other shells.

simonmichael:
Thanks for investigating this. Is there a current bug/problem that this fixes ? If not, I'd like to try and find one.

hseg:
What, the removal of `COLUMNS`?
In trying to substantiate my understanding that it is a bashism, it turns out it is defined by POSIX, section 8.3 https://pubs.opengroup.org/onlinepubs/9799919799/
> This variable shall represent a decimal integer >0 used to indicate the user's preferred width in column positions for the terminal screen or window; see /3.75 Column Position[https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap03.html#tag_03_75]/. If this variable is unset or null, the number of columns shall be set according to the terminal window size (see XSH /tcgetwinsize[https://pubs.opengroup.org/onlinepubs/9799919799/functions/tcgetwinsize.html#]/()[https://pubs.opengroup.org/onlinepubs/9799919799/functions/tcgetwinsize.html#]); if the terminal window size cannot be obtained, the implementation determines the number of columns, appropriate for the terminal or window, in an unspecified manner. When /COLUMNS/ is set, the number of columns in the terminal window size and any terminal-width information implied by /TERM/ are overridden. Users and conforming applications should not set /COLUMNS/ unless they wish to override the system selection and produce output unrelated to the terminal characteristics.
> Users should not need to set this variable in the environment unless there is a specific reason to override the implementation's default behavior, such as to display data in an area arbitrarily smaller than the terminal or window.

However, it is my understanding that in practice, it is managed by the user's shell, which can only set it when it has control - in particular, this is only really useful in interactive sessions, or (for bash >4.3) in scripts with `shopt checkwinsize`. One testcase that should illustrate my understanding is to try polling `COLUMNS` in a long-running program without returning control to bash - it shouldn't update, even if the terminal window is resized in the interim.
Instead, my understanding is that compliant programs should listen for SIGWINCH and resize accordingly. And if one is using those interfaces, it makes more sense to be querying terminfo or ioctl.

27 feb 2025 21:23:05 Simon Michael ***@***.***>:

hseg:
Not sure why the CI is failing, will need to check another day.

27 feb 2025 22:22:19 Gesh ***@***.***>:

hseg:
Ah, you run with -Werror on everything. Fair enough, could've cleaned up the imports. It would've been useful to see the full run had -Werror not killed the run - perhaps -Wwarn combined with checking whether any warning was emitted might be preferable?

27 feb 2025 22:23:16 Gesh ***@***.***>:

thomie:
Commit 372a2d768b is relevant: 

> 2. or a COLUMNS environment variable (NB: not the same as a bash shell var)

The manual also mentions `COLUMNS` as an environment variable.

simonmichael:
@hseg do you not build locally ? That's where I (usually) see warnings before pushing.

simonmichael:
(Actually, ideally I see them in the VS code errors pane as I work.)

simonmichael:
FYI if/when you do set up a local environment, it'll be easy to see the warnings, no need to do anything special. Eg `just ghcid`, `stack build`, or (I think) `cabal build`) should show them.

hseg:
I would've seen the warnings, but I wouldn't've realized they're fatal for the CI run

28 feb 2025 01:02:09 Simon Michael ***@***.***>:

simonmichael:
Now you know. :) I wish every haskeller knew this, in fact spewing warnings is the norm for much of Hackage. :( Yes we could do more in the hledger project to help contributors avoid losing time this way, thanks for the feedback.

hledger-lib/Hledger/Utils/IO.hs
```diff
@@ -413,8 +413,30 @@ hasOutputFile = do
...
+-- the correct entries. (A hack that sometimes works but which isn't robust
+-- enough to be relied upon is to set TERM=dumb -- while this advice does appear
+-- in some places, it's not guaranteed to work)
+--
+-- In any case, $LINES/$COLUMNS should not be used as a source for the terminal
+-- size -- that is a bashism, and in particular they aren't updated as the
+-- terminal is resized.
+--
+-- See #2332 for details
+
```
simonmichael, line 436: Great notes!

hledger/hledger.cabal
```diff
@@ -98,11 +98,6 @@ flag debug
   manual: True
   default: False
 
-flag terminfo
-  description: On POSIX systems, build with the terminfo lib for detecting terminal width
-  manual: False
-  default: True
-
```
simonmichael, line 105: Removing a build flag/variant is always good! I had the impression terminfo could be a useful lib though, is there nothing in there we need ? I guess we have replaced it by other things.

simonmichael (approved):
This looks great. I can report that hledger-2340 in my osh sessions, in terminal windows or emacs shell buffers, now detects resizes and uses the full width where it didn't before. An immediate quality of life improvement for me.

simonmichael:
Merged, plus a few tweaks, as d346e56b7. Thank you!

hseg:
Thanks! @thomie is correct to point out I missed removing references to COLUMNS in files that aren't *.hs - in particular the manuals.

28 feb 2025 01:47:09 Simon Michael ***@***.***>:

simonmichael:
Thanks for the reminder, I'll remove those.

simonmichael:
Done. The functional tests previously required COLUMNS for reproducibility, but they seem to magically still work now, I guess they are nicely defaulting to 80 columns.

hseg:
Hm. It'd be nice to be able to be more confident that that was what was happening - the default width is indeed 80. Perhaps removing COLUMNS was an overzealous mistake? Do we have tests exercising whether a different terminal size (eg set by running resize before the running the test) is picked up?

28 feb 2025 03:30:02 Simon Michael ***@***.***>:

simonmichael:
I just ran `just functest` a few times at different window sizes, with the COLUMNS=80 removed, which in the past would have failed. I'm singing lalala and assuming that hledger when run as a subprocess by shelltestrunner is failing to detect terminal size and falling back on 80.

hseg:
Rebased the branch to be on top of the commit master was on when you merged it
(this was probably unnecessary, but it lets me avoid warnings when deleting the
branch). If you wanted me to rebase before merging, you could've said so.

hseg:
OK, I've learned two things:

- `stty cols NCOLUMNS` will cause the terminal to pretend it's `NCOLUMNS` columns wide (eg `top` is sensitive to this) -- perhaps useful for testing this automatically?
- `Hledger.Cli.CliOptions.getHledgerCliOpts'` isn't used by `main`, and I'm getting a little lost tracking its usages down. Can't figure out where `main` is getting its arguments from. Probably need to install HLS to navigate the codebase better, but that's for another time, then.

PR #2341, opened 2025-03-02
Hledger.Cli.CliOptions: Remove commented-out, unused functions

hseg:
While looking at `Hledger.Cli.CliOptions`, I noticed a couple of bits of dead
code which are duplicated elsewhere. I propose we remove them

- `debugArgs`:
  A variant of it is present as a local definition in `getHledgerCliOpts'`

- `widthFromOpts`:
  As noted in 5c289ac, it is unused elsewhere, and `registerWidthsFromOpts` is right next to it and implements nearly the exact same logic


hseg:
Rebased onto master

simonmichael:
I guess I was thinking non-register commands might need to use width at some point. But YAGNI. Thanks for the cleanup!

PR #2347, opened 2025-03-08
Fix #2345: do not pre-load journal(s) at the start of run/repl, and honor reporting interval changes for --forecast

### LINKED ISSUE #2345: run does not correctly handle forecast with different report dates.
`test.j`
```
2025-01-01  Buy beans
    expenses:food   $2
    assets         -$2

~ every week from 2025-01-01  Buy soylent green
    expenses:food:not people   $1
    assets                    -$1
```

`cmdlist.txt`
```
register --forecast -e 2025-02
register --forecast -e 2025-03
```

`hledger -f test.j run cmdlist.txt` will produce the same output for both commands. This is because it is using the cache, which has already generated forecast transactions, for the second call. The cache of `InputOpts` will need to remember the report span in the cases that forecast transactions are being generated.

```
$ hledger -f test.j run cmdlist.txt 2>&1 | grep cach
readAndCacheJournalFile reading and caching /home/username/test.j:
readAndCacheJournalFile reading and caching /home/username/test.j:
readAndCacheJournalFile using cache for /home/username/test.j:
```

I'll also note that it's attempting to read/run the cache three times, which is more than necessary given there are only two commands. Maybe this can be improved.

adept:
This fixes #2345 

Previosly: both "run" and "repl" did `withJournalCached` at the start to pre-load journal. The returned journal would be dropped on the floor, but when the command will ask for the same journal, it would already be in the cache. What's bad: there is no guarantee that _any_ command will actually ask for this same journal. We are not winning anything by preload, this was just side-effect of multiple refactors.

Now: no journals are preloaded, any reading/parsing will only occur when commands ask for it

Aslo previosly: InputOpts is used in the cache key, and I would blank out `reportspan_` there, thinking that as it is derived from (forecast_ :: Maybe DateSpan) which is also in InputOpts, it is sufficient to rely on `forecast_` changing. What's bad: actually, we are interested in distinguishing inputs with `--forecast` set and different reporting intervals

Now: `reportspan_` is only blanked out when `--forecast` is not set.

I've added one more test. 

Tests demonstrate both of these changes.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Run.hs
+++ b/hledger/Hledger/Cli/Commands/Run.hs
@@ -81,28 +81,28 @@ newtype DefaultRunJournal = DefaultRunJournal (NE.NonEmpty String) deriving (Sho
 -- | The actual run command.
 run :: Maybe DefaultRunJournal -> (String -> Maybe (Mode RawOpts, CliOpts -> Journal -> IO ())) -> [String] -> CliOpts -> IO ()
 run defaultJournalOverride findBuiltinCommand addons cliopts@CliOpts{rawopts_=rawopts} = do
-  withJournalCached defaultJournalOverride cliopts $ \(_,key) -> do
-    let args = dbg1 "args" $ listofstringopt "args" rawopts
-    isTerminal <- isStdinTerminal
-    if args == [] && not isTerminal
-      then do
-        inputFiles <- journalFilePathFromOpts cliopts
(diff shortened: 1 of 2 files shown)


simonmichael:
Just wondering, if a --forecast=FORECASTPERIOD is specified, is it handled right ?

adept:
Yep, it is :)

simonmichael:
Thanks!

PR #2348, opened 2025-03-08
Add script : balance-as-budget-multi.hs

adept:
Run-inspired variant of balance-as-budget. Load two journal files, and run balance-as-budget many times.

I use this to generate "generate a year summary that compares this year to the previous". Might be useful to others.


simonmichael:
Let's add an entry to bin/README.md.
We haven't before, but feel free to, add a small example of output there.
How does this compare to balance-as-budget, is that one still useful ?

adept:
Added it to readme.

Both are still useful, I think.

`hledger-balance-as-budge-multi` is like `hledger run` + `balance-as-budget` - you load journal once, and then run `balance-vs-balance` as many times as you want, but the way it is written it requires all inputs to be in files, which is less convenient that `balance-as-budget`, which takes them from the command line

PR #2349, opened 2025-03-09
imp run: More successful cache hits with forecast

Xitian9:
If forecast start and end dates are explicitly requested with --forecast=, those parts of the reportspan_ can be replaced with the forecast dates, meaning more successful cache hits.


adept:
Very nice catch, thank you! I dont know if this is worthy of a separate test or not

PR #2353, opened 2025-03-12
fix: csv: fix regression in parsing rules containing & (#2352)

### LINKED ISSUE #2352: 1.42: & in CSV rules no longer works, unescapable
`&` in the regex expression of an if block raises error on CSV import unless the `&` is followed by a conditional block. Escaping with backslashes or enclosing `&` in a regex group expression (e.g., `[ \&]*`) has no effect.

**Example CSV rule file**

```
if
%description STOP \& SHOP
    account2 Expenses:Food and Supplies
```

**Error Output**

```
hledger: Error: ...chase-visa.rules:137:21:
    |
137 | %description STOP \& SHOP
    |                     ^
unexpected space
expecting conditional block
```
I'm using hledger-flow import and hledger 1.42, mac-aarch64. I did not have this error prior to 1.42.

thomie:
Fixes #2352.

### DIFF
--- a/hledger-lib/Hledger/Read/RulesReader.hs
+++ b/hledger-lib/Hledger/Read/RulesReader.hs
@@ -723,7 +723,7 @@ regexp end = do
        Left x -> Fail.fail $ "CSV parser: " ++ x
        Right x -> return x
   where
-    double_ampersand = lookAhead . void $ char '&' >> char '&'
+    double_ampersand = lookAhead . void $ string "&&"
 
 -- -- A match operator, indicating the type of match to perform.
 -- -- Currently just ~ meaning case insensitive infix regex match.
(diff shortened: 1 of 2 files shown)


hledger-lib/Hledger/Read/RulesReader.hs
```diff
@@ -1596,9 +1596,14 @@ tests_RulesReader = testGroup "RulesReader" [
...
 
+
   ,testGroup "regexp" [
     testCase "regexp.ends-before-&&" $
-      parseWithState' defrules (regexp empty) "A A && xxx" @?= (Right $ toRegexCI' "A A")
+      parseWithState' defrules (regexp eof) "A A && xxx" @?= (Right $ toRegexCI' "A A")
+   ,testCase "regexp contains &" $
+      parseWithState' defrules (regexp eof) "A & B" @?= (Right $ toRegexCI' "A & B")
+   ,testCase "regexp contains escaped &" $
+      parseWithState' defrules (regexp eof) "A \\& B" @?= (Right $ toRegexCI' "A \\& B")
```
simonmichael, line 1606: I don't think we support any backslash escaping here, so no need to test for it here or in csv.test.

simonmichael:
Thanks!

hledger-lib/Hledger/Read/RulesReader.hs
thomie: Do you want to support escaping double ampersands?

2025-01-01,COFFEE && CODE,100
2025-01-02,COFFEE CODE,200

if
%description COFFEE \&\& CODE
    account2 expenses:bars

That's currently working: the rule only gets applied to the first line. But it's not documented nor tested.

simonmichael: I'm releasing right now and we didn't discuss it, so I'll pass. That could use a bit more thinking about.

Issue #2355, opened 2025-03-14
Allow balance assertions in `hledger add`


voidus:
It would be nice to allow assertions in when adding transactions through the command line.

## Motivation

I have a few accounts that I need to zero every month, and adding the assertion mitigates the risk of typos or other inconsistencies creeping in.

## Proposal
Currently, this is not possible:

```
Amount  1 [10,90€]: -15€ = 0
A valid hledger amount is required. Eg: 1, $2, 3 EUR, "4 red apples".
```

Assuming the balance assertion holds, the flow should continue just as it would without the assertion.
If it doesn't, I think we should alert the user quickly:

```
Amount  1 [10,90€]: -15€ = 0
Balance assertion failed (calculated balance is 1,00€, difference 1,00€)
Amount 1 [10,90€]: -15€ = 0
```
I'd pre-fill the previous text for convenience

## Out of scope

I don't think it is necessary to allow an option to force it, if a user wants that there are other ways.

## Effort

Without knowing the codebase, this feels like it shouldn't be too invasive, but I won't pretend to actually know.

## Impact

I think users who aren't interested in this wouldn't notice anything at all. If I really try to imagine a workflow this would break, I can only see someone relying on the fact that this errors now, which seems like a stretch to me.

reesmichael1:
Personally, I think this makes sense to support. I'd be happy to look at implementing this.

voidus:
I'd also be interested in taking a look, but I can't promise when I'll find time, so don't let that stop you if you're motivated :)

voidus:
Hm. Seems like this isn't quite as straight-forward as I imagined.

I do think that we should validate after every posting. If I make a mistake, I'd want to know early rather than having to go back later.

I guess a transaction can include multiple postings for the same account, even though I'm not sure this is ever a reasonable thing to do, and I'm also not sure how the assertions work in that case.
Personally, I'd assume it to take the journal and all the previously entered postings into account.

---

I just tried some things with multiple transactions for the same account, that's the only thing I can think of where it get's complicated. It seems like the assertion is checked with the previous postings taken into account but not the later ones, i.e. this checks:

```
2025-03-14 asdf
    Aktiva:Bank         -1234€ = 0€ ; this holds with this posting but not at the end of the transaction
    Aktiva:Bank                               20€
    Aktiva:Foo
```

while this reports a failed assertion:

```
2025-03-14 asdf
    Aktiva:Bank                               20€
[...]

PR #2356, opened 2025-03-19
imp: add: Verify balance assertions on each posting (#2355)

reesmichael1:
As requested in #2355, this verifies balance assertions during `hledger add`. Here's an example of the new workflow. With the following journal file:

```
2025-03-18 * Deposit
    Assets:Checking                     $20
    Income:Salary                      $-20
```

```
$ hledger add
Date [2025-03-19]:
Description: Food
Account 1: Expenses:Food
Amount  1: $15
Account 2: Assets:Checking
Amount  2 [$-15]: $-15 == $10
1:6:
  | 2025-03-19 Food
  |     Expenses:Food               $15
3 |     Assets:Checking            $-15 == $10
  |                                     ^^^^^^
  |                                  $0

Balance assertion failed in Assets:Checking
Across all commodities at this point, excluding subaccounts, ignoring costs,
the asserted balance is:        $10
but the calculated balance is:   $5
(difference: $5)
To troubleshoot, check this account's running balance with assertions disabled, eg:
hledger reg -I 'Assets:Checking$'
Amount  2 [$-15]: $-15  == $5
Account 3 (or . or enter to finish this transaction):
2025-03-19 Food
    Expenses:Food               $15
    Assets:Checking            $-15 == $5

Save this transaction to the journal ? [y]: y
Saved.
```

The only thing that might be odd here is including the "To troubleshoot..." lines. I left those in because it's not bad advice for tracking down why an assertion might be failing, but also it isn't really helpful and might be misleading in the middle of an `hledger add`, so I'm open to removing those lines here.

### DIFF
--- a/hledger-lib/Hledger/Data/Balancing.hs
+++ b/hledger-lib/Hledger/Data/Balancing.hs
@@ -18,6 +18,8 @@ module Hledger.Data.Balancing
 , isTransactionBalanced
 , balanceTransaction
 , balanceTransactionHelper
+  -- * assertion validation
+, checkAssertions
   -- * journal balancing
 , journalBalanceTransactions
   -- * tests
@@ -146,6 +148,16 @@ transactionCheckBalanced BalancingOpts{commodity_styles_} t = errs
 -- | Legacy form of transactionCheckBalanced.
 isTransactionBalanced :: BalancingOpts -> Transaction -> Bool
 isTransactionBalanced bopts = null . transactionCheckBalanced bopts
+
+-- | Verify that any assertions in this transaction hold 
+-- when included in the larger journal.
+checkAssertions :: BalancingOpts -> Journal -> Transaction -> Either String Transaction
+checkAssertions bopts j t =
+  if (ignore_assertions_ bopts) || noassertions t then Right t else do
+    j' <- journalStyleAmounts j 
+    fmap (\_ -> t) $ journalBalanceTransactions defbalancingopts j'{jtxns = (t : (jtxns j')) }
+  where
+    noassertions = all (isNothing . pbalanceassertion) . tpostings
 
 -- | Balance this transaction, ensuring that its postings
 -- (and its balanced virtual postings) sum to 0,
--- a/hledger/Hledger/Cli/Commands/Add.hs
+++ b/hledger/Hledger/Cli/Commands/Add.hs
@@ -231,16 +231,27 @@ confirmedTransactionWizard prevInput es@EntryState{..} stack@(currentStage : _) 
       confirmedTransactionWizard prevInput es{esPostings=init esPostings} (dropWhile notPrevAmountAndNotEnterDesc stack)
 
   EnterAmountAndComment txnParams account -> amountAndCommentWizard prevInput es >>= \case
-    Just (amt, comment) -> do
+    Just (amt, assertion, comment) -> do
       let p = nullposting{paccount=T.pack $ stripbrackets account
                           ,pamount=mixedAmount amt
                           ,pcomment=comment
                           ,ptype=accountNamePostingType $ T.pack account
+                          ,pbalanceassertion = assertion
                           }
           amountAndCommentString = showAmount amt ++ T.unpack (if T.null comment then "" else "  ;" <> comment)
           prevAmountAndCmnt' = replaceNthOrAppend (length esPostings) amountAndCommentString (prevAmountAndCmnt prevInput)
           es' = es{esPostings=esPostings++[p], esArgs=drop 1 esArgs}
-      confirmedTransactionWizard prevInput{prevAmountAndCmnt=prevAmountAndCmnt'} es' (EnterNewPosting txnParams (Just posting) : stack)
+          -- Include a dummy posting to balance the unfinished transation in assertion checking
+          dummytxn = nulltransaction{tpostings = esPostings ++ [p, post "" missingamt]
+                                     ,tdate = txnDate txnParams
+                                     ,tdescription = txnDesc txnParams }
+          validated = balanceTransaction defbalancingopts dummytxn >>= checkAssertions defbalancingopts esJournal
+      case validated of
+        Left err -> do
+          liftIO (hPutStrLn stderr err)
+          confirmedTransactionWizard prevInput es (EnterAmountAndComment txnParams account : stack)
+        Right _ -> 
+          confirmedTransactionWizard prevInput{prevAmountAndCmnt=prevAmountAndCmnt'} es' (EnterNewPosting txnParams (Just posting) : stack)
     Nothing -> confirmedTransactionWizard prevInput es (drop 1 stack)
 
   EndStage t -> do
@@ -352,13 +363,14 @@ amountAndCommentWizard PrevInput{..} EntryState{..} = do
                                   ""
                                   (T.pack s)
       nodefcommodityj = esJournal{jparsedefaultcommodity=Nothing}
-      amountandcommentp :: JournalParser Identity (Amount, Text)
+      amountandcommentp :: JournalParser Identity (Amount, Maybe BalanceAssertion, Text)
       amountandcommentp = do
         a <- amountp
         lift skipNonNewlineSpaces
+        assertion <- optional balanceassertionp
         c <- T.pack <$> fromMaybe "" `fmap` optional (char ';' >> many anySingle)
         -- eof
-        return (a,c)
+        return (a, assertion, c)
       balancingamt = maNegate . sumPostings $ filter isReal esPostings
       balancingamtfirstcommodity = mixed . take 1 $ amounts balancingamt
       showamt = wbUnpack . showMixedAmountB defaultFmt . mixedAmountSetPrecision


simonmichael:
Thanks for this!

We should try to explore document how this differs from the behaviour of balance assertions when reading a journal. I think it must be at least a little different, if immediate feedback is desirable (and I agree that it is).

Here are some things to check:

- assertion types: do =, ==, =*, ==* all work ?
- balance assignments: what happens if someone tries to write balance assignment (omitting the posting amount) ?
- multiple assertions in one transaction
- multiple assertions to the same account within one transaction
- posting dates (I think hledger add supports those), which could allow entering multiple assertions to the same account in non date order

reesmichael1:
Sorry for letting this languish for a bit! Some other projects have consumed me. I'm happy to review/document all of this, and will try to do it later this week or early next week.

reesmichael1:
(Let's ignore that it's taken me three weeks to resurface....)

Doing all of this definitely revealed some bugs! I'll explore fixing them, just wanted to get them documented first:

* `=` and `=*` work as expected unless the transaction being entered is on the same day as other transactions with the same account 
* `==` and `==*` also don't work when there are multiple commodities on the right hand side of the assertion (they're not recognized as a valid hledger amount)
* Multiple assertions seem to be fine
* I'll explore posting dates more once the date issues mentioned above are resolved

simonmichael:
Thanks for the update -

> * `==` and `==*` also don't work when there are multiple commodities on the right hand side of the assertion (they're not recognized as a valid hledger amount)

I think we can usually ignore that case - there's no way to write it in journal format. (Explicitly written posting amounts are always single-commodity.)

Re posting dates etc, we like to be flexible but we can impose restrictions for sanity, if needed.

simonmichael:
I see. I think the main thing we'd want is that error reporting and balance calculations are the same both inside and outside an `add` session. 

Until now posting dates didn't affect an `add` session, but I think now they'd need to, to satisfy the above.

As a simplifying limitation, we could disallow combining a posting date with a balance assertion/assignment when using `add`. (On a single posting, or even within the whole transaction if needed.)

reesmichael1:
Sorry, forgot to run the complete test suite instead of my subset. I'll see why this one fails!

hledger-lib/Hledger/Data/Balancing.hs
```diff
@@ -146,6 +148,17 @@ transactionCheckBalanced BalancingOpts{commodity_styles_} t = errs
...
 isTransactionBalanced bopts = null . transactionCheckBalanced bopts
 
+-- | Verify that any assertions in this transaction hold 
+-- when included in the larger journal.
+checkAssertions :: BalancingOpts -> Journal -> Transaction -> Either String Transaction
+checkAssertions bopts j t =
+  if (ignore_assertions_ bopts) || noassertions t then Right t else do
+    j' <- journalStyleAmounts j 
+    let newtxns = sortOn tdate (jtxns j' ++ [ t ])
+    fmap (\_ -> t) $ journalBalanceTransactions defbalancingopts j'{jtxns = newtxns}
```
simonmichael, line 158: I guess that's a shorthand for
case ... of
  Right _ -> Right t
  Left e -> Left e
?

hledger-lib/Hledger/Data/Balancing.hs
```diff
@@ -146,6 +148,17 @@ transactionCheckBalanced BalancingOpts{commodity_styles_} t = errs
 isTransactionBalanced :: BalancingOpts -> Transaction -> Bool
 isTransactionBalanced bopts = null . transactionCheckBalanced bopts
 
+-- | Verify that any assertions in this transaction hold 
+-- when included in the larger journal.
+checkAssertions :: BalancingOpts -> Journal -> Transaction -> Either String Transaction
+checkAssertions bopts j t =
```
simonmichael, line 154: Worth renaming to transactionCheckAssertions ?

hledger-lib/Hledger/Data/Balancing.hs
```diff
@@ -146,6 +148,17 @@ transactionCheckBalanced BalancingOpts{commodity_styles_} t = errs
 isTransactionBalanced :: BalancingOpts -> Transaction -> Bool
 isTransactionBalanced bopts = null . transactionCheckBalanced bopts
 
+-- | Verify that any assertions in this transaction hold 
+-- when included in the larger journal.
```
simonmichael, line 152: I imagine this could be slow if called for a lot of transactions, but we intend it to be called only for one, the one user has just entered during an `add` session. So, minimal user-visible delay even with a big journal and slow machine, hopefully..

hledger/Hledger/Cli/Commands/Add.hs
```diff
@@ -231,16 +231,33 @@ confirmedTransactionWizard prevInput es@EntryState{..} stack@(currentStage : _)
       confirmedTransactionWizard prevInput es{esPostings=init esPostings} (dropWhile notPrevAmountAndNotEnterDesc stack)
 
   EnterAmountAndComment txnParams account -> amountAndCommentWizard prevInput es >>= \case
-    Just (amt, comment) -> do
+    Just (amt, assertion, (comment, tags, pdate1, pdate2)) -> do
+      -- This check is necessary because we cons a ';' in the comment parser above,
+      -- and we don't want to add an empty comment here if it wasn't given.
+      let pcomment = if T.length comment == 1 then "" else comment
```
simonmichael, line 237: This seems a bit fragile, would a check for comment == ";" be clearer and safer ?
Also this seems to prevent users adding such an empty comment if they wanted to. Not the end of the world.

hledger/test/add.test
```diff
@@ -114,7 +114,96 @@ $ rm -f nosuch.journal; hledger -f nosuch.journal add; rm -f nosuch.journal
...
+a
+-500 JPY ==* 500 JPY
+c
+-500 JPY
+.
+$ rm -f nosuch.journal; hledger -f nosuch.journal add; rm -f nosuch.journal
+> /Save this transaction to the journal/
+>2 //
+
+## 13. Assertions with posting dates
```
simonmichael, line 165: Should we also test add-ing an amount with an assertion and a posting date ? I'm not sure if that's supported.

Also what about balance assignments, are they expected to work or be rejected ?

simonmichael (commented):
Looking good! Comments attached.

simonmichael (commented):
Looking good! Comments attached.

hledger/test/add.test
reesmichael1: Good catch! It looks like it's not possible to add an assertion and a posting date such that the assertion would be valid on the day of the posting date. However, this is also not valid in a journal file:
```test.hledger
2025-05-01 x
    a     50 USD
    b    -50 USD

2025-05-02 x2
    a    100 USD  == 50 USD  ; date:2025-05-05
    c   -100 USD

2025-05-10 x3
    a     50 USD  == 200 USD
    d    -50 USD
```

```
$ hledger print -f test.hledger
hledger: Error: /home/michael/builds/hledger/test.hledger:6:19:
  | 2025-05-02 x2
6 |     a         100 USD == 50 USD  ; date:2025-05-05
  |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |     c        -100 USD

This balance assertion failed.
In account:    a
and commodity: USD                             (no other commodities allowed)
this balance was asserted:     50
but the calculated balance is: 150
```

This might be possible to support eventually, but since this is already the existing behavior, I think that's better for a follow-up PR.

Balance assignments aren't supported in stock `hledger`:
```
$ hledger add
Date [2025-05-28]:
Description: Opening Balances
Account 1: Assets:Checking
Amount  1:   = $409.32
A valid hledger amount is required. Eg: 1, $2, 3 EUR, "4 red apples".
Amount  1:
```

If we're interested, I could look at adding those next.

hledger/Hledger/Cli/Commands/Add.hs
reesmichael1: After looking at this more, I was actually mistaken and it's a newline character, not the single `;`, that's coming out of the comment parser. Good suggestion though--I'll change it to `comment == "\n"` for clarity, and I'll edit the (Haskell) comment to match. 

And perhaps suprisingly, stock `hledger add` already doesn't support adding an empty comment.

reesmichael1: Actually, after some more investigation, it turns out that using `postingcommentp` removes all space from the given comment. I figured this out because I thought about using `strip` to remove the newline and this check, decided against it because we'd mess with the user's spacing, tested, and found out it was happening anyway!

We do seem to already strip the comment in the description, but I don't want to introduce that behavior here unnecessarily, so instead I'm switching to using the old comment string given by `amountandcommentp` for `pcomment` while still using the other values from `postingcommentp` for tags and both date fields. This (at least in terms of getting the user's comment string into the file) should match the current behavior exactly.

hledger-lib/Hledger/Data/Balancing.hs
simonmichael: >  this will also check that any future assertions still hold as a result of the transaction

Aiee! Good point, I hadn't thought of that. We should probably mention in the manual, and I should mention in release notes: entering an amount during `add` can now fail if it would break any balance assertion in the journal. Also it can fail if there was a pre-existing failing assertion. Though in that case, `add` probably wouldn't run - unless you run it with -I. We should probably make sure that the new assertion check also respects -I.

hledger/test/add.test
simonmichael: > However, this is also not valid in a journal file:

I wouldn't expect that journal file to pass. But this should be accepted, shouldn't it ?

```
$ hledger -f new.j add
Date [2025-05-28]: 
Description: 
Account 1: (a)
Amount  1: 10
Account 2 (or . or enter to finish this transaction): 
2025-05-28
    (a)              10

Save this transaction to the journal ? [y]: 
Saved.
Starting the next transaction (. or ctrl-D/ctrl-C to quit)
Date [2025-05-28]: 
Description: 
Account 1 [(a)]: 
Amount  1 [10]: 3 = 3  ; date:5/1
```

simonmichael: (Equivalent to this journal:)
```
2025-05-28
    (a)              10

2025-05-28
    (a)               3 = 3  ; date:5/1
```

simonmichael: > Balance assignments aren't supported in stock hledger:

Right you are, current hledger doesn't allow add-ing balance assertions so of course it doesn't allow add-ing balance assignments either. 

Once it allows add-ing assertions, people might expect assignments too, let's mention in the manual that add-ing those isn't supported yet.

hledger-lib/Hledger/Data/Balancing.hs
reesmichael1: > We should probably make sure that the new assertion check also respects `-I`.

I thought that [this check](https://github.com/simonmichael/hledger/blob/45416fbd9e778594e940332bb9e3ca72cb4dd9db/hledger-lib/Hledger/Data/Balancing.hs#L155) would skip any assertions, but a quick test seems to indicate otherwise! I'll check this out as well. I'm happy to update the manual as well, and good call on flagging this in the release notes.

It also appears that I might have been ahead of myself and *already* implemented the "only check if there's an assertion in the transaction" restriction--let me think a bit more about the ramifications of that.

reesmichael1: > I thought that [this check](https://github.com/simonmichael/hledger/blob/45416fbd9e778594e940332bb9e3ca72cb4dd9db/hledger-lib/Hledger/Data/Balancing.hs#L155) would skip any assertions [...]

It would help to not just use `defbalancingopts` everywhere :man_facepalming:

reesmichael1:
Hmmm, not really sure why that test failed? All of them pass on my machine :tm:...

simonmichael:
The failing test was a real puzzle. Finally it just worked for no apparent reason. I re-ran the PR test job and it too passed. My best guess is that caching went wrong somehow and it wasn't running a freshly built hledger.

Now merged. Thanks for seeing this through!

simonmichael:
Somehow I missed a failure due to incorrect `\$` escaping in a test. Fixed by avoiding $ there.

PR #2357, opened 2025-03-23
Make project compile again with ghc-8.10

thomie:
I found I needed to make these changes to make the project compile again with ghc-8.10.7.

I ran into the following doctest failure:
```
Hledger/Data/Posting.hs:119:57: error:
    parse error on input ‘-- ^ transactions generated by a periodic txn rule’
    |
119 | generatedTransactionTagName = "_generated-transaction"  -- ^ transactions generated by a periodic txn rule
    |                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

Tested using  `stack8.10.yaml`, following the same steps as in `.github/workflows/oldest.yml`.


simonmichael:
Thanks for these fixes. I thought we had dropped 8.10.7 support for at least one of the packages, but the base bounds say otherwise. I'll merge them for now, and I'll update tested-with in all cabal files. 

The haddock change for fixing 8.10.7 doctest seems like a backward step and I might revert it. We should re-ask if it's time to drop 8.10.7 support. I like to remove as many barriers as possible for hledger installing and packaging, but there is a cost of carrying support for old GHCs.

Issue #2044, opened 2023-06-03
Provide Nix Flake


alerque:
It would be quite useful for some situations (especially CI runners and comparing differences between processing a project an Different hledger versions) if there was a Flake provided in this repository with a lock file updated and tested to be working on each release. This would enable running any arbitrary version of hledger (at least ones post-dating introduction of the Flake!) with a simple command like `nix run gihub:simonmichael/hledger`.

simonmichael:
We need a resident Nix packager - interested ?

alerque:
I'd help, but I not familiar with how Nix handles Haskell in general. I kind of have a handle on it's Autotools and Lua stuff and am getting acquainted with Rust packaging, but Haskell is not my forte.

As a side note the nixpkgs packaging for hledger is pretty out of date too.

chvp:
I can look into it

simonmichael:
Thanks! We probably need to update or drop the Nix install method at https://hledger.org/install.html#other too.

alerque:
I'd say they need updating not dropping. I don't know what the use of the archive hard coded version is about there, but before I even opened this issue I was playing around with it and an invocation as simple as `nix run nixpkgs#hledger` worked just fine. I assume that archive much be at attempt to get a newer version that isn't released yet, but the best think to do is probably to chase down why it isn't updated in the first place and just recommend the normal ecosystem commands: `nix-env --install hledger` or something basic like that if what you wanted was to suggest the way to drop into a shell with `hledger available`. The archive version being referenced isn't fresh anyway.

alerque:
From reviewing [nixpkgs issues](https://github.com/NixOS/nixpkgs/issues?q=is%3Aopen+sort%3Aupdated-desc+hledger) and asking around Arch packagers why `hledger` is out of date there too it seems the answer is everybody is having trouble dealing with GHC/Stack updates: either they can't update the compilers because some things don't compile or have updated them and some things don't compile.

Specifically for Arch, `hledger-iadd` not building on GHC 9 appears to be the hold up for also updating the Hledger package. That's also what [inspired this PR](https://github.com/hpdeifel/hledger-iadd/pull/74).

For Nix I'm not so sure on details of what is holding up what, but

chvp:
The haskell packages in nixpkgs follow stackage LTS releases, AFAIK.

simonmichael:
[...]

PR #2358, opened 2025-03-23
Drop base compat dependency

thomie:
The `base-compat` dependency was really only used in one module in `hledger-web`.

By replacing a call to `NonEmpty.tail` with `Safe.tailSafe`, we can drop this dependency.

Tested with GHC 8.10.7 (oldest, requires #2357).

It's particularly nice to drop this dependency, because we required base-compat >= 0.14 before (which introduced `NonEmpty.prependList`), while stackage LTS still has 0.13.1. So it should help with fixing the stack9.*.yaml compiles, and it also helps nix users (#2044) who are on stackage LTS.


simonmichael:
base-compat and other -compat packages have been useful in the past to paper over API changes without needing CPP, but if not needed that's even better. Thanks!

PR #2360, opened 2025-03-26
Introduce multi-period Account data type and use it for MultiBalanceReport and BudgetReport.

Xitian9:
This rejigs the `MultiBalanceReport` internals to use an enhanced `Account` data type to save the values. This has a few effects:

- A small speed improvement on large journal files with interval reporting, due to processing the posting list in one pass.
- Simplified program logic, as there was a lot of code to convert back and forth between list and tree representations. This can now be removed.
- Ability to merge `Account` means that `BudgetReport` can be simplified.

There are some small changes in behaviour with respect to budget reports, where it looked like some behaviour was implemented to work around needing to get the budget and actuals into the same shape so they could be merged. This is no longer necessary, but may still be desired for other reasons.

Let me know your thoughts.


hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
...
+applyAccountBalance f a = a{abebalance = f $ abebalance a, abibalance = f $ abibalance a}
+
+-- | Perform an operation on the 'MixedAmount' in two 'AccountBalance'.
+opAccountBalance :: (MixedAmount -> MixedAmount -> MixedAmount) -> AccountBalance -> AccountBalance -> AccountBalance
+opAccountBalance f a b = a{abebalance = f (abebalance a) (abebalance b), abibalance = f (abibalance a) (abibalance b)}
+
+
+-- tests
+
+tests_AccountBalance = testGroup "AccountBalance" [
```
simonmichael, line 134: Keep these separate for clarity if you like, but I suggest grouping them under tests_AccountBalance for export to avoid breaking the consistent pattern (every module exports one similarly named test group)

hledger-ui/hledger-ui.cabal
```diff
@@ -9,7 +9,6 @@ version:        1.42.99
 synopsis:       Terminal interface for the hledger accounting system
 description:    A simple terminal user interface for the hledger accounting system.
                 It can be a more convenient way to browse your accounts than the CLI.
-                This package currently does not support Microsoft Windows, except in WSL.
                 .
```
simonmichael, line 12: This change is in master and will disappear after rebase I think.

hledger/test/balance/balance.test
```diff
@@ -172,27 +171,17 @@ $ hledger -f - balance -NE --declared --tree --depth 2 a
...
+                   0    ac
 
-# ** 11. In list mode we can see that non-leaf declared accounts are excluded.
-<
-account a
-account a:aa
-account a:ab
-account a:ac:aca
-account b
+# ** 11. In flat mode we can see that undeclared accounts are excluded.
```
simonmichael, line 176: Text trivia:

I have been trying to move in the direction of "list" actually as the preferred term actually (for the -l short flag, eg. Possibly should go all the way and add --list long flag ?)

But your change matches the flag used in the command, so maybe it's better..

I feel maybe the old test description here was bad.. I think --declared includes declared accounts, rather than excluding undeclared ones. Perhaps we should drop that text.

simonmichael: I'm ok with the behaviour change here personally. I'd put a ! (= "breaking change") after the first commit message prefix which helps me notice these at changelog/relnotes time. 

ref: is a new prefix meaning "refactoring" I guess. Ok, but perhaps it isn't the right one here since this is not strictly a refactoring. In such cases I often go with a generic dev:.

hledger/test/balance/balance.test
```diff
@@ -217,3 +206,19 @@ $ hledger -f- bal --count -1
                    1  b
 --------------------
                    3  
+
+# ** 16. balance --flat --empty does not display accounts which have not been
+# seen, even if they're implied, but does show accounts that have been seen
+# with 0 balance.
```
simonmichael, line 212: I'm not sure what this means in userese ?

hledger/test/balance/budget.test
```diff
@@ -50,7 +50,6 @@ Budget performance in 2016-12-01..2016-12-03:
                   ||          2016-12-01          2016-12-02           2016-12-03 
 ==================++==============================================================
  assets:cash      || $-10 [ 40% of $-25]  $-14 [56% of $-25]  $-51 [204% of $-25] 
- expenses         ||  $10 [ 40% of  $25]   $14 [56% of  $25]   $51 [204% of  $25]
```
simonmichael, line 53: Another behaviour change, worth noting with a ! in message.

The parent "expenses" account is not shown, because there's no explicit budget goal for it, and because we're in list mode ? So if we want to see aggregated budget performance, tree mode will be needed. Ok I guess.

Why is it still shown in the previous test ?

hledger/test/balance/budget.test
```diff
@@ -747,8 +745,7 @@ Budget performance in 2023-08:
 ==============++=================
  income       || 0 [0% of $-200] 
    employment || 0 [0% of $-100] 
-   gifts      || 0               
-     cash     || 0 [0% of $-100] 
+   gifts:cash || 0 [0% of $-100]
```
simonmichael, line 748: This looks like a consistency fix, squashing these like a normal balance report would. I think I tried and failed to fix this recently. 👍🏻

hledger/test/incomestatement.test
```diff
@@ -296,6 +296,7 @@ Income Statement ..
 =================++====
  Revenues        ||    
 -----------------++----
+ revenues        ||  0
```
simonmichael, line 299: Should drop "leaf" from the test description now I guess.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -733,18 +735,31 @@ nullaccountdeclarationinfo = AccountDeclarationInfo {
...
-  } deriving (Generic)
+  ,abalances                 :: AccountBalances a   -- ^ historical and date-associated account balances
+  } deriving (Generic, Functor)
+
+-- | A component of an 'Account' containing historical 'AccountBalance', and a
+-- map of start dates to 'AccountBalance'.
+data AccountBalances a = AccountBalances {
+   abhistorical :: a            -- ^ historical balance information
+  ,abdatemap    :: IM.IntMap a  -- ^ balance information associated to a start day
+  } deriving (Eq, Functor, Generic)
```
simonmichael, line 754: What's the reason for using a type parameter ? Do we truly need it ?

How does the "start day" map, with Int keys I assume, work here ?

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -733,18 +735,31 @@ nullaccountdeclarationinfo = AccountDeclarationInfo {
 
 -- | An account, with its balances, parent/subaccount relationships, etc.
```
simonmichael, line 736: Perhaps we should change this to "An account within a hierarchy", or something. Representing/navigating a hierarchy of account balances is Account's purpose.

simonmichael (commented):
Thanks! Initial comment.

hledger-lib/Hledger/Data/Types.hs
simonmichael: > What's the reason for using a type parameter ? Do we truly need it ?

For budget report, I guess.

simonmichael:
> A small speed improvement on large journal files

Could we quantify that a little more - eg "balance reports are 1% faster with 1k txns, 5% faster with 10k txns" ?

simonmichael:
Also I wonder if there's any memory impact, `stats` might be a quick way to check (it runs balance reports probably)

hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
...
+-- | The Semigroup instance for 'AccountBalance' will simply take the union of
+-- keys in the date map section. This may not be the result you want if the
+-- keys are not identical.
+instance Semigroup a => Semigroup (AccountBalances a) where
+  AccountBalances h1 as1 <> AccountBalances h2 as2 = AccountBalances (h1 <> h2) $ IM.unionWith (<>) as1 as2
+
+instance Monoid a => Monoid (AccountBalances a) where
+  mempty = AccountBalances mempty mempty
+
+-- | Construct an 'AccountBalance' from a list.
```
simonmichael, line 71: Construct an 'AccountBalances' from a list

hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
...
+accountBalancesFromList :: a -> [(Day, a)] -> AccountBalances a
+accountBalancesFromList h = AccountBalances h . IM.fromList . map (\(d, a) -> (fromInteger $ toModifiedJulianDay d, a))
+
+-- | Get the 'AccountBalance' associated to the period containing a given 'Day'.
```
simonmichael, line 75: Get the account balance information associated ... ?

hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
...
+lookupAccountBalance :: Day -> AccountBalances a -> a
+lookupAccountBalance d (AccountBalances h as) =
+    maybe h snd $ IM.lookupLE (fromInteger $ toModifiedJulianDay d) as
+
+-- | Add the 'AccountBalance' to the appropriate location in 'AccountBalances'.
```
simonmichael, line 80: Add balance information ...

hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
+{-# LANGUAGE CPP #-}
+{-|
+
+
+An 'AccountBalance' consists of a historical balance, along with balances
```
simonmichael, line 5: Maybe clarify that this module contains code for both the AccountBalances and AccountBalance types, which have separate purposes.

hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
...
+lookupAccountBalance d (AccountBalances h as) =
+    maybe h snd $ IM.lookupLE (fromInteger $ toModifiedJulianDay d) as
+
+-- | Add the 'AccountBalance' to the appropriate location in 'AccountBalances'.
+insertAccountBalances :: Semigroup a => Maybe Day -> a -> AccountBalances a -> AccountBalances a
+insertAccountBalances mday b balances = case mday of
+    Nothing  -> balances{abhistorical = abhistorical balances <> b}
+    Just day -> balances{abdatemap = IM.insertWith (<>) (fromInteger $ toModifiedJulianDay day) b $ abdatemap balances}
+
+-- | Performs an operation on the contents of two 'AccountBalances'.
```
simonmichael, line 86: Combines two AccountBalances into one, using the given operation to combine their balance informations.

hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
...
+--
+-- This will drop keys if they are not present in both AccountBalances.
+opAccountBalances :: (a -> b -> c) -> AccountBalances a -> AccountBalances b -> AccountBalances c
+opAccountBalances f (AccountBalances h1 as1) (AccountBalances h2 as2) =
+    AccountBalances (f h1 h2) $ IM.intersectionWith f as1 as2
+
+-- | Performs an operation on the contents of two 'AccountBalances', with
+-- separate treatments for those only in the first, only in the second, or in
+-- both 'AccountBalance's.
```
simonmichael, line 95: Merges two AccountBalances, with separate treatments for balance informations only in ... ?

hledger-lib/Hledger/Data/AccountBalance.hs
```diff
@@ -0,0 +1,159 @@
...
+  AccountBalance n e i <> AccountBalance n' e' i' = AccountBalance (n + n') (maPlus e e') (maPlus i i')
+
+instance Monoid AccountBalance where
+  mempty = AccountBalance 0 nullmixedamt nullmixedamt
+
+-- | Apply an operation to both 'MixedAmount' in an 'AccountBalance'.
+applyAccountBalance :: (MixedAmount -> MixedAmount) -> AccountBalance -> AccountBalance
+applyAccountBalance f a = a{abebalance = f $ abebalance a, abibalance = f $ abibalance a}
+
+-- | Perform an operation on the 'MixedAmount' in two 'AccountBalance'.
```
simonmichael, line 127: Merge two AccountBalance, using the given operation to combine their amounts ?

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -733,18 +735,31 @@ nullaccountdeclarationinfo = AccountDeclarationInfo {
...
-  ,aboring                   :: Bool           -- ^ used in the accounts report to label elidable parents
+  ,asubs                     :: [Account a]        -- ^ this account's sub-accounts
+  ,aparent                   :: Maybe (Account a)  -- ^ parent account
+  ,aboring                   :: Bool                -- ^ used in the accounts report to label elidable parents
   -- balance information
-  ,anumpostings              :: Int            -- ^ the number of postings to this account
-  ,aebalance                 :: MixedAmount    -- ^ this account's balance, excluding subaccounts
-  ,aibalance                 :: MixedAmount    -- ^ this account's balance, including subaccounts
-  } deriving (Generic)
+  ,abalances                 :: AccountBalances a   -- ^ historical and date-associated account balances
```
simonmichael, line 746: It's hard to be clear on this everywhere, but maybe here mention both "end balances" (on a date) and "balance changes" (in a period) are represented, consistent with terminology elsewhere.

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -85,33 +108,47 @@ nullacct = Account
...
+accountTreeFromBalanceAndNames rootname bals as =
+    (accountFromBalances rootname bals){ asubs=map (uncurry accountTree') $ M.assocs m }
   where
     T m = treeFromPaths $ map expandAccountName as :: FastTree AccountName
     accountTree' a (T m') =
-      nullacct{
-        aname=a
-       ,asubs=map (uncurry accountTree') $ M.assocs m'
-       }
+      (accountFromBalances a bals){ asubs=map (uncurry accountTree') $ M.assocs m' }
```
simonmichael, line 151: This doc and function is a bit hard to understand, any clarification welcome

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -130,43 +167,59 @@ treeFromPaths = foldl' mergeTrees (T M.empty) . map treeFromPath
...
-accountsLevels :: Account -> [[Account]]
+accountsLevels :: Account a -> [[Account a]]
 accountsLevels = takeWhile (not . null) . iterate (concatMap asubs) . (:[])
 
 -- | Map a (non-tree-structure-modifying) function over this and sub accounts.
-mapAccounts :: (Account -> Account) -> Account -> Account
+mapAccounts :: (Account a -> Account a) -> Account a -> Account a
 mapAccounts f a = f a{asubs = map (mapAccounts f) $ asubs a}
 
+-- | Apply a function to all 'AccountBalances' within this and sub accounts.
```
simonmichael, line 190: Similar operations in AccountBalances.hs are named "apply" - maybe "map" is best ?

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -130,43 +167,59 @@ treeFromPaths = foldl' mergeTrees (T M.empty) . map treeFromPath
...
     | p a = True
     | otherwise = any (anyAccounts p) $ asubs a
 
+-- | Is the predicate true on all of this account and its subaccounts ?
+allAccounts :: (Account a -> Bool) -> Account a -> Bool
+allAccounts p a
+    | not (p a) = False
+    | otherwise = all (allAccounts p) $ asubs a
+
 -- | Add subaccount-inclusive balances to an account tree.
```
simonmichael, line 206: While you're here, we could improve this one:
Recalculate all the subaccount-inclusive balances in this account tree.

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -130,43 +167,59 @@ treeFromPaths = foldl' mergeTrees (T M.empty) . map treeFromPath
...
-  | otherwise      = a{aibalance=ibal, asubs=subs}
+sumAccounts :: Account AccountBalance -> Account AccountBalance
+sumAccounts a = a{asubs = subs, abalances = setImplicitBalances $ abalances a}
   where
     subs = map sumAccounts $ asubs a
-    ibal = maSum $ aebalance a : map aibalance subs
+    subtotals = foldMap abalances subs
+
+    setImplicitBalances :: AccountBalances AccountBalance -> AccountBalances AccountBalance
+    setImplicitBalances = mergeAccountBalances onlyChildren noChildren combineChildren subtotals
```
simonmichael, line 214: setInclusiveBalances, I think ?

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -224,33 +278,47 @@ pruneAccounts p = headMay . prune
...
 
--- | Sort each group of siblings in an account tree by inclusive amount,
--- so that the accounts with largest normal balances are listed first.
--- The provided normal balance sign determines whether normal balances
--- are negative or positive, affecting the sort order. Ie,
--- if balances are normally negative, then the most negative balances
--- sort first, and vice versa.
-sortAccountTreeByAmount :: NormalSign -> Account -> Account
-sortAccountTreeByAmount normalsign = mapAccounts $ \a -> a{asubs=sortSubs $ asubs a}
+-- | Merge two accounts and their subaccounts.
```
simonmichael, line 291: Merge two account trees ?

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -85,33 +108,47 @@ nullacct = Account
...
-    acctsflattened = flattenAccounts acctswithparents
-  in
-    acctsflattened
+accountsFromPostings :: (Posting -> Maybe Day) -> [Posting] -> [Account AccountBalance]
+accountsFromPostings getPostingDate = flattenAccounts . accountFromPostings getPostingDate
+
+-- | Derive 1. an account tree and 2. each account's total exclusive
+-- and inclusive changes from a list of postings.
+-- This is the core of the balance command (and of *ledger).
+-- The accounts are returned as tree.
```
simonmichael, line 117: a tree

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -85,33 +108,47 @@ nullacct = Account
 -- The accounts are returned as a list in flattened tree order,
 -- and also reference each other as a tree.
 -- (The first account is the root of the tree.)
```
simonmichael, line 110: In this (and maybe the next), let's explain a bit more that this calculates multiple inclusive and exclusive change amounts for each account, associated with the dates produced by the given date-bucketing function (usually representing the end dates of report subperiods)

simonmichael: (It looks like this will enable non-contiguous report periods, which will be cool)

Xitian9:
Thanks for the detailed comments. I'm on holiday at the moment, so it'll be
about 3 weeks before I can make the requested changes, but I will when I
get back.

hledger-lib/Hledger/Data/AccountBalance.hs
Xitian9: I'll follow your later advice and split these apart into separate modules (`AccountBalance` and `AccountBalances`).

hledger-lib/Hledger/Data/AccountBalance.hs
Xitian9: I'll break them apart into separate modules.

hledger/test/balance/balance.test
Xitian9: I don't think it's new: I got it from the suggested prefixes in the developer documentation https://hledger.org/COMMITS.html. But you're correct that it's not just a refactoring, so I'll change to dev.

hledger-lib/Hledger/Data/AccountBalance.hs
Xitian9: Done

hledger-lib/Hledger/Data/AccountBalance.hs
Xitian9: Done

hledger-lib/Hledger/Data/Types.hs
Xitian9: Done

hledger-lib/Hledger/Data/Account.hs
Xitian9: Done

hledger-lib/Hledger/Data/Types.hs
Xitian9: Yes, the main use is for the budget report. But I think it also simplifies things a bit by exposing the functor, foldable, and traversable interfaces for `AccountBalances`, saving having to write out a lot of boilerplate to perform tasks that re-implement that functionality in a monomorphic container.

hledger-lib/Hledger/Data/Account.hs
Xitian9: Done

hledger/test/balance/balance.test
Xitian9: I've changed the test description. Let me know if it's clearer.

hledger/test/incomestatement.test
Xitian9: Done

hledger/test/balance/balance.test
Xitian9: I've modified the test descriptions and marked this as a breaking change.

hledger-lib/Hledger/Data/Types.hs
Xitian9: Done

hledger-lib/Hledger/Data/Account.hs
Xitian9: Done

hledger-lib/Hledger/Data/AccountBalance.hs
Xitian9: Done

hledger-lib/Hledger/Data/Account.hs
Xitian9: Good catch. I must have been pretty tired when I wrote that.

hledger-lib/Hledger/Data/AccountBalance.hs
Xitian9: Done

hledger/test/balance/budget.test
Xitian9: Let me look into this.

Xitian9:
Here is the benchmarking. Marginal change for small journals, but about 5% time savings for the 100k journal, and roughly comparable for my real-life journal (21k transactions, 796 accounts of depth 7).

```
Running 6 tests 5 times with 2 executables at 2025-04-25 14:16:10 AEST:

Best times:
+--------------------------------------------------------------------++------------------+------------------------+
|                                                                    || ./hledger-master | ./hledger-multiaccount |
+====================================================================++==================+========================+
| -f examples/10ktxns-1kaccts.journal balance                        ||             0.82 |                   0.81 |
| -f examples/1ktxns-1kaccts.journal balance --weekly                ||             0.64 |                   0.64 |
| -f examples/10ktxns-1kaccts.journal balance --weekly               ||             7.66 |                   7.57 |
| -f examples/100ktxns-1kaccts.journal balance --yearly              ||            11.35 |                  10.43 |
| balance --value=end @/home/myname/expenses-report.args             ||             1.40 |                   1.31 |
| balance --layout=tidy --daily @/home/myname/assetsliabilities.args ||             1.49 |                   1.46 |
+--------------------------------------------------------------------++------------------+------------------------+
```

`hledger stats` shows no difference in memory usage, but that's not too surprising as it uses a single-period balance report in the backend. The changes will only affect multi-period balance reports.

Xitian9:
It looks like memory use is a bit higher with the new code. Lower heap use, but higher maximum residency.

```
$ ./hledger-master-prof -f examples/100ktxns-1kaccts.journal balance --yearly +RTS -s > /dev/null
  90,879,905,384 bytes allocated in the heap
   8,694,480,976 bytes copied during GC
     559,648,704 bytes maximum residency (15 sample(s))
      13,950,024 bytes maximum slop
            1573 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0     21878 colls,     0 par    2.543s   2.555s     0.0001s    0.0034s
  Gen  1        15 colls,     0 par    2.087s   2.092s     0.1395s    0.2950s

  INIT    time    0.003s  (  0.003s elapsed)
  MUT     time   57.091s  ( 57.175s elapsed)
  GC      time    4.630s  (  4.647s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   61.725s  ( 61.825s elapsed)

  Alloc rate    1,591,840,064 bytes per MUT second

  Productivity  92.5% of total user, 92.5% of total elapsed

$ ./hledger-multiaccount-prof -f examples/100ktxns-1kaccts.journal balance --yearly +RTS -s > /dev/null

  87,899,201,024 bytes allocated in the heap
   8,609,724,032 bytes copied during GC
     603,841,256 bytes maximum residency (15 sample(s))
      13,949,544 bytes maximum slop
            1688 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0     21163 colls,     0 par    2.438s   2.450s     0.0001s    0.0034s
  Gen  1        15 colls,     0 par    2.183s   2.188s     0.1459s    0.3211s

  INIT    time    0.003s  (  0.003s elapsed)
  MUT     time   56.039s  ( 56.113s elapsed)
  GC      time    4.622s  (  4.638s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   60.664s  ( 60.754s elapsed)

  Alloc rate    1,568,534,957 bytes per MUT second

  Productivity  92.4% of total user, 92.4% of total elapsed
```

hledger/test/balance/budget.test
Xitian9: It looks like the reason is as follows:

Unless called with `-E`, then a budget report will count any unbudgeted subaccounts against their earliest budgeted parent. So both `expenses:cab` and `expenses:movies` are rolled up to `expenses`. Even though `expenses` doesn't have a budget itself, it gets the sum of the budgets of its subaccounts.

I'm not sure how I feel about this behaviour, but I think it's out of scope for this PR.

Xitian9:
I think I've responded to all comments. Let me know if you want to discuss further.

hledger-lib/Hledger/Data/Account.hs
```diff
@@ -68,50 +88,73 @@ instance Eq Account where
...
+    tieAccountParents . sumAccounts $ mapAccounts setBalance acctTree
+  where
+    -- The special name "..." is stored in the root of the tree
+    acctTree     = accountTree "root" . HM.keys $ HM.delete "..." accountMap
+    setBalance a = a{abalances = HM.lookupDefault mempty name accountMap}
+      where name = if aname a == "root" then "..." else aname a
+    accountMap   = processPostings ps
+
+    processPostings :: [Posting] -> HM.HashMap AccountName (AccountBalances AccountBalance)
+    processPostings = foldl' (flip processAccountName) mempty
```
Xitian9, line 134: There's a question about whether we get better performance with this as `foldl'` or `foldr`. It seems that `foldl'` is slightly faster, while `foldr` has better memory usage.

Xitian9: Here's the time and memory usage when using the foldr version. I wonder if foldr is the winner here.

```
Running 6 tests 5 times with 3 executables at 2025-04-25 19:44:53 AEST:

Best times:
+--------------------------------------------------------------------++------------------+------------------------+------------------------------+
|                                                                    || ./hledger-master | ./hledger-multiaccount | ./hledger-multiaccount-foldr |
+====================================================================++==================+========================+==============================+
| -f examples/10ktxns-1kaccts.journal balance                        ||             0.81 |                   0.80 |                         0.81 |
| -f examples/1ktxns-1kaccts.journal balance --weekly                ||             0.68 |                   0.65 |                         0.64 |
| -f examples/10ktxns-1kaccts.journal balance --weekly               ||             7.91 |                   7.63 |                         7.94 |
| -f examples/100ktxns-1kaccts.journal balance --yearly              ||            11.29 |                  10.92 |                        10.84 |
| balance --value=end @/home/myname/expenses-report.args             ||             1.39 |                   1.32 |                         1.32 |
| balance --layout=tidy --daily @/home/myname/assetsliabilities.args ||             1.49 |                   1.46 |                         1.46 |
+--------------------------------------------------------------------++------------------+------------------------+------------------------------+
```
 ```
$ ./hledger-multiaccount-foldr-prof -f examples/100ktxns-1kaccts.journal balance --yearly +RTS -s > /dev/null
   87,949,386,664 bytes allocated in the heap
   8,323,439,592 bytes copied during GC
     558,243,136 bytes maximum residency (15 sample(s))
      13,941,192 bytes maximum slop
            1551 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0     21167 colls,     0 par    2.663s   2.680s     0.0001s    0.0100s
  Gen  1        15 colls,     0 par    2.385s   2.394s     0.1596s    0.5332s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time   56.837s  ( 56.914s elapsed)
  GC      time    5.048s  (  5.074s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   61.888s  ( 61.990s elapsed)

  Alloc rate    1,547,396,211 bytes per MUT second

  Productivity  91.8% of total user, 91.8% of total elapsed
```

Comment (Xitian9) [reply hledger-lib/Hledger/Data/Account.hs:134]: Actually, I think I messed up the memory analysis by using profiled versions of the executables. The answer is less dramatic for the normal versions. I've included them here.

```
$ ./hledger-master -f examples/100ktxns-1kaccts.journal balance --yearly +RTS -s
  43,526,689,224 bytes allocated in the heap
   4,818,585,000 bytes copied during GC
     376,952,024 bytes maximum residency (13 sample(s))
       3,414,824 bytes maximum slop
            1041 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0     10559 colls,     0 par    1.651s   1.660s     0.0002s    0.0043s
  Gen  1        13 colls,     0 par    1.711s   1.718s     0.1322s    0.3766s

  INIT    time    0.002s  (  0.001s elapsed)
  MUT     time    7.573s  (  7.582s elapsed)
  GC      time    3.362s  (  3.378s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   10.937s  ( 10.961s elapsed)

  Alloc rate    5,747,632,285 bytes per MUT second

  Productivity  69.2% of total user, 69.2% of total elapsed

$ ./hledger-multiaccount -f examples/100ktxns-1kaccts.journal balance --yearly +RTS -s
  41,742,275,096 bytes allocated in the heap
   4,600,175,152 bytes copied during GC
     388,933,464 bytes maximum residency (13 sample(s))
       2,396,632 bytes maximum slop
             999 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0     10128 colls,     0 par    1.628s   1.637s     0.0002s    0.0034s
  Gen  1        13 colls,     0 par    1.494s   1.499s     0.1153s    0.2873s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time    7.544s  (  7.561s elapsed)
  GC      time    3.122s  (  3.136s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   10.667s  ( 10.698s elapsed)

  Alloc rate    5,533,021,721 bytes per MUT second

  Productivity  70.7% of total user, 70.7% of total elapsed

$ ./hledger-multiaccount-foldr -f examples/100ktxns-1kaccts.journal balance --yearly +RTS -s
  41,751,866,552 bytes allocated in the heap
   4,611,401,952 bytes copied during GC
     381,899,736 bytes maximum residency (13 sample(s))
       3,317,352 bytes maximum slop
            1014 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0     10129 colls,     0 par    1.528s   1.537s     0.0002s    0.0028s
  Gen  1        13 colls,     0 par    1.503s   1.507s     0.1159s    0.3262s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time    7.356s  (  7.368s elapsed)
  GC      time    3.031s  (  3.044s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   10.387s  ( 10.412s elapsed)

  Alloc rate    5,676,066,670 bytes per MUT second

  Productivity  70.8% of total user, 70.8% of total elapsed

./hledger-new -f examples/100ktxns-1kaccts.journal balance --yearly +RTS -s
  41,747,524,560 bytes allocated in the heap
   4,624,839,952 bytes copied during GC
     384,986,456 bytes maximum residency (13 sample(s))
       3,320,528 bytes maximum slop
            1025 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0     10128 colls,     0 par    1.505s   1.515s     0.0001s    0.0026s
  Gen  1        13 colls,     0 par    1.620s   1.627s     0.1251s    0.3372s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time    7.366s  (  7.389s elapsed)
  GC      time    3.125s  (  3.141s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   10.492s  ( 10.531s elapsed)

  Alloc rate    5,667,555,667 bytes per MUT second

  Productivity  70.2% of total user, 70.2% of total elapsed
```

**Comment by simonmichael** [comment]
Hi @Xitian9 - sorry for the slow response. 

For various reasons I have been finding this PR hard to digest, after multiple attempts with both github's web ui and VS code git(hub) uis.

I considered asking you to rewrite the history for maximum simplicity and reviewability, but that would have lost all the github comment context, so I didn't.

I think it'll help me a little if you don't mark comments resolved - let me/the reviewer do that.

I'm having a hard time seeing/confirming the new diffs/resolutions you made, in any of the UIs; sometimes I see things have changed, sometimes I don't. I'm giving up on that and just trusting your "done" comments.

Thanks for the performance data. I honestly can't digest it right now, and can't use it in release notes. For those I'll need to describe the perf changes in a sentence, no more. It seemed to me that with this PR we'll be saying "hledger uses slightly more memory and is very slightly faster". So I think the main value of this PR is cleaner and more flexible code ?

**Comment by simonmichael** [comment]
Thanks @Xitian9 - it's not essential but if you have time, it might be quite helpful to see a simpler history - or possibly just a single squashed commit if it seems appropriate.

**Comment by simonmichael** [comment]
One other thing - should AccountBalance(s) modules be added to package.yaml also ? I think so.

**Comment by Xitian9** [comment]
What would be more helpful? More atomic commits (implementing fewer more self-contained changes, yielding possibly more commits) or more monolithic commits in which more changes can be seen at once?

**Comment by simonmichael** [comment]
> What would be more helpful? More atomic commits (implementing fewer more self-contained changes, yielding possibly more commits) or more monolithic commits in which more changes can be seen at once?

I wasn't sure, so.. here's a draft of a single squashed commit: 
https://github.com/simonmichael/hledger/tree/sm-multiaccount-1

I added modules to package.yaml also.

**Comment by simonmichael** [comment]
There's something funny going on with benchmarking, for me (on mac m1, compiling with ghc 9.10.1). I can't seem to get a stable measurement with the big 100k file - the relative speeds seem to alternate, and by a lot, seemingly no matter how many times I try. Can you replicate the following ?
```
~/src/hledger$ j bench -w hledger-1.42.1,hledger-2360 -n3 -N4 -f bench10k.sh 
Running quick benchmarks (times are approximate, can be skewed):
Running 3 tests 3 times with 2 executables at 2025-05-08 09:17:25 HST:

Best times 1:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           0.87 |         1.02 |
| -f examples/10ktxns-1kaccts.journal register ||           3.94 |         3.28 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.94 |         0.87 |
+----------------------------------------------++----------------+--------------+

Best times 2:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           0.91 |         0.92 |
| -f examples/10ktxns-1kaccts.journal register ||           3.90 |         3.54 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.84 |         0.82 |
+----------------------------------------------++----------------+--------------+

Best times 3:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           0.88 |         0.87 |
| -f examples/10ktxns-1kaccts.journal register ||           3.57 |         3.44 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.82 |         0.85 |
+----------------------------------------------++----------------+--------------+

Best times 4:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           0.85 |         0.89 |
| -f examples/10ktxns-1kaccts.journal register ||           3.47 |         3.44 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.90 |         0.85 |
+----------------------------------------------++----------------+--------------+
~/src/hledger$ j bench -w hledger-1.42.1,hledger-2360 -n1 -N4 -f bench100k.sh 
Running quick benchmarks (times are approximate, can be skewed):
Running 3 tests 1 times with 2 executables at 2025-05-08 09:20:02 HST:

Best times 1:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||           8.46 |         8.48 |
| -f examples/100ktxns-1kaccts.journal register ||          59.43 |        81.54 |
| -f examples/100ktxns-1kaccts.journal balance  ||           7.69 |         7.98 |
+-----------------------------------------------++----------------+--------------+

Best times 2:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||          10.63 |        10.11 |
| -f examples/100ktxns-1kaccts.journal register ||         100.75 |        85.51 |
| -f examples/100ktxns-1kaccts.journal balance  ||          10.78 |        10.77 |
+-----------------------------------------------++----------------+--------------+

Best times 3:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||          20.57 |        12.53 |
| -f examples/100ktxns-1kaccts.journal register ||          94.98 |       132.27 |
| -f examples/100ktxns-1kaccts.journal balance  ||          22.26 |        14.28 |
+-----------------------------------------------++----------------+--------------+

Best times 4:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||          17.18 |        17.49 |
| -f examples/100ktxns-1kaccts.journal register ||         124.83 |        85.31 |
| -f examples/100ktxns-1kaccts.journal balance  ||           8.68 |         9.34 |
+-----------------------------------------------++----------------+--------------+
```

**Comment by simonmichael** [comment]
Perhaps a 2 minute run time is just too long for casual benchmarking on this machine.

**Comment by simonmichael** [comment]
A few more benchmark runs, being slightly more careful:

10k txns:
```
~/src/hledger$ j bench -w hledger-1.42.1,hledger-2360 -n3 -N4 -f bench10k.sh 
Running quick benchmarks (times are approximate, can be skewed):
Running 4 cycles of 3 tests best of 3 times with 2 executables at 2025-05-08 13:26:18 HST:

Best of 3 times, cycle 1:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           1.01 |         1.04 |
| -f examples/10ktxns-1kaccts.journal register ||           3.15 |         3.09 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.97 |         0.98 |
+----------------------------------------------++----------------+--------------+

Best of 3 times, cycle 2:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           1.01 |         1.03 |
| -f examples/10ktxns-1kaccts.journal register ||           3.27 |         3.07 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.98 |         1.00 |
+----------------------------------------------++----------------+--------------+

Best of 3 times, cycle 3:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           1.02 |         1.04 |
| -f examples/10ktxns-1kaccts.journal register ||           3.02 |         3.06 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.97 |         1.00 |
+----------------------------------------------++----------------+--------------+

Best of 3 times, cycle 4:
+----------------------------------------------++----------------+--------------+
|                                              || hledger-1.42.1 | hledger-2360 |
+==============================================++================+==============+
| -f examples/10ktxns-1kaccts.journal print    ||           1.04 |         1.04 |
| -f examples/10ktxns-1kaccts.journal register ||           3.11 |         3.21 |
| -f examples/10ktxns-1kaccts.journal balance  ||           0.98 |         1.01 |
+----------------------------------------------++----------------+--------------+
```

100k txns:
```
~/src/hledger$ j bench -w hledger-1.42.1,hledger-2360 -n2 -N4 -f bench100k.sh 
Running quick benchmarks (times are approximate, can be skewed):
Running 4 cycles of 3 tests best of 2 times with 2 executables at 2025-05-08 13:28:38 HST:

Best of 2 times, cycle 1:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||           7.52 |         8.00 |
| -f examples/100ktxns-1kaccts.journal register ||          44.09 |        48.28 |
| -f examples/100ktxns-1kaccts.journal balance  ||           6.91 |         7.00 |
+-----------------------------------------------++----------------+--------------+

Best of 2 times, cycle 2:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||           7.72 |         7.79 |
| -f examples/100ktxns-1kaccts.journal register ||          66.86 |        39.35 |
| -f examples/100ktxns-1kaccts.journal balance  ||           6.70 |         6.84 |
+-----------------------------------------------++----------------+--------------+

Best of 2 times, cycle 3:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||           7.41 |         7.65 |
| -f examples/100ktxns-1kaccts.journal register ||          38.12 |        39.67 |
| -f examples/100ktxns-1kaccts.journal balance  ||           6.51 |         6.74 |
+-----------------------------------------------++----------------+--------------+

Best of 2 times, cycle 4:
+-----------------------------------------------++----------------+--------------+
|                                               || hledger-1.42.1 | hledger-2360 |
+===============================================++================+==============+
| -f examples/100ktxns-1kaccts.journal print    ||           7.69 |         8.02 |
| -f examples/100ktxns-1kaccts.journal register ||          37.01 |        39.94 |
| -f examples/100ktxns-1kaccts.journal balance  ||           6.75 |         6.87 |
+-----------------------------------------------++----------------+--------------+
```

With these commands I can't reproduce any speedup, indeed it almost looks slower.

**Comment by Xitian9** [comment]
Something's odd here. There should be no difference between the two for print or register reports, as the changed code is not used in either of them (I guess there's a minor change in the register code, but it shouldn't have any practical effect). In addition, there shouldn't be any appreciable difference for a single-period balance report, as the different algorithm should be identical when it's all being put into a single period.

I'll rebase this evening and run a comparison against current master (with the other strictness changes that have been made in master, the benchmarking may have changed).

**Comment by Xitian9** [comment]
```
Running 7 tests 2 times with 2 executables at 2025-05-09 19:36:15 AEST:

Best times 1:
+-------------------------------------------------------++------------------+--------------------------------+
|                                                       || ./hledger-master | ./hledger-multiaccount-rebased |
+=======================================================++==================+================================+
| -f examples/10ktxns-1kaccts.journal print             ||             0.89 |                           0.89 |
| -f examples/10ktxns-1kaccts.journal register          ||             3.11 |                           2.89 |
| -f examples/10ktxns-1kaccts.journal balance           ||             0.85 |                           0.82 |
| -f examples/100ktxns-1kaccts.journal print            ||             8.80 |                           8.78 |
| -f examples/100ktxns-1kaccts.journal register         ||            29.54 |                          26.84 |
| -f examples/10ktxns-1kaccts.journal balance --weekly  ||             7.52 |                           7.67 |
| -f examples/100ktxns-1kaccts.journal balance --yearly ||            10.78 |                          11.54 |
+-------------------------------------------------------++------------------+--------------------------------+

Best times 2:
+-------------------------------------------------------++------------------+--------------------------------+
|                                                       || ./hledger-master | ./hledger-multiaccount-rebased |
+=======================================================++==================+================================+
| -f examples/10ktxns-1kaccts.journal print             ||             0.88 |                           0.88 |
| -f examples/10ktxns-1kaccts.journal register          ||             2.95 |                           2.74 |
| -f examples/10ktxns-1kaccts.journal balance           ||             0.84 |                           0.82 |
| -f examples/100ktxns-1kaccts.journal print            ||             8.59 |                           8.63 |
| -f examples/100ktxns-1kaccts.journal register         ||            29.02 |                          26.85 |
| -f examples/10ktxns-1kaccts.journal balance --weekly  ||             7.68 |                           8.05 |
| -f examples/100ktxns-1kaccts.journal balance --yearly ||            10.48 |                          10.81 |
+-------------------------------------------------------++------------------+--------------------------------+

Best times 3:
+-------------------------------------------------------++------------------+--------------------------------+
|                                                       || ./hledger-master | ./hledger-multiaccount-rebased |
+=======================================================++==================+================================+
| -f examples/10ktxns-1kaccts.journal print             ||             0.87 |                           0.88 |
| -f examples/10ktxns-1kaccts.journal register          ||             3.19 |                           2.83 |
| -f examples/10ktxns-1kaccts.journal balance           ||             0.85 |                           0.82 |
| -f examples/100ktxns-1kaccts.journal print            ||             8.65 |                           8.66 |
| -f examples/100ktxns-1kaccts.journal register         ||            30.50 |                          28.16 |
| -f examples/10ktxns-1kaccts.journal balance --weekly  ||             7.95 |                           7.82 |
| -f examples/100ktxns-1kaccts.journal balance --yearly ||            11.07 |                          11.07 |
+-------------------------------------------------------++------------------+--------------------------------+

Best times 4:
+-------------------------------------------------------++------------------+--------------------------------+
|                                                       || ./hledger-master | ./hledger-multiaccount-rebased |
+=======================================================++==================+================================+
| -f examples/10ktxns-1kaccts.journal print             ||             0.88 |                           0.89 |
| -f examples/10ktxns-1kaccts.journal register          ||             3.16 |                           2.82 |
| -f examples/10ktxns-1kaccts.journal balance           ||             0.84 |                           0.82 |
| -f examples/100ktxns-1kaccts.journal print            ||             8.61 |                           8.77 |
| -f examples/100ktxns-1kaccts.journal register         ||            28.59 |                          31.16 |
| -f examples/10ktxns-1kaccts.journal balance --weekly  ||             8.12 |                           7.16 |
| -f examples/100ktxns-1kaccts.journal balance --yearly ||            11.38 |                          10.77 |
+-------------------------------------------------------++------------------+--------------------------------+
```

Running the benchmarks again. There may be an effect here on the balance report, but it seesaws back and forth depending on the run. I'll call this a wash for time. There are fewer heap allocations overall, but memory usage seems to increase faster with more intervals, leading to greater memory usage when there are a very large number of intervals

Oddly, there seems to be a relatively consistent improvement in the register report. I'm confused by that, as I didn't think the new code had any effect there. Looking at the memory use, there seems to be a significant decrease in slop for the register report with the new code. Maybe this is a benchmarking artifact.

I would probably summarise this as:
Small improvements in heap usage for the balance reports. There can be greater overall memory usage when there are a large number of intervals. (Maybe speed improvements for the register report, but not certain)

**Comment by Xitian9** [comment]
> More draft change notes, and questions:
> 
> Balance reports' --declared flag now also shows "boring" zero-balance parents, if they are declared.
> 
> Budget reports (balance --budget) no longer show boring zero-balance parent accounts.
> 
> Improve the type of mixedAmountApplyValuationAfterSumFromOptsWith, removing an error case.
> 
> > hledger-web/Hledger/Web/Widget/Common.hs	Modification of “isInteresting” logic using historical balances
> 
> What's happening here ?

This is just a change necessitated by the fact that the data type `Account` was changed. It formerly had a field `aebalance` and this code checked whether that field was zero. But `Account` no longer has a field `aebalance`, rather it has a field `abalances` which holds an `AccountBalances`, which itself has fields `abhistorical` and `abdatemap`, each of which are an `AccountBalance` with field `abebalance`. So instead of checking whether `aebalance` is zero, we check whether all the `abebalances` in `abdatemap . abalances` are zero (only one of them will be populated in this case).

So overall the code does the same thing, it just had to be rewritten to change the fact that the datatype `Account` no longer has the same structure.

> 
> > hledger-lib/Hledger/Reports/PostingsReport.hs	Updated balance calculation to use historical balances from abalances
> 
> And here ? I don't see that the old code forced historical balances, are we sure there's no change of behaviour here ?

This is the same as above. Previously we got the balance from either `aebalance` or `aibalance` in `Account`, but we can't do that anymore. Instead we store everything in the `abhistorical` field of `abalances` (that's what the `accountsFromPostings (const Nothing)` does), and then pull that out. 

> 
> ```haskell
> -- | A component of an 'Account' containing historical 'AccountBalance', and a
> -- map of start dates to 'AccountBalance'.
> data AccountBalances a = AccountBalances {
>    abhistorical :: a            -- ^ historical balance information
>   ,abdatemap    :: IM.IntMap a  -- ^ balance information associated to a start day
> ```
> 
> A note for possible future cleanups: The use of start date feels a little odd here, since a balance is usually valid as of a period's end date. And it seems to rely on information stored elsewhere (this type doesn't know the duration of the report periods).

This is definitely worth discussing. I made this change very explicitly, but it does have downsides.

The advantage of this setup is that illegal states are unrepresentable. The start dates of the periods impose a partition of time, and given a date there is exactly one bin that it can fit into. The previous code, where both the start and end times were stored, could result in illegal setups, like [DateSpan 2023-01-01 2023-02-01, DateSpan 2023-01-02, 2023-02-02]. We either had to write error-checking code to make sure the partition was valid, assume that no invalid states could arise, or write extra cases which raised programming errors like "This state is illegal". With this new representation, we don't need to worry about those cases.

The disadvantage is what you say: we need to do extra work to get the end dates of periods. I've gone back and forth on whether there's a improvement to be made here that would keep illegal states unrepresentable, while also having that information easy to access. It could conceivable be made a field of `AccountBalance`, but this would result in a lot of redundant values floating around. And it would result in its own illegal states: what's to guarantee that every `AccountBalance` in an `Account` has the same end-date recorded? This should be a feature of the `Account` itself, not of every individual element.

If you have any ideas, I'd love to hear them.

> 
> > hledger-lib/Hledger/Data/Json.hs	Adapted JSON serialization to the new Account structure
> 
> I wonder if this changes our JSON output enough to break command line scripts and hledger-web clients like MoLe.

Quite possibly. I'm not familiar enough with them to be able to say.

**Comment by simonmichael** [comment]
Thanks for those explanations. 

I appreciate the work and thought and refactorings here, and as usual I expect I'll see the benefits more clearly before long. I'm not quite there yet, so I will contemplate the changes some more. 

I am starting to see some independent bits that we could extract as easier-to-discuss PRs, if we need to.

I'd like to see whether this has reduced complexity/complication overall, or just moved it to another place. I think the balance report and budget report code got simpler while the Account and new AccountBalances/AccountBalance types have got complex. They make perfect sense from an implementation point of view, but I'd like to feel sure they are still sufficiently comprehensible/documented/maintainable. 

(If I'm being over-cautious, well it's no great harm to move a little slower when contemplating long-lasting code changes, especially if we're not obviously fixing something broken.)

Some more eyes would be helpful. I have pinged a random few other devs who have worked on hledger internals, in case they have time to give an opinion on this pr 2360. Here's a new draft of overall change notes:

> dev: Account: store the amounts/balances for multiple periods
> 
> Previously, Account stored amounts (two of them: subaccount-inclusive and subaccount-exclusive)
> representing the balance change in a single period, or the end balance on a single end date.
> Now, it can store these amounts for all the periods in a multi-period report,
> and also, for the pre-report period. These are held in new helper types:
> 
> - AccountBalance: the inclusive amount, exclusive amount, and posting count for a single period
> - AccountBalances: an AccountBalance for each report period and the pre-report period.
> 
> Additionally, Account and AccountBalances have been made polymorphic,
> so things other than single amounts can be stored; eg, the budget report stores
> two AccountBalance values, representing the goal and actual amounts for each period.
> 
> This consolidates and centralises common functionality that was
> previously in multiBalanceReport and budgetReport,
> and should make it easier to create new kinds of multi-period report.
> 
> We think performance is about the same as before, possibly with a small speedup in register report.
> 
> Some other changes have come along with this:
> 
> The logic for displaying zero balance parent accounts has changed slightly.
> Budget reports no longer show them.
> And balance reports' --declared flag now does show them, if they are declared.
> 
> There is a new test for balance --flat --empty,
> showing that implied accounts with no postings are not displayed,
> but accounts with zero balance and actual postings are.
> 
> markAccountBoring has been refactored, rephrasing everything in terms of boringness for clarity.
> 
> mixedAmountApplyValuationAfterSumFromOptsWith's type has been improved, removing an error case.

@Xitian9 are any of those last items independent/extractable ?

Re markAccountBoring: usually I've had a specific meaning of "boring parent accounts" in mind, deriving from "the ones that `ledger balance` combines into the same line as their parent". markAccountBoring used "interesting" and now "boring" in a slightly broader sense including depth-clipped subaccounts, which is fine but I just wanted to mention it in case we find confusion elsewhere.

**Comment by simonmichael** [comment]
PS at times I feel I could contribute by adding or editing code docs, or trying different arrangements/extractions of commits, but I don't quite know how to do it without messing up the attribution of your work. I have been refreshing my git/github/vscode/jujutsu-fu. Perhaps there's some more efficient workflow to be discovered ?

Re AccountBalances: is there a benefit to making this a separate type, rather than keeping its fields in Account ?

**Comment by simonmichael** [comment]
> Some more eyes would be helpful. I have pinged a random few other devs who have worked on hledger internals, in case they have time to give an opinion on this pr 2360. 

And anyone I didn't ping, please feel free to take a look! More eyes helps.

**Comment by Xitian9** [comment]
> Re AccountBalances: is there a benefit to making this a separate type, rather than keeping its fields in Account ?

There are two main benefits:
1. By allowing `Account` to contain different types it enables more flexible use and better type information. The main place that's been used here is that we can contain `These AccountBalance AccountBalance` for balance report, allowing the same data structure to represent the more complicated type of a pair of balances (one actual, the other budget) without the need for writing a new data type and associated code. In theory we could also use it for better constraint of the data structure (for example, store only exclusive balances before we run `sumAccounts`, and then have `sumAccounts :: Account OnlyExclusive -> Account AccountBalances`). I haven't done that here because I don't think it's worthwhile, but if we thought something like this were helpful at some point we could do it.
2. By representing `Account` as a container of `AccountBalance`s, it gives us access to a lot of machinery for free through polymorphic instances. For example, `Account AccountBalance` makes it clear that you can alter the `AccountBalance` through the functor instance, and we don't need to write all the boilerplate code to do the same thing as `fmap` for a monomorphic type. We similarly get the `Foldable` and `Traversable` instances for `AccountBalances`, which saves a lot of cognitive burden and opens up a lot of library functions to us. (We could in theory also implement a `Foldable` and `Traversable` instance for `Account`. I've chosen not to do that because we don't use it anywhere.)

**Comment by simonmichael** [comment]
> There are two main benefits:

Thanks for the explanations, and sorry if I'm re-asking things already answered.

Would you mind rebasing this against master - I think there's one test failing when you do that.

What would you think of the following name/doc tweaks ?
If you'd prefer we discuss this as a followup PR, that's ok too.

```haskell
-- Goals:
--  Use names and descriptions matching the generality/specificity of the types
--  Use distinctive names not easily confusable with each other
--  Reduce ambiguities, eg "balance" implying end balances, and singulars vs plurals

-- | An account within a hierarchy, with references to its parent
-- and subaccounts if any, and with per-report-period data of type a.
-- Only the name is required; the other fields may or may not be present.
data Account a = Account {
   aname            :: AccountName        -- ^ full name
  ,adeclarationinfo :: Maybe AccountDeclarationInfo  -- ^ optional extra info from account directives
  ,asubs            :: [Account a]        -- ^ subaccounts
  ,aparent          :: Maybe (Account a)  -- ^ parent account
  ,aboring          :: Bool               -- ^ used in some reports to indicate elidable accounts
  ,adata            :: PerPeriod a        -- ^ associated data per report period
  } deriving (Generic, Functor)

-- Data values for zero or more report periods, and for the pre-report period.
-- Report periods are assumed to be contiguous, and represented only by start dates
-- (as keys of an IntMap). XXX how does that work, again ?
data PerPeriod a = PerPeriod {
   pppre     :: a
  ,ppperiods :: IM.IntMap a
  } deriving (Eq, Functor, Generic)
-- other names: PeriodData a, PeriodsData a, PeriodDatums a

-- | Data that's useful in "balance" reports:
-- subaccount-exclusive and -inclusive amounts,
-- typically representing either a balance change or an end balance;
-- and a count of postings.
data BalanceData = BalanceData {
   bdexcludingsubs :: MixedAmount
  ,bdincludingsubs :: MixedAmount
  ,bdnumpostings   :: Int
  } deriving (Eq, Generic)
-- other names:  AccountBalance a, AccountBalances a, AccountAmounts a, AccountBalanceData a
```

simonmichael (commented):
-

Xitian9:
Sorry for the delay. I've rebased against master and pushed a quick fix for the failing test. I think there's a better fix involving a bit more work, but I want to avoid making this PR heavier than it already is.

I've been a bit busy, so haven't yet had the chance to create a separate squashed PR. Please let me know if you still feel that's valuable and I'll try to devote time to it when I can.

Those doc tweaks look good to me. I'll incorporate them as a separate commit rather than rebasing, if you don't mind.

simonmichael:
Thanks for the updates. I don't mind merging this as a merge commit if it saves time.

We are close to (ideal) release date. I don't want to cause more delay, but prudence would probably say merge this after release so we have more time to flush out surprises. On the other hand it's passing tests. I could probably go either way, do you have a preference ?

Xitian9:
Yes, definitely wait. I'm in no hurry, and some extra time for testing would be appropriate for a refactoring of the core code like this.

Xitian9:
I've renamed the data constructors and their record names according to your suggestions above. The one exception is I chose one of your alternative names `PeriodData` instead of `PerPeriod`. I'm happy to be overruled on that call.

simonmichael:
This was merged as #2395.

Issue #2141, opened 2023-12-18
timeclock: allow multiple open sessions


simonmichael:
hledger's timeclock format doesn't allow clocking in to multiple named sessions concurrently, at least not within a single timeclock file. I think Ledger does allow this. We'd like to be able to freely clock in and out of different named sessions, tracking them individually. This means clock outs need to be matched with the right open sessions.

simonmichael:
[Related discussion on #2142](https://github.com/simonmichael/hledger/issues/2142#issuecomment-1866857286)

reesmichael1:
I've been working on this, and I have a functioning implementation ready to open a PR. However, I believe this is impossible to implement without breaking some backwards compatibility. Consider this case from test 1 in `timeclock.test`:

```
i 2009/1/1 08:00:00
o 2009/1/1 09:00:00 stuff on checkout record  is ignored
```

I am pairing entries by the account name. If there is no account name on the `o` record, I'm closing the most recently opened `i` record. This works well in every other case. However, as the quoted test shows, we were previously ignoring the account name on the checkout record, which no longer works.

How should we handle this case? I'm happy to open my PR if the discussion would be better continued there.

simonmichael:
Our [time clock format spec](https://hledger.org/1.42/hledger.html#timeclock) doesn’t mention extra text on the o records at all; it was added for ledger compatibility and not documented I guess. 

The proposal will make life harder for some timeclock users migrating from ledger to hledger. But adds power that some hledger timeclock users want. I think it’s ok to change it.

reesmichael1:
Makes sense to me! I'll clean up a few things and open the PR soon.

simonmichael:
This was added in 1.43.

schoettl:
Hi there, I just wanted to note that I need to use `--timeclock-old`. Without, my two timeclock files fail to parse already at line 93 of 6754 and 205 of 17120 respectively.

I don't really understand why it fails because I always clock-out after an clock-in without an account name. So there should be no nested sessions.

With this two scripts, I make sure that `i` and `o` lines are really always alternating and that `o` really never has an account name.

```
awk '{curr = $1; if (curr == prev) {print "error in", NR; exit 1}; prev = curr;}' timelog.timeclock
awk '$1 == "o" && $4 != "" { print "error in", NR; exit 1; }' timelog.timeclock
```

simonmichael:
Shrinking your data should reveal the cause..

Issue #2142, opened 2023-12-18
timeclock: better support for collecting data in multiple files ?


simonmichael:
This came up in a mastodon/matrix chat:

nobodyinperson:
> The merge conflicts I encountered with timeclock came up because my (emphasis on my) checkin/out commands appended to the same file. I would frequently checkin and out from different devices (e.g. checkout from phone in the elevator down, no reception there), but then I couldn't get the newest additions of the day due to lack of internet. Local checkout commit conflicts with all new commits of the day because all just append to the same file. Same with timewarrior. Randomizing locations in the file could reduce but not fully prevent conflicts.
> 
> The basic append workflow works if one has internet always everywhere - a  weird requirement if one has a distributed offline system like git underneath.
> 
> IIRC, timeclock also can't track simultaneous things, same as timewarrior.

sm
> with current hledger, it seems you can only have one open session per file. Multiple open sessions split across files works.
> 
> secondly, I see that you can't have a clockout in a file without having a preceding clockin in that same file
> 
> thirdly I think the clockout account is ignored, it will close any preceding clockin in that file (same as lack of multi-session support)
> 
> fourthly I see how updating the same file in two repos and then merging gives a git conflict (my brain was seeing it as an "obvious merge" that should work...)
> 
> I can do a merge manually by sorting both files: sort a.timeclock b.timeclock | sponge a.timeclock, in which case I don't get the commit history from b's repo

The multiple sessions limitation is noted as #2141. 

This issue is more about making hledger timeclock work more smoothly when you are logging time in multiple files / on multiple devices. Including clocking in on one and device and clocking out on another. There are two sub-cases: 

1. Not using version control, or only one file using version control
2. All files using version control, and a wish to preserve and merge the commit histories from all devices
3.

jamescooke:
Could you help clarify the following?

> hledger won't read the multiple timeclock files unless you merge them first

However, I've logged all my work in 2023 with one timeclock file per week (for example `2023_w01_20230102.timeclock`). Those are unified in a journal file `2023.jourmal` which has:

```hledger
include 2023_*.timeclock
```

I can then run `hledger -f 2023.journal bal` just fine. Is this what you mean by "merge them first"?

simonmichael:
Yup, I meant with "Clockins/clockouts split across files". I think you aren't clocking in in one file and clocking out in another. As long as you keep clockins/outs nicely paired in each file, including them works fine as you say.

simonmichael:
[...]

PR #2362, opened 2025-04-01
feat: timeclock: Add support for multiple clocked in sessions (#2141)

reesmichael1:
As requested in #2141, this adds support for having multiple sessions clocked in. These are paired by account name if given on the out entry, and otherwise an out closes the most recent in entry.

Note that this breaks some backwards compatibility, in that we previously ignored the description on the clock out entry.

(Also, I tried to update the documentation with `Shake manuals`, but that seems to have made lots of other changes...not sure if I didn't do something correctly or if those files were just out of date.)

### DIFF
--- a/hledger-lib/Hledger/Data/Timeclock.hs
+++ b/hledger-lib/Hledger/Data/Timeclock.hs
@@ -10,10 +10,12 @@ converted to 'Transactions' and queried like a ledger.
 
 module Hledger.Data.Timeclock (
    timeclockEntriesToTransactions
+  ,timeclockEntriesToTransactionsSingle
   ,tests_Timeclock
 )
 where
 
+import Data.List (partition, sortBy, uncons)
 import Data.Maybe (fromMaybe)
 import qualified Data.Text as T
 import Data.Time.Calendar (addDays)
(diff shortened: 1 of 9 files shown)


simonmichael:
Don't bother updating the generated man pages; I usually do that before release with `./Shake manuals -c`. Just updating doc source files like hledger.m4.md is fine.

You're right that this could break for existing hledger users who for whatever reason have text on their o records. Ie, it doesn't only affect new users migrating from Ledger as I was thinking. 

In such cases it's good to have an easy workaround ready. I wonder if we need to provide a flag to restore the old behaviour temporarily. Or we could say in such cases "downgrade hledger", I guess. But that's unpleasant.

hledger-lib/Hledger/Data/Timeclock.hs
```diff
@@ -47,47 +48,88 @@ instance Read TimeclockCode where
...
+
+data Sessions = Sessions
+    { completed :: [Session],
+      active :: [TimeclockEntry]
+    }
+
+-- Find the relevant clockin in the actives list that should be paired with this clockout.
+-- If there is a session that has the same account name, then use that.
+-- Otherwise, if there is an active anonymous session, use that.
+-- Otherwise, raise an error.
```
simonmichael, line 64: Nice noting of the error raising, let's do that in pairClockEntries' haddock as well.

simonmichael: And timeclockEntriesToTransactions.

hledger/hledger.m4.md
```diff
@@ -4606,12 +4606,19 @@ i 2015/03/30 09:00:00 some account  optional description after 2 spaces ; option
...
+o 2015/04/02 15:00:00 another:account
 ```
 
 hledger treats each clock-in/clock-out pair as a transaction posting
-some number of hours to an account. Or if the session spans more than
-one day, it is split into several transactions, one for each day. For
-the above time log, `hledger print` generates these journal entries:
+some number of hours to an account. Entries are paired by the account
+name if the same name is given for a clock-in/clock-out pair. If no 
+name is given for a clock-out, then it is paired with the most recently
```
simonmichael, line 4618: Typo: most recent

hledger/test/cli/addons.test
```diff
@@ -1,60 +0,0 @@
-# * addon command subsystem tests
```
simonmichael, line 1: Accidental file deletion ?

simonmichael (commented):
Looking good.

simonmichael:
> In such cases it's good to have an easy workaround ready. I wonder if we need to provide a flag to restore the old behaviour temporarily.

Eg, under "General input/data transformation flags", or perhaps better, a hidden flag: --timeclock-single-session or --timeclock-old ?

Also, do you know if this PR will also resolve/help #2142 ? I don't recall that one well.

reesmichael1:
Thanks for the review! I'll clean those things up (not sure what happened with that file deletion...). 

I like the idea of the hidden flag to restore the old behavior. I'll get that added as well. Is it worth adding that flag to the manual (e.g., `"hledger versions prior to 1.33 parsed timeclocks differently [blah blah blah]. To restore that behavior, use --timeclock-old.`)?

I think this is a good first step towards #2142 but haven't tested applying it towards that myself. I could continue with that next. (Also, while working on this, I also found two unrelated bugs with timeclock parsing that I'll open separate issues for /should be easy fixes.)

simonmichael:
> Is it worth adding that flag to the manual

I wouldn't even bother at this stage... the person who'll need it likely won't find it there and it's clutter for most, let's hope it's never needed.

reesmichael1:
I restored the deleted file and (although I'm guessing it probably wasn't necessary) rolled back the manuals to match the ones on `master`.

simonmichael:
When you get a chance, rewrite the commits without changes to the generated manuals and force push, to make things clearer.

reesmichael1:
Sure thing! Do you want just one commit or leave as two commits, just without the manuals? (My default is always one commit but some projects prefer to keep them separate.)

simonmichael:
One is fine! Thanks.

simonmichael:
I may regret not documenting the flag but press on! Looks great, thanks!

Issue #2367, opened 2025-04-09
GHC 9.12 support: unwanted "uncaught exception" output


simonmichael:
hledger in master builds cleanly with GHC 9.12, eg using stack912.yaml.

But error output has changed again. I'm seeing that `errorWithoutStackTrace` now also prints

    PROGNAME: Uncaught exception ghc-internal:GHC.Internal.Exception.ErrorCall: 

(when compiled; not in GHCI).

We need to suppress that output.
[GHC 9.12 release notes](https://downloads.haskell.org/ghc/9.12.1/docs/users_guide/9.12.1-notes.html) say some unclear things about call stacks and exceptions changes in base 4.21.

simonmichael:
More context:

Traditional `error` output (which in hledger we call via `error'`):

```
$ hledger-ghc98 nosuchcommand
hledger-ghc98: Error: command nosuchcommand is not recognized, run with no command to see a list

```

GHC 9.10/base 4.20 added a stack trace by default, so we switched to `errorWithoutStackTrace`. We couldn't suppress the two extra newlines though:

```
$ hledger-ghc910 nosuchcommand
hledger-ghc910: Error: command nosuchcommand is not recognized, run with no command to see a list



```

GHC 9.12/base 4.21 fixed the two newlines, but adds this "uncaught exception" output:

```
$ hledger-ghc912 nosuchcommand
hledger-ghc912: Uncaught exception ghc-internal:GHC.Internal.Exception.ErrorCall:

Error: command nosuchcommand is not recognized, run with no command to see a list

```

simonmichael:
#2369 aims to fix this.

simonmichael:
Fixed in master.

PR #2369, opened 2025-04-10
consistent console error output independent of GHC version

simonmichael:
- **fix:cli,ui,web: consistent console error output independent of GHC version [#2367]**

Hledger.Utils.IO helpers have been updated and new ones have been
added (exitOnExceptions, exitWithError) to allow consistent display of
program errors whether compiled with GHC <9.10, GHC 9.10, or GHC >9.10.
The trailing newlines added by GHC 9.10 are gone,
and so is the "uncaught exception" output added by GHC 9.12.

### DIFF
--- a/hledger-lib/Hledger/Read.hs
+++ b/hledger-lib/Hledger/Read.hs
@@ -143,10 +143,9 @@ import Data.Time (Day)
 import Safe (headDef, headMay)
 import System.Directory (doesFileExist, getHomeDirectory)
 import System.Environment (getEnv)
-import System.Exit (exitFailure)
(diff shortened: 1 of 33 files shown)


simonmichael:
This looks ok in my testing so far, but I'm going to use it for a while. Any extra testing appreciated. I believe there's still some inconsistent error output from hledger, eg `hledger print -f nosuchfile` doesn't call error' and probably should.

simonmichael:
Some this may be working around GHC bugs, so there's a chance this will break again with GHC 9.12.3.

PR #2373, opened 2025-04-19
Apply amt:/cur: queries after valuation, not before (to fix boolean queries)

### LINKED ISSUE #2371: boolean queries can give wrong results
I found an example that's easy to reproduce:
```
$ hledger -f examples/sample2.journal reg expr:'(checking and amt:>0) or credit'
2025-01-01 starting balances    assets:bank:checking   1000.00 USD   1000.00 USD
                                li:credit card         -400.00 USD    600.00 USD
2025-01-01 pay rent             assets:bank:checking   -800.00 USD   -200.00 USD
2025-01-02 salary               assets:bank:checking   1000.00 USD    800.00 USD
2025-01-03 pay half of credi..  assets:bank:checking   -200.00 USD    600.00 USD
                                li:credit card          200.00 USD    800.00 USD
2025-01-04 shopping             assets:bank:checking   -250.00 USD    550.00 USD
```

This should not be showing the negative checking postings. I expected to see:
```
2025-01-01 starting balances    assets:bank:checking   1000.00 USD   1000.00 USD
                                li:credit card         -400.00 USD    600.00 USD
2025-01-02 salary               assets:bank:checking   1000.00 USD   1600.00 USD
2025-01-03 pay half of credi..  li:credit card          200.00 USD   1800.00 USD
```

The problem is with the [filterQuery](https://github.com/simonmichael/hledger/blob/f54677f5c0be8bdbb4dc6dc8a44288c57580fe18/hledger-lib/Hledger/Query.hs#L583-L593) calls [in journalValueAndFilterPostingsWith](https://github.com/simonmichael/hledger/blob/f54677f5c0be8bdbb4dc6dc8a44288c57580fe18/hledger-lib/Hledger/Reports/ReportOptions.hs#L603-L604). It doesn't know how to handle a non-trivial boolean query, and ends up altering it:

```haskell
ghci> filterQuery queryIsAmtOrSym $ Or [ And [ Acct ( toRegex' "checking" ) , Amt Gt 0 ] , Acct ( toRegex' "credit" ) ]
Any
ghci> filterQuery (not.queryIsAmtOrSym) $ Or [ And [ Acct ( toRegex' "checking" ) , Amt Gt 0 ] , Acct ( toRegex' "credit" ) ]
Or
    [ Acct
        ( Regexp "checking" )
    , Acct
        ( Regexp "credit" )
    ]
```

simonmichael:
This is a draft for discussion. It reverses #1625 (matching amt:/cur: before valuation) to fix #2371 (some boolean queries not working). 

Here's the current description:

> Until now, at least since #1625 (2021), cur: and amt: matched the
> pre-cost/value-conversion commodity symbols and amounts.
> 
> To do that, we were separating the cur:/amt: parts from the rest of
> the report query. But with boolean queries #1989 (2023), it's not
> possible to do that reliably without changing the query's meaning (?).
> 
> So we no longer do that, which means that cur: and amt: now match the
> post valuation amounts. This is a breaking change for reports:
> 
> - register and all balance reports (using journalValueAndFilterPostingsWith)
>   and aregister, hledger-ui & hledger-web register views (using accountTransactionsReport)
>   no longer pre-filter by amt:/cur:. amt:/cur: with these reports now
>   match the post-valuation amounts and commodity symbols, not the
>   pre-valuation ones.
> 
> - print, codes, descriptions, notes, hledger-web journal view (using entriesReport)
>   now apply all queries after valuation, not before it
>   (for consistency with the above).

- [Old and new amt:/cur: tests](https://github.com/simonmichael/hledger/blob/eeb701140df47260125cf7d26bd746482f89abd8/hledger/test/journal/amounts-and-commodities.test#L300-L361)
- [fixed boolean query test](https://github.com/simonmichael/hledger/blob/eeb701140df47260125cf7d26bd746482f89abd8/hledger/test/query-expr.test#L161-L166)

This fixes boolean queries, or at least some of them. There are other cases where reports transform the report query (eg with filterQuery), which have not been fully audited. A common case is removing the depth: limit from the query. It's not yet clear if these cases can also break boolean queries - more auditing, testing and producing repros will help.

"Early amt:/cur:" (before valuation) has been the status quo for years. So this needs more thought and ideas about whether and how to preserve compatibility. As it stands, this would break some existing reports, and might make those reports hard or impossible to produce at all (testing needed).

Some options to consider ?

1. Try to match with amt:/cur: both before and after valuation.
   (We do this with acct:, trying to match both before and after account name aliasing.)
   Even if advisable, I'm not sure how feasible it would be to make this robust.
   Eg, where A amounts have been converted to value in B:
   - `reg -V cur:A` matches the B amounts (because they were formerly A)
   - `reg -V cur:B` also matches the B amounts
   - `reg -V expr:'(cur:A AND cash) OR checking'` is still problematic; cur:A is part of a more complex query, can't extract and prefilter with it

2. Keep the old behaviour for simple queries, but disable it when a problematic boolean query is detected (one with amt:/cur: in subexpressions).
   Inconsistent behaviour would probably be bad.
   Eg: 
   - `reg -V expr:'cur:A AND cash'` cur:A is in a top level AND, can extract it and prefilter with it (and exclude it from the postfilter so it doesn't exclude the B result)
   - `reg -V expr:'(cur:A AND cash) OR checking'` cur:A is in a subquery, can't extract/prefilter with it; must leave it in the postfilter (it will exclude the B result)

3. Disallow amt:/cur: in boolean queries or subqueries.
   (We already disallow date: in OR subexpressions.)
   This might be over-restrictive.
   Eg:
   - `reg -V expr:'cur:A AND cash'` "Error: amt: and cur: are not supported in boolean queries"
   - `reg -V expr:'(cur:A AND cash) OR checking'` "Error: amt: and cur: must be part of a top-level AND query"
   - `reg -V cur:A cash` works as before

4. Think harder about 
   - more safe/limited/reasonable ways to extract the amt:/cur: queries for pre-filtering, preserving the old behaviour without altering boolean queries.
   - why those reports were doing that pre-filtering, and if it's really needed. Reports often make a variant of the user's query, eg removing a depth limit, or changing the period in order to calculate starting balances, while preserving the rest of the query as-is.
   - how to ease the transition, if behaviour is changed

### DIFF
--- a/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
+++ b/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
@@ -103,10 +103,12 @@ accountTransactionsReport rspec@ReportSpec{_rsReportOpts=ropts} j thisacctq = it
     -- Queries on currency or amount are also ignored at this stage; they are handled earlier, before valuation.
     reportq = simplifyQuery $ And [aregisterq, periodq]
       where
-        aregisterq = filterQuery (not . queryIsCurOrAmt) . filterQuery (not . queryIsDepth) $ _rsQuery rspec
+        -- #2371: don't remove amt:/cur: from the register query, it breaks boolean queries
+        -- aregisterq = filterQuery (not . queryIsCurOrAmt) . filterQuery (not . queryIsDepth) $ _rsQuery rspec
+        aregisterq = filterQuery (not . queryIsDepth) $ _rsQuery rspec
         periodq = Date . periodAsDateSpan $ period_ ropts
-    amtq = filterQuery queryIsCurOrAmt $ _rsQuery rspec
-    queryIsCurOrAmt q = queryIsSym q || queryIsAmt q
+    -- amtq = filterQuery queryIsCurOrAmt $ _rsQuery rspec
+    -- queryIsCurOrAmt q = queryIsSym q || queryIsAmt q
     wd = whichDate ropts
 
     -- Note that within this function, we are only allowed limited
@@ -129,8 +131,13 @@ accountTransactionsReport rspec@ReportSpec{_rsReportOpts=ropts} j thisacctq = it
         -- maybe convert these transactions to cost or value
         . journalApplyValuationFromOpts rspec
         . traceOrLogAtWith 5 (("ts2:\n"++).pshowTransactions.jtxns)
+
+        -- #2371: don't try to prefilter (before valuation) with just the amt:/cur: from the register query; it breaks with boolean queries.
+        -- Use the whole query.
+        -- . (if queryIsNull reportq then id else filterJournalAmounts reportq)
         -- apply any cur: or amt: filters in reportq
-        . (if queryIsNull amtq then id else filterJournalAmounts amtq)
+        -- . (if queryIsNull amtq then id else filterJournalAmounts amtq)
+
         -- only consider transactions which match thisacctq (possibly excluding postings
         -- which are not real or have the wrong status)
         . traceOrLogAt 3 ("thisacctq: "++show thisacctq)
--- a/hledger-lib/Hledger/Reports/EntriesReport.hs
+++ b/hledger-lib/Hledger/Reports/EntriesReport.hs
@@ -38,8 +38,9 @@ entriesReport rspec@ReportSpec{_rsReportOpts=ropts} =
       sortBy (comparing $ transactionDateFn ropts)
     . map  (if invert_ ropts then transactionNegate else id)
     . jtxns
-    . journalApplyValuationFromOpts (setDefaultConversionOp NoConversionOp rspec)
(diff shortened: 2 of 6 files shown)


simonmichael:
I forgot about this and haven't been testing it further. But I probably wouldn't have noticed any problem. I think that people will notice only if it's shipped in a release, probably after a few months' delay.

To do:
- Add a few more tests demonstrating the changed behaviour of amt: and cur:
- Post these examples widely (mail list, chat, mastodon..) and ask for comment.

simonmichael:
Or we could:

5. Change cur: and amt: to work after valuation, so they can be used safely in boolean queries. But also add ocur: and oamt: (origcur: and origamt: ?) which work the old way, ie before valuation, and are not allowed in boolean queries. This would give an escape hatch when someone finds the new behaviour breaking their setup.

simonmichael:
Abandoned in favour of #2387, which also fixes #2371.

PR #2378, opened 2025-04-27
Fix: "balanced" check is enabled only with "check" command

### LINKED ISSUE #2377: A "balanced" argument with any command enables the balanced check
Noting a small issue:

The [`balanced` check](https://hledger.org/dev/hledger.html#strict-checks), which disallows implicit-cost conversion transactions, is normally enabled only in strict mode. But hledger also enables it if a "balanced" argument appears in the command line; there is some logic in rawOptsToInputOpts that does this to accommodate the `hledger check` command. This was safe enough with the check's original name, `balancednoautoconversion`, but since then the check was renamed to `balanced`. It means that since hledger 1.31 (2023), if someone runs any command with "balanced" as an argument, they are also enabling this check.

samahri:
fixes #2377


simonmichael:
Ha nice one ! 😅

Issue #1322, opened 2020-08-05
Ledger format API endpoint(s)


repomaa:
I think adding API endpoints or at least one for adding entries in the ledger text format would be very helpful. I'll try to explain why:

I'm currently writing a simplified web-frontend to use for quickly entering transactions into ledger. From the user perspective this should be as easy to use as possible. My current main grievance with the hledger-web json api is the amounts. Having to specify it using decimal mantissa, commodity formatting options etc. requires me to basically duplicate the functionality already done in hledger cli where the parsing is implemented. A quick example. I might have defined commodities like this: `1.000,00 €` and `1.000,0000 "SOME ETF"`. Now when a user enters an amount to a transaction, I don't want to force any specific format on them as I can't know what they have defined (or not defined) in their journals. So what I'm left with is having to either make assumptions of the format or making it configurable in the frontend.

From my naive point of view this could be avoided if the server would just check the `Content-Type` and/or `Accept` header of the incoming request and use the current json format for `application/json` (or unset for backwards compatibility) and run it through the hledger cli code when it's `text/plain`. So, and correct me if I'm wrong, I think this is fairly low hanging fruit and would help prevent duplication in frontend(s) and server.

repomaa:
In a perfect world a json format like the following or similar would be even better i think:

``` json
{
  "description": string,
  "comment": maybe(string),
  "postings": array(posting)
}
```

with posting being something like:

``` json
{
  "account": string,
  "comment": maybe(string),
  "amount": maybe(string),
}
```

(Having amount being nilable would allow entering transactions that auto-balance, like in the text format)

simonmichael:
This sounds very sensible but also quite like what hledger-web's add form does - I don't remember all the routes involved here but can't you use the one the add form uses ?

simonmichael:
Checking.. here's the [routes](https://github.com/simonmichael/hledger/blob/master/hledger-web/config/routes) file; /add is what the add form posts to. 
Here's [add-form.hamlet](https://github.com/simonmichael/hledger/blob/master/hledger-web/templates/add-form.hamlet) showing the form fields.
Here's [postAddR](https://github.com/simonmichael/hledger/blob/master/hledger-web/Hledger/Web/Handler/AddR.hs#L30), route handler. 
They use the form widget [AddForm.hs](https://github.com/simonmichael/hledger/blob/master/hledger-web/Hledger/Web/Widget/AddForm.hs) to do validation.

repomaa:
[...]

Issue #1195, opened 2020-02-22
Very large numbers in API response for inferred transaction prices


matteodelabre:
When using hledger’s [inferring of transaction prices](https://hledger.org/journal.html#transaction-prices) for unit prices that cannot be exactly represented in decimal, the values that show up in API responses can be very large (up to 255 decimal digits). Consider for example the following transaction:

```
2042/02/22 Problematic transaction
    acc1    10.20 USD
    acc2    10.20 USD
    acc3   -13.97 CAD
```

Running `hledger-web` on this transaction, the following response is produced by the API (excerpt):

```json
[
  {
    "tpostings": [
      {
        "pamount": [
          {
            "aprice": {
              "tag": "UnitPrice",
              "contents": {
                "aprice": null,
                "acommodity": "CAD",
                "aquantity": {
                  "decimalPlaces": 255,
                  "decimalMantissa": 684803921568627450980392156862745098039215686274509803921568627450980392156862745098039215686274509803921568627450980392156862745098039215686274509803921568627450980392156862745098039215686274509803921568627450980392156862745098039215686274509803921568627
                },
                "aismultiplier": false
              }
            },
            "acommodity": "USD",
            "aquantity": {
              "decimalPlaces": 2,
              "decimalMantissa": 1020
            },
            "aismultiplier": false
          }
        ]
      }
    ]
  }
]
````

Although this is technically valid JSON, it is far outside the range of usual numeric datatypes in most languages. (An unsigned long value would only allow for up to 19 decimal digits.) _This notably breaks MoLe_ because it tries to parse all numbers as Java longs.

Do you think this is better addressed in here or that clients should be prepared to handle very large numbers?

---

* Version: 1.16.2
* Platform: Arch Linux

simonmichael:
Ooh, good point. I expect that's an irrational number or repeating decimal. We don't round these based on display precision, as the UIs do.

We want the JSON to be usable by others. We'd like to minimise imprecision. Ideally we'd like hledger -> JSON -> hledger roundtripping to not be lossy.

What trade off would you suggest ?

matteodelabre:
Right, it's definitely a repeating decimal—and if I'm not mistaken, we should never get irrational numbers because the unit price is computed by dividing two decimal numbers. Because of this, an option would be to give the unit price as a fraction. For example, it could be something like this for the previous example:

```json
"aprice": {
  "tag": "UnitPrice",
  "contents": {
    "aprice": null,
    "acommodity": "CAD",
[...]

Issue #1005, opened 2019-04-07
Add API for modifying existing postings


real-dam:
As mentioned in #316, it would be great if there was an API for modifying existing postings. Perhaps something like `PUT /add?id=42`?

Not sure how to protect from overwriting changes from multiple parties - for example a direct modification of the underlying ledger file may change the transaction ordering and thus IDs. Perhaps some form of hash instead of plain ordering number? (MoLe uses a hash over the transaction's contents including its ID for a similar purpose)

(/accounts, /transactions and /add APIs work great so far!)

simonmichael:
Glad to hear those are working. 

The current transaction id is just sequence based as you note - ids should be re-read after any change to existing transactions. I believe it's possible - in practice, not just in theory - to have multiple identical transactions, even on the same day. Eg transactions sometimes get retried or duplicated. Because of this I think some kind of ordering or file position based information will always be needed in the transaction id, so I'm not sure if a more complex hash really helps much.

New hledger-lib functionality is needed to support this long-desired feature: the ability to modify or delete an existing transaction, referenced by id, not just in memory but also in the journal file on disk, without disturbing the file content before and after it.

real-dam:
Oh, sure. Identical transactions happen and should work.
That's why I thought about using a hash for unique identification. A hash calculated over the content of the transaction plus the sequential number that is now served as id. This way moving a transaction around would change its hash, and several identical transactions would still have unique hashes due to their different positions in the file.

Of course this only solves the problem of unique identification.

As for the problem of consistency while modifying the file, I was hoping there is some kind of serialization already for the case when several browser requests to `/add` come at the same time :)

real-dam:
Short answer: parallel users.

Imagine two MoLe or web app users had retrieved the transaction list with 10 transactions. Not long later the first user requests deletion of transaction 5. Three seconds later the second user requests deletion of transaction 6. The end result would be removal of transactions 5 and **7** in the initial numbering. 

Perhaps an UUIDs would also work. But then hledger would need some kind of storage.

simonmichael:
I that adding more information to the hash (besides file position/sequence no.) is not a full solution, because it doesn't help in the case of identical transactions on the same day (in which case, literally the only thing different is their sequence). 

I think a full solution would be (eg) always doing these as an atomic transaction: 1. read the latest ids 2. modify the entries you want 3. write the updated journal. 

[...]

PR #2390, opened 2025-05-23
Web: OpenAPI doc improvements

bsima:
This PR serves the openapi.yaml doc as an endpoint at `openapi.json` and improves the docs therein.

I'm creating this because Open-WebUI recently created these [tool servers](https://docs.openwebui.com/openapi-servers/), which are like MCP but uses OpenAPI instead of yet another API format. 

Using this, I was able to interact with my ledger via claude 4 sonnet and gpt-4o, unfortunately doing anything useful is not possible yet because the responses are way too large. The API responses need to be filtered more precisely to fit in the LLM context windows. I will investigate this filtering next.


simonmichael:
Looks great! Only one note, hledger-web is the only correct spelling of this app (not Hledger Web). It's not a blocker, I'll merge soon.

simonmichael:
(After you remove the WIP: prefix I guess)

bsima:
I fixed the naming and force pushed, this can be merged now.

I'm not seeing any way to filter API queries at first look. There seem to be a few issues:

1. The endpoints don't have any query params for filtering on the output
2. The returned representations are very verbose
3. Not all hledger functionality is implemented in the API

With this in mind I think it is a larger project to make hledger-web usable as an API, but its something I'd like to see accomplished. I'll work on API design over the weekend and try to come up with something. There are a few API-related tickets that maybe we could roll into a redesign (#1322, #1005, #1195).

simonmichael:
Thank you very much!

You are right; the web API is rudimentary, and little used so far. (Only by the MoLe android app, AFAIK.)

PR #2395, opened 2025-06-03
!dev: lib: Allow Account to store date-indexed balances.

Xitian9:
This upgrades Account to enable it to store a multiperiod balance, with a separate balance for each date period. This enables it do the hard work in MultiBalanceReport.

Some new types are created to enable convenient operation of accounts.
- `BalanceData` is a type which stores an exclusive balance, inclusive balance, and number of postings. This was previously directly stored in Account, but is now factored into a separate data type.
- `PeriodData` is a container which stores date-indexed data, as well as pre-period data. In post cases, this represents the report spans, along with the historical data.
- Account becomes polymorphic, allowing customisation of the type of data it stores. This will usually be `BalanceData`, but in `BudgetReport` it can use `These BalanceData BalanceData` to store both actuals and budgets in the same structure. The data structure changes to contain a `PeriodData`, allowing multiperiod accounts.

Some minor changes are made to behaviour for consistency:
- --declared treats parent accounts consistently.
- --flat --empty ensures that implied accounts with no postings are not displayed, but accounts with zero balance and actual postings are.

This is a squashed version of #2360.


simonmichael:
Great, let's rebase this for merge, superseding #2360. 

I tried to resolve the conflict through github ui with 0119843984 but I don't like the result, could you force-push a clean rebase instead.

Xitian9:
Rebased.

simonmichael:
Thank you very much ! Merged.

simonmichael:
@Xitian9, the use of Data.Foldable1 requires base 4.18/ghc 9.6. Until now we have supported back to base 4.14/ghc 8.10.7 (though I was ready to bump this at least to ghc 9.0). Bumping to ghc 9.6 would be very pleasant, but does rule out quite a few ghc versions released in 2021-2023 (see https://wiki.haskell.org/Base_package). If you see an easy quick fix I would take it, otherwise I'll require ghc 9.6.

simonmichael:
It looks like I can just omit that Foldable1 instance, or use CPP.

Xitian9:
That sounds reasonable to me. I think I added those instances because I
could, but if I recall correctly they're not used anywhere, but I can
confirm when I'm at my computer.

simonmichael:
I did that, but after hitting several more paper cuts around the code base - I believe it's time to require GHC >=9.6 / base >=4.18 !

aragaer:
Not sure if this change is the culprit, but MoLe crashes expecting to find aebalance/aibalance fields in /accounts. These seem to be declared to be required in openapi.yaml, but I'm not a web-person, so again I am not sure if that means anything.

simonmichael:
Thanks for the report @aragaer. Is there a MoLe tracker where it could be reported ?

aragaer:
Can't find anything other than author's e-mail. I think once the new API version is finalized it would make sense to ask the author to implement support for that.

simonmichael:
Let's do that once #2459 is resolved.

Issue #2396, opened 2025-06-03
1.43: print cur: has broken


simonmichael:
```
2025-06-03
 a  1 A
 a  1 B
 b
```

Old behavior:
```
$ hledger -f /tmp/test.journal bal cur:A -N
                 1 A  a
                -1 A  b
```

New behavior:
```
$ hledger bal -f /tmp/test.journal cur:A -N
                 1 A  a
                -1 A
                -1 B  b
```

_Originally posted by @aragaer in https://github.com/simonmichael/hledger/issues/2387#issuecomment-2936241641_

Yikes! You found a hole in our tests. Thanks @aragaer!

simonmichael:
(Please claim your https://hledger.org/REGRESSIONS.html#regression-finder-bounty when ready.)

aragaer:
Actually it also broke `amt:` queries. Using the same journal file:
```
$ hledger bal -f /tmp/test.journal 'amt:>0' -N
                 1 A
                 1 B  a
                 1 B  b
$ hledger bal -f /tmp/test.journal 'amt:<0' -N
                -1 B  a
                -1 A
                -1 B  b
```

simonmichael:
I would expect amt: and cur: to pass or fail together, but I haven't been able to reproduce your amt: failure, strangely.
```
$ cat a.j
2025-06-03
  a  1 A
  a  1 B
  b
$ hledger --version
hledger 1.43-g0d7c251a6-20250601, mac-aarch64
$ hledger -f a.j bal -N 'amt:>0'
                 1 A
                 1 B  a
```

simonmichael:
Notes:

1.42: handles some boolean queries involving amt:/cur: wrongly:
```haskell
-- | Like 'journalValueAndFilterPostings', but takes a 'PriceOracle' as an argument.
journalValueAndFilterPostingsWith :: ReportSpec -> Journal -> PriceOracle -> Journal
journalValueAndFilterPostingsWith rspec@ReportSpec{_rsQuery=q, _rsReportOpts=ropts} j =
    -- Filter by the remainder of the query
      filterJournal reportq
    -- Apply valuation and costing
    . journalApplyValuationFromOptsWith rspec
    -- Filter by amount and currency, so it matches pre-valuation/costing
      (if queryIsNull amtsymq then j else filterJournalAmounts amtsymq j)
  where
    -- with -r, replace each posting with its sibling postings
[...]

PR #2398, opened 2025-06-04
fix cur: filtering of multicommodity amounts in posting-based reports

simonmichael:
This is a fix for #2396. It adds a filterJournalAmounts pass to filter out any unmatched amounts from multicommodity amounts.

- **;dev: query-expr.test: a simpler test [# 2371]**
- **;lib: note an issue with filterQuery**
- **lib:Hledger.Query: queryIsAmtOrSym**
- **dev: refactor journalValueAndFilterPostingsWith a bit [# 2396]**
- **fix: cur: in posting-based reports filters multicommodity amounts again [# 2396]**


simonmichael:
This was merged.

Issue #1975, opened 2023-01-06
Auto postings: Allow referring to the matched account [💰$150]


lukasbestle:
## Feature

In an auto posting rule, it should be possible to refer to the account that matched the rule.

There are two common use cases for this:

- It allows to modify the amount of the original posting (see my example below).
- Another different use case I've seen is for budgeting. E.g. you could use `(Budget:$account)` to create virtual postings that contain the account name of the original expense.

## Prior work

`ledger` supports the special [`$account` string](https://www.ledger-cli.org/3.0/doc/ledger3.html#Referring-to-the-matching-posting_0027s-account) for that. This string can be used anywhere within the account part of the auto posting(s) and will automatically be replaced with the original account name.

## Use case description

For tax reporting (German VAT "Umsatzsteuer" in my case), I have built auto posting rules that react on a `tax` tag (defined directly in the transaction or as a default in the account definition). A simple transaction could look like this:

```
2023-01-06 * A client
  Income:Revenue  -1190 €
    ; tax: 19%
  Assets:Bank
```

The client pays me 1000 € + 19 % VAT = 1190 €. But of course the VAT is not really income, so it needs to be separated. That's where the auto posting rule comes in (using the special `$account` string that is supported by `ledger`):

```
= acct:Income tag:tax=19%
  $account        *-0,15966387 ; 1-100/119
  Liabilities:VAT  *0,15966387 ; 1-100/119
```

The expected generated result is this:

```
2023-01-06 * A client
  Income:Revenue   -1000 €
  Liabilities:VAT  -190 €
  Assets:Bank
```

In this example I could have created a separate auto posting rule for each income account (there are not that many that would apply). But the issue gets more pronounced with expense accounts. There are *a lot* of expense accounts and it's not viable to create a separate auto posting rule for each one. So having a way to refer to the matched account is really vital.

## Alternatives?

Maybe there is another way to implement the auto posting rule so that it modifies the amount of the original posting. If you have an idea, I'd be very grateful to hear it.

## Bounty

I am ready to pay $20 to the person who implements this feature. Should multiple people be involved, the bounty will go to whoever's contribution is largest. The bounty will be due on a stable hledger release that includes this feature.

The hledger project will [round up the bounty to $150](https://github.com/simonmichael/hledger/issues/1975#issuecomment-2967722409).

lukasbestle:
This would also contribute to increased ledger file support (https://github.com/simonmichael/hledger/issues/1962).

[...]

PR #2412, opened 2025-06-18
imp: lib: Allow special string %acct in auto-posting rules (#1975)

Xitian9:
This allows using the special string `%acct` in auto posting rules. When run, this will be substituted with the account name of the matched posting.

This is a minimal implementation of #1975. More bells and whistles can be added if we decide what features exactly we want to support here.

We may want to use the string `%(acct)` or `%(account)` instead of `%acct`.


simonmichael:
Thanks for this. I have seen the %(...) syntax somewhere, but do we use it anywhere in hledger ? %account seems better to me.

Is the example in the test correct ? 19% of 1190 should be 226.10 I though.

Xitian9:
Happy for you to make the call on exactly which format to use. The
advantage of the brackets is that strings like `%accountfoo` are hard to
parse from a human perspective, while `%(account)foo` makes clear where the
special string lies, bit this is clearly not a common use.

My understanding of the example (taken from OP) is that you are selling an
item costing 1000, and 19% VAT is applied, bringing the total amount of
money you receive to 1190. However, you owe 190 of that to the government.
You are therefore seeking to recover the original 1000, which means
multiplying by 1/1.19, and the VAT is (1 - 1/1.19) times the original
amount.

simonmichael:
Ah I see. I think for consistency I'd prefer to support just %account for now nevertheless. I think it will be rare for people to use %account immediately followed by letters.

Xitian9:
Updated.

simonmichael:
Thank you very much!

Issue #2411, opened 2025-06-18
Incorrect columns in output when using --summary-only & csv


guysimard:
The income statement report includes incorrect column headers in export when using the --summary-only option and csv as an output format.

Steps to reproduce:
```
% hledger --version
hledger 1.43.2, mac-aarch64
% cat f.journal 
2025-01-01 Store
    expenses:food       20
    assets:bank

2025-02-01 Store
    expenses:food       25
    assets:bank

2025-03-01 Store
    expenses:food       50
    assets:bank

% hledger -f f.journal is --monthly --average --summary-only       
Monthly Income Statement 2025Q1

               || Average 
===============++=========
 Revenues      ||         
---------------++---------
---------------++---------
               ||       0 
===============++=========
 Expenses      ||         
---------------++---------
 expenses:food ||      32 
---------------++---------
               ||      32 
===============++=========
 Net:          ||     -32 
% hledger -f f.journal is --monthly --average --summary-only -O csv
"Monthly Income Statement 2025Q1","","","",""
"Account","Jan","Feb","Mar","Average"
"Revenues","","","",""
"Total:","0"
"Expenses","","","",""
"expenses:food","32"
"Total:","32"
"Net:","-32"
lmcgsim@EMB-Q0EDQ6LR 2025 % 

```

csv output includes the following headers:
```
"Account","Jan","Feb","Mar","Average"
```

While they it should actually be (as can be seen in the regular, non csv, output format):
```
"Account","Average"
```

simonmichael:
Thanks for the report.

simonmichael:
This should be fixed by #2414.

PR #2414, opened 2025-06-25
fix: csv: Make --summary-only work with csv output (#2411)

Xitian9:
Don't display date headers with --summary-only in csv output.


simonmichael:
Thanks!

Issue #1995, opened 2023-01-31
(at least areg): optionally do not abbreviate account names


jmtd:
The areg documentation says

```-   the names of all the other account(s) involved in this transaction
    (probably abbreviated)```

In my experience my account names are always abbreviated - perhaps that's a function of their length, it does not seem to be influenced by the width for default output format.

It would be great if areg -- and any other command which abbreviates command names -- could be told not to.

simonmichael:
Thanks. I believe aregister is the only command which abbreviates always, regardless of width; it just assumes it will need to. I agree some smarter behaviour would be nice. I guess you are willing to use a very wide terminal, or panning with a pager or similar, to view unelided wide reports.

jonestristand:
Agree, would love to see this fixed. It's more tricky and error prone to get the report I want from register, but for sending stuff to my accountant, having the full account name is essential. Any progress?

simonmichael:
No progress. Is there anyone who'd like to look at this ? It involves looking at how register does it in Register.hs and porting some of that to Aregister.hs.

savanto:
I encountered this issue recently, specifically with `areg`'s CSV output, which also always had abbreviated account names. I don't mind the abbreviations when outputting to text/terminal, but I do post-processing on the CSV and then supply it to an accountant, and I needed the account names to be unabridged. I was working around this by building a mapping of `abbrev->full` account names in my post-processing tool, but it seems to me that machine-readable output should not be applying an abbreviation strategy at all, and leaving decisions like that to the post-processor. 

I implemented a small fix to this (sub-)problem in #2416 by simply not abbreviating accounts in CSV/HTML/FODS output, while leaving text output unchanged.

Since I have this piece of the codebase in my head right now, I'll note that `aregister` uses [`accountSummarisedName`](https://github.com/simonmichael/hledger/blob/5d8b30ba371b03dbf08054e88a2bd1673865c74e/hledger-lib/Hledger/Data/AccountName.hs#L146) to trim all account components but the last one down to two characters no matter what, which is a pretty blunt approach. In contrast `register` appears to use [`elideAccountName`](https://github.com/simonmichael/hledger/blob/5d8b30ba371b03dbf08054e88a2bd1673865c74e/hledger-lib/Hledger/Data/AccountName.hs#L318) which not only uses a width parameter to determine whether to abbreviate at all, but decides whether to abbreviate or elide entirely.

[...]

PR #2416, opened 2025-06-27
imp: areg: Do not abbreviate other accounts in machine-readable output (#1995)

savanto:
The `aregister` report always abbreviates "other accounts" in the transaction by taking just the first two characters of all account components but the last one. Thus `asset:cash:bank:checking` always becomes `as:ca:ba:checking`. This is a reasonable strategy when outputting to text/terminal, as space is likely to be limited, but is less useful when outputting to machine-readable formats (CSV, HTML, FODS, etc).

#### This change makes `aregister` output unabridged account names to machine-readable formats:
- Do not turn "other accounts" into a comma-delimited string when the report is constructed, but pass `AccountName`s up to the caller. This requires modifying the `AccountTransactionsReportItem` type to contain `[AccountName]` rather than `Text`.

- Move the account name summarization (`Hledger.Data.AccountName.accountSummarizedName`) closer to the rendering of the report, so that different report formats can choose their own summarization strategy. The summarization algorithm for `aregister` isn't changed even though eg. `register` uses a different one.

- Continue to summarize as before for terminal/text output (ie. human-readable). Do not summarize any more for machine-readable output (csv/html/fods).

> [!WARNING]
> This has the potential to be a breaking change for users, if they are relying on abbreviated account names in their CSV/etc output. Is this a concern?

#### Other approaches I considered:
1. Instead of converting other accounts into `[AccountName]` and doing summarization in the caller, pass the desired summarization strategy into `accountTransactionsReportItem`. This felt like poor encapsulation, as rendering details would leak into report construction. I thought that instead the report builder should provide structured data, and the renderer should decide how to present them.

2. Collect user preferences for summarization on the command line. This would still leave (1) unresolved, since you'd need to then decide when to apply the preference, in the renderer, or in the report builder. I also thought that not abridging the account names in machine output is the "sensible" default, since that way you provide the full output and allow the post-processing tool to take its own approach to presentation. One further complication is that there are several confusing approaches to summarization (shorten/elide/ellipsize), which can be used in combination with each other, so figuring out a way to present the right options to the user is non-trivial.

I realize that "sensible" is subjective, and that adding a flag instead may be necessary to preserve established workflows. I'm open to implementing such a flag, with input on how best to name it/what it should control.

#### Testing notes
- I was able to verify through manual testing that `areg` on the command line continues to produce abbreviated account names, while CSV & FODS reports had full account names.
- JSON output appears to have the full structured transactions/postings even for summary reports like `aregister`, so the full account names are unchanged.
- I was unable to test HTML output for some reason. Both the build off my branch, off `master`, and even release v1.43.2, produced a header row but no other rows of the html table. This appears to be a regression, as release v1.42.2 produces html output just fine.

### DIFF
--- a/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
+++ b/hledger-lib/Hledger/Reports/AccountTransactionsReport.hs
@@ -27,7 +27,6 @@ import Data.List (mapAccumR, nub, partition, sortBy)
 import Data.List.Extra (nubSort)
 import Data.Maybe (catMaybes)
 import Data.Ord (Down(..), comparing)
-import Data.Text (Text)
 import qualified Data.Text as T
 import Data.Time.Calendar (Day)
 
@@ -80,7 +79,7 @@ type AccountTransactionsReportItem =
    Transaction -- the transaction, unmodified
   ,Transaction -- the transaction, as seen from the current account
   ,Bool        -- is this a split (more than one posting to other accounts) ?
-  ,Text        -- a display string describing the other account(s), if any
+  ,[AccountName] -- the other account(s), if any
   ,MixedAmount -- the amount posted to the current account(s) (or total amount posted)
   ,MixedAmount -- the register's running total or the current account(s)'s historical balance, after this transaction
   )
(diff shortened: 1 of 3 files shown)


simonmichael:
Great. I'm ok with changing this behaviour for the non-`text` output formats, without a flag. I think no-one loves these abbreviations.

I guess no docs are affected, and tests aren't super needed.

So I'll merge this soon if nothing else comes up.

simonmichael:
Thank you!

Issue #2169, opened 2024-02-12
Support for per-posting payee


hieu-van:
There are often cases that we might want to record payee on a per-posting basis. For example:

```
2024-1-2 an absurd transaction
        a:b:c                          $1     ; payee: Mr A
        a:b:c                          $2     ; payee: Mr B
        a:b:c                          $3     ; payee: Mr C
        wallet                        -$6
```

AFAIK this is possible with Ledger, but not with hledger. To achieve this, we can use a custom tag for recording payee, but that does come with its own limitations.

simonmichael:
My 2c: this seems like a niche use case. When I have multiple payees, I usually feel separate transactions models that best. I have separate relationships/obligations with each of them which shouldn't be complected.

hieu-van:
You are right. We should hear about the others' opinion on this

pwseo:
While I agree this would be a niche feature, it still seems like a good one because one can already do this with the `date:` posting tag. Ledger compatibility would also be a nice bonus.

simonmichael:
Consistency is a nice quality, but as de facto implementor/maintainer, you can probably understand if I'm reluctant to add complexity just for that, without solving practical problems. Ledger compatibility is also nice but I have experienced that you can take that too far; and I think it becomes less important for us over time.

But as always, a PR increases the chance that at least a few of us would give it a try.

simonmichael:
Something to ask ourselves: does this exist in double entry bookkeeping ? Is it something a professional bookkeeper would do ?

pwseo:
Simon, I agree with you:
- it probably makes more sense to split the transactions / it solves a single rarely occurring problem
- adding complexity can harm hledger's ability to thrive
- Ledger compatibility is less and less important
- there are much more important things to do with your (maintainer) time

I was just providing additional reasoning that I didn't think was explicit in the comments above :)

savanto:
I have a slightly different usecase for this feature, which is arguably not so niche in small business bookkeeping:

My business has customers and issues invoices when work is completed:
```journal
2025-01-01 CustA
    asset:accounts receivable    $500.00
    revenue:sales

2025-01-01 CustB
    asset:accounts receivable   $1500.00
    revenue:sales

2025-01-01 CustC
    asset:accounts receivable   $1500.00
    revenue:sales

2025-01-01 CustD
    asset:accounts receivable   $4000.00
    revenue:sales
```

Payments get deposited (data imported from bank as CSV and processed):
```journal
[...]

PR #2418, opened 2025-06-30
imp: pivot: Tag name shadowing field name takes precedence (#2169)

savanto:
Please see #2169 for the rationale behind this change.

---

Tags shadowing field names (eg. `payee`) take precedence during pivoting. For example, before this change,
```
2025-01-01 CustA
    asset       $100    ; payee:CustA
    asset       $200    ; payee:CustB
    revenue
```
we expect `bal asset --pivot=payee` to produce:
```
                $300  CustA
--------------------
                $300
```

But after this change, we expect:
```
                $100  CustA
                $200  CustB
--------------------
                $300
```

More specific tags take precedence over less specific tags (ie. posting tags take precedence over transaction tags):

```
2025-01-01 CustA  ; payee:CustB
    asset       $100    ; payee:CustA
    asset       $200    ; payee:CustB
    revenue
```
still results in
```
                $100  CustA
                $200  CustB
--------------------
                $300
```

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -1145,8 +1145,10 @@ pivotAccount fieldortagname p =
 -- "comm" and "cur" are accepted as synonyms meaning the commodity symbol.
 -- Pivoting on an unknown field or tag, or on commodity when there are multiple commodities, returns "".
 -- Pivoting on a tag when there are multiple values for that tag, returns the first value.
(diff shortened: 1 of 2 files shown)


simonmichael:
Comments added on #2169.

savanto:
Thanks for your thoughtful comments, you're probably right that this doesn't warrant such a global change. I solved this problem for myself by implementing a custom hledger plugin that allows tags to take precedence for the kind of report I needed to generate (aging summary).

PR #2419, opened 2025-06-30
fix: use the precompiled regex in gexMatchTextGroups (case insensitivity)

jneubrand:
Currently, the CSV reader treats regexes as case-insensitive, but the logic in `hledger-lib/Hledger/Utils/Regex.hs` treats regexes as case-sensitive when evaluating `regexMatchTextGroups`.

A test case says more than words:

```bash
hledger print -f csv:<(echo 'A,b
a,b') --rules=<(echo '
separator ,
date 0000-01-01

account1 X
amount1 $ 2
account2 Y

if %1 (A)
    comment matcher hit!
    account1 \1
')
```

output at `simonmichael/hledger:master`:

```ledger
0000-01-01  ; matcher hit!
    A             $ 2
    Y

0000-01-01  ; matcher hit!
    Y
```

output with this fix:

```diff
 0000-01-01  ; matcher hit!
     A             $ 2
     Y
 
 0000-01-01  ; matcher hit!
+    a             $ 2
     Y
```


simonmichael:
I see what you mean - and more than that, it seems if you reference a match group (by writing \NUM in the value) that did not match with exact case (as above), or does not exist (\9 when there's no 9th parenthesised pattern), it supresses that posting entirely - which is highly unintuitive. Would you like to have a look at that larger issue as well, in this PR ?

jneubrand:
Yes, kind of—it seems the error is handled by generating an empty string, which _then_ leads to providing an empty string to the `account1` rule, which _then_ leads to the posting being skipped. (To convince yourself of this, replace `account1 \1` with `account xyz\1` in the example.)

It seems that two different cases could generate warnings:

1. a nonexistent match group is referenced and expands to an empty string
2. a posting has a nonempty comment / amount / commodity / ..., _but no account_, so it is skipped

Anyway, Hyrum’s law applies:

1. I think some users might actually rely on behavior 1—namely, when there are multiple OR-chained field matchers which contain different numbers of matching groups, they may be able to write more compact rules this way.
2. I know for a fact that some of my rule files rely on behavior 2.

(The same parsing can still be done, of course, but to avoid triggering warning 1. if-rules will need to be split up, and to avoid 2. additional commodity-/comment-/amount-/...-rules will need to be added.)

I’d be more than happy to implement one/both of these warnings, I'm just not sure whether they’ll annoy more people than they help :)

simonmichael:
Thanks for the analysis and fix. I'm going to merge this fix and carefully back away from the other issue for now.

PR #2422, opened 2025-07-02
Fix: timeclock: allow overlapping sessions with same account name

### LINKED ISSUE #2417: 1.43: timeclock no longer accepts overlapping sessions with same account name
After upgrading to hledger version 1.43.2 (mac-aarch64) from 1.42.2, my existing timeclock file no longer works due to what appears to be an unintended breaking change. This differs from the documented breaking change in the [release notes](https://hledger.org/relnotes.html#hledger-143) regarding the significance of `o` entry suffixes.

Example test.timeclock file:
```
i 2024-04-10 13:00:00 test  test desc
o 2024-04-10 14:30:00
i 2024-04-10 14:30:00 test  new desc
o 2024-04-10 15:00:00
```

Expected / was:
```
$ hledger -f test.timeclock balance --timeclock-old
               2.00h  test
--------------------
               2.00h
```

Actual: an error occurs
```
$ hledger -f test.timeclock balance
hledger: Error: /Users/frank/test.timeclock:3:1:
3 | i 2024-04-10 14:30:00 test  new desc
  | ^

Encountered clockin entry for session "test" that is already active.
```

brymut:
This PR fixes a regression that broke existing timeclock files containing overlapping sessions with the same account name. The issue was in `pairClockEntries` function which added a validation that prevented clock-in entries for accounts that already had active sessions. 

However, the existing `findInForOut` logic already had handling for multiple sessions with the same account name using `partition` to correctly pair clock-ins with clock-outs. 

Closes #2417


brymut:
Hi @simonmichael, made an attempt at #2417 here. 
Edit: Apologies, may have found some issues with my approach. Will Draft this PR until I figure it out.

simonmichael:
For next release at least, the regression has been resolved by documenting the limitation, so I'm closing this. We can revisit, see #2417 for discussion.

PR #2425, opened 2025-07-07
fix: use date2 when --date2 is present for reg --csv

### LINKED ISSUE #1206: Q: is there a way to have date2 displayed in the hledger reg --output=csv export
The `hledger hreg` help says

> With --date2, it shows and sorts by secondary date instead. 

But i've been unable to have the `date2` column show when doing a csv export.

For example with 2 journal entries with the following dates

```text
2019/12/31=2020/01/02
    ...

2020/01/01=2020/01/02
   ...
```

Then the following command with `--date2` correctly finds those

```bash
$ hledger reg -Ocsv --date2 Business-CY:Assets:Girokonto:Revolut:EUR date2:2020-01-02
"txnidx","date","code","description","account","amount","total"
"7500","2019/12/31",...
"7502","2020/01/01",...
```

But I need the _effective date_ (date2) to be shown in the csv export. 

So filtering seems to work, but only `date` is shown.

Is there a way to affect the csv output to include `date2`?

jneubrand:
fixes #1206

draft: haven't verified that this…

- handles html/ods output correctly
- handles row sorting correctly

### DIFF
--- a/hledger/Hledger/Cli/Commands/Register.hs
+++ b/hledger/Hledger/Cli/Commands/Register.hs
@@ -101,36 +101,38 @@ register opts@CliOpts{rawopts_=rawopts, reportspec_=rspec} j
     rpt = postingsReport rspec j
     render | fmt=="txt"  = postingsReportAsText opts
            | fmt=="json" = toJsonText
-           | fmt=="csv"  = printCSV . postingsReportAsCsv
-           | fmt=="tsv"  = printTSV . postingsReportAsCsv
+           | fmt=="csv"  = printCSV . postingsReportAsCsv wd
+           | fmt=="tsv"  = printTSV . postingsReportAsCsv wd
            | fmt=="html" =
                 (<>"\n") . Lucid.renderText . styledTableHtml .
                 map (map (fmap Lucid.toHtml)) .
-                postingsReportAsSpreadsheet oneLineNoCostFmt baseUrl query
+                postingsReportAsSpreadsheet oneLineNoCostFmt wd baseUrl query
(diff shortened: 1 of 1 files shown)


simonmichael:
Forgive this off the cuff reaction, but it’s this: secondary dates were a mistake and I’d rather not have them consume any more time and attention…

jneubrand:
I can survive adding this to the list of stuff I'm maintaining in [here](https://github.com/jneubrand/hledger/tree/main), I suppose ;)

It'd be nice to have consistency between `areg` & `reg` in this regard, though.

Thoughts on <code>date2</code>:

I saw the "use posting dates instead" suggestion on hledger.org, but it doesn't really address the transaction/post date divide I want to keep track of. That is: a transaction can occur on Monday, but arrive at my bank on Tuesday and post on Wednesday. The bank provides me with two different dates in CSV columns (that I put into date/date2), and a third (POS-system-reported) date+time *can* be given via a text field (which is then piped into the Expense posting date tag).

With inter-bank transfers, there can be cases where a single transaction with two postings needs **four** different associated dates which are both "interesting" for different purposes (available balance, balance reported on bank statement, assertions, …). I guess I just don't want to use another tag to keep track of this, so I've been sticking to `date2`.

edit: oh, and the posting date can be *before or after* the transaction date at my bank ^^

simonmichael:
Ok; I'll merge this when ready. Consistency is good.

Generally, the use case seems a bit niche. There's no real limit to the number of dates someone might want to track; they'll probably need to use tags no matter what.

simonmichael:

Issue #2407, opened 2025-06-13
1.32-1.43.1: aregister does not round all amounts to display precision


simonmichael:
aregister does not apply display precision consistently. This has been broken since 1.32.

PR #2435, opened 2025-08-01
fix: perform hard rounding in hledger close

aragaer:
Back in 2019 I bought some yen and recorded that with a "@-style" transaction. Now every time I use hledger-close it prints one of my accounts with 4 digit precision.

This particular fix is similar to #2407, so it shouldn't break much but some existing tests rely on the current rounding logic and thus fail.

P.S. Couldn't run tests myself -- don't have `fd` binary on my system and don't know what that is supposed to be. Didn't find a mention in the documentation either. I also don't have `ghc` in $PATH, only available as `stack ghc`.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Close.hs
+++ b/hledger/Hledger/Cli/Commands/Close.hs
@@ -257,7 +257,7 @@ close CliOpts{rawopts_=rawopts, reportspec_=rspec0} j = do
 
   -- print them
   -- allow user-specified rounding with --round, like print
-  let styles = amountStylesSetRoundingFromRawOpts rawopts $ journalCommodityStyles j
+  let styles = amountStylesSetRoundingFromRawOpts rawopts $ journalCommodityStylesWith HardRounding j
   maybe (pure ()) (T.putStr . showTransaction . styleAmounts styles) mclosetxn
   maybe (pure ()) (T.putStr . showTransaction . styleAmounts styles) mopentxn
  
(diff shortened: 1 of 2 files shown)


simonmichael:
Like print, close is expected to produce precise balanced journal entries, so probably shouldn't be hard-rounded, at least not by default ?

aragaer:
Is there any way to somehow reduce precision when it only adds zeroes?

simonmichael:
I don't have the full context but is that like print's --round=soft ?

aragaer:
```
$ cat /tmp/test.journal
commodity $1.00
commodity JPY  1000.
2025-06-12
  assets:a  100000 JPY @ $0.006631
  assets:a
$ hledger -f /tmp/test.journal close --round=soft -e 2026
2025-12-31 closing balances  ; clopen:
    assets:a                                $663.10 = $0.00
    assets:a                            JPY -100000 = JPY 0
    equity:opening/closing balances
```
Yeah, that suits me.

simonmichael:
Thanks for the test case. 
```
commodity $1.00
commodity JPY  1000.
2025-06-12
  assets:a  100000 JPY @ $0.006631
  assets:a
```
For the record: some might wonder (I did) why $ gets six decimals by default here:
```
$ hledger close
2025-07-31 closing balances  ; clopen:
    assets:a                            $663.100000 = $0.000000
    assets:a                            JPY -100000 = JPY 0
    equity:opening/closing balances
```

It's because the second posting amount is inferred as $-663.100000, as if the entry had been written as:
```
$ hledger print -x
2025-06-12
    assets:a    JPY 100000 @ $0.006631
    assets:a              $-663.100000
```
And print-like commands always try to show entries as they were written. Unless you use --round, then the display precision (declared by `commodity` as two digits in this case) is used. --round=hard can unbalance entries; --round=soft is safer, as it will only add or remove trailing zeros.

PR #2440, opened 2025-08-13
dev: fix Dockerfile

lfos:
Sync Docker GHC version with Stack configuration, update dependencies, and replace deprecated Dockerfile syntax.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM haskell as dev
+FROM haskell:9.12.2 AS dev
 
 RUN mkdir /root/hledger
 WORKDIR /root/hledger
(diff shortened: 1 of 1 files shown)


simonmichael:
Thank you!

One of these days, stack912.yaml will disappear for a while (it will become the default stack.yaml). So ideally/normally the docker file could rely on the default stack.yaml. We'll tweak it when the time comes.

lfos:
> Thank you!
> 
> One of these days, stack912.yaml will disappear for a while (it will become the default stack.yaml). So ideally/normally the docker file could rely on the default stack.yaml. We'll tweak it when the time comes.

As long as the GHC version needs to exactly match what's in the Stack config, I think it'd be best to continue hardcoding both the Docker tag and the stack version. Is it possible to keep the version-specific Stack config files and simply make `stack.yaml` a symlink that's updated to whatever version is the new default?

simonmichael:
I wish it was, unfortunately that breaks the build on Windows.

simonmichael:
I could keep two copies of the default stack.yaml. The cost is more clutter and places to maintain.

Issue #2434, opened 2025-07-30
master: multi-balance reports with a depth limit omit some amounts


simonmichael:
There's a bug in master, caused by #2395 unfortunately. I wasn't been able to minimise it more than this yet (the custom account names seem required). Note how with depth limit 1, the A amount is wrongly excluded.

```
account sm:assets               ; type:A
account sm:revenues             ; type:R

2025-01-01
    sm:revenues
    sm:assets                                     $1

2025-01-02
    sm:revenues                                 A -2
    sm:assets                                    A 2


comment

Limited to depth 2: correct

$ hledger -f a.j is -2
Income Statement 2025-01-01..2025-01-02

             || 2025-01-01..2025-01-02 
=============++========================
 Revenues    ||                        
-------------++------------------------
 sm:revenues ||                $1, A 2 
-------------++------------------------
             ||                $1, A 2 
=============++========================
 Expenses    ||                        
-------------++------------------------
-------------++------------------------
             ||                      0 
=============++========================
 Net:        ||                $1, A 2 



Limited to depth 1: not correct

$ hledger -f a.j is -1
Income Statement 2025-01-01..2025-01-02

          || 2025-01-01..2025-01-02 
==========++========================
 Revenues ||                        
----------++------------------------
 sm       ||                     $1 
----------++------------------------
          ||                     $1 
==========++========================
 Expenses ||                        
----------++------------------------
----------++------------------------
          ||                      0 
==========++========================
 Net:     ||                     $1 
```

@Xitian9 might you have some time to look at this ?

Xitian9:
That's a strange one. Let me take a look at it and get back to you.

Also, hope you weren't too badly affected by the tsunami warnings.

Xitian9:
Okay, I think I've found it. Proposed solution in #2447.

simonmichael:
Fixed by #2447.

PR #2447, opened 2025-08-27
fix: balance: Clip account names in generateMultiBalanceAccount, not before (#2434)

Xitian9:
Previously accounts were clipped in getPostings, however compound balance reports re-use the output of getPostings for the different subreports. This caused a problem when clipping erased the information needed to determine the account type, as would be used by e.g. incomestatement.

Add some extra tests for --count.


simonmichael:
Thank you! Merged as ef5c220690ae27243ed4d8da60fdd1e88ca2f2b8.

Issue #2445, opened 2025-08-26
using a future-dated P directive in value reports is unintuitive ?


simonmichael:
As reported in #hledger, a valuation date can be sometimes be chosen that's in the future. 
(To work around, add an explicit report end date, like `-e today`.)

Eg
```
P 3000-01-01 A  2 B

2025-01-01
  (a)      1 A
```

A market price declaration dated in the future is quite unusual, but let's say you have one.
Running `hledger print -V` today, we might expect the 3000-dated P to be ignored, but it isn't:
```
$ hledger print -V
2025-01-01
    (a)             2 B
```

3000-01-01 is chosen as the valuation date, as shown by `--debug=2`.
The manual's Valuation date section was "clarified" in September 2023 to say:

> For single period reports (including normal print and register reports):
> ...
> - Otherwise the latest transaction date or P directive date is used (even if it's in the future)

This was catching up with a behaviour change made in 2021, arising from some complicated discussion at #1495, #1416 and originally #1405 . I think the goal was to fix unintuitive ignoring of P directives in value reports.

But we might have overlooked the case of a P directive dated in the future, which is an uncommon situation.

Perhaps in this case it would be better to pick:
"Otherwise the latest transaction date; otherwise the latest P directive date that's not in the future"
?

simonmichael:
#2449 is a draft fix for this; it needs discussion, as I'm not sure if it's correct/better.

simonmichael:
#2449 has been updated, and is looking better.

PR #2449, opened 2025-08-27
valuation: future P directives don't set valuation date

simonmichael:
This is a proposed fix for #2445:
For end-value reports, a P directive later than the latest transaction
will set the valuation date - even if that P directive is in the future.

With this PR, future-dated P directives are ignored; only non-future P directives
can influence valuation date.

It also tries to clarify report span calculation, and various related functions.

The changes to tests show that the P directives can also extend the report end date,
which seems to be new behaviour. I'm not sure if the old or new behaviour is
more correct/intuitive/straightforward, so this needs some review and discussion.


simonmichael:
After several attempts, I have pushed a new fix that implements just the change described above, while seemingly not changing old behaviour otherwise. This is a confusing rabbit hole, but I think this is the right fix.

Issue #2443, opened 2025-08-19
--summary-only doesn't work with --budget


entropitor:
version:
```
hledger 1.40, mac-aarch64
```

When adding the `--budget` flag, the `--summary-only` flag seems to be ignored

simonmichael:
Thanks for the bug report.

simonmichael:
I believe this was fixed in 1.50.

PR #2450, opened 2025-08-30
fix: budget: Respect --summary-only flag. (#2443)

Xitian9:
Budget reports will now respect the --summary-only flag.


simonmichael:
Awesome, thanks!

PR #2455, opened 2025-09-09
fix: Don't crash on nonsense date span

### LINKED ISSUE #2452: 1.50: Unexpected error when no transaction is in period
Hi,

## How to reproduce
1. The `test.ledger`  content:

```
2021-01-01 Test transaction
  Assets:Bank   -100USD
  Expenses:Food   100USD
```

2. Then run command below and set a period that matches no transaction:

`hledger -f test.ledger  bal Assets -p 'from 2030-10-01'`

or

`hledger -f test.ledger  bal Assets -p 'to 2000-10-01'`

#### Expected result:
Show zero, because no transaction is found. There should not be any error.

```
--------------------
                   0
```

#### Actual result:
Result is an error which is not very clear. Why should both start and end be specified? The same period works if there are matching transactions and only error is thrown when no transaction is in the period. 

`hledger: Error: balanceDataPeriodEnds: expected all non-initial spans to have start and end date`

## Info
`hledger --version` output: hledger 1.50, linux-x86_64
OS: Archlinux

alt-romes:
Fixes #2452

### DIFF
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -314,16 +314,23 @@ periodDataValuation ropts j priceoracle colspans =
 
     balanceDataPeriodEnds :: PeriodData Day
     balanceDataPeriodEnds = dbg5 "balanceDataPeriodEnds" $ case colspans of  -- FIXME: Change colspans to nonempty list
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks! I see this removes one of the three (!) error calls in this function. Tip, you can run the main tests locally with `just functest`.

simonmichael:
I'm going to accept #2456 instead. Thank you for the pr!

PR #2456, opened 2025-09-10
fix: balance: Correctly handle empty journals (#2452)

Xitian9:
Eliminate several partial functions.

There is more work to be done creating fewer error states in the function `splitSpan`, but I'll leave this to a separate PR.

### DIFF
--- a/hledger-lib/Hledger/Data/PeriodData.hs
+++ b/hledger-lib/Hledger/Data/PeriodData.hs
@@ -9,12 +9,18 @@ Report periods are assumed to be contiguous, and represented only by start dates
 -}
 module Hledger.Data.PeriodData
 ( periodDataFromList
+, periodDataToList
 
 , lookupPeriodData
+, lookupPeriodDataLE
 , insertPeriodData
 , opPeriodData
 , mergePeriodData
 , padPeriodData
+
+, periodDataToDateSpans
+, maybePeriodDataToDateSpans
+, dateSpansToPeriodData
 
 , tests_PeriodData
 ) where
@@ -24,18 +30,17 @@ import Data.Foldable1 (Foldable1(..))
 #else
 import Control.Applicative (liftA2)
 #endif
+import Data.Bifunctor (first)
 import qualified Data.IntMap.Strict as IM
-import qualified Data.IntSet as IS
 #if !MIN_VERSION_base(4,20,0)
 import Data.List (foldl')
 #endif
 import Data.Time (Day(..), fromGregorian)
 
-import Test.Tasty (testGroup)
-import Test.Tasty.HUnit ((@?=), testCase)
-
 import Hledger.Data.Amount
+import Hledger.Data.Dates
 import Hledger.Data.Types
+import Hledger.Utils
 
 
 instance Show a => Show (PeriodData a) where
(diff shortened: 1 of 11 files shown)


hledger-lib/Hledger/Data/PeriodData.hs
```diff
@@ -73,18 +78,29 @@ instance Monoid a => Monoid (PeriodData a) where
...
+periodDataFromList h = PeriodData h . IM.fromList . map (\(d, a) -> (dayToInt d, a))
+
+-- | Convert 'PeriodData' to a list of pairs.
+periodDataToList :: PeriodData a -> (a, [(Day, a)])
+periodDataToList (PeriodData h as) = (h, map (\(s, e) -> (intToDay s, e)) $ IM.toList as)
 
--- | Get account balance information to the period containing a given 'Day'.
+
+-- | Get account balance information to the period containing a given 'Day',
+-- and the historical data if this day lies in the historical period.
```
simonmichael, line 89: Would "for the period" and "or the historical data" be clearer ?

hledger-lib/Hledger/Data/PeriodData.hs
```diff
@@ -73,18 +78,29 @@ instance Monoid a => Monoid (PeriodData a) where
...
 lookupPeriodData :: Day -> PeriodData a -> a
-lookupPeriodData d (PeriodData h as) =
-    maybe h snd $ IM.lookupLE (fromInteger $ toModifiedJulianDay d) as
+lookupPeriodData d pd@(PeriodData h _) = maybe h snd $ lookupPeriodDataLE d pd
+
+-- | Get account balance information for the period containing a given 'Day',
+-- along with the start of the period, or 'Nothing' if this day lies in the
+-- historical period.
+lookupPeriodDataLE :: Day -> PeriodData a -> Maybe (Day, a)
```
simonmichael, line 96: What's LE ?

hledger-lib/Hledger/Data/PeriodData.hs
```diff
@@ -103,10 +119,38 @@ mergePeriodData only1 only2 f = \(PeriodData h1 as1) (PeriodData h2 as2) ->
   where
     merge = IM.mergeWithKey (\_ x y -> Just $ f x y) (fmap only1) (fmap only2)
 
--- | Pad out the datemap of an 'PeriodData' so that every key from a set is present.
-padPeriodData :: Monoid a => IS.IntSet -> PeriodData a -> PeriodData a
-padPeriodData keys bal = bal{pdperiods = pdperiods bal <> IM.fromSet (const mempty) keys}
+-- | Pad out the datemap of an 'PeriodData' so that every key from another 'PeriodData' is present.
```
simonmichael, line 122: an -> a

simonmichael (changes requested):
Looking good, thank you!

hledger-lib/Hledger/Data/PeriodData.hs
Xitian9: Less than or equal to, named in parallel with the `lookupLE` function in IntMap. I agree it could be better named. Perhaps this could be `lookupPeriodData` and the previous `lookupPeriodData` could be `lookupPeriodDataOrHistorical`? I'm open to suggestions.

simonmichael: Those sound good.

simonmichael:
Thank you!

Issue #2454, opened 2025-09-08
1.50: `hledger-ui` does not hide 0 accounts with zero/nonzero toggle


nicmollel:
running in docker
```shell
 $ hledger-ui --version
hledger-ui 1.50-g75c3b2cfe-20250903, linux-x86_64
```

minimal example
```journal
2025-09-07 * opening
    [checking:budget:item1]             $100.00
    [checking:budget:item2]              $25.00
    [checking:budget:allocated]        $-125.00
    checking                            $125.00
    equity:opening balance             $-125.00

2025-09-07 * expense
    [checking:budget:item1]            $-100.00
    [checking:budget:allocated]         $100.00
    expenses:item1                      $100.00
    checking                           $-100.00
```

viewing this journal with `hledger-ui` and toggling nonzero 

<img width="879" height="230" alt="Image" src="https://github.com/user-attachments/assets/c20258d4-8300-48ae-b1a4-2d2368f62a17" />

`checking:budget:item1` account is visible. Expected behavior would be for the account to be hidden in **nonzero** mode

simonmichael:
Thanks for the report @nicmollel ! This is a regression since 1.50. Claim your https://hledger.org/REGRESSIONS.html#regression-finder-bounty when ready.

Here's a more minimal example:
```
2025-09-07
    equity                 $-10
    bank                    $10

2025-09-07
    bank                   $-10
    expenses                $10
```

With `hledger-ui-1.50 --all --tree`, adding -E flag or pressing Z key should toggle the zero-balance `bank` account, but it doesn't. -E with `hledger-1.50 bal -t` does. So I think this is hledger-ui-specific, and likely related to this release note:

> - Some edge cases in balance report behaviour were changed for internal consistency:
> 
>   - balance --declared now shows undeclared parents of declared subaccounts.
> 
>   - balance --flat --empty now ensures that implied accounts with no postings are not displayed,
>     but accounts with zero balance and actual postings are.
> 
>   (Stephen Morgan, #2360, #2395)

simonmichael:
I confirm that this was caused by 80cf1d199.

Xitian9:
Okay, I've found it. It's not a problem with hledger-ui, but rather with the `--no-elide` option (which hledger-ui uses). You can see the same problem on your minimal example with `hledger balance --tree --no-elide`. Fix incoming.

simonmichael:
Fixed by #2460.

PR #2457, opened 2025-09-12
Eliminate error calls in report periods

Xitian9:
This eliminates the last few error calls that can occur in calculating the report spans of a multi-balance report.

1. We introduce a newtype wrapper `DayPartition` for `PeriodData Day`. By not exporting the constructor, we can ensure that all `DayPartition`s are well-formed.
2. `splitSpan` is moved into `Hledger.Data.DayPartition` and now produces `Maybe DayPartition` directly, rather than passing through `[DateSpan]`.

That said, this introduced a fair amount of code churn in order to eliminate the last few error calls. If you feel that the benefit is not worth the cost, let me know.

### DIFF
--- a/hledger-lib/Hledger/Data/DayPartition.hs
+++ b/hledger-lib/Hledger/Data/DayPartition.hs
@@ -11,16 +11,21 @@ module Hledger.Data.DayPartition
 
 , dayPartitionToNonEmpty
 , dayPartitionToList
-, dayPartitionToPeriodData
+, dayPartitionSpans
 , dayPartitionToDateSpans
+, dayPartitionToPeriodData
 , maybeDayPartitionToDateSpans
-, dateSpansToDayPartition
+
+, splitSpan
+, intervalBoundaryBefore
+
+, tests_DayPartition
 ) where
 
 import qualified Data.IntMap.Strict as IM
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NE
-import Data.Time (Day, addDays)
+import Data.Time (Day(..), addDays, addGregorianMonthsClip, addGregorianYearsClip, fromGregorian)
 
 import Hledger.Data.Dates
 import Hledger.Data.PeriodData
@@ -56,9 +61,10 @@ isValidDayPartition (DayPartition pd) = case ds of
 
 -- | Construct a 'DayPartition' from a non-empty list of boundary days.
 boundariesToDayPartition :: NonEmpty Day -> DayPartition
-boundariesToDayPartition xs =
-    DayPartition $ periodDataFromList (addDays (-1) b) $ zip (b:bs) (map (addDays (-1)) bs)
-  where (b:|bs) = NE.nub $ NE.sort xs
+boundariesToDayPartition xs = DayPartition . periodDataFromList (addDays (-1) b) $ case bs of
+    []  -> [(b, b)]  -- If only one boundary is supplied, it ends on the same day
+    _:_ -> zip (b:bs) $ map (addDays (-1)) bs  -- Guaranteed non-empty
+  where b:|bs = NE.nub $ NE.sort xs
 
 -- | Construct a 'DayPartition' from a list of boundary days, returning
 -- 'Nothing' for the empty list.
@@ -93,6 +99,12 @@ dayPartitionToNonEmpty (DayPartition xs) = NE.fromList . snd $ periodDataToList 
 dayPartitionToList :: DayPartition -> [(Day, Day)]
 dayPartitionToList = NE.toList . dayPartitionToNonEmpty
 
+-- | Return the whole day range spanned by a `PeriodData Day`.
+dayPartitionSpans :: DayPartition -> (Day, Day)
+dayPartitionSpans (DayPartition (PeriodData _ ds)) =
+    -- Guaranteed not to error because the IntMap in non-empty.
+    (intToDay . fst $ IM.findMin ds, snd $ IM.findMax ds)
+
 -- | Convert 'DayPartition' to a list of 'DateSpan's.
 --
 -- Note that the end date of each period will be equal to the start date of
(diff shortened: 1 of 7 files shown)


simonmichael:
I'll hold this one for now as I don't have an open bug that needs it.

If you have time, #2454 would be a great one to look at next for 1.50.1.

simonmichael:
Hola @Xitian9..

This PR has two commits, so I don't understand git log's showing me just one when I check out this branch:

- 97a04945a (grafted, HEAD -> splitSpan, Xitian9/splitspan, pr/Xitian9/2457) dev!: lib: Refactor splitSpan to return `Maybe DayPartition`. (Stephen Morgan, Sat Sep 13 09:24)

could you rebase when you have a chance.

Xitian9:
Rebased.

hledger-lib/Hledger/Data/DayPartition.hs
```diff
@@ -0,0 +1,269 @@
...
+-- Get the natural start for the given interval that falls on or before the given day,
+-- when applicable. Works for Weeks, Months, Quarters, Years, eg.
+intervalBoundaryBefore :: Interval -> Day -> Day
+intervalBoundaryBefore i d =
+  case dayPartitionToNonEmpty <$> splitSpan True i (DateSpan (Just $ Exact d) (Just . Exact $ addDays 1 d)) of
+    Just ((start, _) :| _ ) -> start
+    _ -> d
+
+
+intToDay = ModifiedJulianDay . toInteger
```
simonmichael, line 228: > The Modified Julian Day is a standard count of days, with zero being the day 1858-11-17.

This sounds like it might not fit with our support of all years ? (CE currently, maybe BCE some day)

simonmichael (commented):
AI gave this overview of the latest date types:

**Date-related Types Overview**

  Basic date types (Types.hs:103-123):

  - EFDay - exact or flexible date (Exact Day | Flex Day)
  - DateSpan - optional start/end dates (DateSpan (Maybe EFDay) (Maybe EFDay))

  Report period types (Types.hs:129-158):

  - Period - high-level periods (DayPeriod, MonthPeriod, etc.)
  - Interval - report intervals (Days, Weeks, Months, Quarters, Years, etc.)

  Period data structures:

  - PeriodData a (Types.hs:756) - data for report periods + pre-period
    - pdpre :: a - historical/pre-report data
    - pdperiods :: IntMap a - data per period (Day as IntMap key)
  - DayPartition (DayPartition.hs:37) - well-formed time partition
    - Newtype wrapping PeriodData Day
    - Opaque constructor enforces contiguous spans
    - Invariants: non-empty, contiguous, end = start-1 of next

  Key functions:

  - splitSpan :: Interval -> DateSpan -> Maybe DayPartition - split span into
  periods
  - lookupDayPartition :: Day -> DayPartition -> (Maybe Day, Day) - find period
   containing day
  - dayPartitionToDateSpans :: DayPartition -> [DateSpan] - convert to list of
  spans

Does it sound about right to you ?

DayPartition is PeriodData with mandatory smart constructors to guarantee invariants. Is it valuable to still have the separate PeriodData type ? 

If so we should probably expand the description of both types, making things more approachable.

I have installed this locally and will try to exercise it as well.

simonmichael:
Now also available as https://github.com/simonmichael/hledger/releases/tag/nightly binaries.

hledger-lib/Hledger/Data/DayPartition.hs
Xitian9: Yes, this is from an earlier commit, and yes it technically does not support all years.

- On a 64-bit system this will support all years between 252 quadrillion BC (well before the development of currency) and 252 quadrillion AD (at which point we'll be more worried about powering our computers since all stars will have become black holes or white dwarfs).
- On a 32-bit system, it will support all years between 5 877 751 BC and 5 881 468 AD.

We could extend support to all integers. This would incur a small performance penalty as we'd have to switch from IntMap to Map internally. I personally think the supported date ranges are fine, but happy for you to make a call.

Xitian9:
Yes, I think that description is basically accurate, though I think only `splitSpan` winds up being a key function in practice.

We would still need the separate `PeriodData` constructor, as it also holds balance information in the `Account` data type. I'll have a go at expanding the descriptions.

hledger-lib/Hledger/Data/DayPartition.hs
simonmichael: Oh wow. So 0 to 1858 is not a problem then. Great.

hledger-lib/Hledger/Data/Types.hs
```diff
@@ -750,13 +750,16 @@ data Account a = Account {
...
--- | Data values for zero or more report periods, and for the pre-report period.
--- Report periods are assumed to be contiguous, and represented only by start dates
--- (as keys of an IntMap). XXX how does that work, again ?
+-- | A general container for storing data values associated to zero or more
+-- report periods, and for the pre-report period. Report periods are assumed to
+-- be contiguous, and represented only by start dates.
+--
+-- Data is stored in an 'IntMap' for efficiency, where Days are stored as as
+-- Int representing the underlying modified Julian date.
 data PeriodData a = PeriodData {
```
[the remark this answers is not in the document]
simonmichael: Still, let's spell out how exactly the Int key represents the date.

hledger-lib/Hledger/Data/DayPartition.hs
```diff
@@ -0,0 +1,275 @@
...
+import Hledger.Utils
+
+
+-- | A partition of time into contiguous spans, along with a historical period
+-- before any of the spans.
+--
+-- This is a newtype wrapper around 'PeriodData Day', where the start dates are
+-- the keys and the end dates are the values. Spans are stored in inclusive format
+-- [start, end]. Note that this differs from 'DateSpan' which uses [start, end)
+-- format.
```
simonmichael, line 42: Thanks for expanding these descriptions. I think we can do better still; as a person reading this, not familiar with PeriodData (or even if I am), I find it not too clear. "Start dates are keys, end dates are values." I don't see it.

hledger-lib/Hledger/Data/DayPartition.hs
```diff
@@ -0,0 +1,275 @@
...
+--
+-- The constructor is not exported so that we can ensure the spans are valid
+-- partitions of time.
+newtype DayPartition = DayPartition { dayPartitionToPeriodData :: PeriodData Day } deriving (Eq, Ord, Show)
+
+-- Developer's note. All constructors must guarantee that:
+--   1. The value stored in pdperiods has at least one key.
+--   2. The value stored in pdpre equals one day before the smallest key in pdperiods.
+--   3. The value stored in each entry of pdperiods equals one day before the
+--      next largest key, except for the value associated to the largest key.
```
simonmichael, line 52: Who needs to follow these rules - a user of this module ? When using which functions ? Would it be better to mention these as invariants, in DayPartition's doc ?

hledger-lib/Hledger/Data/Types.hs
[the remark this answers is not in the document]
Xitian9: I'm not sure this is the best place to do that, as that's more or less an internal detail of the `Data.Time` module. In the unlikely event that they decide to change their internal representation, our documentation would be out of date.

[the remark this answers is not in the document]
simonmichael: Ok, a pointer to the authoritative doc would be fine then. So a person wanting to understand this type can do so, right ?

simonmichael:
I'm going to merge this and maybe follow up with another pass at haddocks. Thank you!

PR #2460, opened 2025-09-13
fix: balance: Don't display zero rows with --no-elide (#2454)

Xitian9:
Zero rows with --no-elide should only be displayed if either --empty is called or they have non-zero children at some depth.


Xitian9:
I'm not sure what that build failure is. It doesn't seem to be related to this PR.

simonmichael:
Thank you! I'm not sure either. I'm travelling tomorrow but will sort it out.

simonmichael:
Tested locally, looks good here. (And the CI workflow succeeded when I reran it.)

PR #2463, opened 2025-09-21
Forbid solving for hashtables 1.3

hseg:
On modern gccs, this sometimes causes build failures, see
https://github.com/gregorycollins/hashtables/issues/97

### DIFF
--- a/hledger-lib/hledger-lib.cabal
+++ b/hledger-lib/hledger-lib.cabal
@@ -148,7 +148,7 @@ library
     , extra >=1.7.11
     , file-embed >=0.0.10
     , filepath
-    , hashtables >=1.2.3.1
+    , hashtables >=1.2.3.1 && <1.3 || >=1.4.0
     , lucid
     , megaparsec >=7.0.0 && <9.8
     , microlens >=0.4
--- a/hledger-lib/package.yaml
+++ b/hledger-lib/package.yaml
@@ -68,7 +68,7 @@ dependencies:
 - encoding >=0.10
 - file-embed >=0.0.10
 - filepath
-- hashtables >=1.2.3.1
+- hashtables >=1.2.3.1 && <1.3 || >=1.4.0
 - lucid
 - megaparsec >=7.0.0 && <9.8
 - microlens >=0.4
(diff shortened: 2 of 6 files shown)


hseg:
Forgot to patch `package.yaml`, added it (probably worth squashing when you merge this, though)

hseg:
Am afk, forgot to update `stack98.yaml` to contain this `hashtables`, trying to do so from web

hseg:
OK, that should be all of them. Of course, I'd rather these all be squashed before merge.

simonmichael (commented):
Thanks @hseg !

At first read of https://github.com/gregorycollins/hashtables/issues/97 it sounded like only version 1.3.1 was affected; I think your changes exclude some other versions as well. Is that intended/ok ? (The rest of the discussion gets pretty complicated.)

Your stack yaml updates are thorough but let's not touch the ones I've recently marked with .old - they're now officially unmaintained and I've kept them there only in case someone really needs to revive them. I'll comment them more clearly. Although, since you've done the changes I guess it's no harm to keep them; either is fine.

Can you squash the changes together yourself and force-push the PR branch again, that will be helpful.

simonmichael:
Also, the hashtables dependency appears three times in hledger-lib.cabal; while it may work as-is, it's probably better to have them all consistent, which I usually ensure by 1. changing the package.yaml 2. regenerating the cabal file with `./Shake cabalfiles` (usually I do it as a follow-on commit with `./Shake cabalfiles -c`). Updating all the places manually is ok too if that's easier.

hseg:
It affects 1.3 and 1.3.1, and there are no other releases in the 1.3.* version. Unfortunately, afaict cabal doesn't support blacklisting single versions in this way. I _guess_ I could narrow the gap if necessary, I'll check if cabal supports exclusive lower bounds. Won't be able to squash tonight anymore, but will try to do so tomorrow - will be otherwise unavailable until Thursday.

hseg:
I'm not used to using Shake, will do so in the future

simonmichael:
Alright, in that case I'll just merge this and add any fixes that seem right. Thank you!

hseg:
Thanks for that, I see you managed to keep the attribution information around.

PR #2466, opened 2025-09-28
;examples: add Vietnamese translations for top-level accounts

ooker777:


simonmichael:
Yay, thanks!

For consistency would you mind also translating the comment at the top. (No need for the "machine translated" line unless it's applicable.)

simonmichael:
Thanks!

PR #2474, opened 2025-10-03
Help ui --watch catch changes from editors that overwrite files

### LINKED ISSUE #1617: hledger-ui --watch does not auto-update when journal file changes
Further to a [recent thread on reddit about methods for reconciliation](https://www.reddit.com/r/plaintextaccounting/comments/oiol2g/reconciliation_workflow_any_tips_for_a_n00b/), there appears to be a bug with `hledger-ui` where it does not update automatically when changes to the journal file are saved. The workaround is to manually refresh using `g`.

### System specifications
Arch Linux 5.12.13-arch1-2
hledger 1.21

### Steps to reproduce
1. Edit a journal file
`$ vim myfile.journal`
Change the flag on any unreconciled transaction in `Account A` by changing `<blank>` to `*`.
Do not save yet.

2. Open a new terminal window and start `hledger-ui`
$ `hledger-ui -f myfile.journal --watch`
Navigate to `Account A` and list the transactions so that the just-edited transaction is visible. It should _not_ show a `*` flag.

3. Return to the editor window and Save the file.

### Expected behavior
The edited transaction is updated in the `hledger-ui` window and now shows `*`

### Actual behavior
The edited transaction is _not_ updated in the `hledger-ui` window.

### Workaround
In the `hledger-ui` window, refresh the underlying data by pressing `g`. The edited transaction is updated and now correctly shows as being reconciled `*`.

alerque:
Not all editors trigger the `Modified` event for files because they don't edit the bits of an existing file, they flush their internal buffer to a new temporary file, then clobber the existing file on disk. This means many editors trigger the `Added` event even when saving changes to an existing journal file.

Tested with `gedit`.

Closes #1617

### DIFF
--- a/hledger-ui/Hledger/UI/Main.hs
+++ b/hledger-ui/Hledger/UI/Main.hs
@@ -44,7 +44,7 @@ import Lens.Micro ((^.))
 import System.Directory (canonicalizePath)
 import System.Environment (withProgName)
 import System.FilePath (takeDirectory)
-import System.FSNotify (Event(Modified), watchDir, withManager, EventIsDirectory (IsFile))
+import System.FSNotify (Event(Added, Modified), watchDir, withManager, EventIsDirectory (IsFile))
 import Brick hiding (bsDraw)
 import Brick.BChan qualified as BC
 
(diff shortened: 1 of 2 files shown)


simonmichael:
Love this fix! I'm just going to merge, what could possibly go wrong. Thank you.

PR #2477, opened 2025-10-10
DayPartition doc edits, cleanups

simonmichael:
- **;dev: PeriodData, DayPartition: haddock updates**
- **;dev: dayPartitionSpans -> dayPartitionStartEnd**
- **;dev: lookupDayPartition -> dayPartitionFind**

@Xitian9 I tried to clarify some details to myself - let me know if you see any mistakes.

(Followup for #2457.)

### DIFF
--- a/hledger-lib/Hledger/Data/DayPartition.hs
+++ b/hledger-lib/Hledger/Data/DayPartition.hs
@@ -15,7 +15,7 @@ module Hledger.Data.DayPartition
 -- * operations
 , unionDayPartitions
 , dayPartitionStartEnd
-, lookupDayPartition
+, dayPartitionFind
 , splitSpan
 , intervalBoundaryBefore
 -- * tests
@@ -130,12 +130,12 @@ dayPartitionStartEnd (DayPartition (PeriodData _ ds)) =
   -- Guaranteed not to error because the IntMap is non-empty.
   (intToDay . fst $ IM.findMin ds, snd $ IM.findMax ds)
 
-lookupDayPartition :: Day -> DayPartition -> (Maybe Day, Day)
-lookupDayPartition d (DayPartition xs) = lookupPeriodDataOrHistorical d xs
 -- | Find the start and end dates of the period within a 'DayPartition' which contains a given day.
 -- If the day is after the end of the last period, it is assumed to be within the last period.
 -- If the day is before the start of the first period (ie, in the historical period),
 -- only the historical period's end date is returned.
+dayPartitionFind :: Day -> DayPartition -> (Maybe Day, Day)
+dayPartitionFind d (DayPartition xs) = lookupPeriodDataOrHistorical d xs
 
 -- | Split a 'DateSpan' into a 'DayPartition' consisting of consecutive exact
 -- spans of the specified Interval, or `Nothing` if the span is invalid.
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -295,7 +295,7 @@ calculateReportAccount rspec@ReportSpec{_rsReportOpts=ropts} j priceoracle (Just
         mapPeriodData (padPeriodData mempty (dayPartitionToPeriodData colspans)) $
         accountFromPostings getIntervalStartDate ps
 
-    getIntervalStartDate p = fst $ lookupDayPartition (getPostingDate p) colspans
+    getIntervalStartDate p = fst $ dayPartitionFind (getPostingDate p) colspans
     getPostingDate = postingDateOrDate2 (whichDate (_rsReportOpts rspec))
 
 -- | The valuation function to use for the chosen report options.
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -674,7 +674,7 @@ journalApplyValuationFromOptsWith rspec@ReportSpec{_rsReportOpts=ropts} j priceo
       where
         postingPeriodEnd d = fromMaybe err $ case interval_ ropts of
           NoInterval -> fmap (snd . dayPartitionStartEnd)    . snd $ reportSpan j rspec
-          _          -> fmap (snd . lookupDayPartition d) . snd $ reportSpanBothDates j rspec
+          _          -> fmap (snd . dayPartitionFind d) . snd $ reportSpanBothDates j rspec
         -- Should never happen, because there are only invalid dayPartitions
         -- when there are no transactions, in which case this function is never called
         err = error' "journalApplyValuationFromOpts: expected all spans to have an end date"


simonmichael:
If I and the machines calculated rightly, PeriodData can work with dates from (estimated) -25,252,734,927-11-25 to +25,252,734,930-09-12. I'm curious to know the behaviour at those limits, specifically caused by PeriodData's Int key. Current hledger supports only positive CE dates, so I tried some large ones. No obvious issues with this:

```
;25,252,734,930
25252734930-09-12 Starbucks
   (expenses:coffee)   10

25252734931-01-01 Starbucks
   (expenses:coffee)   10
```
```
$ hledger -f a.j reg -O csv  -Y
"txnidx","date","code","description","account","amount","total"
"0","25252734930-01-01","","","expenses:coffee","10","10"
"0","25252734931-01-01","","","expenses:coffee","10","20"
```
Adding something many years apart, like
```
26111222333-01-01 Restaurant At The End Of The Universe
   (expenses:dinner)   1000
```
causes it to churn for too long, but that's expected and probably not caused by PeriodData.

Xitian9:
I added a couple of comments for slight inaccuracies, but overall it looks good to me!

simonmichael:
Thanks for the corrections, I have applied them.

I'm wondering what we should do about the [wraparound bug](https://github.com/simonmichael/hledger/pull/2477#discussion_r2422551356) with periodic reports involving far future dates. At least document it I guess. We could avoid it by using Integer or Day as map keys, giving up some time/space efficiency I guess.

Xitian9:
True, but my opinion is that 254 quadrillion years is more than sufficient for our purposes. Unless you're worried about the wraparound causing some sort of security problem, perhaps?

Xitian9:
By the way, this is the approach taken by the [thyme](https://hackage.haskell.org/package/thyme) package.

simonmichael:
> Unless you're worried about the wraparound causing some sort of security problem, perhaps?
> 

It sounds possible. It would be safer to disallow the dates we can't handle, with an error in the date parser.

It's also troubling that it's not a fixed range, but system-dependent. Eg if hledger is running in some 8 bit embedded context, the range of dates that works right in periodic reports shrinks to 1858-07-12..1859-03-24.

simonmichael:
Likewise, on a 16 bit system, the range would be 1769-02-28..1948-08-04. And if we make it an error, other dates wouldn't be supported at all.

I guess <32 bit haskell is a bit hypothetical right now but it seems a shame not to be usable if and when that's a thing.

simonmichael:
And here's the limits for other word sizes, for the record:

- On a 32 bit machine, it is about 5 million years into past and future
  (-5877752-05-08 to 5881469-05-27).
- On a 64 bit machine, the range is about 25 quadrillion years into past and future
  (-25252734927764696-04-22 to 25252734927768413-06-12).

(MicroHS can target 32 bits)

simonmichael:
A more limited safe handling strategy could be: raise an informative error from the PeriodData code if a greater-than-Int date is seen. That would allow non-periodic reports to keep working. But any use of PeriodData would be partial.

simonmichael:
For experimentation, I have pushed a version using Integer keys at 21089d186.
(And the corresponding Int version is 6578c6cde).

That fixes the bug:
```
$ hledger-int -f examples/farfuture.j reg -O csv -Y
"txnidx","date","code","description","account","amount","total"
"0","-25252734927764696-11-10","","","expenses","6","6"
$ hledger-integer -f examples/farfuture.j reg -O csv -Y
"txnidx","date","code","description","account","amount","total"
"0","25252734927768413-01-01","","","expenses","6","6"
"0","25252734927768414-01-01","","","expenses","4","10"
```

With no report interval, the Integer version seems faster (somehow..):
```
$ quickbench -n2 -N2 -w hledger-int,hledger-integer 'hledger -f examples/10ktxns-1kaccts.journal reg'
Running 2 cycles of 1 tests best of 2 times with 2 executables at 2025-10-11 07:31:04 HST:

Best of 2 times, cycle 1:
+-----------------------------------------++-------------+-----------------+
|                                         || hledger-int | hledger-integer |
+=========================================++=============+=================+
| -f examples/10ktxns-1kaccts.journal reg ||        3.15 |            3.04 |
+-----------------------------------------++-------------+-----------------+

Best of 2 times, cycle 2:
+-----------------------------------------++-------------+-----------------+
|                                         || hledger-int | hledger-integer |
+=========================================++=============+=================+
| -f examples/10ktxns-1kaccts.journal reg ||        3.26 |            3.04 |
+-----------------------------------------++-------------+-----------------+
```

And with a report interval (generating 27 subperiods in this example) I see no speed difference:
```
$ quickbench -n2 -N2 -w hledger-int,hledger-integer 'hledger -f examples/10ktxns-1kaccts.journal reg -Y'
Running 2 cycles of 1 tests best of 2 times with 2 executables at 2025-10-11 07:35:18 HST:

Best of 2 times, cycle 1:
+--------------------------------------------++-------------+-----------------+
|                                            || hledger-int | hledger-integer |
+============================================++=============+=================+
| -f examples/10ktxns-1kaccts.journal reg -Y ||        3.40 |            3.41 |
+--------------------------------------------++-------------+-----------------+

Best of 2 times, cycle 2:
+--------------------------------------------++-------------+-----------------+
|                                            || hledger-int | hledger-integer |
+============================================++=============+=================+
| -f examples/10ktxns-1kaccts.journal reg -Y ||        3.45 |            3.35 |
+--------------------------------------------++-------------+-----------------+
```

And with or without a report interval, Integer's memory use looks slightly better:
```
$ gtime hledger-int -f examples/10ktxns-1kaccts.journal reg >/dev/null
2.49user 0.23system 0:02.73elapsed 99%CPU (0avgtext+0avgdata 130400maxresident)k
0inputs+0outputs (0major+8868minor)pagefaults 0swaps
$ gtime hledger-integer -f examples/10ktxns-1kaccts.journal reg >/dev/null
2.49user 0.23system 0:02.75elapsed 99%CPU (0avgtext+0avgdata 129408maxresident)k
0inputs+0outputs (1major+8807minor)pagefaults 0swaps

$ gtime hledger-int -f examples/10ktxns-1kaccts.journal reg -Y >/dev/null
2.74user 0.24system 0:02.99elapsed 99%CPU (0avgtext+0avgdata 127472maxresident)k
0inputs+0outputs (0major+8685minor)pagefaults 0swaps
$ gtime hledger-integer -f examples/10ktxns-1kaccts.journal reg -Y >/dev/null
2.75user 0.24system 0:03.00elapsed 99%CPU (0avgtext+0avgdata 127392maxresident)k
0inputs+0outputs (0major+8681minor)pagefaults 0swaps
```

(+RTS -s shows similar results. A smaller -D report interval, generating ~500 subperiods, shows similar results).

Looking good ? Go figure!

simonmichael:
More background from #Haskell chat:

> There should be a mention in the IntMap docs that it's actually a different kind of tree, and you should test with both to see which performs better for your use case
> The implementation is based on big-endian patricia trees. (…) compared to a generic size-balanced map implementation (see Data.Map).
> (they say it was faster in their testing, but it's not actually unusual that Data.Map is faster for at least some uses)
> note in particular that the IntMap docs say that union, intersection, and insertion and deletion were faster in simple tests, but nothing about comparing lookup time

simonmichael:
Just using `Map Day a` performs even slightly better. Guessing you'll approve, I'll just go ahead and switch to that.

simonmichael:
By the way I had seen this text somewhere:

> Note the underlying representation is now for disjoint spans, whereas
> previously the end date of a span was equal to the start date of the
> next span, and then was adjusted backwards one day when needed.

I think that's out of date now; we still support only contiguous spans, right ?

Xitian9:
Technically still true (all contiguous spans are disjoint), but could certainly be strengthened to contiguous.

Issue #2479, opened 2025-10-11
1.50-1.50.2: periodic reports are wrong with dates larger than machine's word size


simonmichael:
This was a new bug added in 1.50. See https://github.com/simonmichael/hledger/pull/2477#issuecomment-3393006666 for the full discussion, but for example, on 64 bit machines:

25252734927768413-06-12 PeriodData's max date
   (expenses)   1

25252734927768414-01-01 next year past PeriodData's max date
   (expenses)   2

$ hledger reg -O csv --yearly
"txnidx","date","code","description","account","amount","total"
"0","-25252734927764696-11-10","","","expenses","1","1"

PR #2480, opened 2025-10-11
Fix 1.50's far-future periodic report bug

simonmichael:
This fixes a periodic report regression with far future dates, discussed in #2477 and #2479.

- **fix:PeriodData: use Integer keys to avoid date wraparound bugs**
- **fix:PeriodData: simplify, use Day keys**


Xitian9:
It certainly makes the code more elegant and avoids a layer of developer thought needed. An important lesson to always benchmark before you optimise.

PR #2483, opened 2025-10-17
;docs:manual: add info about the -c command line option in the commodity directy section

ooker777:

### DIFF
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -920,7 +920,7 @@ $ hledger print -c '$1.000,0'
 
 This option can be repeated to set the display style for multiple
 commodities/currencies. Its argument is as described in 
-the [commodity directive](#commodity-directive).
+the [commodity directive](#commodity-directive). Note that omitting the commodity symbol will set the display style for just the no-symbol commodity, not all commodities.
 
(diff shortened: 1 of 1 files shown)


hledger/hledger.m4.md
```diff
@@ -920,11 +920,13 @@ $ hledger print -c '$1.000,0'
...
 This option can be repeated to set the display style for multiple
 commodities/currencies. Its argument is as described in 
-the [commodity directive](#commodity-directive).
+the [commodity directive](#commodity-directive). Note that omitting the commodity symbol will set the display style for just the no-symbol commodity, not all commodities.
 
 In some cases hledger will adjust number formatting to improve their parseability
 (such as adding [trailing decimal marks](#trailing-decimal-marks) when needed).
 
+To ignore a commodity in all reports by default, try adding `not:cur:commodity-name` in your [config file](#config-files).
+
```
simonmichael: I feel this tip is a little out of place here, also I'm not sure it won't break some commands if people just naively put it at the top of the config file - I think maybe just omit it here

simonmichael (commented):
Thanks! Some comments attached

simonmichael:
Thank you!

PR #2484, opened 2025-10-23
Fix nested tables and broken HTML on export

jokesper:
A HTML export results in a table which has a stylesheet and another table nested inside. This is not valid HTML and gets auto corrected by closing the first table and opening another table. The result is the heading of the table can expand further than the remaining table.

This results in a few notable changes (as seen below):

- The date is in a far bigger cell and it's very noticeable it's centered compared to the amounts
  (we may want to right align the date, though I don't have any particular thoughts about this)
- We no longer have two consecutive lines with black background and the start
- The table is as big as the heading (+ the default left margin)

Using the following small example journal:

```journal
D 1,000.00€

account foo  ; type: A
account bar  ; type: A

2025-01-01 Test
	foo   1,000.00€
	bar  -1,000.00€
```

We get the following difference:

Old | New
---|---
Old | New

In case we do not want this new look, I still recommend fixing the tables But in this case this can simply be done by dropping the top table and only have the heading. I do not recommend this since the table not extending as far as the heading looks (especially on a longer list of accounts) rather weird.

### DIFF
--- a/hledger/Hledger/Cli/CompoundBalanceCommand.hs
+++ b/hledger/Hledger/Cli/CompoundBalanceCommand.hs
@@ -16,6 +16,7 @@ module Hledger.Cli.CompoundBalanceCommand (
 
 import Control.Monad (guard)
 import Data.Bifunctor (second)
+import Data.Foldable (traverse_)
 import Data.Function ((&))
 import Data.List.NonEmpty (NonEmpty((:|)))
 import Data.Maybe (fromMaybe, mapMaybe, maybeToList)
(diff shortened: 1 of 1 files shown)


jokesper:
Depending on what people do with the HTMLs this may also be considered breaking

simonmichael:
I'm sorry I've been slow on PRs lately; I will look at this soon. 🙏🏻

simonmichael:
Sigh, such a fine simple PR, such a slow review. Thank you @jokesper !!

PR #2487, opened 2025-10-27
aregister: respect display order when picking the first-matched account

simonmichael:
This is more consistent with the accounts and balance reports,
and allows more customisation of what 'areg SHORTNAME' matches.


simonmichael:

Issue #2427, opened 2025-07-09
accounts: in tree mode, should show parents when needed


simonmichael:
`hledger accounts --tree PATTERN` will show only the accounts matching PATTERN, indented according to their depth, but unlike `balance` it doesn't show the unmatched parents required to see the tree structure. So the output can be confusing. I think it should show those parents for clarity, even though the query doesn't match them directly.

simonmichael:
Fixed by #2488.

PR #2488, opened 2025-10-29
imp: Show parent accounts in tree mode for context

samahri:
Fixed issue #2427 where `hledger accounts --tree PATTERN` would show only matching accounts without their parent accounts, making the tree structure confusing and hard to understand.

Before:
```
$ hledger accounts --tree groceries
groceries
```

After:
```
$ hledger accounts --tree groceries
expenses
  food
    groceries
```


simonmichael:
Thanks @samahri !

PR #2496, opened 2025-11-17
;bin: sortandmergepostings: Overhaul for more robust determinism

alerque:
This is some stuff I added to the implementation of this I use for my own projects a while back. The exact sorting order is just my personal preference, but the major change that will help anybody here is that the results are much more deterministic. The previous iteration submitted here will flip-flop the order of postings if the account name and direction were the same on each run (e.g. Euro and Dollar deductions from the same account would reverse order each time the script was run).

* Avoids non-deterministic flip-flopping when the alphabetical account sort has multiple commodities
* Sorts postings commodities so commodities are in the same order across transactions
* Sorts postings with matching commodity by posting amount

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -233,12 +233,14 @@ $ watchaccounts -f time.journal client1 date:thismonth -l
 ### sortandmergepostings
 
 [`sortandmergepostings`](https://github.com/simonmichael/hledger/blob/master/bin/sortandmergepostings)
-is an adventuresome awk script intended to clean up and merge similar postings in a transaction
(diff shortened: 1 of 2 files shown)


simonmichael:
Sounds great. Any updates needed at https://hledger.org/scripts.html#sortandmergepostings or is it good enough ?

alerque:
I forgot there were extra docs there too, I just amended them a little to reflect the extra sorts.

simonmichael:
That failing test is my fault. Thanks!

PR #2502, opened 2025-12-05
fix all the scripts in ./bin. Fixes #2497

### LINKED ISSUE #2497: Do we love scripts in ./bin or do we hate them? (hledger-balance-as-budget breakage)
Hi!

I've noticed that scripts in ./bin become broken regularly (if we want a concrete example, commit
80cf1d199597038ea8e56bd1e0af90102fbd4960 removed `combineBudgetAndActual`, that two scripts depend on, and I happen to use and love them.

I also think that those same scripts are not built by CI pipeline (which would explain how breakage occurs, and goes unnoticed).

At the same time, the existence of scripts is mentioned in docs, and some, i think, even get an explicit shout-out by name.

@simonmichael , what's the best long-term plan?

- Shoud scripts be built in CI pipeline?
- Or should they be declared best-effor only, and removed on breakage?
- Should the removal of API endpoints be frowned upon if scripts depend on them, or endorsed?
- Something else, like maybe encouraging to quickly convert scripts into hledger command, even at the risk of proliferation of niche narrow-scope commands as a result?

My current issue is that 80cf1d199597038ea8e56bd1e0af90102fbd4960 is quite a large commit, and it wont be a "swap this API call for that API call kind of fix" (at a first glance). This feels sad.

adept:
I've fixed a bunch of package imports, cause ghc 9.12 does not like when `qualified` comes after the package name.

I've also re-implemented `combineBudgetWithActual` locally in the affected scripts.

The whole of ./bin should compile now (via `cd bin && ./compile.sh`)

### DIFF
--- a/bin/hledger-balance-as-budget-multi.hs
+++ b/bin/hledger-balance-as-budget-multi.hs
@@ -12,6 +12,8 @@ import Data.Text.Lazy.IO qualified as TL
 import System.Environment (getArgs)
 import Hledger.Cli.Script
 import Hledger.Cli.Commands.Balance
+import qualified Data.Map as Map
+import Data.List (sortOn)
 
 ------------------------------------------------------------------------------
 cmdmode = hledgerCommandMode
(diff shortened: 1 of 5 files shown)


simonmichael:
Ah thank you. Is it ImportQualifiedPost language pragma that's needed ? I remember converting the whole codebase to this format for easier import aligning.

simonmichael:
Hmm no this looks like something different related to PackageImports. Merging, thank you.

PR #2513, opened 2025-12-26
fix:lib:Hledger.Query: words'': Parse escaped quotes in quoted patterns

### LINKED ISSUE #649: hledger-web doesn't show accounts with apostrophes correctly
When clicking on an account (on the left hand column) that contains an apostrophe like "Mimi's Cafe", the result that appears is always zero. This applies to all sub-accounts too.

Other accounts without apostrophe like "Bella Italia" doesn't have this problem.

Thanks!

TuongNM:
Fixes #649 by parsing the escaped quotes for account 2 and 4 from https://github.com/simonmichael/hledger/issues/649#issuecomment-374239882 that hledger-web produces.

I saw `parseBooleanQuery` use the same code as `words''` for quoted patterns so I extracted the functions to reduce code duplication.

### DIFF
--- a/hledger-lib/Hledger/Query.hs
+++ b/hledger-lib/Hledger/Query.hs
@@ -86,7 +86,7 @@ import Data.Text (Text)
 import Data.Text qualified as T
 import Data.Time.Calendar (Day, fromGregorian )
 import Safe (headErr, readMay, maximumByMay, maximumMay, minimumMay)
-import Text.Megaparsec (between, noneOf, sepBy, try, (<?>), notFollowedBy)
+import Text.Megaparsec (between, noneOf, sepBy, try, (<?>), notFollowedBy, anySingle)
 import Text.Megaparsec.Char (char, string, string')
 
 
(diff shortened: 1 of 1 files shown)


simonmichael:
Seems good! Thank you @TuongNM!

simonmichael:
I plan to include this fix in the next 1.51 minor release. I realised that it is more impactful than I thought, affecting not just hledger-web, so I have done some extra testing and review. I couldn't find any impact on the CLI, but it should make both hledger-ui and hledger-web more robust with quote handling. In the release branch I have documented it in more detail, as a685c39f1.

simonmichael:
PS I think this could be a breaking change, since querying for a literal backslash now requires writing two backslashes. I'd like to understand how likely users are to notice this, but I ran out of steam for testing it, any help is welcome.

simonmichael:
Cc'ing the description:

> fix!: ui,web,lib: handle single/quotes in data more robustly [https://github.com/simonmichael/hledger/issues/649], [https://github.com/simonmichael/hledger/pull/2513] (Tuong Nguyen Manh, Simon Michael)
> This is a more detailed description for Tuong Nguyen Manh's commit in master:
> https://github.com/simonmichael/hledger/commit/de900e5b24fa4fbd4c6aa17814e6711f153842c5 fix:lib:Hledger.Query: words'': Parse escaped quotes in quoted patterns
> 
> In Hledger.Query, words'' and parseBooleanExpression now allow
> backslash-escaping of single quotes within a single-quote-enclosed string,
> or of double quotes within a double-quote-enclosed string.
> In fact they allow backslash-escaping of any character.
> 
> This seems not to have much visible impact on the CLI, but it should
> fix some issues when working with data containing single or double quotes,
> in hledger-web, hledger-ui and the hledger-lib API.
> Eg, when viewing this with hledger-web:
> 
> 2025-12-21 "little" john | john's apples
>     "borrowed" food
>     gifts:will o'gamwell                           2 apple
>     assets                                         2 apple
> 
> clicking the '"borrowed" food' account in the sidebar now properly
> shows that account's register (bug https://github.com/simonmichael/hledger/issues/649, reported in 2017-11-21).
> 
> And generally in hledger-ui and hledger-web's search/filter fields it
> should now be possible to search for text containing quotes,
> by backslash-escaping the quotes.
> 
> Note since backslash in queries now has special meaning,
> to query for a literal backslash, you should write \\ rather than \.
> This might be a visible breaking change for some.

TuongNM:
I've tested it a bit in hledger-web with an account that's just a backlash:

```
2026/1/1
    (\)  1
```

- Clicking on it works since it creates a query without quotes: `inacct:\ `
- Putting it in quotes doesn't work anymore: `inacct:"\"`
- As you've described, you now have to escpae the backslash: `inacct:"\\"`

I guess it depends on how someone uses hledger-web to notice this. If you just click on the account links, then you probably won't notice this much. If you input your queries manually, then you might notice this change.

TuongNM:
During further testing I did notice that I introduced a bug. Using account 3 from the comment in the original issue as an example and adding a backslash:

```
2018/1/3
    (3:expenses:le'shan's \cafe)  1
```

With my changes if there are backslashes in the account name, hledger-web won't display results. For the above hledger-web creates the query `inacct:"3:expenses:le'shan's \cafe" ` that now fails to find the account.
The backslash now needs to be escaped: `inacct:"3:expenses:le'shan's \\cafe" `

One way to fix this would be to compose `escapeDoubleQuotes` with `escapeBacklash`.

So turn this:
https://github.com/simonmichael/hledger/blob/1a5f10f928a1715631860c7009be5b39079ca240/hledger-lib/Hledger/Utils/Text.hs#L140

into this:
```
escapeDoubleQuotes = T.replace "\"" "\\\"" . escapeBackslash
```

But I'm not sure if `escapeDoubleQuotes` should escape backslashes generally or if that should only be done inside hledger-web.

simonmichael:
@TuongNM I appreciate this extra testing. Quoting can be very tricky. I think this would be noticed and would have to be handled as a regression. So I'm going to revert it in master and release branch for the moment.

PR #2518, opened 2025-12-31
fix #2514, and improve auto-creating behaviour

### LINKED ISSUE #2514: 1.50.3-1.51.1: add with no journal and no options gives an error, should pick default path
In hledger 1.50.3-1.51.1, when LEDGER_FILE is unset and ~/.hledger.journal does not exist, `hledger add` fails instead of creating that file. This is more fallout from the recent `include`/glob changes.

```
~$ HOME=/tmp LEDGER_FILE= hledger add
hledger: Error: neither LEDGER_FILE nor "/tmp/.hledger.journal" was found
```

simonmichael:
Fix #2514, and improve auto-creating behaviour.

- **fix:add,import: autocreate missing journal files again (but later) [#2514]**
- **imp: -f now errors if given a glob matching no files, like LEDGER_FILE**
- **imp:add,import: also create directories when autocreating journal file**

### DIFF
--- a/hledger-lib/Hledger/Read.hs
+++ b/hledger-lib/Hledger/Read.hs
@@ -151,9 +151,9 @@ import Data.Text qualified as T
 import Data.Text.IO qualified as T
 import Data.Time (Day)
 import Safe (headDef)
-import System.Directory (doesFileExist)
+import System.Directory (createDirectoryIfMissing, doesFileExist)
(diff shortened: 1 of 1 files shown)


simonmichael:
Now included in the https://github.com/simonmichael/hledger/releases/tag/testbin binaries.

PR #2519, opened 2026-01-03
feat: prices: add the --quoted option to filter prices by the commodity they're quoted in

basvandijk:
If you have prices of a commodity quoted in multiple currencies like:

```
hledger prices cur:EUR -p 2025-12-31              
P 2025-12-31 EUR CHF 0.93060
P 2025-12-31 EUR USD 1.17450
```

It's useful to be able to filter them on the currency they're quoted in:

```
hledger prices cur:EUR -p 2025-12-31 --quoted CHF
P 2025-12-31 EUR CHF 0.93060
```

```
hledger prices cur:EUR -p 2025-12-31 --quoted USD 
P 2025-12-31 EUR USD 1.17450
```

I'm not sure about the option name. `--quoted-in` might read better.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Prices.hs
+++ b/hledger/Hledger/Cli/Commands/Prices.hs
@@ -27,6 +27,7 @@ pricesmode = hledgerCommandMode
   [flagNone ["costs"]                (setboolopt "infer-market-prices") "deprecated, use --infer-market-prices instead"
   ,flagNone ["inverted-costs"]       (setboolopt "show-reverse") "deprecated, use --show-reverse instead"
   ,flagNone ["infer-reverse-prices"] (setboolopt "show-reverse") "deprecated, use --show-reverse instead"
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks!

How necessary is this, compared to alternatives like `hledger prices cur:EUR | grep CHF` ?

simonmichael:
Ping. I'm wondering if this feature is needed.

basvandijk:
> How necessary is this, compared to alternatives like hledger prices cur:EUR | grep CHF ?

In my scripts I currently indeed do the filtering outside of hledger so it's not like not having this feature is blocking me from anything. I was just surprised this didn't exist because most other commands in hledger allow you to filter on the currency. 

But this might not be worth the complexity.

simonmichael:
I might be wrong, but it feels a little bit unintuitive and not that much needed. I think I'll hold off for now. We might hear more support, or think of another solution. (A new query type, `othercur:`, useful in other contexts ? Probably more complicated.)

basvandijk:
Works for me. Thanks.

PR #2527, opened 2026-01-13
fix just action to generate sample journals

markokocic:
fix just samplejournals [#2528]

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -478,6 +478,7 @@ SHELLTEST := STACK + ' exec -- shelltest --execdir --threads=32'
 
 # build hledger warning-free and run functional tests, with any shelltest OPTS. (after mktestaddons)
 @functest *STOPTS:
+    #!/usr/bin/env bash
     $STACK build --ghc-options=-Werror hledger
     time (({{ SHELLTEST }} --exclude=/_ --hide {{ if STOPTS == '' { '' } else { STOPTS } }} \
         hledger/test/ bin/ \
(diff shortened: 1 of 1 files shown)


markokocic:
I use Debian stable on WSL.

I tried executing the `bash` command before running `just` and setting the `SHELL` variable to /usr/bin/bash, but for some reason `just` decides to run these snippets using sh, which fails.

I also tried to put a space between `(`, but I get the same error as before.

The only way to make it work is by forcing it to use bash by putting `#!/usr/bin/env bash`

`sh` version on my system is the default one that comes with Debian stable. It's a symlink to `dash`:
``` sh
marko@host ~/s/c/hledger (just) [2]> ls -l /bin/sh
lrwxrwxrwx 1 root root 4 Feb  4  2025 /bin/sh -> dash*
marko@host ~/s/c/hledger (just)> ls -l /usr/bin/sh
lrwxrwxrwx 1 root root 4 Feb  4  2025 /usr/bin/sh -> dash*
marko@host ~/s/c/hledger (just)> dash -version
dash: 0: Illegal option -r
marko@host ~/s/c/hledger (just) [2]> dpkg -l dash
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-=================================
ii  dash           0.5.12-12    amd64        POSIX-compliant shell
```

simonmichael:
I see; just isn't affected by the shell you run it from; it always uses `sh` as its default shell. And on your system that is `dash`, which doesn't like something about that recipe.

I think the best thing is to make it use bash as default shell (and require bash to be installed). I think we just need to add `set shell := ["bash", "-uc"]` at the top, could you confirm ?

simonmichael:
I went ahead and did that. If issues remain, let me know.

PR #2529, opened 2026-01-13
Preserve and print Ledger-style lot annotations

simonmichael:
- **dev: add a CostBasis field to Amount**
- **feat: preserve Ledger-style lot info, and show it in print's json & txt output**
- **feat:print: show lot info in beancount output; update docs**

### DIFF
--- a/examples/investing/export-lots-workflow/beancount-gains-howto.md
+++ b/examples/investing/export-lots-workflow/beancount-gains-howto.md
@@ -1,0 +1,249 @@
+# How to Report Realized Gains with Beancount
+
+Generated: 2026-01-18, Claude
+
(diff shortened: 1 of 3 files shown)


simonmichael:
It seems to work; any review is welcome.
I haven't tested for time/space impact of the new field in Amount.

simonmichael:
I tested performance with
```
j bench -w hledger,hledger-cb -n2 -N2
j bench-throughput hledger; j bench-throughput hledger-cb
j perftest
```
and all looks normal. Also, I've added 

- [examples/investing/export-lots-workflow](https://github.com/simonmichael/hledger/tree/master/examples/investing/export-lots-workflow), a how-to for lot reporting with other apps
- in the same directory, claude-generated general how-tos for gains reporting with ledger/beancount/rustledger

PR #2536, opened 2026-01-20
Small Justfile cleanup

markokocic:


simonmichael:
The Justfile has had some mac-isms and sm-isms for my convenience over the past years. Thanks for the fixes.

Issue #2528, opened 2026-01-13
just test/functest/doctest issues


markokocic:
I had multiple issues when trying to run `just functest`

1. it requires /usr/bin/time - hard to spot, because the error message is not clear, since my fish shell already have time as built-in. Figured out that I need to install time package only after switching temporarily to bash.
2. looks like there's an issue with Just syntax for functtest and perftest:
```
marko@host ~/s/c/hledger (just)> just functest
hledger-1.51.99: unregistering (local file changes: /home/marko/src/cvstree/hledger/.git/index)
hledger> build (lib + exe) with ghc-9.12.2
Preprocessing library for hledger-1.51.99...
Building library for hledger-1.51.99...
[ 6 of 42] Compiling Hledger.Cli.CliOptions [/home/marko/src/cvstree/hledger/.git/index changed]
[ 9 of 42] Compiling Hledger.Cli.Commands.Stats [/home/marko/src/cvstree/hledger/.git/index changed]
Preprocessing executable 'hledger' for hledger-1.51.99...
Building executable 'hledger' for hledger-1.51.99...
hledger> copy/register
Installing library in /home/marko/src/cvstree/hledger/.stack-work/install/x86_64-linux-tinfo6/4adf8bb2b6ddceca88962ce280ef3f37ec29c6c8a4e6c37418e12d788ac807ed/9.12.2/lib/x86_64-linux-ghc-9.12.2-167a/hledger-1.51.99-7LzA5FA3oRZHPpFCYRyB0k
Installing executable hledger in /home/marko/src/cvstree/hledger/.stack-work/install/x86_64-linux-tinfo6/4adf8bb2b6ddceca88962ce280ef3f37ec29c6c8a4e6c37418e12d788ac807ed/9.12.2/bin
Registering library for hledger-1.51.99...
sh: 1: Syntax error: "(" unexpected (expecting ")")
error: Recipe `functest` failed on line 486 with exit code 2
marko@host~/s/c/hledger (just) [2]> just --version
just 1.46.0
marko@host~/s/c/hledger (just)>
```

simonmichael:
https://github.com/casey/just?tab=readme-ov-file#shell says just's default shell `sh -cu`.  Is that fish on your system ? 

I can run `just functest` and `just perftest` without error from osh, bash, zsh or fish shells here.

markokocic:
My default shell is `fish` shell, but I also tried with `bash` and `sh`, the results are the same.
[...]

PR #2538, opened 2026-01-21
feat: register: Implement --drop flag mirroring functionality

alerque:
This implements the same --drop functionality as balance command for register subcommand, allowing users to trim leading account name components from register output.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Register.md
+++ b/hledger/Hledger/Cli/Commands/Register.md
@@ -17,6 +17,7 @@ Flags:
                             description closest to DESC
   -r --related              show postings' siblings instead
      --invert               display all amounts with reversed sign
+     --drop=N               omit N leading account name parts
      --sort=FIELDS          sort by: date, desc, account, amount, absamount,
                             or a comma-separated combination of these. For a
                             descending sort, prefix with -. (Default: date)
@@ -71,6 +72,8 @@ $ hledger register checking -b 2008/6 --historical
 ```
 
 The `--depth` option limits the amount of sub-account detail displayed.
+
+The `--drop` option will trim leading segments from account names.
 
 The `--average`/`-A` flag shows the running average posting amount
 instead of the running total (so, the final number displayed is the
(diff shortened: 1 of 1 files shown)


simonmichael:
Nice! Would you mind mentioning it briefly in the body of Register.md also. (The flags list at the top will be regenerated before release, so updating that is optional.)

alerque:
Something like that?

simonmichael:
Great. There's still a build warning; hopefully `just functest` reproduces it on your machine. (If not, try a `stack clean --full` first.)

alerque:
Unfortunately none of the `just` tasks or default invocations of `stack` work for me at all (due to old `ghc`, lack of `ghcid`, etc.) so part of the reason to get the PR open was to get CI results.

simonmichael:
> Unfortunately none of the just tasks or default invocations of stack work for me at all (due to old ghc

Could you help me understand this kind of situation better ? What ghc version(s), installed by what method, do you have ?

simonmichael:
PR looks good - thank you!

alerque:
> Could you help me understand this kind of situation better ? What ghc version(s), installed by what method, do you have ?

GHC 9.6.6 installed by system packages on Arch. Yes it is ancient, yes there is work being done to update it, but our whole Haskell stack being dynamically linked is a huge pain point. GHCID is just not packaged at all, probably because it wasn't compatible with whatever ghc we had a while back and had to get dropped to unblock a ghc update. I see it does actually build at the moment. I'll check on it again when the GHC stack gets fully up to date and see if it can get added.

simonmichael:
Ah Arch. (It's always Arch. :) But, GHC 9.6 is still supported by hledger, so it shouldn't be a big deal to use it. It might be enough to change this line in Justfile (make it use stack96.yaml):
```
#STACK := 'stack --stack-yaml=stack912.yaml'
```

ghcid is not something you need, probably. I use it only when HLS in VS Code isn't working.  

We do want it to be easy for everyone to be able to run `just test` or `just functest` though.

simonmichael:
Related: recent discussion of Justfile/test portability and dev docs at #2528.

PR #2541, opened 2026-01-23
feat: areg: Add --matching flag

### LINKED ISSUE #2540: Add a flag for areg to show the queried account, not the *other* accounts
Consider the following ledger:

```ledger
2020-01-01 Vendor A
  Expenses:Foo                                        $1000.00
  Expenses:BankFee                                       $5.00
  Assets:Account

2020-01-01 Vendor B
  Expenses:Bar                                        $1000.00
  Expenses:BankFee                                       $5.00
  Assets:Account
```

Here is what `register` puts out:

```console
$ hledger reg Expenses
2020-01-01 Vendor A              Expenses:Foo               $1000.00      $1000.00
                                 Expenses:BankFee              $5.00      $1005.00
2020-01-01 Vendor B              Expenses:Bar               $1000.00      $2005.00
                                 Expenses:BankFee              $5.00      $2010.00
```

This is accurate and expected, but for some purposes pretty noisy. If I'm trying to review expenses, having the bank fees split out separately is not always a useful view, but if you were to filter them out with `not:BankFee` then the running totally would get off pretty quick.

For this use case, `aregister` is the closer tool for the job:

```console
$ hledger areg Expenses
2020-01-01 Vendor A              As:Account                 $1005.00      $1005.00
2020-01-01 Vendor B              As:Account                 $1005.00      $2010.00
```

This is less noisy and has an accurate running total, but we've lost the relevant sub account information.

The `--related` flag for the regular register does something interested too:

```console
$ hledger reg Expenses --related
2020-01-01 Vendor A              Assets:Account            $-1005.00     $-1005.00
2020-01-01 Vendor B              Assets:Account            $-1005.00     $-2010.00
```

This also groups the postings from each transactions resulting in just two register entries and an accurate running total. But still not the relevant account info. What I really want is something equivalent to the reverse of `--related` that would match the account tree I actually queried for *not the other accounts*, sort of like this:

```console
$ hledger areg Expenses --someflagnamehere
2020-01-01 Vendor A              Expenses:Foo               $1005.00      $1005.00
2020-01-01 Vendor B              Expenses:Bar               $1005.00      $2010.00
```

Unfortunately I can't get this using `areg` by querying account because there are other things going on there and multiple sources. Real life isn't as simple as this MWE. I also can't just merge the postings in the actual ledger data because I *do* need it split out for other purposes.

alerque:
Closes #2540.

Works like the inverse of the --related flag on register. In fact there is probably some better name for this flag that somehow connects it to that functionality.

This is a POC at this point, I haven't evalutaed whether this affects anything else.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Aregister.hs
+++ b/hledger/Hledger/Cli/Commands/Aregister.hs
@@ -55,6 +55,7 @@ aregistermode = hledgerCommandMode
   -- ,flagNone ["average","A"] (setboolopt "average")
   --    "show running average of posting amounts instead of total (implies --empty)"
   -- ,flagNone ["related","r"] (setboolopt "related") "show postings' siblings instead"
+  ,flagNone ["matching"] (setboolopt "matching") "show the matched account instead of related accounts"
   ,flagNone ["invert"] (setboolopt "invert") "display all amounts with reversed sign"
   ,flagReq  ["heading"] (\s opts -> Right $ setopt "heading" s opts) "YN"
      "show heading row above table: yes (default) or no"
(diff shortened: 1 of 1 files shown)


simonmichael:

PR #2542, opened 2026-01-24
Fix double dash

schoettl:
This fixes this problems:

- discard `--` from arguments to `add` command

It does not fix the that `-5` gets translated `--depth 5` even after `--`.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Add.hs
+++ b/hledger/Hledger/Cli/Commands/Add.hs
@@ -123,7 +123,7 @@ add opts j
         showHelp
         let today = opts^.rsDay
             state = defAddState{asOpts=opts
-                              ,asArgs=listofstringopt "args" $ rawopts_ opts
+                              ,asArgs=dropFirstDoubleDash $ listofstringopt "args" $ rawopts_ opts
                               ,asToday=today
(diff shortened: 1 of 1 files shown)


schoettl:
- [x] run hledger tests locally
- [x] `just functest` – 5 failed test cases on NixOS, same on master
- [ ] how about other commands?

I think the problem that double dash is passed as arg to some commands is hard to fix globally because the args are taken within the command's function (e.g. within `add :: CliOpts -> Journal -> IO ()` here).

schoettl:
> It does not fix the that `-5` gets translated `--depth 5` even after `--`.

Looks like this isn't a problem at all:

```
$ hledger bal assets: -- -1
          123 €  assets:bank:yyy
          456 €  assets:prepaid:gifts
--------------------
          123456 €  

```

So in this example, `-1` isn't interpreted as option because it's after a double dash. Hence it isn't converted to `--depth 1`.

But the `balance` command as well misinterprets the double dash. It should be discarded there, too.

simonmichael:
Sorry for losing track of this. Could you explain the original problem, or link the original bug report ?

simonmichael:

PR #2543, opened 2026-01-24
Add --drop flag to aregister

alerque:
Pretty much the same as #2538, but for aregister as well.


simonmichael:
Sorry I lost track of this. Sounds good!

Issue #2544, opened 2026-01-25
"Negative number not allowed in -z" error (unclear error from less)


sutehachi:
OS: macOS 26.2
hledger version: hledger 1.51.2, mac-aarch64

All hledger command shows following message.
```
$ hledger
Negative number not allowed in -z
Press RETURN to continue
```

simonmichael:
Goodness. What do these say

`hledger --version`
`which hledger`
`/usr/bin/env hledger --version`
`hledger -n`
`hledger setup`

sutehachi:
I hope these help you.

$ hledger --version
hledger 1.51.2, mac-aarch64

$ which hledger
/opt/homebrew/bin/hledger

$ usr/bin/env hledger --version
hledger 1.51.2, mac-aarch64

$ hledger -n
Negative number not allowed in -z
Press RETURN to continue

$ hledger setup
Checking your hledger setup..
Legend: good, neutral, unknown, warning

hledger
* is a released version ?                  yes  hledger 1.51.2, mac-aarch64
* is up to date ?                          yes  1.51.2 installed, latest is 1.51.2
* is a native binary for this machine ?    yes  aarch64
* is installed in PATH (this version) ?    yes  /opt/homebrew/bin/hledger
* has a system text encoding configured ?  yes  UTF-8, data files must use this encoding
* has a user config file ?                  no
* has a local config file ?                 no

terminal
* the NO_COLOR variable is defined ?        no
* --color is configured by config file ?    no
* hledger will use color by default ?      yes
* the PAGER variable is defined ?          yes  less
* --pager is configured by config file ?    no
* hledger will use a pager when needed ?   yes  /opt/homebrew/bin/less
* the LESS variable is defined ?           yes
* the HLEDGER_LESS variable is defined ?    no
* adjusting LESS variable for color etc. ? yes
* tables will use box-drawing chars ?       no  you can use --pretty to enable them

journal
* the LEDGER_FILE variable is defined ?     no
* a default journal file is readable ?     yes  /Users/apy/.hledger.journal
* it includes additional files ?           yes  27 files
* all commodities are declared ?           yes  12 commodities
* all accounts are declared ?              yes  133 accounts
* all accounts have types ?                 no  4 accounts without types
* accounts of all types exist ?            yes  ALERXCV accounts detected
* commodities/accounts are being checked ?  no  you can use -s to check them
* balance assertions are being checked ?   yes  you can use -I to ignore them

simonmichael:
[...]

PR #2546, opened 2026-01-30
run less more robustly; don't add --quit-at-eof

simonmichael:
This adds more robust handling of various less and pager failure mode. Generally when anything goes wrong, it should always fall back to printing the output unpaged.

Based on feedback from chat, I have removed --quit-at-eof from the options that hledger adds to LESS when using less as pager. It meant that when you scrolled to the bottom of long output, you would often end up exiting less by accident, so you couldn't scroll back up. Now when less is invoked (because output is longer than the window), you'll always see its end of output indicator at the bottom and you'll always have to press q to exit. 

Of course you can override this default by setting your preferred less options in HLEDGER_LESS (or by setting PAGER to something else or by using --pager=no).

Related: #2544


simonmichael:
This is in today's https://github.com/simonmichael/hledger/releases/tag/testlatest binaries, all testing including on old environments is welcome.

PR #2550, opened 2026-02-08
fix:ui: Display All Transactions correctly

### LINKED ISSUE #2476: UI won't display "All transactions"
Starting `hledger-ui`, then going to "All  accounts", then tapping `-` to reduce the depth until at level 0 and "All" is the only thing left, then hitting enter to see the transaction list shows ... dots. Just a page full of dots, and "all transactions (1/0)" as a header.

At first I thought this was a regression I caused in #2473, but it turns out this is in the last released version as well.

TuongNM:
Even if there there are no displayitems left, there are still blankitems in the List. With this the List is not technically empty and will therefore keep showing a selection.
This is fixed by actively removing the list selection if there are no displayitems left.

Also, if there are only blankitems left, the colwidths cacluation in the RegisterScreen produces negative widths. The fitText function would then see that the 0-length of the blankitems is greater than the negative widths and ellipsify.
This is fixed in the RegisterScreen by ensuring that no negative widths are passed and more generally in the fitText function by treating negative max widths as 0.

Fixes #2476

### DIFF
--- a/hledger-lib/Hledger/Utils/Text.hs
+++ b/hledger-lib/Hledger/Utils/Text.hs
@@ -211,7 +211,7 @@ fitText mminwidth mmaxwidth ellipsify rightside = clip . pad
     clip s =
       case mmaxwidth of
         Just w
-          | realLength s > w ->
+          | realLength s > max 0 w ->
             if rightside
               then textTakeWidth (w - T.length ellipsis) s <> ellipsis
               else ellipsis <> T.reverse (textTakeWidth (w - T.length ellipsis) $ T.reverse s)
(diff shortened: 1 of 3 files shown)


TuongNM:
Technically, it's only necessary to either not pass negative widths in the RegisterScreen or have fitText treat negative max widths as 0. Both instances seemed unintended to me so I adjusted them both. But if there is a reason I'm unaware of where one of these behaviors needs to be kept as is, it would be possible.

simonmichael:
This was merged, as b0383bfd2 (added the pr number to the message). See also discussion on #2476. Thanks!

Issue #2168, opened 2024-02-09
review/update hledger-ui theme names/selection


simonmichael:
Related to #2167: I find the existing theme names a bit unclear. Here are my notes:

- default: normally shows me a light background and some colour. But when run in a VS Code terminal, shows me a darkish gray background. Would be nice if it were more consistent, and then it could be renamed to light ?
- greenterm: shows monochrome green text on black background. Rename to greenonblack ?
- terminal: shows monochrome text using the terminal's current foreground/background colours. Is there a clearer name for  this - minimal, transparent, mono.. ?
- dark (added by #2167): shows dark gray text on black background, and a little colour for selection and negatives.

jmtd:
It may be worth ensuring we have at least one theme that honours the user's terminal colour choices (i.e., a theme which, if they've got light-text-on-dark, so does the theme, and vice-versa); at least one defined in indexed colours  (i.e. colours 0-7) which might be redefined by the user* ; and one or more  that uses specific colour values (24 bit?) so it looks consistent no matter the terminal settings.

I'm not very experienced with Brick (yet) so I don't know whether it gives us sufficient control. I recently read this blog post -- [I Just Wanted Emacs to Look Nice — Using 24-Bit Color in Terminals](https://chadaustin.me/2024/01/truecolor-terminal-emacs/) -- which was a nice summary of how this all works in Today's terminals.

* I'm fairly sure it's only the 4-bit colours and the greyscale ramp that are typically overriden by users who specify  custom palettes in terminals. I just tried cycling through the palette options in gnome-terminal whilst watching the output of `256colors2.pl` and only the 0-7 indexed "system colours" and grayscale ramp changed, not the colour cube (=8 bit)

jmtd:
@jmtd
>> terminal
> Perhaps a synonym or reference to the fact it delegates the colour decision to the terminal. I'll ponder on that.

"inherit" is the best I've got so far.

jmtd:
> default: normally shows me a light background and some colour. But when run in a VS Code terminal, shows me a darkish gray background. Would be nice if it were more consistent, and then it could be renamed to light ?

[...]

Issue #1911, opened 2022-08-10
hledger-ui bigger jumps (like 10 rows at a time) with K and J


aron-87:
My journal has 50 accounts, scrolling from upper accounts (assets) to lower (income, liabilities) in hledger-ui takes forever. It would be  nice to jump 10 rows at a time with `Shift + k`, `Shift + j`. Currently these hotkeys are not used.

simonmichael:
That might be nice. Adding more keys to remember can be a trade-off. Do you know about page-up/page-down ?

simonmichael:
Also, scrolling with mouse wheel/touchpad drag ?

simonmichael:
I also envisioned a "jump to item" feature, like searching with `/` except instead of filtering it would jump to the next matching item..

aron-87:
The idea is to not lift fingers from `h,j,k,l` keys I'm already on.  PgUp/PgDn are far away, mouse is even more far away.

Fuzzy search would be cool too (it is already mentioned in wishes somewhere around here).

PR #2551, opened 2026-02-09
fix:ui: implement J/K jumps and rename default theme to light [#1911] [#2168]

RahulShankarV52:
This PR addresses the UI and navigation issues to the TUI experience:

#1911: Implemented 10-row jumps using capital J and K in both the AccountsScreen and RegisterScreen.

#2168: Renamed the "default" theme to "light" for better consistency with the "dark" theme option.

#2175: Updated selection bar highlights to Cyan for better contrast in both light and dark terminal profiles.

Tested locally with linux mint OS.


simonmichael:
Sorry for losing track of this. Three separate PRs would have been ideal, but this is ok. But we need to make sure that the J/K keys work on all list screens.

simonmichael:
Actually, since I might need to release soon, I'll merge then check that. Thank you!

RahulShankarV52:
No Probs. You're Welcome

Issue #2459, opened 2025-09-13
hledger-web 1.50 doesn't follow own openapi spec


aragaer:
The #2395 had removed `aibalance`, `aebalance` and `anumpostings` fields from Account class and from its toJSON representation. However all these three fields are declared as required in the current openapi.yaml.

Also the `aboring` field is still present in the class, but not in the json representation and it's also a required field according to the schema.

simonmichael:
https://github.com/simonmichael/hledger/blob/master/hledger-web/config/openapi.yaml needs updating for 1.50, thanks for the report.

PR #2554, opened 2026-02-26
doc:web: update openapi.yaml for 1.50

n0vdd:
Update the OpenAPI specs because I wanted to try the web module and saw #2459.

### DIFF
--- a/hledger-web/config/openapi.yaml
+++ b/hledger-web/config/openapi.yaml
@@ -111,6 +111,7 @@ paths:
                               required:
                               - acommodity
                               - acost
+                              - acostbasis
                               - aquantity
                               - astyle
                               properties:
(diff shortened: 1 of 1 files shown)


simonmichael:
Thanks for the update! I would prefer it without the CHANGES.md changes. I update those as part of the release process.

n0vdd:
Now is it right?

simonmichael:
Not quite; you have added a commit deleting CHANGES.md. You should have reverted the CHANGES.md part of the original commit and amended that commit with that reversal (it's hard to describe since there are lots of git tools). End result: a single commit just changing openapi.yaml. Perhaps easier: abandon both commits and redo the openapi one, then force-push to this PR branch.

n0vdd:
Thank you for your patience; I am not used to doing this kind of stuff.
Now I think it is correct

n0vdd:
One minor doubt, if you do not care: was this the correct way to update the docs, or should I have done it on the original open issue?

simonmichael:
Looks good - thank you! A pull request like this was the best way.

PR #2560, opened 2026-03-03
;fix: fancyassertions: Nicer show call

Guvante:
- As part of its output fancyassertions writes relevant balances: and then the balances that caused the failure
- Now it uses the showMixedAmount function instead to properly format the amount as expected by users


simonmichael:
Thanks!

PR #2569, opened 2026-03-28
;feat: fancyassertions: Add -s/--strict

Guvante:
- Adding a strict flag to match normal hledger commands

### DIFF
--- a/bin/hledger-check-fancyassertions.hs
+++ b/bin/hledger-check-fancyassertions.hs
@@ -117,7 +117,7 @@ import Hledger.Data qualified as H
 import Hledger.Query qualified as H
 import Hledger.Read qualified as H
 import Hledger.Utils.Parse qualified as H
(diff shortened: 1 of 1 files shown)


Guvante:
- Skipped CI since there is a TODO to test these addon scripts
- I like to run an Hledger script as a "build" command to verify I didn't mess anything up, I swapped to fancyassertions since it allows some powerful extra checks (like make sure my checking account won't empty with the upcoming transactions...) but it was missing the strict checks
- I could accomplish this with two passes but that seems wasteful and this seemed generally useful

simonmichael:
It would be good to give this more of the standard hledger command-line parsing, but I'll take this! (Soon.) Thank you, it's good to see check-fancyassertions is used and getting some maintenance.

simonmichael:
Merged as fcf42a61e. Thanks!

PR #2580, opened 2026-05-06
imp: support zed & sublime edit file at position

zbrox:
This adds support for opening a file from the UI at the right position if one is using Zed or Sublime Text.

I didn't see a mention about this in other parts of the documentation so I haven't added anything extra in the docs.


simonmichael:
Thanks!

PR #2582, opened 2026-05-09
Widen balance amount column for long commodity names

### LINKED ISSUE #1148: Inconsistent alignment for currencies with long names
Hi!

## Situation

I am using `hledger` for inventory/asset management in my research project. It works great to keep track of where _categories_ of things are and to a degree even for tracking _individual_ objects (via identifying tags). The best solution for my application turned out to be **quoted long currency names** for _types_ of things (better readability and searchability via `cur:'.*keyword.*'`).

## Problem

However, `hledger balance` seems to have a hard-coded (minimum?) currency column width which works fine for short currencies (like `USD` or `EUR`) and otherwise calculate the currency indentation per-account. This leads to weird indentations and breaks the nice accounts tree view. I didn't find anything concerning this in the [online documentation](https://hledger.org).

### Example

Cooked-up example journal `test.hledger`:

```ledger
2019/12/09 Outdoor
    [room:workshop]      5 "80 watt solar panel" @@ 500.00 EUR
    [room:workshop]      1 "rugged outdoor laptop" @@ 1000.00 EUR
    [store]

2019/12/09 Car
    [room:garage]     4 "car tires" @@ 300.00 EUR
    [store]
```

```sh
lsb_release -a
# Distributor ID:	Ubuntu
# Description:	Ubuntu 18.04.3 LTS
# Release:	18.04
# Codename:	bionic
# docker image https://hub.docker.com/r/dastapov/hledger
hledger --version
# hledger 1.16.1
hledger -f test.hledger balance
#   5 "80 watt solar panel"
#             4 "car tires"
# 1 "rugged outdoor laptop"  room
#        4 "car tires"    garage
#   5 "80 watt solar panel"
# 1 "rugged outdoor laptop"    workshop
#         -1800.00 EUR  store
# --------------------
#   5 "80 watt solar panel"
#              -1800.00 EUR
#             4 "car tires"
# 1 "rugged outdoor laptop"
```

I'd rather expect an output like this (all currencies indented like in the total summary):

```bash
hledger -f test.hledger balance
#   5 "80 watt solar panel"
#             4 "car tires"
# 1 "rugged outdoor laptop"  room
#             4 "car tires"    garage
#   5 "80 watt solar panel"
# 1 "rugged outdoor laptop"    workshop
#              -1800.00 EUR  store
# --------------------
#   5 "80 watt solar panel"
#              -1800.00 EUR
#             4 "car tires"
# 1 "rugged outdoor laptop"
```

## Solution

For example:

- calculating the longest currency name in advance and using that as a currency column size
- let user specify a forced currency column width via command-line flag

> PS: `hledger` is really awesome!

adamsardo:
Fixes #1148.

## Summary
- widen the default single-column balance amount field to the widest visible rendered amount
- keep the historic 20-character minimum for ordinary reports
- add a shell regression for long quoted commodity names in flat balance output

## Tests
- `stack build hledger:exe:hledger`
- `stack exec -- shelltest --execdir --hide hledger/test/balance/1148.test`
- `stack exec -- shelltest --execdir --hide hledger/test/balance/1148.test hledger/test/balance/layout.test hledger/test/balance/balance.test`

### DIFF
--- a/hledger/Hledger/Cli/Commands/Balance.hs
+++ b/hledger/Hledger/Cli/Commands/Balance.hs
@@ -491,23 +491,55 @@ balanceReportAsCsv opts =
 
 -- | Render a single-column balance report as plain text.
 balanceReportAsText :: ReportOpts -> BalanceReport -> TB.Builder
-balanceReportAsText opts ((items, total)) = case layout_ opts of
+balanceReportAsText originalopts report@((items, total)) = case layout_ originalopts of
     LayoutBare | iscustom -> error' "Custom format not supported with commodity columns"  -- PARTIAL:
-    LayoutBare -> bareLayoutBalanceReportAsText opts ((items, total))
+    LayoutBare -> bareLayoutBalanceReportAsText originalopts report
     _ -> unlinesB ls <> unlinesB (if no_total_ opts then [] else [overline, totalLines])
   where
+    opts = widenDefaultBalanceLineFormat originalopts report
     (ls, sizes) = unzip $ map (balanceReportItemAsText opts) items
     -- abuse renderBalanceReportItem to render the total with similar format
-    (totalLines, _) = renderBalanceReportItem opts ("",0,total)
+    (totalLines, _totalSizes) = renderBalanceReportItem opts ("",0,total)
     -- with a custom format, extend the line to the full report width;
-    -- otherwise show the usual 20-char line for compatibility
-    iscustom = case format_ opts of
+    -- otherwise show the amount column's width.
+    iscustom = case format_ originalopts of
         OneLine       ((FormatField _ _ _ TotalField):_) -> False
         TopAligned    ((FormatField _ _ _ TotalField):_) -> False
         BottomAligned ((FormatField _ _ _ TotalField):_) -> False
         _ -> True
-    overlinewidth = if iscustom then sum (map maximum' $ transpose sizes) else 20
+    isdefault = format_ originalopts == defaultBalanceLineFormat
+    overlinewidth
+      | iscustom  = sum (map maximum' $ transpose sizes)
+      | isdefault = defaultBalanceAmountColumnWidth originalopts report
+      | otherwise = 20
     overline   = TB.fromText $ T.replicate overlinewidth "-"
+
+-- | The default balance format historically used a 20-character amount
+-- column. Keep that as the minimum, but widen it when any rendered amount
+-- needs more room.
+widenDefaultBalanceLineFormat :: ReportOpts -> BalanceReport -> ReportOpts
+widenDefaultBalanceLineFormat opts report
+  | format_ opts == defaultBalanceLineFormat = opts{format_ = setTotalMinWidth amountwidth defaultBalanceLineFormat}
+  | otherwise = opts
+  where
+    amountwidth = defaultBalanceAmountColumnWidth opts report
+    setTotalMinWidth w (BottomAligned (FormatField l _ m TotalField : rest)) =
+      BottomAligned (FormatField l (Just w) m TotalField : rest)
+    setTotalMinWidth _ fmt = fmt
+
+defaultBalanceAmountColumnWidth :: ReportOpts -> BalanceReport -> Int
+defaultBalanceAmountColumnWidth opts ((items, total)) =
+    maximum' $ 20 : map renderedAmountWidth visibleAmounts
+  where
+    visibleAmounts =
+      map (\(_,_,_,amt) -> amt) items ++
+      [total | not (no_total_ opts)]
+    renderedAmountWidth =
+      wbWidth . showMixedAmountB noCostFmt
+        { displayCommodity = layout_ opts /= LayoutBare
+        , displayMinWidth = Just 0
+        , displayColour = False
+        }
 
 -- | Render a single-column balance report as plain text with a separate commodity column (--layout=bare)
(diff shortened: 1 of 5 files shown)


adamsardo:
Opened this for #1148. If the fix matches what you had in mind for the bounty, happy to coordinate payout however you prefer.

adamsardo:
Pushed a follow-up fix after the first CI run exposed a regression in `bcexample.test`.

The original implementation let regular long rendered amounts widen the default balance separator/output shape. The follow-up narrows the widening trigger to genuinely wide commodity symbols, keeps the historical 20-character default separator where no widening is needed, and renders width calculations without ANSI color codes.

Verification now passing locally:

- `stack build hledger:exe:hledger`
- `COLUMNS=80 stack exec -- shelltest --execdir --hide hledger/test/balance/bcexample.test hledger/test/balance/1148.test`
- `COLUMNS=80 stack exec -- shelltest --execdir --hide hledger/test/balance`
- `COLUMNS=80 stack exec -- shelltest --execdir -j16 hledger/test -x /_ -x /addons -x ledger-compat/ledger-baseline -x ledger-compat/ledger-regress -x ledger-compat/ledger-collected`

The remote GitHub Actions `ci` check is also passing now.

simonmichael:
Thank you!

It looks like you used OpenAI's Codex. 
Our current https://hledger.org/AI.html#rules-of-engagement discourage this.
I'll accept it this time, but other options are preferred for hledger work.

Claude had these suggestions:

> 1. Scope is narrower than the issue (Balance.hs:537-542). defaultBalanceAmountColumnWidth only widens when some amount has a commodity wider than 18 chars. Wide quantities with short commodities (e.g. 1234567890.00 USD) still misalign — the title of #1148 is about long commodity names, but a more general fix would just be maximum' (20 : map renderedAmountWidth visibleAmounts). Worth considering whether to drop the hasWideCommodity filter and simplify.
> 
> 2. Magic 18 (Balance.hs:542). This is 20 - 2 (one digit + one space). Without a comment it reads as arbitrary; if the threshold logic stays, a one-line note on the relationship to the 20-char minimum would help future readers.
> 
> 3. defaultBalanceAmountColumnWidth computed twice — once in widenDefaultBalanceLineFormat and once for overlinewidth (Balance.hs:513,525). Cheap, but trivially shareable.
> 
> 4. Manual mention (Balance.md:248) still describes %20(total) as the default. Technically still accurate (the format string is unchanged; only the min-width is dynamically overridden at render time), but a brief note about the dynamic widening for long commodity names would improve discoverability of the new behavior.
> 
> 5. Commit messages — both are bare summaries with no category:  prefix. Project convention is e.g. `imp: bal: ...`. Worth squashing to one commit and reformatting before merge.

It sounds like you changed 1 intentionally - I wonder why ?

simonmichael:
Also, it would be great if you can include some rough estimate of AI usage for this work (model and approx tokens used, eg output tokens, or approx time spent chatting with/running AI). In the commit message is fine.

You can squash these into one commit, or I can when merging.

simonmichael:
An example of 1 (still misaligned if the number itself is wide):
```
$ cat y.j
2024-01-01
  assets:big       12345678901234567890.00 USD
  assets:small     1.00 USD
  equity:opening
$ stack exec -- hledger -f y.j bal
12345678901234567890.00 USD  assets:big
            1.00 USD  assets:small
-12345678901234567891.00 USD  equity:opening
--------------------
                   0  
```

adamsardo:
Thanks, that broader interpretation makes sense. I pushed a revised/squashed commit: `b2dede085` (`imp: bal: widen default amount column`).

Changes from the previous version:

- Dropped the wide-commodity-only guard, so the default balance amount field now widens to the widest visible rendered amount, with 20 columns kept as the minimum.
- Added a regression case for the wide-number example you posted, using `--no-total` to avoid adding new trailing-whitespace expectations.
- Updated `bcexample.test` for the expected wider default column.
- Added a short manual note that `%20(total)` is the default minimum and can widen automatically when displayed amounts need more room.
- Squashed the branch to one project-style commit message and included the requested AI usage estimate there.

Verification on the squashed commit:

- `stack build hledger:exe:hledger`
- `COLUMNS=80 stack exec -- shelltest --execdir --hide hledger/test/balance/1148.test`
- `COLUMNS=80 stack exec -- shelltest --execdir --hide hledger/test/balance/bcexample.test hledger/test/balance/1148.test`
- `COLUMNS=80 stack exec -- shelltest --execdir --hide hledger/test/balance` -> 214 passed, 0 failed
- `git diff --check`

On AI usage: I used OpenAI Codex to investigate, edit, and verify the patch. I do not have exact token accounting for just this repository work, but the rough estimate in the commit message is about 2-3 hours of interactive agent time and about 15k-25k output tokens.

adamsardo:
One more update: the first CI run on the squashed commit exposed one additional functional golden output in `hledger/test/journal/precision.test`, from the same broader widening behavior. I updated that expected alignment and force-pushed `c425de8f5`.

Additional local verification now passing:

- `COLUMNS=80 stack exec -- shelltest --execdir --hide hledger/test/journal/precision.test` -> 20 passed, 0 failed
- `COLUMNS=80 stack exec -- shelltest --execdir --hide hledger/test/balance hledger/test/journal/precision.test` -> 234 passed, 0 failed
- `git diff --check`

simonmichael:
Thanks for the fix @adamsardo !

Issue #1640, opened 2021-08-05
Pivoting over multiple tags: Making `--pivot` much more powerful


nobodyinperson:
Hi everyone,

Tags in combination with pivoting add an incredible amount of depth to an `hledger` journal. They effectively enable looking at the transactions in an arbitrary amount of different layers of categorization.

I have an idea to push this even further while maintaining backwards-compatibility.

## 👀 Current `--pivot` behaviour

As far as I understand (and the docs state it), `--pivot=TAG` causes each account name to be replaced by the (possibly non-existant) value of the given `TAG`. So basically pivoting is actually just a string manipulation action at a time where all tags attached to the transaction are known.

Currently, this string manipulation action can only do one thing: replacing the existing account name with the value of a *single* tag.

## ✨ Proposed new `--pivot` behaviour

I propose to make pivoting much more powerful by adding new simple string manipulations. A tag name looks like `[[:word:]-]+`, so pretty much all special characters are free to have a special meaning in `--pivot`, for example:

### ⤵ `--pivot=TAG1|TAG2|TAG3` - OR - fallback to other tag value

Replace the account name with value of `TAG1`, but if this one is empty or doesn't exist, replace it with `TAG2`'s value. If `TAG2` is equally empty, try `TAG3`. 

This is useful to fill empty space when pivoting in `hledger register` over a description/notes tag (be it the built-in `desc` or another tag one uses for additional info like `mynotes`):  `hledger register checking --pivot='mynotes|desc'`. A CSV import for example can only really set the `desc`-tag, which for real-world-accounts will be something like the gibberish transaction details, but I find myself to often add additional info in human-readable form to the transactions for nice presentation in `hledger register`. This ”or” operator could be used to fall back to certain tag values.

This operator would have the lowest priority, i.e. it would be evaluated last.

### ➕ `--pivot=TAG1:TAG2` - concatenation with colon

Replace the account name with the given tags' values concatenated with a colon `:`.

[...]

PR #2583, opened 2026-05-11
Add pipe fallback syntax for pivot components

arbazkhan971:
/claim #1640

This extends --pivot so each colon-separated component can try multiple field/tag names left-to-right with |, returning the first non-empty value. The existing colon-delimited composition from #2050 stays intact; this only adds the missing fallback behavior from the original issue proposal.

Validation:
- git diff --check
- static review of the changed pivot code and regression test (no Haskell toolchain is installed in this environment)


simonmichael:
Thanks for the PR; this one currently does not pass tests, has not responded to 1640 discussion, and is superseded by #2585.

PR #2585, opened 2026-05-12
Support pipe fallback in pivot components

zzhang82:
/claim #1640

## Summary
- Add `|` fallback within each colon-delimited `--pivot` component.
- Use the first non-empty field/tag value from left to right, eg `--pivot 'mynotes|desc'`.
- Preserve existing colon composition behavior, so expressions like `acct:category|desc` work per component.

## Validation
- Built `hledger-lib` and `hledger` from source in Docker with GHC 9.6.7 using `stack --stack-yaml stack96.yaml build hledger hledger-lib`.
- Manually verified new fallback cases:
  - tag present uses tag value
  - absent tag falls back to `desc`
  - three-way fallback chooses first non-empty value
  - fallback works inside existing colon composition
  - all alternatives absent preserves empty-string pivot behavior
- Manually verified existing colon composition regression:
  - `--pivot acct:kind:project` still produces `expenses:equipment:job1`, `expenses:equipment:job2`, and `expenses:fee:job2`.
- Ran `git diff --check`.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -1441,13 +1441,24 @@ pivotAccount :: Text -> Posting -> Text
 pivotAccount fieldortagname p =
   T.intercalate ":" [pivotComponent x p | x <- T.splitOn ":" fieldortagname]
 
--- | Get the value of the given field or tag for this posting.
+-- | Resolve one colon-separated pivot component for a posting.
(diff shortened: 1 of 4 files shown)


simonmichael:
Thanks for the PR. Currently this is blocked on my two latest comments on #1640.

simonmichael:
I'll close this due to no followup, not meeting our AI policy then or now, and the bounty owner requiring non-AI solutions.

PR #2586, opened 2026-05-12
Support plus-joined pivot components

newmattock:
Refs #1640.

This implements the non-duplicative `+` portion of the pivot enhancement proposal: within each existing colon-delimited pivot component, `TAG1+TAG2+...` resolves each field/tag, drops empty values, and joins the remaining values with spaces. Colon composition continues to work as before, so expressions can mix hierarchy and same-level labels.

Changes:
- add plus-joined pivot component resolution in `pivotComponent`
- add a regression case for `--pivot kind+project`
- update the manual source plus generated text/man/info outputs

Verification:
- `git diff --check`

I could not run the Haskell test suite locally in this environment because no `ghc`/`stack`/`cabal` toolchain is installed and the local Docker daemon is not running.

### DIFF
--- a/hledger-lib/Hledger/Data/Journal.hs
+++ b/hledger-lib/Hledger/Data/Journal.hs
@@ -1441,13 +1441,22 @@ pivotAccount :: Text -> Posting -> Text
 pivotAccount fieldortagname p =
   T.intercalate ":" [pivotComponent x p | x <- T.splitOn ":" fieldortagname]
 
+-- | Resolve one colon-delimited pivot component. A component may contain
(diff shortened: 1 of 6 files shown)


newmattock:
Thanks for clarifying. Since #2586 was AI-assisted and I disclosed that above, I will withdraw it rather than add review burden or keep it in consideration for a bounty now limited to 100% human contributions.

simonmichael:
Thanks for the PR. I appreciate your followup and haven't ruled out merging it, as it's small. But currently it is blocked on my latest comment on #1640.

PR #2587, opened 2026-05-13
journal: Respect posting tag overrides

### LINKED ISSUE #1950: Inherited Tag Values are not Overwritten

Wheezily:
Fixes #1950.

Posting tag matching now treats explicitly set posting tags as overriding inherited transaction tags with the same name. Same-level repeated tag values are still kept; only inherited same-name tags are filtered out.

Validation:
- `cabal test hledger-lib:unittest -j1 --test-options='--pattern=override'` in `haskell:9.8`
- `cabal test hledger-lib:unittest -j1 --test-options='--pattern=Query'` in `haskell:9.8`
- `cabal test hledger-lib:unittest -j1` in `haskell:9.8`
- `git diff --check`

### DIFF
--- a/hledger-lib/Hledger/Data/Posting.hs
+++ b/hledger-lib/Hledger/Data/Posting.hs
@@ -101,7 +101,7 @@ import Data.Foldable (asum)
 import Data.Function ((&))
 import Data.Map qualified as M
 import Data.Maybe (fromMaybe, isJust, mapMaybe)
-import Data.List (sort, union)
+import Data.List (sort)
 #if !MIN_VERSION_base(4,20,0)
 import Data.List (foldl')
 #endif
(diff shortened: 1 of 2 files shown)


simonmichael:
Thanks for the PR, I have responded on #1950.

simonmichael:
Hi @Wheezily, are you still interested in this PR ? These things would help it move forward:

- AI disclosure in the commit message, like https://hledger.org/AI.html#rules-of-engagement > 4.
- Tests/clarifications as mentioned in https://github.com/simonmichael/hledger/issues/1950#issuecomment-4480637343 and https://github.com/simonmichael/hledger/issues/1950#issuecomment-4558033876

simonmichael:
See also https://github.com/simonmichael/hledger/issues/1950#issuecomment-5140836443.

Issue #1825, opened 2022-02-03
slowly increasing cpu load with hledger-ui --watch


simonmichael:
An idle hledger-ui with `--watch`, left running for days, may show gradually increasing CPU usage - 1%, 2%, 3%, 4%...
I have always seen this on mac, and @the-solipsist sees it on ubuntu gnu/linux. I believe it is more apparent when you have a lot of accounts / transactions. I seem to remember, or I assume, that it's a problem with the underlying C file-watching library (which might be different on each platform). As a workaround, you can occasionally quit and restart hledger-ui (`q`, `C-p`, `enter`), or suspend and resume it when not in use (`C-z`, `fg`, `enter`). It's quite unfortunate.

simonmichael:
hledger-ui --watch uses https://hackage.haskell.org/package/fsnotify, which uses 

| platform | haskell lib | C lib | Problem confirmed |
| - | - | - |- |
| GNU/Linux | [hinotify] | the linux kernel's inotify feature | Yes |
| OpenBSD, FreeBSD | [hinotify] | inotify-tools port |
| Mac | [hfsevents] | Cocoa CoreServices | Yes |
| WIndows | [Win32] | Windows API ([this?]) |

[hinotify]: https://hackage.haskell.org/package/hinotify
[hfsevents]: https://hackage.haskell.org/package/hfsevents
[Win32]: https://hackage.haskell.org/package/Win32
[this?]: https://docs.microsoft.com/en-us/windows/win32/fileio/obtaining-directory-change-notifications

simonmichael:
This was also discussed on #836 in 2019.

simonmichael:
Update: with current hledger-ui, the memory leak can be easily reproduced by loading a largeish file (1k transactions will do), and pressing `g` repeatedly while watching top. Observed on mac and on ubuntu.

simonmichael:
Great! Thanks for the help. Yes you should be able to reproduce easily on linux. Eg
```
just samplejournals
hledger-ui -f examples/1ktxns-10accts.journal --watch --all
```

simonmichael:
Fair warning, UI code is hard on the brain!

simonmichael:
With a [user pain score](https://hledger.org/ISSUES.html#prioritising) of 16, this is currently our highest priority open bug. I'm declaring some hledger [issue bounties](https://hledger.org/FINANCE.html#issue-bounties): 

On each of the GNU/linux, mac, and windows platforms:
[...]

PR #2590, opened 2026-05-16
fix 1825: refresh only active hledger-ui screen on reload

lymeow233-wq:
## Summary

This changes `hledger-ui --watch` reload handling so a file reload refreshes only the active screen instead of regenerating the whole cached screen stack. Hidden screens are regenerated when they become active again, so they do not keep old journal-dependent screen data alive across reloads.

It also gives the transaction screen a real `tsUpdate` path, preserving the selected transaction where possible after journal reloads, and includes a small UTF-8 `embedStringFile` fix needed for reliable Windows local builds.

## Root cause

The leak path I could reproduce was not just watcher event volume. Each watched reload rebuilt the active screen and the hidden screen stack from the new journal, while the previous screen states could still hold old journal-derived structures. In register and transaction views this caused memory to keep climbing after successive file changes, and CPU stayed elevated after the reload work should have settled.

This patch makes watched reloads update the current screen only. Normal UI option/date changes still regenerate the full stack, while back-navigation regenerates the newly active parent screen from the latest journal.

## Validation

- `stack build hledger-ui --fast --no-run-tests`
- `git diff --check origin/main..HEAD`
- `stack exec -- hledger --version`
- `stack exec -- hledger-ui --version`

Windows 11 / Windows Terminal `hledger-ui --watch` register-screen repro:

- `main`, 30 file changes: working set grew to about 610-633 MB with sustained idle CPU afterward.
- this branch, same 30 file changes: working set was about 128.6 MB immediately after the changes, then settled to about 87 MB idle; idle CPU settled around 2-4% after reload work completed.
- no-change 30s sanity run: working set stayed around 67-69 MB.

Transaction-screen reload sanity check:

- before reload: `screen=TS count=4000 selectedPos=4000 selectedTxnIndex=3976`
- after appending transactions: `screen=TS count=4003 selectedPos=4000 selectedTxnIndex=3976`

## Notes

This intentionally does not add watcher debounce/rate limiting. It addresses the reload/screen-state retention path, including the register and transaction screens discussed in #1825.

AI assistance disclosure: implemented and validated with OpenAI Codex assistance. I reviewed the resulting changes and am responsible for the commits. Rough usage: one extended interactive implementation/debugging session; exact output-token count was not available in the local UI.

Refs #1825

### DIFF
--- a/hledger-lib/Hledger/Utils/IO.hs
+++ b/hledger-lib/Hledger/Utils/IO.hs
@@ -132,11 +132,12 @@ import           Data.Colour.RGBSpace (RGB(RGB))
 import           Data.Colour.RGBSpace.HSL (lightness)
 import           Data.Colour.SRGB (sRGB)
 import           Data.Encoding (DynEncoding)
-import           Data.FileEmbed (makeRelativeToProject, embedStringFile)
+import           Data.FileEmbed (makeRelativeToProject)
 import           Data.Functor ((<&>))
 import           Data.List hiding (uncons)
 import           Data.Maybe (isJust, catMaybes)
 import Data.Text qualified as T
+import Data.Text.Encoding qualified as T
 import           Data.Text.Encoding.Error (UnicodeException)
 import Data.Text.IO qualified as T
 import Data.Text.Lazy qualified as TL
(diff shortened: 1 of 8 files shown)


simonmichael:
Thanks for the PR.
For future work, please avoid using OpenAI per our AI policy.

simonmichael:
Some comments here: https://github.com/simonmichael/hledger/issues/1825#issuecomment-4798872151

simonmichael:
Closing in favour of #2652; see latest discussion at #1825.

PR #2591, opened 2026-05-16
fix 1825: Fix infinite memory usage by allowing old data to be garbage collected

### LINKED ISSUE #1825: slowly increasing cpu load with hledger-ui --watch

alerque:
Re-opening of #2473. Fixes #1825.

### DIFF
--- a/hledger-ui/Hledger/UI/RegisterScreen.hs
+++ b/hledger-ui/Hledger/UI/RegisterScreen.hs
@@ -245,7 +245,9 @@ rsHandle ev = do
               where
                 p = reportPeriod ui
 
-            e | e `elem` [AppEvent FileChange, VtyEvent (EvKey (KChar 'g') [])] -> uiReload copts d ui >>= put'
+            e | e `elem` [AppEvent FileChange, VtyEvent (EvKey (KChar 'g') [])] -> do
+              ui' <- uiReload copts d ui
+              put' $ regenerateScreens (ajournal ui') d ui'
 
             VtyEvent (EvKey (KChar 'I') []) -> uiToggleBalanceAssertions d ui
             VtyEvent (EvKey (KChar 'a') []) -> suspendAndResume $ clearScreen >> setCursorPosition 0 0 >> add (cliOptsDropArgs copts) j >> uiReloadIfFileChanged copts d j ui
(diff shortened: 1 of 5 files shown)


simonmichael:
Hi @alerque, sorry for the delay on this. Have you been using it much ?

I got these warnings from opus:

### 1. Regression: regenerateScreens no longer updates aPrevScreens

  Old UIState.hs:381:
  
      ui{ajournal=j, aScreen=screenUpdate opts d j s, aPrevScreens=map (screenUpdate opts d j) ss}
  
  New:
  
      let !ui' = ui{ajournal=j, aScreen=s'}
          !s' = case s of ...
  
  aPrevScreens is silently dropped from regeneration. After reload on a
  deep screen, popping back will show stale data. The XXX comment was
  edited to only mention the error screen, but the breakage is wider —
  every prev screen on the stack is now stale. This needs to be restored
  (with the same $!-style strictness, if desired).

### 2. Redundant regeneration in rsHandle

  RegisterScreen.hs:248:
  
  
      e | e `elem` [...] -> do
        ui' <- uiReload copts d ui
        put' $ regenerateScreens (ajournal ui') d ui'
  
  uiReload already calls regenerateScreens j d ui internally
  (ErrorScreen.hs:177), so this regenerates twice on every g press /
  file change — exactly the hot path the PR is trying to optimize. And
  if reload failed (error screen pushed), this re-regenerates against
  the old journal. Should just be uiReload copts d ui >>= put' like the
  other screens.

### 3. Top-level type signature removed

  UIScreens.hs:245:
  
      -rsUpdate :: UIOpts -> Day -> Journal -> RegisterScreenState -> RegisterScreenState
      -rsUpdate uopts d j rss@RSS{_rssAccount, _rssForceInclusive, _rssList=oldlist} =
      +rsUpdate uopts d j rss@RSS{_rssAccount, _rssForceInclusive} =
  
  -Wall will warn here.

### 4. tsUpdate fallback may show stale data


      updatedTxn = case find (\t -> tindex t == tindex currentTxn) (jtxns j) of
        Just t  -> t
        Nothing -> currentTxn  -- fallback to current if not found

  If the user deletes the displayed transaction externally and
  reloads, the screen will silently keep showing the old
  transaction. The XXX caveat comment acknowledges index shifts but
  not deletions. Worth considering popping back to the register screen
  on Nothing instead

simonmichael:
Closing in favour of #2652; see latest discussion at #1825.

PR #2593, opened 2026-05-16
ui: Fix scrolling past last item

### LINKED ISSUE #2278: hledger-ui keeps scrolling at the end of lists
Hi! Is there any reason or real usefulness for making the lists keep scrolling once the cursor is at the end of the list?
I personally find this behavior annoying but also unexpected since it's different from any other TUI application or even graphical interfaces. Specially when the entire list of items fits in one page, it's odd that part of it becomes off screen. And even when it doesn't fit, I can see no point in not using as much screen as possible to display information instead of blank space.

I can see this is not a default behavior in brick, but instead made explicit for example here: https://github.com/simonmichael/hledger/blob/659e493b306bfd32e62390efbe3c0398a5031d74/hledger-ui/Hledger/UI/AccountsScreen.hs#L319-L321

I could try fixing this and send a PR unless there's a good reason or personal attachment to this behavior. 😄

waj:
Finally made a fix for #2278. Now the lists don't keep scrolling when the end of the list is reached.

Also I did a small change in how the lists are initialized and centered, so there's no blank space at the bottom for lists with enough items to fill the screen.

I feel this behavior is now more familiar and matches other TUI tools with scrolling lists.

Closes #2278

### DIFF
--- a/hledger-ui/Hledger/UI/AccountsScreen.hs
+++ b/hledger-ui/Hledger/UI/AccountsScreen.hs
@@ -332,8 +332,8 @@ asHandleNormalMode ass ev = do
       else
         put' ui{aScreen=AS ass{_assList=l1}}
 
-    -- DOWN key when selection is at the last item: scroll instead of moving, until maximally scrolled
-    VtyEvent e | e `elem` moveDownEvents, isBlankItem mnextelement -> vScrollBy (viewportScroll $ l^.listNameL) 1
+    -- DOWN key when selection is at the last item: do nothing
+    VtyEvent e | e `elem` moveDownEvents, isBlankItem mnextelement -> return ()
       where mnextelement = listSelectedElement $ listMoveDown l
 
     -- Any other vty event (UP, DOWN, PGUP etc): handle with List's default handler.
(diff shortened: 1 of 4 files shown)


simonmichael:
Thanks for the PR.

simonmichael:
I agree this is more familiar behaviour, and I can't find many issues caused by it just now, so this sounds good to me. I see just one problem:

- open a long accounts list, longer than screen height
- move selection to the last item, and press C-l ("recenter selection"). 
  Nothing happens; this is expected now, since we're not allowing further scrolling.
- move selection to a nearby item just above, say the second or third from the bottom, and press C-l.
  I think nothing should happen here either, but instead it scrolls so that the selected item is at bottom of screen, pushing the lower items offscreen. This seems inconsistent.

We'll need to rebase against latest main also.

simonmichael:
Actually I have a fix for that (making scrollSelectionToMiddle more accurate), which I can add later.

simonmichael:
PS, but if you are able to rebase this PR against current main before I get to it, please go ahead, it will probably make for a cleaner PR merge.

simonmichael:
Merged. Thank you!

PR #2594, opened 2026-05-17
fix 1825: resolve hledger-ui --watch memory leak and CPU hogging

### LINKED ISSUE #1825: slowly increasing cpu load with hledger-ui --watch

bomlinux92-byte:
## Summary
Fix for [BOUNTY $150 x 3 platforms] hledger-ui --watch consumes more CPU and RAM over time — Issue #1825

## Root cause
Haskell's lazy evaluation was retaining thunks (unevaluated expressions) from previous data states. Each time  reloaded the journal, the old UI state data structures were kept alive because:

1.  was reusing the same UIState record with lazy field updates, keeping old thunk chains alive
2. Register screen wasn't calling  after reload
3. Transaction screen was just doing `seq` instead of actually updating

This caused memory to grow indefinitely (one thunk chain per reload), and CPU to accumulate because the Haskell runtime kept accumulating work.

## Fix (ported from alerque's PR #2473)
- **UIState.hs**: `regenerateScreens` now uses strict field updates () to force evaluation of new data structures, allowing old ones to be garbage collected
- **UIScreens.hs**: Display items are now strictly evaluated, and transaction screen properly updates to show current transaction data after reload
- **RegisterScreen.hs**: Now calls `regenerateScreens` after `uiReload` to ensure all screens are properly refreshed
- **TransactionScreen.hs**: Proper update logic that finds the current transaction in the new journal
- **hledger-ui.m4.md**: Removed documentation of the now-fixed bug

## Changes
```
hledger-ui/Hledger/UI/RegisterScreen.hs    |  4 ++-
hledger-ui/Hledger/UI/TransactionScreen.hs | 47 ++----------------------------
hledger-ui/Hledger/UI/UIScreens.hs         | 47 ++++++++++++++++--------------
hledger-ui/Hledger/UI/UIState.hs           | 29 +++++++++---------
hledger-ui/hledger-ui.m4.md                |  4 ---
5 files changed, 45 insertions(+), 86 deletions(-)
```

## Testing
- Reproduction: press `g` repeatedly while watching memory usage — old version grows unboundedly, new version stabilizes
- Multiple testers confirmed memory stabilizes after the fix (see issue comments)
- Windows testing confirmed by @lymeow233-wq in PR #2590

Closes #1825

### DIFF
--- a/hledger-ui/Hledger/UI/RegisterScreen.hs
+++ b/hledger-ui/Hledger/UI/RegisterScreen.hs
@@ -245,7 +245,9 @@ rsHandle ev = do
               where
                 p = reportPeriod ui
 
-            e | e `elem` [AppEvent FileChange, VtyEvent (EvKey (KChar 'g') [])] -> uiReload copts d ui >>= put'
+            e | e `elem` [AppEvent FileChange, VtyEvent (EvKey (KChar 'g') [])] -> do
+              ui' <- uiReload copts d ui
+              put' $ regenerateScreens (ajournal ui') d ui'
 
             VtyEvent (EvKey (KChar 'I') []) -> uiToggleBalanceAssertions d ui
             VtyEvent (EvKey (KChar 'a') []) -> suspendAndResume $ clearScreen >> setCursorPosition 0 0 >> add (cliOptsDropArgs copts) j >> uiReloadIfFileChanged copts d j ui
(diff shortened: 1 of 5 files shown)


simonmichael:
Thanks for the PR; closed as a duplicate (see #1825).

PR #2596, opened 2026-05-19
print: beancount: convert underscores in account names to CamelCase

waj:
Underscore-separated subparts within an account name component are now capitalised and joined, so e.g. `assets:wells_fargo:fima_usd` becomes `Assets:WellsFargo:FimaUsd` instead of the previous hex-encoded `Assets:WellsC5ffargo:FimaC5fusd`. Underscores in hledger account names are conventionally used as a space substitute, so treating them as word boundaries produces much cleaner Beancount output.

### DIFF
--- a/hledger-lib/Hledger/Write/Beancount.hs
+++ b/hledger-lib/Hledger/Write/Beancount.hs
@@ -256,7 +256,9 @@ type BeancountAccountNameComponent = AccountName
 -- It replaces spaces with dashes and other non-supported characters with C<HEXBYTES>;
 -- prepends the letter A to any part which doesn't begin with a letter or number;
 -- adds a second :A part if there is only one part;
--- and capitalises each part.
+-- and capitalises each part (also capitalising each underscore-separated
+-- subpart within a part and joining them, so e.g. "wells_fargo" becomes
+-- "WellsFargo").
 -- It also checks that the first part is one of the required english
 -- account names Assets, Liabilities, Equity, Income, or Expenses, and if not
 -- raises an informative error.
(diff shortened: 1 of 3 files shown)


simonmichael:
Thanks for the PR.

simonmichael:
I use hyphens rather than underscores to separate account name sub-parts; and of course spaces are the other common separator. Here's how they're all handled currently:

```
~/src/hledger$ cat a.j
2000-01-01
  Assets
  Expenses:foo_bar  1
  Expenses:foo-bar  1
  Expenses:foo bar  1

~/src/hledger$ hledger -f a.j print -O beancount
...
2000-01-01 *
    Assets:A
    Expenses:FooC5fbar            1 CC
    Expenses:Foo-bar              1 CC
    Expenses:Foo-bar              1 CC
```

Would it be simpler and more consistent to replace `_` with `-` ?

https://hledger.org/1.52/hledger.html#beancount-account-names currently says

> hledger will adjust your account names to make valid Beancount account names by capitalising each part, replacing spaces with -, replacing other unsupported characters with `C<HEXBYTES>`, prepending A to account name parts which don't begin with a letter or digit, and appending :A to account names which have only one part.

It would be nice to always use `-` and never `C<HEXBYTES>`; but I think we can't do that, because then you could quite easily get names containing many hyphens, likely to overlap.

simonmichael:
I think the simplest, most consistent fix is to replace underscore with hyphen, as we do for spaces. Would you like to follow up on this @waj ?

simonmichael:
I went with that alternate fix. Thanks for raising the issue.

Issue #2420, opened 2025-07-01
Computing TWR: corner-cases,complications and considerations


adept:
I've decided to create this item to document all the corner-cases and complications that affect `roi`'s TWR computations so that we have something to refer back to.

*How to compute TWR*:  Wikipedia has a [nice explanation](https://en.wikipedia.org/wiki/Time-weighted_return#Time-weighted_return_compensating_for_external_flows)

simonmichael:
Thanks for these notes. I have applied the WISH label as they seem to be more of that than a current BUG.

Issue #18, opened 2013-04-08
register, --display and --empty don't play nicely


simonmichael:
_Original author: sampoin...@googlemail.com (December 19, 2009 04:42:28)_

borked.ledger:
2009-12-01 \* foo
    foo    £1.00
    bar

2009-12-15 \* bar
    foo    £1.00
    baz

2009-12-21 \* baz
    bar    £2.00
    foo

hledger -f borked.ledger register foo -W -E --display=d&gt;=[2009-12-07]
2009/11/30 - 2009/12/06                                           0  
£1.00
2009/12/07 - 2009/12/13                                           0  
£1.00
2009/12/14 - 2009/12/20         foo                           £1.00  
£2.00
2009/12/21 - 2009/12/27         foo                          £-2.00  
    0

This looks wrong to me. register should not display entries where the
entire date range is excluded by a --display predicate. Instead, I'd expect
this:
2009/12/07 - 2009/12/13                                           0  
£1.00
2009/12/14 - 2009/12/20         foo                           £1.00  
£2.00
2009/12/21 - 2009/12/27         foo                          £-2.00  
    0

_Original issue: http://code.google.com/p/hledger/issues/detail?id=18_

PR #2624, opened 2026-05-30
`roi` rework (TWR part)

adept:
Issue #2420 outlined a number of issues with existing time-weighted return (TWR) implementation, chief of which was that:
- it is hard to value investment at arbitrary point in time, and very hard to value it intraday (when multiple transactions happen on a given day)
- it is hard to handle days with multiple events as different types of entries (transactions vs P directives, for example) are not explicitly orderable in the way the user might want them to be

This PR attempts to address some (or maybe all?) of these issues by switching TWR computation from "true" TWR (which requires ability to value investment precisely after any cashflow) to BAI (Bank Administration Institute) Linked IRR, which requires valuations at the start and end of periods. Any reporting period would be split into "windows" on known valuation dates (where either P directives exist, or balance-modifying PnL-altering transactions exist), each window would be valued using IRR, and resulting rates geometrically linked.

This method approximates "true" TWR in a more robust way than before, and the more valuation points you have, the better approximation is going to be. This explains slight differences in test outputs.

Docs for `roi` updated accordingly, and somewhat cleaned up overall, especially the section that explains IRR vs TWR and when to use what.


simonmichael:
This sounds great. Here are some notes from opus 4.7-high, in case they're of interest. I'll merge when you are ready.

## Review of PR #2624 — `roi` rework (TWR part)

### Overview

Replaces the existing TWR implementation (true TWR subdividing on each cashflow with valueAfterDate adjustments) with **BAI Linked IRR**: the period is split into windows at known valuation dates (price directives + PnL transaction dates), each window's return is solved via Ridders' IRR, and the windows are geometrically chained. This is a meaningful semantic improvement - TWR no longer needs to value the portfolio "between" cashflows on the same day, which was fragile.

Documentation is significantly rewritten and much clearer. A regression test (#18) for the P-directive fencepost case is added.

### Correctness

- **`computeWindowReturn` is called twice per window when `--cashflow` is set**: once to build `windowRets`, and again inside the table cell in the `showCashFlow` block. For windows with cashflows, that re-runs Ridders root-finding. Cheap fix: bind `(rate, w) <- zip windowRets windows` and reuse `rate`.
- **`segmentIntoWindows` drops cashflows landing exactly on the window's start date**: `inside = filter (\(cd,_) -> cd > s) here`. Those cashflows aren't included in any window. They *are* implicitly captured in the next window's `awVStart` via `valueAfterDate s`, so the math works out, but it's a subtle invariant worth a comment.
- **`OneSpan` no longer uses `_valueBeforeAmt` / `_valueAfterAmt`** in `timeWeightedReturn`. They are still computed by the caller for IRR and the summary table, so this is fine - but worth confirming none of the displayed "Value (begin)" / "Value (end)" columns in the outer report shift behavior. (The diffs in `roi.test` only show TWR cells changing, so this looks safe.)
- **Window length `d`**: `fromIntegral (awEnd w \`diffDays\` awStart w)`. Since `sortedValDates` is `nub`+`sort`ed, consecutive dates can't be equal, so `d > 0`. Good - but `segmentIntoWindows` is a general helper and a caller that passes duplicate dates would hit a divide-by-zero in `windowNPV`. Consider adding a precondition comment.
- **Empty/single-date list cases**: `segmentIntoWindows []` and `[_]` return `[]`, so `twr = foldl' ... 0 []` = 0. Reasonable, but the original code distinguished "no cashflows" from "non-trivial value change" by computing `(valueAfter - valueBefore)/valueBefore`. New behavior reports 0% when there are no value-date breakpoints - but in practice `begin` and `end-1` always seed `sortedValDates`, so the list has at least 2 entries. OK in practice.

### Style / project conventions

- **Doc/code wording mismatch**: `Roi.md` says "Cannot compute sub-period **TWR** for a window" but the code emits "Cannot compute sub-period **IRR** for window …" in `subPeriodIRR`. Pick one; "TWR window" reads better for users.
- **Naming inconsistency**: the data type is `ActivityWindow` and code uses "window" everywhere, but the helper is `subPeriodIRR`. Renaming to `windowIRR` (or `activityWindowIRR`) would be consistent.
- **Trailing whitespace** on the `-- Compute Net Present Value of the ActivityWindow given discount rate r ` line, and "dont" (missing apostrophe) in the `segmentIntoWindows` comment.
- **`computeWindowReturn` guard `awVStart w == 0, all ((== awEnd w) . fst) (awCashFlows w) = 0`** is terse but a bit cryptic. A one-line comment ("zero starting value and all flows on the end date → 0% return, IRR is undefined") would help future readers.
- <s>**`Roi.txt` not regenerated**: command help text in `hledger/Hledger/Cli/Commands/Roi.txt` is generated from `Roi.md` via `./Shake cmddocs -c`. Not in the diff. Should be run before merging.</s>

### Performance

- Double Ridders call per window with `--cashflow` (noted above).
- `valueAt` (= `valueAfterDate`) recomputes from scratch on each call, and `segmentIntoWindows` calls it on the same date repeatedly (once as `awEnd` of window N, once as `awStart` of window N+1). For long histories with many price directives this could be slow. A `Data.Map` memoization would be a straightforward follow-up but not blocking.

### Test coverage

- Test #18 nicely covers the "P directive fencepost" regression that motivated the rework.
- The numeric drift in tests 5, 6, 8, 9, 12, 16, 17 (e.g. `10.00%` → `12.73%`, `-90.00%` → `-95.75%`) is expected and explained in the PR description, but it's a user-visible change. Worth a CHANGES note when merged.
- No test exercises the new `--cashflow` table layout (column headers changed). A small fixture confirming the new format would prevent accidental regressions.

### Risks

- **Backward incompatibility for users with locked-in TWR numbers**: anyone tracking TWR over time will see a step-change. The PR description acknowledges this but a release note is warranted.
- **Error paths now fire per window**: a single ill-conditioned window aborts the whole report. For long histories that's a real possibility - consider degrading to 0%/NaN for that window and flagging it, rather than `error'`.

### Summary

Solid algorithmic improvement with clearer docs. Main concerns are the double `computeWindowReturn` call when displaying, the doc/code TWR-vs-IRR wording mismatch, <s>and the need to regenerate `Roi.txt`</s>. None are blocking.

adept:
Thanks, that's helpful, and I adressed most of them, except most notably the duplicate `valueAt` calls: It is O(N) vs O(2N) concern which I think is negligible in practice.

I've also refenerated Roi.txt (which I forgot initially), and fixed a couple of extra typos i spotted on re-read. 

From my perspective, ready to merge now - unless something new comes up on your end :)

adept:
I've done another review pass and realised that the annualization code tried to account for leap years, but did this incorrectly, so I replaced that with a more widely used "year has 365.25 days" which is dead-simple and way more defensible than the old code.

adept:
In terms of Changelog, I think something along these lines could be said: "roi: time-weighted return computations changed to use algorithm that is more robust and works better with sparse valuations, and an issue with annualization has been fixed. This could lead to the output of" roi" changing for existing data"

simonmichael:
Thanks!

Issue #316, opened 2016-01-08
web: split up into REST api server and a js client


TomK32:
I had a look through old issues concerning hledger-web and little progress seems to be made, and my limited knowledge of haskell keeps me from doing more for it. Which is why I'd like to propose a very radical step:
- Cut down the yesod app to serve as a REST API
- Move all logic and templates into a javascript app (i don't care wether we use Angular, Backbone or Ember)

This move would bring the web-clients language to the browsers native language javascript, allowing easier modifications for the plenty numbers of javascript programmers out there. Logic and view templates could be separated far better than they are right now. For remaining server, a wider and purer REST api would allow inter-machine JSON requests, imagine a POS or webshop sending details about your sales directly into your bookkeeping!

For reference, there's an unfinished(?) project to do the same for ledger using python and AngularJS: https://groups.google.com/forum/#!msg/ledger-cli/2gRLc9kCPCc/Cw_4KG1dJqwJ -- [source](https://github.com/conservancy/accounting-api) -- [docs](http://npoacct.sfconservancy.org/accounting-api/restapi.html)

and another old project, for ledger: https://github.com/slashdotdash/node-ledger-web
##

simonmichael:
Yes, this has been on the wishlist for a long time. I'm leaning towards this approach: create a new hledger-api package containing

1. a series of simple example client-side web apps
2. the simplest possible web API server (I'm trying to avoid misusing "REST" here) that makes them work. I'd give servant a try.

The examples should always work, and drive the evolution of the server.

simonmichael:
We now have a [minimal API server and client-side example](https://github.com/simonmichael/hledger/tree/master/hledger-api). Please build more examples/apps and let me know what API methods you need added to the server (pull requests for that also welcome).

TomK32:
I'd prever `/accounts/` to return a tree.

TomK32:
The next resource i need is the balances for an account.
`GET /accounts/Expenses:Food/balances`
to return
```json
[
{ "name": "USD", "amount": 200 },
{ "name": "€", "amount": 440 }
]
```

TomK32:
Okay, I did some work on my branch, sidebar with the accounts is a happy little sidebar now. What's not working yet is the transactions for one account. either we change `/api/v1/accounts/:name` to return a hash rather than an array or we rename it to `/api/v1/accounts/:name/transaction`. I'd be happy with either.

simonmichael:
[...]

Issue #200, opened 2014-08-27
Usability improvement proposals for hledger-web


markokocic:
Issue 1:
Huh, I was wandering for quite some time why sidebar is removed, until I clicked on a ? button next to Search button to learn about keyboard shortcut. I always thought it was about search filtering help, and thus never read it, since I use command line regularly and didn't need to look up the syntax.

Issue 2:
it's nice that hledger-web supports keyboard shortcuts, but it would be better that it still has a link somewhere to toggle sidebar, as well links for other shortcuts. May make it less awkward on mobile.

Issue 3:
Both sidebar and entries should have their own scroll bars. Would make it easier to navigate.

Issue 4:
Might be good that sidebar contains a few toggles, like show/hide empty accounts. It would be bonus points if these were remembered in session.

Issue 5:
It would be nice if search filter from the box would be also applied to sidebar. That way sidebar could be really useful for analysis, of, for example income statement or balance sheet in some time period or whatever is defined in filter. Again, there could be a check box whether to display full sidebar, or it should follow the search filter.
Going one step further from that, i may make sense to provide different views for sidebar, similar to bal, bs and inc views in command line.
##

maisammusthafa:
I also agree on issue 4, especially about the ability to show/hide empty accounts.
hledger-web has the --empty flag, but since hledger-web always displays empty accounts, that flag is not meaningful. Maybe you could add a --hide-empty flag instead?

misev:
+1 for "show/hide empty accounts"

AidenTHu:
I'm wiling to implement this feature request and made a [video](https://drive.google.com/file/d/12_QX5CXSR1Wt2eO8rPwWAg8gLsD4Brin/view?usp=sharing) of what I'm proposing to change. I've also  reviewed aciander's work and it looks good, so I'll build off his code and continue from where he left off.

simonmichael:
Thanks for the offer and very clear video (how did you make that ?)
We haven't had review bandwidth to keep up with PRs lately; those will be reviewed ASAP.
If you do more work, please avoid using OpenAI per our AI policy.

Current hledger-web is intentionally simple and works reasonably well, or at least consistently, on all browsers/devices/window sizes. So we will keep it around for now.
Your proposals look good and I expect will need a lot of testing and stabilising. I suggest to call yours something else, for now - maybe hledger-web2 (fits with hledger 2.x's naming).

AidenTHu:
[...]

Issue #33, opened 2013-04-08
it should be possible to use hledger-web with javascript disabled


simonmichael:
_Original author: simon@joyful.com (December 17, 2010 16:23:35)_
_Original title: edit form always displays in non-javascript browsers, like elinks_

Reported by djp. The web ui should be usable, at least for browsing, in text-mode browsers and when javascript is disabled or not available.

_Original issue: http://code.google.com/p/hledger/issues/detail?id=33_

simonmichael:
Update: I believe that the above is no longer an issue (since the edit form was disabled), and that hledger-web ui is usable in read-only mode with javascript disabled. It's still not possible to add entries without javascript, and we'd like to make that possible. I'll repurpose this issue for that.

PR #2644, opened 2026-06-06
imp: web: preserve sidebar position during navigation

acinader:
This is an exploratory draft PR for hledger 2.x `main`.

When reviewing a large chart of accounts in hledger-web, clicking account links currently reloads the full page and resets the account sidebar position. This patch keeps the existing server-rendered links and progressively enhances same-origin `/journal` and `/register` navigation when JavaScript is available:

- keep ordinary `href`s for no-JS / failed-JS fallback behavior
- fetch the target page as HTML
- replace only `#main-content`
- refresh the sidebar account table so the active account highlight updates
- preserve the existing sidebar container and `scrollTop`
- use `history.pushState` / `popstate` for URLs and browser back/forward

I also made `#sidebar-menu` a bounded vertical scroll container.

Related context:

- #200 includes "Both sidebar and entries should have their own scroll bars."
- #33 says hledger-web should remain usable with JavaScript disabled; this keeps normal links as the fallback.
- #316 was a larger REST API / JavaScript client idea; this intentionally avoids that scope.

Checks run locally:

- `node -c hledger-web/static/hledger.js`
- `git diff --check`
- `cabal v2-build hledger-web --enable-tests --allow-newer --constraint='persistent >=2.17'`
- `cabal v2-test hledger-web:apptest --allow-newer --constraint='persistent >=2.17'`

The Cabal build/test needed the `persistent >=2.17` constraint because this machine only has GHC 9.14.1, and the default plan selected `persistent-2.11.0.4`, which did not compile with this toolchain. With the local dependency workaround, hledger-web built and its test suite passed: 4 examples, 0 failures.

Manual smoke check:

- Ran the patched hledger-web against a generated local journal with 81 accounts.
- Confirmed the patched `hledger.js` and `hledger.css` were served.
- Manually tried sidebar account navigation in the in-app browser and found it worked well in first use.

Still worth checking before considering this ready:

- more real-world use on larger account trees
- browser back/forward behavior
- register chart rendering after AJAX navigation
- add-transaction modal behavior when hledger-web is not in view-only mode
- sidebar hidden/shown cookie behavior
- behavior with a non-root `--base-url`

AI assistance disclosure:

This patch and PR text were prepared with OpenAI/Codex assistance. Approximate usage: several interactive Codex turns for research, patch drafting/refinement, local testing, and PR preparation. I understand the project currently prefers not to accept OpenAI-assisted work; I am opening this as a draft for review/discussion and to learn whether this direction is useful.

### DIFF
--- a/hledger-web/CHANGES.md
+++ b/hledger-web/CHANGES.md
@@ -29,6 +29,9 @@ Improvements
 
 - Allow yesod-static 1.6.1.1 and later again.
 
+- In hledger-web, account/register/journal links now preserve the sidebar
+  scroll position when JavaScript is available.
+
 - Uses hledger 1.99.2.
 
 
--- a/hledger-web/static/hledger.css
+++ b/hledger-web/static/hledger.css
@@ -92,7 +92,9 @@ ul {
 }
 
 #sidebar-menu {
-    overflow:hidden;
+    max-height: calc(100vh - 5.5em);
+    overflow-x:hidden;
+    overflow-y:auto;
     border-right: 1px solid #ebebeb;
 }
 #sidebar-menu .main-menu {
--- a/hledger-web/static/hledger.js
+++ b/hledger-web/static/hledger.js
@@ -4,35 +4,12 @@
 // STARTUP
 
 $(document).ready(function() {
+  hledgerInitGlobal();
+  hledgerInitPage();
+  hledgerInitAjaxNavigation();
+});
 
-  // add form helpers XXX move to addForm ?
-
-  // date picker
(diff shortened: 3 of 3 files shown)


acinader:
I have been using it for a few weeks now.

simonmichael:
Thanks for the PR. 
For future work, please avoid using OpenAI per our AI policy.

acinader:
Hi @simonmichael, thanks for the feedback. Challenge joyfully accepted! Closing this for now. I will do a clean reproduction with claude and report back  🫡

PR #2647, opened 2026-06-12
Note limitation for CSV-related matching

sten0:
Hi,

I'm expecting to need to rebase and will add the suggested commit message prefix then.  Meanwhile, I'm still not convinced that this is an intended hledger design decision rather than a bug.  Please let me know.

### DIFF
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -3836,7 +3836,7 @@ This can be confusing, so let's start with an overview:
   then its value will be automatically assigned to that hledger field.
 
 - For a CSV with the following fields:
-  accounttype, accounted, `date`,, type, payee, `amount`, for example:
+  accounttype, accountid, `date`,, type, payee, `amount`, for example:
   ```csv
   Chequing,12345-1111111,5/1/2026,,"WWW TRF DDA - 2651",,-50.00,,
   Savings,12345-2222222,5/1/2026,,"Transfer","WWW TRANSFER - 1743 ",50.00,,
--- a/hledger/hledger.m4.md
+++ b/hledger/hledger.m4.md
@@ -3865,7 +3865,7 @@ This can be confusing, so let's start with an overview:
   WWW TRF DDA
     account2 equity:transfers
   if HYDRO
-  account2 expenses:utilities:hydro
+    account2 expenses:utilities:hydro
   ```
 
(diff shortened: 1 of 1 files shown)


hledger/hledger.m4.md
```diff
@@ -3835,6 +3835,39 @@ This can be confusing, so let's start with an overview:
...
+  rather then the above rules, the following (which are scoped to single CSV field) must be used:
+  ```rules
+  if
+  Transfer
+  WWW TRF DDA
+    account2 equity:transfers
+  if HYDRO
+    account2 expenses:utilities:hydro
+  ```
+
```
simonmichael, line 3870: Thanks for the doc feedback. There's no bug here, but the doc may need to be clearer.

There's a balance between adding detailed examples of everything, and keeping the verbosity under control; when docs are too long people can't digest them and they become useless.

If an example is needed here, we should put it after  the list of overview points, not in the middle of it. 

But I think it's not needed, at least not this one exactly. No-one should expect `if Payment HYDRO` to succeed, because the literal text "Payment HYDRO" does not appear in the data (either in the CSV record that is matched, or in the constructed hledger description that you thought was matched). A correct pattern would be `if Payment.*HYDRO`. Also, nearby it says "hledger fields can only be written to, not read."

hledger/hledger.m4.md
```diff
@@ -4198,6 +4231,10 @@ they can express many matchers and field assignments in a more compact tabular f
 
 ```rules
 if,HLEDGERFIELD1,HLEDGERFIELD2,...
+TODO: /\ This is wrong if hledger is supposed to only match CSV fields
+         Alternatively, this documentation is correct, and hledger is buggy.
+         The resolution would then be to matching for a %description that is spans multiple contatenated values.
+         Alternatively hledgerfield is wrong because there is a middle-processing step where description="type string,payee string".  In this case this middle step must be noted in the documentation, becaues hledgerfield=description is what appears in the journal and what the user expects.
```
simonmichael, line 4237: if tables are confusing and hard to explain, indeed. But the doc is correct: only the first field on each line is a csv matcher; the other fields are values being assigned to hledger fields.

simonmichael:
I have rewritten 
https://hledger.org/dev/hledger.html#csv-fields-vs-hledger-fields
and https://hledger.org/dev/hledger.html#if-table , hopefully it helps.

PR #2651, opened 2026-06-26
web: keep the account sidebar's scroll position when switching accounts

acinader:
## What

When reviewing a long chart of accounts in hledger-web, clicking an account reloads the page scrolled to the top, so the account sidebar jumps back to the top and you lose your place while working down the list.

This makes the sidebar and the main content scroll **independently** (on wider screens), so the window itself no longer scrolls, and it remembers the sidebar's scroll position across navigations (per-tab, in `sessionStorage`). The account you clicked stays exactly where it was — even after scrolling a long register. Most noticeable with a long account list and/or long transaction lists, where both panes extend below the viewport. (Also addresses the "both sidebar and entries should have their own scroll bars" note in #200.)

## Testing

- `stack test hledger-web` (apptest) passes.
- Manually and via Playwright against a generated ~600-transaction / 45-account journal: scrolled the sidebar, clicked down the account list, and confirmed the clicked account stays put (clicking the same spot twice lands on the same account) even after scrolling the transactions; the sidebar toggle, hide-empty-accounts, and mobile/off-canvas behaviour are unaffected.

## Supersedes

Supersedes #2644 — a clean reimplementation taking a deliberately smaller, no-AJAX approach (that PR progressively enhanced navigation with AJAX; this one keeps full-page navigation and just preserves scroll).

## AI disclosure

AI usage: Claude Opus 4.8, ~1M output tokens


acinader:
I added a video that explains the issue and demonstrates this fix.

https://github.com/user-attachments/assets/9ec90894-b34c-496d-9c53-0936ff88f213

simonmichael:
Apologies for the slow reponse as I work through PR backlog @acinader. Thanks for this one - and the excellent video! It looks great, I will review more soon.

simonmichael:
@acinader it looks good for now. 

Not a blocker, but it would be great to find a way to prevent the visual pop when clicking to a new account while the accounts list is scrolled up.

If you are around, would you mind rewriting the AI disclosure line in your commit message to 

    AI usage: Claude Opus 4.8, ~1M output tokens

(I have hopes to report AI usage based on these, later).

I could do it but I suspect Github will detect the PR merge better if you do it.

acinader:
I added a commit that eliminates the flicker by moving the script out of `document.ready()` callback and making it inline. see https://github.com/simonmichael/hledger/pull/2651/commits/4d15c53a57b3ecc296cab68500e0a5af988e8f7e.

Updated ai attribution in the pr description.

simonmichael:
Sorry, I meant in the commit message(s). I have tweaked them to preferred style and rebased to avoid a merge commit. Now merged as bc624be48, thank you very much!

Issue #2429, opened 2025-07-16
balance with full account paths and unused accounts


thielema:
I want to do the following: Write a balance report to a FODS file, load this FODS into LibreOffice Calc and append the table to another table, that accesses the numbers from the balance report. I want to access the balance report numbers with the LibreOffice function VLOOKUP (in the German version: SVERWEIS). For proper lookup, VLOOKUP needs the full account name, as provided by the list mode of hledger-balance. However, the list mode misses the amounts for parent accounts. Thus what I want is all the lines and numbers from the tree mode, but the account names written as in the list mode.

This wish is similar to https://github.com/simonmichael/hledger/issues/1550 but I do not want indentation. But I guess, VLOOKUP can be made to tolerate the indentation.

For now I work around the missing functionality by creating multiple balance reports in list mode, but with depths 2, 3, 4.

What I also need is that unused accounts are reported anyway, but with amount zero. This is for accounts that are unused in one year, but exist for transactions in other years. I consider an account existing if there is any transaction that refers to that account or if there is an explicit account declaration.

thielema:
I have wondered, whether I should add full account paths as option to tree mode or whether I should add parent accounts to list mode. It turns out that adding paths to tree mode in the code is much simpler.
What would be a good name for such an option?

thielema:
Let me give a simplified example. The tree view of accounts in CSV/FODS may look like this:
~~~~
Expense, -1000 Euro
  Rent, -1000 Euro
Revenue, 800 Euro
  Rent, 600 Euro
    Doe, 300 Euro
    Miller, 300 Euro   
  Operating, 200 Euro
    Doe, 100 Euro
    Miller, 100 Euro
~~~~
The list view instead looks like this:
~~~~
Expense:Rent, -1000 Euro
Revenue:Rent:Doe, 300 Euro
Revenue:Rent:Miller, 300 Euro   
Revenue:Operating:Doe, 100 Euro
Revenue:Operating:Miller, 100 Euro
~~~~
I want to include the FODS table into a LibreOffice calc spreadsheet and access, say, the Revenue:Rent number, i.e. 600 Euro, using the `VLOOKUP` function. The List View misses that number, whereas the Tree View has the number, but only with unqualified name `Rent`, which is ambiguous.

The `--full-path` option adds the full account paths to the account names in the Tree Mode.

simonmichael:
[...]

PR #2661, opened 2026-07-14
balance: --full-path option

thielema:
This is a draft for #2429. If you accept it, it still needs some documentation. It serves my needs for a year now, however I am unhappy that we may add a lot of these options in the future that alter an aspect or another of selecting accounts and adapting their formatting. Hledger started with a set of configurations that were useful but now it turns out that sometimes you need to alter individual items in the configurations.

### DIFF
--- a/hledger-lib/Hledger/Reports/ReportOptions.hs
+++ b/hledger-lib/Hledger/Reports/ReportOptions.hs
@@ -148,6 +148,7 @@ data ReportOpts = ReportOpts {
     ,date2_            :: Bool
     ,empty_            :: Bool
     ,no_elide_         :: Bool
+    ,full_path_        :: Bool
     ,real_             :: Bool
     ,format_           :: StringFormat
     ,balance_base_url_ :: Maybe T.Text
@@ -217,6 +218,7 @@ defreportopts = ReportOpts
     , date2_            = False
     , empty_            = False
     , no_elide_         = False
+    , full_path_        = False
     , real_             = False
     , format_           = def
     , balance_base_url_ = Nothing
@@ -291,6 +293,7 @@ rawOptsToReportOpts d usecoloronstdout rawopts =
           ,date2_            = boolopt "date2" rawopts
           ,empty_            = boolopt "empty" rawopts
           ,no_elide_         = boolopt "no-elide" rawopts
+          ,full_path_        = boolopt "full-path" rawopts
           ,real_             = boolopt "real" rawopts
           ,format_           = format
           ,balance_base_url_ = T.pack <$> maybestringopt "base-url" rawopts
--- a/hledger/Hledger/Cli/Commands/Balance.hs
+++ b/hledger/Hledger/Cli/Commands/Balance.hs
@@ -351,6 +351,7 @@ balancemode = hledgerCommandMode
     ,flagNone ["summary-only"] (setboolopt "summary-only") "display only row summaries (e.g. row total, average) (in multicolumn reports)"
     ,flagNone ["no-total","N"] (setboolopt "no-total") "omit the final total row"
     ,flagNone ["no-elide"] (setboolopt "no-elide") "in tree mode, don't squash boring parent accounts"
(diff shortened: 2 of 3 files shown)


simonmichael:
I'm afraid I don't really understand this or #2429. And I'm a bit worried about flag proliferation. Draft docs / examples / use case will be helpful.

thielema:
I hoped that #2429 would explain well enough what I need the option for. I'll try to add a simplified example. Summarised I need the option for using VLOOKUP in LibreOffice for accessing numbers provided by Hledger in a FODS table.

I agree with your worries about flag proliferation. I do not have very good ideas for that problem other than making the flags more principled. We could also make full-path the default for FODS and CSV export, since these are intended for machine-processing. However, I tried hard to unify export to HTML, CSV, FODS before.

simonmichael:
That helped, thanks. 

This new general report flag only affects tree mode balance reports' FODS output, I see.

simonmichael:
More precisely: currently it's a new balance command flag, affecting only the tree-mode FODS output, but it's stored in the general ReportOpts.

simonmichael:
In its present form it feels a bit inconsistent. One might expect it to affect all reports that can show a tree of account names.

Would this be another way to get the same info: make list mode balance reports include all the normally-elided parent accounts, when --no-elide is used ?

thielema:
As I have written in #2429, I also tried adding parent accounts to List Mode, but adding full paths to Tree Mode turned out much simpler.
I added the `--full-path` option where I found the List/Tree switch. Is there a better place?
The `--full-path` option affects HTML, CSV, FODS output. It seems to leave Text output unaffected, although this was not intended. However, since Text is not suitable for further automatic processing, this could be feature not a bug.

thielema:
You can see here my first attempt to add parent accounts to the List View:
https://github.com/thielema/hledger/tree/balance-include-parent

simonmichael:
I see. From a UX perspective, it sounds nice if we could reuse --no-elide, since it's already there, has the right sort of meaning, and currently has no effect on list mode reports.

simonmichael:
I can have a look at it also if helpful.

thielema:
On Wed, 15 Jul 2026, Simon Michael wrote:

> [10144?s=20&v=4] simonmichael left a comment (simonmichael/hledger#2661)
> 
> I see. From a UX perspective, it sounds nice if we could reuse 
> --no-elide, since it's already there, has the right sort of meaning, and 
> currently has no effect on list mode reports.

That is, --no-elide was left unimplemented for list mode for lack of time 
or so and it would be the preferred option for what I intended with the 
--include-parent option?

I remember that List Mode with --include-parent did not quite do the right 
thing for me, but Tree Mode with --full-path did. I have to look again, 
what the difference actually was.

simonmichael:
Yes I think reusing --no-elide would be preferred, if it would give the required info and it's not too much of a pain to implement. 

It was added for tree mode. At the time I saw no particular effect it should have on list mode. But now we may have found one, which means we could convert this niche mode-specific flag into a slightly more general one.

thielema:
On Wed, 15 Jul 2026, Simon Michael wrote:

> It was added for tree mode. At the time I saw no particular effect it 
> should have on list mode. But now we may have found one, which means we 
> could convert this niche mode-specific flag into a slightly more general 
> one.

Then I will try to implement --no-elide for List Mode.

thielema:
On Wed, 15 Jul 2026, Simon Michael wrote:

> [10144?s=20&v=4] simonmichael left a comment (simonmichael/hledger#2661)

> Yes I think reusing --no-elide would be preferred, if it would give the 
> required info and it's not too much of a pain to implement.

I see in my records that I even tried to get the desired account selection 
using the --no-elide option in the past.

simonmichael:
For the record: this PR is on hold, we are pursuing a different solution for #2429 in #2662.

simonmichael:
I marked this as draft to exclude it from my https://readyprs.hledger.org list, as it's probably not the one we'll merge.

simonmichael:
Closed in favour of #2688.

PR #2662, opened 2026-07-16
balance: enhanced --no-elide 

thielema:
I have mapped `--no-elide` to what I formerly implemented for the `--include-parent` option.
I now see the problem: The included parent accounts do not show the sum of the subaccounts.
This is what worked, when I added the `--full-path` option to the Tree Mode.

### DIFF
--- a/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
+++ b/hledger-lib/Hledger/Reports/MultiBalanceReport.hs
@@ -333,14 +333,20 @@ markAccountBoring ReportSpec{_rsQuery=query,_rsReportOpts=ropts}
 
         qdepth = fromMaybe maxBound . getAccountNameClippedDepth depthspec $ aname acct
         balance = maybeStripPrices . case accountlistmode_ ropts of
-            ALTree | d == qdepth -> bdincludingsubs
-            _                    -> bdexcludingsubs
+            ALTree ->
+                if d == qdepth
+                  then bdincludingsubs
+                  else bdexcludingsubs
+            ALFlat ->
+                if no_elide_ ropts
+                  then bdincludingsubs
+                  else bdexcludingsubs
 
     -- Accounts which don't have enough interesting subaccounts
     isBoringParent :: Account a -> Bool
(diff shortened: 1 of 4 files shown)


thielema:
I do not yet see why the subaccounts are not summed up.
But if they do like intended, I wonder whether it is sensible to enable this by `--no-elide`. The user might expect that `--no-elide` onlys show some more accounts but not different numbers.

thielema:
It seems I must also conditionally enable `bdincludingsubs` in `buildReportRows.balance.ALFlat`.

thielema:
On Thu, 16 Jul 2026, thielema wrote:

> [898989?s=20&v=4] thielema left a comment (simonmichael/hledger#2662)
> 
> It seems I must also conditionally enable bdincludingsubs in buildReportRows.balance.ALFlat.

I have included this change in the pull-request.

thielema:
I can now reproduce the problem. The former version of hledger-ui listed

~~~~
assets:checking  $1480
~~~~

and thus pressing "cursor right" twice descended into "assets:checking", showing three transactions.

With the new effect on "--no-elide" in list mode, we get instead:

~~~~
assets           $1480
assets:checking  $1480
~~~~

and pressing "cursor right" twice descends into "assets" which is empty.


I remember that hledger-ui uses "--empty" per default.

What do you propose as the right behavior, the right way to fix the issue?

thielema:
I have satisfied the tests by flipping the no_elide flag in sync with the `accountlistmode` in `UIState.toggleTree`.
That change should restore the original behavior of hledger-ui, I hope have not overlooked something else.

The Haddock explanation of `toggleTree` reads:
"Toggle between flat and tree mode. If current mode is unspecified/default, assume it's flat."
Is the statement about unspecified modes still valid? I cannot see handling of a default mode.

thielema:
I added the empty account `assets` to the output in `test/balance/219.test` since this is now the expected behavior.

thielema:
You may merge both commits in order to not have an interim commit that violates the test suite. Might be better for a later `git bisect`.

simonmichael:
Closed in favour of #2688.

PR #2671, opened 2026-07-26
web: many improvements for 2.0

### LINKED ISSUE #200: Usability improvement proposals for hledger-web

AidenTHu:
I did not use AI in this PR.

feat: resizable sidebar
feat: sticky header
fix:  #200 
fix: mobile site no longer requires horizontal scrolling
fix: pass 100% lighthouse report for accessibility
dev: drop IE support
dev: Upgrade Bootstrap to latest version
dev: Replace jQuery with modern javascript except for flot chart dependency
doc: bump version to hledger-web 2.0


acinader:
Super nit below. Looks good. Poking around.... 

note aspect ratio of graphs x-axis is stretched when you hide the accounts. Current behavior does not increase the width of the graph.
image

acinader:
I added a video https://github.com/simonmichael/hledger/pull/2651#issuecomment-5093474098 demonstrating the behavior that https://github.com/simonmichael/hledger/pull/2651 is trying to address. This pr does not fix the issue.

acinader:
I lightly reviewed the code and the ui. It looks good and exposes more of hledger's capabilities to a novice hledger user (such as myself), which I think is a good goal.

This is a solid incremental improvement in the functionality of hledger-web. My $.02 is to be a little more ambitious and add a home page with links to at least the journal, balance sheet, and income statement. 

The benefit would be to help a new user see the capabilities more easily, and it would also be generally useful with little added complexity.

simonmichael:
Thanks @AidenTHu for this PR, and @acinader for reviewing.

Some code warnings need to be fixed to pass CI tests. You can see those in the CI log, or locally when you build hledger-web, or HLS will show them.

The PR seems large, so I'm guessing we might want to split it up. But first, I look forward to also reviewing and giving feedback, when I can.

simonmichael:
Also, I notice a commit with description "[Merge branch 'codex/hledger-web-sidebar-state' of ../hledgerOther]" (2c076c2c556ab03607a75e2927346fc4fe0495e0), which suggests some AI use ?

acinader:
> Also, I notice a commit with description "[Merge branch 'codex/hledger-web-sidebar-state' of ../hledgerOther]" ([2c076c2](https://github.com/simonmichael/hledger/commit/2c076c2c556ab03607a75e2927346fc4fe0495e0)), which suggests some AI use ?

that commit traces back to me.

2c076c2c5 is a merge of my commit e2b1c4ae6 from my old PR #2644; codex/hledger-web-sidebar-state was my branch there, and that PR was AI-assisted. It was superseded by the clean-room rewrite in #2651, now merged (7b386e5b6 + bc624be48).

AidenTHu:
> Thanks @AidenTHu for this PR, and @acinader for reviewing.
> 
> Some code warnings need to be fixed to pass CI tests. You can see those in the CI log, or locally when you build hledger-web, or HLS will show them.
> 
> The PR seems large, so I'm guessing we might want to split it up. But first, I look forward to also reviewing and giving feedback, when I can.

Hi, I just got back from vacation. The CI log just says that the code needs to be rebased. I'll do that. Do you have a copy of the old CI log so I can fix the warnings? As for splitting up the PR, that may take more work than I have time for since school starts in September. Can you accept as is based on my testing and @acinader's testing and then log bugs against what needs to change?

acinader:
I did some more testing and found some issues:

1. Loading any `register?...#N` URL directly (bookmark, new tab, middle-click) throws during startup. Register rows have numeric IDs, and `querySelector('#3')` is an invalid selector (hledger.js:111, outside try/catch). Results in a half-initialized page, so things like the date picker, sidebar handlers, and AJAX nav don't work.
2. Adding additional posting rows is broken
3. Account/description autocomplete is gone — typeahead.js was removed, and `accountsCompleter` is never defined, so the guard never passes.
4. The new client-side search validation rejects documented query syntax: `date:2025/1/5, amt:<100`, quoted phrases all pop "Invalid search query", and the check silently disappears after the first AJAX navigation, so behavior is inconsistent.
5. Sidebar clicks never use the new AJAX navigation — the clickable-row handler calls preventDefault() + location.href, so the interceptor bails and the sidebar-position preservation this PR adds doesn't engage for the sidebar itself -- this is likely why I found [this](https://github.com/simonmichael/hledger/pull/2671#issuecomment-5093510582).
6. Scroll-to-hash after AJAX navigation is a no-op — the layout scrolls #main-content, but the code calls window.scrollTo().

I opened https://github.com/AidenTHu/hledger/pull/1 with some tests that exercise the bugs.

I'm concerned that fixing 5 will cause some now-passing tests I added to start failing, but that's just a hunch.

Happy to re-test as things get fixed. Also, I will switch to using your branch for my daily accounting work so I can provide feedback from actual use.

AI usage: testing and test-writing done with Claude Code; findings verified by hand.

hledger-web/static/hledger.js
```diff
@@ -124,12 +124,22 @@ function hledgerInitPage() {
   var addform = document.getElementById('addform');
   if (addform) {
     addform.addEventListener('focus', function(e) {
-      if (e.target.classList.contains('amount-input') && 
+      if (e.target.classList.contains('amount-input') &&
```
acinader: put all white space edits in a single, stand alone commit (even if its just this one).

hledger-web/templates/add-form.hamlet
```diff
@@ -25,14 +25,16 @@
     jQuery('input[name=account]'    ).typeahead({ highlight: true }, { source: globalThis.accountsCompleter.ttAdapter() });
   });
 
-  const utf8textdecoder = new TextDecoder();
+  if (typeof utf8textdecoder === 'undefined') {
```
acinader, line 19: #2702 elimintates this encoding procedure, so we're in conflict  https://github.com/plaintextaccounting/hledger/pull/2697/changes#diff-2ba4bdc6a3c1d2745594618f53b1c0563f7a6f6b964d98afe22846165a23df3eR49-R66

hledger-web/static/hledger.css
```diff
@@ -147,6 +147,10 @@ ul {
     vertical-align:bottom;
 }
 
+#sidebar-menu .main-menu tr.hide {
```
acinader, line 179: I submit that we should seek to eliminate `!important` from our styling and not accept any new use of it without elaborating on alternatives considered.

(ps. this references the line below. sorry)

.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
acinader, line 1: I think you meant to delete these from the repository, but they are still here. If you now have `.vscode` in your `.gitignore` you will need to turn that off, and then make a commit to delete all of this from the repo. Once that's done, you can add vscode back into your ignore.

hledger-web/Hledger/Web/Handler/RegisterR.hs
```diff
@@ -99,10 +99,10 @@ decorateLinks :: [(acct, ([char], [char]))] -> [(Maybe acct, char)]
 decorateLinks = concatMap $ \(acct, (name, comma)) ->
     map (Just acct,) name ++ map (Nothing,) comma
 
--- | Generate javascript/html for a register balance line chart based on
+--- | Generate javascript/html for a register balance line chart based on
```
acinader: was this intentional?

hledger-web/Hledger/Web/Handler/RegisterR.hs
```diff
@@ -99,10 +99,10 @@ decorateLinks :: [(acct, ([char], [char]))] -> [(Maybe acct, char)]
...
 -- the provided "AccountTransactionsReportItem"s.
-registerChartHtml :: Text -> String -> [(CommoditySymbol, [AccountTransactionsReportItem])] -> HtmlUrl AppRoute
-registerChartHtml q title percommoditytxnreports = $(hamletFile "templates/chart.hamlet")
+registerChartHtml :: Text -> String -> [(CommoditySymbol, [AccountTransactionsReportItem])] -> Widget
```
acinader: I see that it is formatting for the chart, can you help me understand why you decided to change it? I think we want to think about the chart. It presents a number of problems: jquery, its old, ugly, and presents a large surface area :). 

My thinking is that we should just touch it as little as possible and instead weigh some modern alternatives. My medium term hope is that once we have cleaned up the status quo, we will be able to add some charts and graphs that would take advantage of hledger's current investment reporting as well as the 2.0 lot reporting potential. It's worth thinking about and getting right.

hledger-web/Hledger/Web/Settings/StaticFiles.hs
```diff
@@ -28,4 +28,6 @@ staticSite =
             -- putStrLn "Using built-in web files" >> hFlush stdout
             return $(embed staticDir))
 
+-- In development mode, staticDevel serves all files from static dir including subdirectories
```
acinader: Can you help me understand this change? If you had to dig here and leave a comment, I'm curious.

hledger-web/Hledger/Web/App.hs
```diff
@@ -159,30 +159,22 @@ instance Yesod App where
...
+      -- Bootstrap 5 CSS
       addStylesheet $ StaticR css_bootstrap_min_css
-      addStylesheet $ StaticR css_bootstrap_datepicker_standalone_min_css
+      addStylesheet $ StaticR css_bootstrap_icons_css
       -- load these things early, in HEAD:
       toWidgetHead [hamlet|
-        <script type="text/javascript" src="@{StaticR js_jquery_min_js}">
-        <script type="text/javascript" src="@{StaticR js_typeahead_bundle_min_js}">
+        <script type="text/javascript" src="@{StaticR js_bootstrap_bundle_min_js}">
+        <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/typeahead.js@0.11.1/dist/typeahead.bundle.min.js">
```
acinader, line 176: we're in conflict here. I propose replacing typeahead with a native `<datalist>`. I give details in the pr body. #2702. But in either case, I do not think we should depend on an external cdn and we should package everything with `hledger-web`. I opened an issue that requests the addition of `Content-Security-Policy-Report-Only` which is an http header that would report on potentially unsafe external javascript. The end goal is to evently deny all external javascript and only run inline script with a [`nonce`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce). See #2703. 

I welcome your feedback.

hledger-web/Hledger/Web/App.hs
```diff
@@ -159,30 +159,22 @@ instance Yesod App where
...
       |]
-      addScript $ StaticR js_bootstrap_min_js
-      addScript $ StaticR js_bootstrap_datepicker_min_js
-      addScript $ StaticR js_jquery_url_js
-      addScript $ StaticR js_jquery_cookie_js
-      addScript $ StaticR js_jquery_hotkeys_js
```
acinader: Looks like we may be in conflict here. Could merge? In any event, I'll look into it.

hledger-web/templates/chart.hamlet
```diff
@@ -1,162 +1,218 @@
...
 <label #register-chart-label style=""><br>
-<div #register-chart style="height:150px; margin-bottom:1em; display:block;">
+<canvas id="registerChartCanvas"></canvas>
 <script type=text/javascript>
-  \$(document).ready(function() {
-    var $chartdiv = $('#register-chart');
-    if ($chartdiv.is(':visible')) {
-      \$('#register-chart-label').text('#{charttitle}');
-      var seriesData = [
+  (() => {
```
acinader, line 4: inline script would need a `nonce` to comply with #2703. Could some of this javascript go in hledger.js?

hledger-web/templates/chart.hamlet
```diff
@@ -1,162 +1,218 @@
-$# If $ is the first character in a line, it must be \-escaped to hide it from hamlet.
 <label #register-chart-label style=""><br>
-<div #register-chart style="height:150px; margin-bottom:1em; display:block;">
+<canvas id="registerChartCanvas"></canvas>
```
acinader, line 2: Can you write a test to cover this change? Using either the fancy new `hledger-web` browser tests (#2697) or the good ole ci run [server tests](https://github.com/plaintextaccounting/hledger/blob/main/hledger-web/Hledger/Web/Test.hs). See: https://hledger.org/TESTS.html. I'd be glad to help review any tests.

or, since this is for the chart, can we just punt?

hledger-web/templates/chart.hamlet
```diff
@@ -1,162 +1,218 @@
...
-        data: [
-          $forall i <- reverse items
-            [
-            #{dayToUtcNoonTimestamp $ triDate i},
-            #{simpleMixedAmountQuantity $ triCommodityBalance c i},
-            '#{showZeroCommodity $ triCommodityAmount c i}',
-            '#{showZeroCommodity $ triCommodityBalance c i}',
-            '#{concat $ intersperse "\\n" $ lines $ T.unpack $ showTransaction $ triOrigTransaction i}',
-            #{tindex $ triOrigTransaction i}
+          const colorIndexVal = #{colorForCommodity c};
```
acinader, line 22: fair warning that I have a css scrub I am working on that is similar to my js scrub #2702 but for css. I'm pretty excited about it. Still in the oven, I think this change will probably overlap.

A change like this would be a good thing to cover with a test.

hledger-web/templates/add-form.hamlet
```diff
@@ -1,38 +1,31 @@
 <script>
 
-  jQuery(document).ready(function() {
+  document.addEventListener('DOMContentLoaded', function() {
```
acinader: also in conflict with #2702. but if we keep it, it would need a `nonce` for inline script to comply with my proposed #2703. I'm working on the `nonce` plumbing, but it is also in the oven with the css scrub. Can we punt on this?

hledger-web/templates/chart.julius
```diff
@@ -0,0 +1,156 @@
+$(document).ready(function() {
```
acinader, line 1: stuff like this should have test coverage that tries to inject xss. see: https://github.com/plaintextaccounting/hledger/pull/2697/changes#diff-2ba4bdc6a3c1d2745594618f53b1c0563f7a6f6b964d98afe22846165a23df3eR36

The two big bennefits is a) you are sure you're not creating a problem. b) if someone in the future touches this code, we can catch a failure to properly escape untrusted text.

hledger-web/templates/default-layout.hamlet
```diff
@@ -1,43 +1,72 @@
...
-      <span .glyphicon.glyphicon-align-left.tgl-icon>
+<script src="/static/js/chart.umd.min.js">
+<script src="/static/js/chartjs-adapter-date-fns.bundle.min.js">
 
-<div#topbar .col-md-8 .col-sm-8 .col-xs-10>
-  <h1>#{takeFileName (journalFilePath j)}
+<div#top-left-controls>
+  <button .btn.btn-default.close-sidebar-arrow-desktop type="button" onclick="sidebarToggle()" title="Close Sidebar&#10Keyboard shortcut: s" aria-label="Close Sidebar">
+    <span .bi.bi-arrow-left aria-hidden="true">
+  <button .btn.btn-default.open-sidebar-arrow type="button" onclick="sidebarToggle()" title="Open Sidebar&#10Keyboard shortcut: s" aria-label="Open Sidebar">
```
acinader: `onclick` handlers can't be "nonced" which wouldn't be allowed if #2703 is accepted.

hledger-web/.eslintrc.js
```diff
@@ -2,7 +2,6 @@ module.exports = {
 	"extends": "eslint:recommended",
 	"root": true,
 	"env": {
-		"browser": true,
-		"jquery": true
```
acinader: #2702 keeps jquery (FOR NOW) pending investigation of chart alternatives.

hledger-web/.version
```diff
@@ -1 +1 @@
-1.99
+2.0
```
acinader, line 1: my assumption is that incrementing the version bump would be outside of a pull request and part of the release process, but I haven't looked into .

hledger-web/Hledger/Web/Handler/RegisterR.hs
```diff
@@ -99,10 +99,10 @@ decorateLinks :: [(acct, ([char], [char]))] -> [(Maybe acct, char)]
...
+registerChartHtml q title percommoditytxnreports = $(whamletFile "templates/chart.hamlet")
```
acinader: I can't follow. Why the change here? what does the w represent? 

Image

              ☝️me

hledger-web/static/hledger.css
```diff
@@ -141,33 +137,657 @@ ul {
...
+    }
+}
+
 .negative {
     color: #a94442;
 }
 
 .date {
-    whitespace: nowrap;
+    white-space: nowrap;
```
acinader, line 802: yup, nice.

hledger-web/static/hledger.css
```diff
@@ -220,8 +899,32 @@ ul {
...
+        transition: left 0.3s ease-out !important;
+        background: white !important;
+        overflow-y: auto !important;
+        box-shadow: 2px 0 5px rgba(0,0,0,0.1) !important;
+        margin-top: 0 !important;
+    }
+    
+    #sidebar-menu.sidebar-offcanvas.active {
+        left: 0 !important;
+    }
```
acinader, line 932: ¡Muy importante!

hledger-web/static/hledger.js
```diff
@@ -3,148 +3,618 @@
 //----------------------------------------------------------------------
 // STARTUP
 
-$(document).ready(function() {
+document.addEventListener('DOMContentLoaded', function() {
+  hledgerInitGlobal();
+  hledgerInitPage();
+  hledgerInitAjaxNavigation();
```
acinader, line 9: The use of of ajax is my biggest concern in this pr. I don't think ajax buys you enough to justify using it, you can get everything you want without it, and it makes the potential attack surface area enormous. 

If performance were the issue you were trying to address, we can probably get better bang for our buck by  doing some performance testing and monitoring to find slow stuff.

My first pass at an hledger-web change (#2644) introduced ajax too, but I believe that that was a mistake and I closed it. I replaced it with (#2651) which can pass the strictest `Content-Security-Policy` when stamped with a `nonce` and I'm convinced it gives 100% of the peformance and usability of an ajax solution without having to write a ton of code to do things that the browser and server are super well suited to do.

hledger-web should be bullet proof.

I didn't appreciate this when I opened #2644. Complicated, bespoke js is hard to write well. The good news is that making `hledger-web` bullet proof is an achieveable, valuable goal and I think we're well on our way.

I'm convinced that with some time and effort you could implement the useful ui changes that this pr introduces in a way that is in accord with #2703.

hledger-web/templates/balance-report.hamlet
```diff
@@ -12,13 +12,16 @@ $forall (acct, adisplay, aindent, abal) <- items
       <div .ff-wrapper>
         \#{indent aindent}
         <a.acct-name href="@?{(registerR, [("q", replaceInacct qparam $ accountQuery acct)])}"
-           title="Show transactions affecting this account and subaccounts">
+           title="Show transactions affecting this account and subaccounts"
+           data-account-name="#{acct}">
```
acinader, line 16: escaping in template 👍

hledger-web/templates/journal.hamlet
```diff
@@ -1,35 +1,32 @@
...
-    Add a transaction
-
-<div .table-responsive>
+<section .table-responsive>
   <table .transactionsreport .table .table-condensed>
     <thead>
       <th .date style="text-align:left;">Date
       <th .description style="text-align:left;">Description
-      <th .account style="text-align:left;">Account
+      <th .account .desktop-only style="text-align:left;">Account
```
acinader, line 6: 👍

hledger-web/templates/journal.hamlet
```diff
@@ -1,35 +1,32 @@
...
-        <td colspan=2>
+        <td .title-description>
           #{textElideRight 60 (tdescription torig)}
+        <td>
         <td .amount style="text-align:right;">
 
       $forall Posting { paccount = acc, pamount = amt } <- tpostings torig
         <tr .posting>
           <td>
+            <span .mobile-account>
```
acinader: i haven't simulated this. I think it would be a good idea to break the mobile fixes out into a single pull request.

hledger-web/templates/manage.hamlet
```diff
@@ -1,7 +1,7 @@
 <h2>
   Your journal's files
 
-<div.row>
+<section.row>
```
acinader, line 4: semantic markup 👍

hledger-web/.eslintrc.js
acinader: FYI, removin jquery breaks typeahead, so typehaead isn't currently working on this branch. Try for yourself.

acinader (changes requested):
While I had poked around a bit before, this code review is my attempt at a full, thorough, code review.

I am reviweing this as your contributing colleague. I am am not associated with the plaintext accounting org or the hledger project other than having contributed a few commits, and recently helping identify a few security issues (see the "Security and Quality" tab for two vulnerablitites that I helped Identify and fix over the last few days that are directly related to the code this pull request impacts).

I also have an open pull request (#2702) that has merge conflicts with this pr. Ideally you and I can come to a common view on how to resolve so the collaborators who will have to approve any change don't have to develop a point of view to resolve our conflicts.

I've posted some thoughts on the issues that this pr raises in https://github.com/plaintextaccounting/hledger/issues/200 and also https://github.com/plaintextaccounting/hledger/issues/2703.

One of the contributors left a comment for you on this pull request that the scope of this pr should be reduced. My suggestion is to focus on the key ui changes you introduce here, which are valuable and I think will help a new user succefully understand how to use hledger and hledger-web. I would be glad to help you by reviwing the changes to get them set for a collaborator to review for merge.

On a hunch, I set up a load test on our two pr (which I think of as my jsscrub and your ajax pr). As it currently stands, the ajax navigation doesn't properly release dom object on page reloads, so the memory usage of the browser continues to grow as you use hledger-web. An end user who uses hledger-web through a typical work day would end up with something like the results below: 

**Navigation stress test: 2,000 link clicks per build**

Same journal (~250 transactions), same Playwright harness driving headless Chrome through the 13 register/journal pages in a cycle; memory sampled every 100 clicks after a forced GC; server process sampled with `ps`.

| metric (2,000 clicks) | full page loads (#2702 JS scrub) | AJAX navigation (#2671) |
|---|---|---|
| navigation latency p50 / p90 / p99 | 100 / 133 / 151 ms | 74 / 97 / 128 ms |
| full page loads counted | 2,001 | 1 |
| JS heap, start → end | 2.0 → 1.5 MB (flat) | 2.7 → **245 MB** |
| retained DOM nodes | ~1,500 (flat) | **1.48 million** (~1.47M detached) |
| live event listeners | 19 (flat) | **24,241** |
| `<script>` tags in `<head>` | 2 (flat) | **3,850** |
| server RSS | ~54 MB, flat | ~51 MB, flat |
| server CPU per navigation | ~12.9 ms | ~8.1 ms |

The AJAX approach is ~25% faster per navigation and does ~40% less server CPU (it skips the static-asset requests; the server still renders the full HTML page either way). But the current implementation leaks linearly with no plateau — roughly 120 KB of heap, 740 detached DOM nodes, 12 listeners, and 2 `<head>` script tags per click. Each AJAX navigation appends the fetched page's scripts to `<head>` without removing the previous ones, and re-running page init registers handlers whose closures keep the previous page's DOM alive. The leak is an implementation issue rather than something inherent to AJAX navigation.

A detailed review of this pull request was really helpful for me to get better acquainted with the codebase. I use hledger and hledger-web to maintain an org's books, complicated partnership interestes, and public security investments. The bullet proof security of the web interface is very important to me, so that's the angle I am looking at changes through. The priority of the collaborators may be different, so my feedback is really just from my perspective.

Finally, I'd be super appreciative if you looked at and gave me your feedback on:

#2703, #2702, and the discusson on #200.

The most useful thing you could do for me, is build the head on #2702 and do some user testing, compare it to your branch and main and give feedback.

The two security issues I mentioned:
https://github.com/plaintextaccounting/hledger/security/advisories/GHSA-vq7r-8w52-jv84
https://github.com/plaintextaccounting/hledger/security/advisories/GHSA-538p-cvc4-4qjm

I hope you find the review helpful. It was a useful process for me.

Best,

Arthur

hledger-web/templates/chart.hamlet
acinader: see #2705

hledger-web/static/hledger.js
AidenTHu: I fixed memory leaks related to re-registering event handlers in the Ajax implementation. Further testing is appreciated to see if I caught them all.

Advantages of Ajax implementation:

1. Navigating to new links in the sidebar doesn't move the position in the sidebar. 
2. 25% faster according to acinader's test
3. It's actually tested and working now, which is the biggest one for me because school starts next week and I'll have a lot less time to devote to this project.

hledger-web/Hledger/Web/Handler/RegisterR.hs
AidenTHu: I tried removing jQuery completely from the code base, but the charts required me to put it back in. I even tried upgrading to the latest version of chart.js, but couldn't get it to work and so backed out. I made the minimum number of changes to get the existing charts working. Fixing the charts properly as you say should be a different issue and PR. I for one would like a checkbox to allow the user to hide the charts if they don't want to see it, but again, that's a different issue and PR.

.vscode/launch.json
AidenTHu: It was accidentally removed from the gitignore for a time. For users who use Visual Studio Code, it is useful to keep. For those who don't, it's easily ignorable and poses no harm.

hledger-web/Hledger/Web/Handler/RegisterR.hs
AidenTHu: No, thanks for catching this.

hledger-web/templates/journal.hamlet
AidenTHu: It passes alpha testing. Separating the PRs into separate ones would require more time than I can devote to this project before school starts next week. If there are bugs though that come up in the future against the mobile version, I'm willing to fix them. We should resolve what to do with the charts first though as that will affect mobile too.

hledger-web/Hledger/Web/Settings/StaticFiles.hs
AidenTHu: Removed those comments from the latest push. I forgot to delete them.

hledger-web/Hledger/Web/App.hs
AidenTHu: I agree with what you suggest, so we should merge this PR first and then you can override your additions over  this one.

hledger-web/Hledger/Web/App.hs
AidenTHu: Yeah, merge this and then override with further cleanup. I did not clean up everything, but made good progress towards it.

hledger-web/templates/chart.hamlet
AidenTHu: This is legacy code. When we replace the chart library with a new one, we should fix the security issues too. It's not in the scope of this (already large) PR.

hledger-web/templates/chart.hamlet
AidenTHu: This is for the chart. Punt to a new chart issue. I question whether it's even useful to have a chart at all. I don't use it and it takes up a lot of screen real estate.

hledger-web/templates/chart.hamlet
AidenTHu: After this PR is merged, then I vote that we merge your clean up work. Then decide on what to do with the charts.

hledger-web/templates/chart.julius
AidenTHu: I don't have expertise or experience with Playwright, so I'll leave test writing to you. Thanks!

hledger-web/.version
AidenTHu: @simonmichael said this would be released with 2.0 version of hledger, so I've been calling it 2.0 version.

hledger-web/Hledger/Web/Handler/RegisterR.hs
AidenTHu: The 'w' in 'whamletFile' stands for Widget. Instead of using the HtmlUrl AppRoute, then turning it to a Yesod Widget via ToWidget(). This sends it as a Widget directly.

In hindsight, this change wasn't really necessary and the issues that lead me to make the change aren't there anymore, so I reverted the change for code clarity.

.vscode/launch.json
acinader: I disagree. 

hledger is pushing twenty years old. It has 178 contributors who have opened almost 1,000 pull requests over 60 releases.

Developer tooling has changed alot in that time, and the various contributors probably use a pretty good mix of environments. If even a small percent checked in their own developer tooling, there'd be alot of it in the repo. There's no documentation or test coverage, so really no guarantee it'll even work for any one other than you.

hledger-web/static/hledger.js
acinader: **Navigation stress test: 2,000 link clicks per build**

Journal (646 transactions, 41 accounts), headless Chrome, 13 register/journal pages in a cycle. Memory sampled every 100 clicks after a forced GC; server process sampled with `ps`. Both builds measured in the same session, on the
same machine, minutes apart.

- before: `upstream/main` @ 6c718b9d8 with the js scrub (#2702)
- after: #2671 @ 578d93865

| metric (2,000 clicks) | full page loads (#2702) | AJAX navigation (#2671) |
|---|---|---|
| navigation latency p50 / p90 / p99 | 67 / 100 / 134 ms | 50 / 82 / 133 ms |
| slowest navigation | 218 ms | 206 ms |
| full page loads counted | 2,001 | 1 |
| JS heap, start -> end | 2.1 -> 1.5 MB | 2.7 -> **264.4 MB** |
| retained DOM nodes | 7,117 -> 2,079 | 6,636 -> **2,819,375** |
| live event listeners | 32 -> 19 | 107 -> **16,883** |
| `<script>` tags in `<head>` | 2 -> 2 | 3 -> **3,850** |
| server RSS | flat | flat |
| server CPU per navigation | ~12.4 ms | ~9.6 ms |

Still retains about 135 KB of heap, 1410 DOM nodes,
8.4 event listeners and 1.93 `<script>` tags, and the growth is linear
with no plateau over the whole run.

**What the latency numbers measure.** The clock starts just before the click and stops once the URL matches the target, `load` has fired, and one animation frame has been painted. For full page loads that includes fetching, parsing and executing every subresource.

Note that it is not the same journal as the run I did this morning, so they are not apples to apples. Also this was a quick and dirty attempt to measure it, so I am not confident that the end measures are exactly the same from an end user perception standpoint, but probably pretty close. 

But none of that matters. The key takeaway is not that ajax is 25% faster or whatever, the key takeway is that the thing is blazingly fast either way. One good side effect of looking into this is that I checked into if we are caching static stuff. And we aren't, so I opened another issue. #2708.

You did reduce the memory leaks, but didn't eliminate them.

The performance test is interesting, but more important is that the web tests don't pass on this branch.

My conclusion. This pr is unlikely to get merged unless scope is reduced as the maintainer requested. 

I don't think you can make an argument to focus the effort on ajax for navigation. The speed gain is imperceptible in exchange for the problems it raises around security, memory use, and lack of tests for that stuff. See my comments above for some highlights.

BUT, there is a gem in your changes, and if you focus on that, I think you've got something worth considering.

I put all of the visible changes from your branch in a series of before after photos below. Some have promise. I think there is something to the nav changes you made on the side panel and the top nav of the journal or register. 

If you concentrate on polishing one or two of the changes, starting from main, and using the existing external libraries, that would be a good first pr.

01_journal
02_register
03_register-chart
04_add-form
05_help
06_manage
07_sidebar-hidden
08_journal-mobile
09_register-mobile
10_addform-mobile

acinader: > 1. Navigating to new links in the sidebar doesn't move the position in the sidebar.

this is already in main as 7b386e5b61ebdcc4cd155d77471248ca9656ed14

simonmichael:
I'm sorry that I haven't had time to keep up with and re-test this very large PR.

So I'm not sure yet - but FYI, it sounds much too invasive and radical to merge straight into hledger-web, especially given @AidenTHu is about to get busy elsewhere. hledger-web aims to be very stable and robust and affordable to maintain. See the [hledger-web goals](https://github.com/plaintextaccounting/hledger/issues/200#issuecomment-5384968371) on #200.

Again, I'm not sure yet, but this always sounded more suitable as an alternate web UI, to be maintained separately and proven in practice, to start with at least.

hledger-web/static/hledger.js
simonmichael: The screenshots are huge time savers. 🙏🏻

simonmichael:
@AidenTHu, sorry but we won't be able to merge this PR. Perhaps smaller parts of it would make sense as new PRs to improve the latest hledger-web in main. Either way, thank you very much for the explorations.

PR #2676, opened 2026-08-01
imp: print: allow omitting trailing decimal marks for declared commodities

### LINKED ISSUE #2664: Can we avoid trailing decimal marks ?
hledger's `print`-like commands will show (and in the case of `add` and `import`, write to the journal) a trailing decimal mark (with no decimal digits after it), when needed to disambiguate an amount containing a single digit group separator.
<https://hledger.org/1.52/hledger.html#trailing-decimal-marks>

This can sometimes be bothersome. Eg, for people working with commodities which never (or practically never) have decimal digits, like Japanese users working with JPY amounts.
<https://x.com/diprex83/status/2061188851935383629>

I'm not sure *how* bothersome the current behaviour is in practice, especially given that most people work with more than one currency/commodity sooner or later. But there are other ways we could handle this, and there might be one that's more convenient and intuitive.

This touches on: journal format, parsing of journal files, reparsing of hledger output, entries written by add and import, digit group mark and decimal conventions in different countries and with different currencies, etc. So it's hard to foresee all impacts on the various data and usage configurations. But I am noting the issue and creating a placeholder for any new discussion/designs.

JoeJoeflyn:
## Summary
- For commodities whose styles are declared in the journal (`commodity` directives), `print` and friends now omit the trailing disambiguating decimal mark, since the declarations re-emitted at the top of the output make the amounts unambiguous when re-parsed.
- Adds `displaySuppressDecimalMark` to `AmountFormat`; `print` and `add` pass the journal's declared commodity set through `showTransactionWithLayout`.
- `print` re-emits the journal's declared commodity style directives as a header so output round-trips.
- `close`, `import`, and `rewrite` keep the previous behaviour (no suppression) since they don't emit the directive header.

Closes #2664.

AI usage: Devin (GLM-5.2 High), ~10k output tokens.

#### Test plan
- [ ] `print` on a journal with a declared commodity using digit group marks (e.g. JPY) shows no trailing decimal mark
- [ ] `print` output round-trips through `hledger -f - print`
- [ ] `add`/`import` write amounts without trailing decimal marks for declared commodities
- [ ] `close`/`import --dry-run`/`rewrite` behaviour unchanged

### DIFF
--- a/hledger-lib/Hledger/Data/Amount.hs
+++ b/hledger-lib/Hledger/Data/Amount.hs
@@ -253,9 +253,14 @@ data AmountFormat = AmountFormat
                                           --   Also, causes 0s to be generated for any commodities which are not present
                                           --   (important for tabular reports).
   , displayDigitGroups      :: Bool       -- ^ Whether to display digit group marks (eg thousands separators)
-  , displayForceDecimalMark :: Bool       -- ^ Whether to add a trailing decimal mark when there are no decimal digits 
-                                          --   and there are digit group marks, to disambiguate
-  , displayOneLine          :: Bool       -- ^ Whether to display on one line.
+   , displayForceDecimalMark :: Bool       -- ^ Whether to add a trailing decimal mark when there are no decimal digits
+                                           --   and there are digit group marks, to disambiguate
+   , displaySuppressDecimalMark :: S.Set CommoditySymbol
+                                           -- ^ Commodities for which the trailing disambiguating decimal mark is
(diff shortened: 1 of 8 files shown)


simonmichael:
@JoeJoeflyn thanks for the PR and the AI disclosure; but we don't accept AI-generated PRs from first-time contributors.
Here's our AI policy (our pull request template instructs your AI to show it to you):
https://hledger.org/AI.html#rules-of-engagement

If you contribute another PR without using AI, then we could take another look at this one.

JoeJoeflyn:
sure, ill do it

Issue #1489, opened 2021-02-22
Underscore as thousands separator


baryluk:
```
commodity 1_000.00 PLN
```

would be nice to have. At least for inputs, but also for outputs.

simonmichael:
@baryluk, we researched this a little and couldn't find any country that uses this convention. Do you know who uses it, or is it just a notation you like ?

adept:
@simonmichael ocaml allows underscores in numeric literals (https://caml.inria.fr/pub/docs/manual-ocaml/lex.html), so it is often used as a convenient separator in source code (`let magic_number = 1_000_000`). Given ocaml's pedigree, maybe it is something that exists in France?

baryluk:
> @baryluk, we researched this a little and couldn't find any country that uses this convention. Do you know who uses it, or is it just a notation you like ?

Just a notation I like and prefer. It is relatively country-neutral, and is cleaner than using other separators.

Xitian9:
GHC has also supported this notation since 8.6.1: https://typeclasses.com/ghc/numeric-underscores

ShrykeWindgrace:
> @simonmichael ocaml allows underscores in numeric literals (https://caml.inria.fr/pub/docs/manual-ocaml/lex.html), so it is often used as a convenient separator in source code (`let magic_number = 1_000_000`). Given ocaml's pedigree, maybe it is something that exists in France?

In France - no, it does not.

simonmichael:
We're discussing this again in chat, but I don't think we should add it without more motivation, since every added feature has a cost. Some examples of it used for accounting/data-tracking-type things in the wild could help.

raingloom:
This is the format Erste's CSV export uses.

the-solipsist:
NBSP is already mentioned in https://github.com/simonmichael/hledger/issues/273#issuecomment-1189003093.  I'd suggest closing this issue in favour of #273.

simonmichael:
Closing for now.

simonmichael:
This is now supported in main, as part of @kfkonrad's #2679.

PR #2679, opened 2026-08-01
feat: journal: accept _ and ' as digit group marks (#273, #1489)

### LINKED ISSUE #273: Can't use a quote as digit grouping character
The documentation says:

>Numbers […] can also have digit group separators at any position (eg thousands separators) which can be comma or period - whichever one you did not use as a decimal point.

However sometimes other characters are used as a separator, e.g. in Switzerland it's common to use the single quote. Would it be possible to support this? Either arbitrary characters or simply expanding the list of separators.

kfkonrad:
This PR adds support for underscores and apostrophies as digit group marks. I've added tests where I deemed them relevant, please let me know if you'd like to see tests added in other places (or any other changes I should make)

Closes #273, #1489.

side note: I haven't used any AI generated code in the commit per your policy for first time contributors. I have however used Claude to review my code and have added that as AI usage to my commit message


simonmichael:
Thank you!

PR #2682, opened 2026-08-03
fix: --tree --transpose flags now show leaf accounts only

### LINKED ISSUE #1941: --transpose --tree gives misleading output
When you transpose a tree-mode balance report, the indented account names are displayed as column headings as indentation is lost. This can be confusing because the amounts shown are hierarchical tree-mode amounts, but there's no longer any visual sign of the account tree, so it looks like flat mode. I wonder if we should disallow --tree with --transpose.

JoeJoeflyn:
fix: --tree --transpose flags now show leaf accounts only.
closes #1941


simonmichael:
I tested this a while back, and it looked like an improvement. Thanks @JoeJoeflyn !

JoeJoeflyn:
Thanks Simon

PR #2683, opened 2026-08-04
holdings: new investment assets and performance report

simonmichael:
Building on the hledger 2.x lots support, here is our long-awaited holdings report - similar to the one in Beancount (and in all brokerage UIs).  It is specified in doc/SPEC-holdings.md.

- One row per holding account, or per lot with `--lots`; columns:
  Date, Age, Units, Unit/Avg cost, Cost, Price, Value, Weight,
  Gain (unrealised), Rgain (realised), XIRR.
- Standard display modes and options: `--tree`/`--depth`/`--drop`/
  `--no-elide`, `-S` value sorting, `-N`, `--round`, pager, `-o`.
- Valuation from P directives or `--infer-market-prices`, in each
  holding's cost commodity by default; `-V`/`-X`/`--value` supported.
- Output formats: txt, csv, tsv, html (with per-column css classes and
  a sample hledger.css), fods, json.
- Docs: command doc, manual lot-reporting integration, spec updates,
  roi-with-lots interop notes and comparison examples.
- Also includes a lots fix (store the styled/widened cost basis) and
  26 new functional tests.

AI usage: Claude Fable 5, ~521k output tokens

lots-entries.journal and irr.journal contain notes on comparing the roi and holdings reports. (They seem to pretty much agree, when you run them just right.)


doc/SPEC-holdings.md
```diff
@@ -0,0 +1,243 @@
...
+See also
+- SPEC-lots.md
+- hledger manual: Lot reporting
+
+Status: phase 1, layout mockup.
+
+## Goal
+
+A standard, convenient report answering "what investments do I hold right now,
+what did they cost, what are they worth, how are they doing ?".
```
acinader, line 15: extra whitespace before '?'

doc/SPEC-holdings.md
```diff
@@ -0,0 +1,243 @@
...
+
+Default (list mode, lot subaccounts hidden):
+
+```
+$ hledger holdings
+Holdings on 2026-03-31
+
+                      ||       Date  Age     Units  Avg cost   Cost  Price  Value           Gain
+======================++========================================================================
+ assets:broker:funds  || 2026-02-15  44d    5 MSFT   $400.00  $2000   $410  $2050    $50  (+2.5%)
```
acinader: why is the gain in ()?

acinader: Ok. I see. The gain is a value, and the percentage that that represents is in ().

My weakly held $.02 is that Gain %  or even just % in the heading row would make for a cleaner report than repeating the () on each row.

doc/SPEC-holdings.md
simonmichael: I usually write space before ?.

doc/SPEC-holdings.md
simonmichael: I have separated the percentage column. Thanks for taking a look.

PR #2684, opened 2026-08-04
web: --port 0 lets the OS choose a free port

### LINKED ISSUE #2559: hledger-web: Randomize port
On a complex dev machine (hi), round-numbered ports are hard to come by. There's always multiple things on X000.

Allow randomizing the port. (Most software allows specifying port 0, and then they'll tell you what port they ended up on.)

acinader:
Closes #2559.

With this change, `hledger-web --port 0` now asks the operating system for a free port. The chosen port is reported in the startup message and used in the default base url, so scripts (and humans) can discover it:

```
$ hledger-web --serve --port 0
Serving web UI and json API at IP address 127.0.0.1 (local access), port 60019
with base url http://127.0.0.1:60019
```

Implementation notes, as discussed on the issue:

- To learn the chosen port we bind the listening socket ourselves and hand it to warp via `runSettingsSocket`. The binding uses streaming-commons' `bindRandomPortTCP` (the same function warp's own `openFreePort` builds on), so `--host` values are interpreted exactly as warp does. streaming-commons was already in the build plan as a transitive dependency; it's now a direct one.
- An explicit `--base-url` is left untouched; only the default host:port base url is rebuilt with the chosen port.
- `--port 0` with `--serve-browse` reports a clear error suggesting `--serve`/`--serve-api`, since browse mode needs a known port up front to open the browser at. (Supporting it would require reworking wai-handler-launch; deliberately out of scope, per the issue discussion.)
- Flag help and the manual are updated.

Tested by hand: `--serve` and `--serve-api` with `--port 0` (banner, base url, and HTTP responses on the reported port), the `--serve-browse` error, `--base-url` passthrough, and normal explicit-port operation.

AI usage: Claude Fable 5, ~50k output tokens (also noted in the commit message).

### DIFF
--- a/hledger-web/Hledger/Web/Main.hs
+++ b/hledger-web/Hledger/Web/Main.hs
@@ -29,6 +29,7 @@ import Control.Exception (bracket)
 import Control.Exception.Backtrace (setBacktraceMechanismState, BacktraceMechanism(..))
 #endif
 import Control.Monad (when, void)
+import Data.Streaming.Network (bindRandomPortTCP)
 import Data.String (fromString)
 import Data.Text qualified as T
 import Network.Socket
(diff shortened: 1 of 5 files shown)


simonmichael:
It looks good - thanks @acinader!

PR #2687, opened 2026-08-11
instance Show Account also shows sub-accounts

thielema:
I was confused for days because code comments talk about account trees and according functions morally have to deal with trees not individual accounts. However, the type is named `Account` and it does not show sub accounts in debug output.

I added showing the sub-accounts to `instance Show Account`. But maybe, for some debug situations this is too noisy?


simonmichael:
Should we close this, or try some more compact output ?

thielema:
On Mon, 21 Sep 2026, Simon Michael wrote:

> [10144?s=20&v=4] simonmichael left a comment (hledgerorg/hledger#2687)
> 
> Should we close this, or try some more compact output ?

We can close it for now.

PR #2697, opened 2026-08-23
dev:web: browser end-to-end tests for the web UI

acinader:
See #200

## Why

`Hledger/Web/Test.hs` (yesod-test) never runs JavaScript, so nothing currently covers the parts of the web UI that only exist in a browser: the add form's posting rows, autocomplete, the date picker, keyboard shortcuts, sidebar show/hide, scroll state, and hash highlighting.

## What is here

- `webui.spec.js` — the current behavior of page startup, the add form, the edit form, search, keyboard shortcuts, and the sidebar. It includes a regression test for #2651 (sidebar scroll position) and one for `register?...#N`.
- `security.spec.js` — journal data (descriptions, account names) renders as text and not markup, including the data handed to the autocomplete's javascript. `fixture.journal` deliberately carries html/javascript payloads for this. Only a browser can show that nothing executes.
- `Hledger/Web/Test.hs` — CSRF tokens and restricting file access to the journal's own files. These need no browser, so they belong here, where they already run in CI. Each negative test has a positive control posting the same request with a valid token; without one they pass whether or not the protection exists.

24 browser tests, ~8s. 15 server tests, ~0.05s.

## Notes / questions

- **The browser tests are not part of the build.** Nothing in `test/e2e` touches `stack build` or `stack test`; node is needed only to run that suite. The `Test.hs` additions are ordinary `stack test` tests.
- **CI is a separate question.** Running the browser suite in CI means installing node and a browser, which costs time on every run. Not wired up in this PR.
- Selectors depend on current markup, so this will conflict with UI PRs in flight. I am happy to rebase around #2671 rather than the other way round.

AI usage: Claude Fable 5, ~70k output tokens

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -478,6 +478,12 @@ SHELLTEST := STACK + ' exec -- shelltest --execdir --threads=40'
 # --test so that subsequent `stack test` won't recompile everything
 # --no-run-tests to avoid running the slow doctest suite every time
 
+# run hledger-web's browser tests against the current build, with any playwright OPTS (needs setup, see hledger-web/test/browser/README.md)
+@browsertest *PWOPTS:
+    {{ STACK }} build hledger-web
+    cd hledger-web/test/browser && \
+        HLEDGER_WEB="{{ STACK }} exec -- hledger-web" pnpm test {{ PWOPTS }}
+
 # too fragile:
 #    echo
 #    just perftest {{ STOPTS }}
--- a/hledger-web/Hledger/Web/Test.hs
+++ b/hledger-web/Hledger/Web/Test.hs
@@ -44,6 +44,11 @@ module Hledger.Web.Test (
 import Data.String (fromString)
 import Data.Function ((&))
 import Data.Text qualified as T
+import Data.Text.IO qualified as TIO
+import Data.Text.Lazy qualified as TL
+import Data.Text.Lazy.Encoding qualified as TLE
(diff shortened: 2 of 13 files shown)


hledger-web/test/e2e/security.spec.js
```diff
@@ -0,0 +1,167 @@
...
+  test('the edit form shows journal text as text', async ({ page }) => {
+    await page.goto('/manage');
+    await page.locator('a.btn', { hasText: 'Edit' }).first().click();
+    expect(await page.locator('textarea').inputValue()).toContain(PAYLOAD);
+    expect(await xssFired(page)).toBe(false);
+  });
+
+});
+
+test.describe('write requests require a csrf token', () => {
```
acinader: move csrf tests to server

hledger-web/test/e2e/security.spec.js
```diff
@@ -0,0 +1,167 @@
...
+      form: { text: '2025-05-01 E2eCsrfEdit\n  a  1\n  b\n' },
+      failOnStatusCode: false,
+    });
+    expect(await res.text()).not.toContain('Saved journal');
+    expect(fs.readFileSync(process.env.E2E_JOURNAL, 'utf8')).toEqual(before);
+  });
+
+});
+
+test.describe('file paths are restricted to the journal files', () => {
```
acinader: move traversal tests to server

PR #2699, opened 2026-08-24
Pass addon arguments through verbatim, as documented

### LINKED ISSUE #2696: Arguments to an external addon aren't passed through as documented
When using an hledger addon such as `hledger-argvdump` (see code below) hledger does not pass through arguments in as documented in https://github.com/simonmichael/hledger/blob/2b28a75dcc78af2fe1e1e1d26d1d08158b9091f3/hledger/Hledger/Cli.hs#L443

Instead flags that any hledger subcommand uses are scanned and `--` detection seems to be broken, too. When such a flag used with hledger (or one of the builtin subcommands) requires an argument hledger prints an error message rather than dispatch to the addon as I'd expect.

## Reproducing the Bug(s)

To reproduce the following examples create a file called `hledger-argvdump`, place it on your PATH and make it executable. This is the content of that file:

```sh
#!/bin/sh
printf 'argv:'; for a in "$@"; do printf ' [%s]' "$a"; done; echo
```

With that set up I ran the following tests:

```sh
sh-5.3$ hledger-argvdump --sort
argv: [--sort]
sh-5.3$ hledger argvdump --sort
hledger: Error: --sort needs a value, none provided (use -h to see usage)
sh-5.3$ hledger argvdump --sort ""
argv: [--sort]
```

```sh
sh-5.3$ hledger-argvdump -- --sort
argv: [--] [--sort]
sh-5.3$ hledger argvdump -- --sort
hledger: Error: --sort needs a value, none provided (use -h to see usage)
sh-5.3$ hledger argvdump -- --sort ""
argv: [--sort]
```

```sh
sh-5.3$ hledger argvdump -n
argv:
sh-5.3$ hledger argvdump -- -n
argv:
```

## Findings

This shows several things:

1. Not all arguments are actually passed to the command, e.g. the empty string after `--sort`
1. hledger parses flags passed to addons, requiring them to conform to hledger's requirements for the flag if a flag of the same name exists for any subcommand
1. hledger checks whether flags requiring an argument have that argument supplied regardless of whether the subcommand being used supports that flag, e.g.:

   ```sh
   sh-5.3$ hledger help --sort 
   hledger: Error: --sort needs a value, none provided (use -h to see usage)
   sh-5.3$ hledger help --sort ""
   hledger: Error: Unknown flag: --sort
   * while parsing the following args, final command line:
   *  help --sort
   ```

   (This is a stronger version of 2. and fixing 3. would likely also fix 2.)

   This could also create an issue if in the future two subcommands use the same flag with different arguments/semantics

1. `--` is filtered as the code comment promises but arguments following it are not passed verbatim. They are still subject to hledger's flag-argument-check and arguments may still be suppressed. Furthermore only the first `--` should get filtered. I.e. instead of

   ```sh
   sh-5.3$ hledger argvdump -- foo -- bar
   argv: [foo] [bar]
   ```

   I would've expected

   ```sh
   sh-5.3$ hledger argvdump -- foo -- bar
   argv: [foo] [--] [bar]
   ```

   so that an addon can make use of `--` as well

## Further Notes

I chose to report these findings as one issue since they're all closely related. Please let me know if you'd prefer them as separate issues

I ran these tests against a current build of hledger's main branch, `hledger 1.99-g2b28a75dc-20260822, linux-x86_64`.

## Workarounds

Two workarounds come to mind:

1. Run `hledger-argvdump` (or any "real" plugin) directly rather than through hledger (`hledger argvdump`)
1. Avoid using any flag used by any hledger subcommand in a plugin

The first workaround is recommended for affected users of hledger plugins.

## Next steps

I'd like to get your opinion on these findings @simonmichael. Would you be okay with fixing all findings in one PR (in separate commits)? Or would you prefer splitting them into separate PR?

kfkonrad:
Fixes #2696: pass addon arguments through verbatim, as documented.

As we discussed in chat and in #2696 I split the changes into four commits in the following order:

1. The required-value flag pre-check ran against the union of every subcommand's flags and scanned past `--`, so `hledger foo --sort` and `hledger help --sort` failed with "--sort needs a value". It is now given only the flags relevant to the identified command, and stops at the first `--`.
2. `filter (/="--")` removed every `--` from the addon's arguments. Only the escaping one is consumed now; later ones are passed on.
3. `-n`/`--no-conf`/`--conf VAL` were stripped from the whole addon argument list. They are now stripped only before the first `--`, so `hledger foo -- -n` delivers `-n`. Without a `--` they keep their current meaning as hledger's own options; the manual says so now.
4. Addons are run with `rawSystem` rather than a shell command line, so empty and shell-significant arguments survive.

Tests: A total of 12 new cases in hledger/test/cli/addons.test and cli.test, using a new `hledger-argvdump` dummy addon (as demoed in #2696). All tests passed on my machine for each commit.

Note that commit 4 only fixes the issue on unix-like systems. For Windows we still shell out to cmd, which presumably swallows empty arguments (`""`). I can try to verify this on my Windows machine soon. I'm mostly using a Mac or Linux these days. So no regression on Windows but also presumably no fix. If I can verify that behavior I'll work around that too in a 5th commit.

I also noticed that `hledger run` may be subject to similar bugs regarding empty or shell-significant arguments (like `"'"`). I may file a separate issue if I can confirm that so that heldger and hledger run behave the same when using addons. As I don't use hledger run myself I'm not sure if changing its argument parsing or dispatch might brake existing use cases.

AI usage: written with some help from Claude Opus, from a plan created with Claude Fable. Some finer adjustments in the plan were my doing and I made sure to test and review each commit thoroughly. Usage in output token is denoted on each commit. btw: I chose to keep with the existing pattern of only documenting use of one model per commit, which is why the plan's AI usage is included in the first commit and the implementation is divvied up among the rest.

### DIFF
--- a/hledger/Hledger/Cli.hs
+++ b/hledger/Hledger/Cli.hs
@@ -280,6 +280,13 @@ main = handleExit $ withGhcDebug' $ do
     mbuiltincmdaction = findBuiltinCommand cmdname
     effectivemode = maybe (mainmode []) fst mbuiltincmdaction
 
+    -- The required-value flags to pre-check on the final command line.
+    -- For an addon command, only hledger's general flags; any other flags are the addon's.
+    -- Otherwise, the flags supported by the command's mode.
+    finalreqvalflagargs
+      | isaddoncmd = generalReqValFlagArgs
+      | otherwise  = modeReqValFlagArgs effectivemode
+
   dbgio "cli args with command first and no cli-specific opts" cliargswithcmdfirstwithoutclispecific
   when isaliascmd $
     dbg1IO "expanded command alias" (cmdarg, (effectivecmdarg, aliasargs))
@@ -354,7 +361,7 @@ main = handleExit $ withGhcDebug' $ do
 
   -- Run cmdargs on command name + supported conf general args + conf subcommand args + cli args to get the final options.
   -- A bad flag or flag argument will cause the program to exit with an error here.
-  let rawopts = cmdargsParse "final command line" (mainmode addons) finalargs
+  let rawopts = cmdargsParseWith finalreqvalflagargs "final command line" (mainmode addons) finalargs
 
   ---------------------------------------------------------------
   seq rawopts $  -- order debug output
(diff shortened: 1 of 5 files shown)


hledger/Hledger/Cli.hs
```diff
@@ -443,22 +452,27 @@ main = handleExit $ withGhcDebug' $ do
     -- 6.5. external addon command found - run it,
     -- passing any cli arguments written after the command name
     -- and any command-specific opts from the config file.
-    -- Any "--" arguments, which sometimes must be used in the command line
-    -- to hide addon-specific opts from hledger's cmdargs parsing,
-    -- (and are also accepted in the config file, though not required there),
-    -- will be removed.
-    -- (hledger does not preserve -- arguments)
+    -- The first "--" argument, which sometimes must be used in the command line
```
acinader: If I'm understanding this right, our choice of marker is: `--`.Could we have use another random string like '@@' instead? Or am I missing the significance of `--` in this context?

I know it wasn't your choice, I'm just trying to follow along.

PR #2730, opened 2026-09-13
AmountFormatExt: central place to explain why we use the somewhat redundant allCommodities parameter

thielema:
The type highlights the connection between `allCommodities` and `AmountFormat`.

`multiBalanceRowAsCsvText`: swap order of parameters since allCommodities plays the role of AmountFormat here.

This is my proposal to address this objection in a non-invasive way:
https://github.com/hledgerorg/hledger/pull/2728#issuecomment-5648263278

Since `allCommodities` is only needed for `LayoutBareWide` we could add an according parameter to the `LayoutBareWide` constructor. But then we would likely need a type parameter for `Layout` for the `LayoutBareWide`-associated data since we do not always want to carry a `[CommoditySymbol]`.

We could split `AmountFormat` into a smaller type for report generating functions and a larger type for `showMixedAmountWith` & Co. The larger type could carry the smaller one in a record field.

Both are pretty invasive changes, though.

### DIFF
--- a/hledger/Hledger/Cli/Commands/Balance.hs
+++ b/hledger/Hledger/Cli/Commands/Balance.hs
@@ -771,8 +771,9 @@ multiBalanceReportAsCsv opts@ReportOpts{..} report =
     rawTableContent $ header ++ body ++ totals
   where
     (header, body, totals) =
-        multiBalanceReportAsSpreadsheetParts machineFmt opts
-            (allCommoditiesFromPeriodicReport $ prRows report) report
+        multiBalanceReportAsSpreadsheetParts
+            (machineFmt, allCommoditiesFromPeriodicReport $ prRows report)
+            opts report
 
 
 multiBalanceReportNumHeaderColumns :: Layout -> Int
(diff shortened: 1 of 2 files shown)


simonmichael:
Sorry, I don't think the tuple adds clarity here. I have pushed some comments instead.

If you think the private module is a win, let's try that as a new PR. I think typical haskell naming would be .Internal ? I'm not quite seeing the benefit of complicating our module tree, given that we can control visibility with exports (and we don't have many library users).

thielema:
On Mon, 21 Sep 2026, Simon Michael wrote:

> [10144?s=20&v=4] simonmichael left a comment (hledgerorg/hledger#2730)
> 
> Sorry, I don't think the tuple adds clarity here. I have pushed some comments instead.
> 
> If you think the private module is a win, let's try that as a new PR. I 
> think typical haskell naming would be .Internal ? I'm not quite seeing 
> the benefit of complicating our module tree, given that we can control 
> visibility with exports (and we don't have many library users).

We have some functions, that shall be used by functions in 
Cli.Commands.Balance and CompoundBalanceCommand but shall not be 
externally visible. How would you solve this without an additional private 
module?

simonmichael:
I see. I could also just export from Balance with special naming and haddock, as a special case. But I see your point.

PR #2734, opened 2026-09-17
hledger/package.yaml.dependencies: remove shakespear

thielema:
it is apparently unused


acinader:
I think I'm about to use it, but if it is currently unused, we should remove it; I can add it back when we use it, and its use is covered.

simonmichael:
I'd guess you are too, and perhaps in hledger-lib as well.

Issue #2006, opened 2023-02-19
Potential contribution: Bank API libraries


livmackintosh:
Dear hledger users and devs,

I've been using hledger for a few years to run my personal and company accounts and I think it is superb. But one shortcoming vs commercial solutions has been that there's no native bank integrations beyond downloading CSV/OFX files and translating them.

Being a bit lazy, I wrote a couple of API libraries for N26 (popular EU internet bank) and Monzo (popular UK internet bank) to retrieve transactions and translate them into the native Transaction type and it works rather well and want to float the idea of contributing them as optional packages.

Does anybody have any thoughts as to if this is a good and welcome idea? I have thought that the set of (hedger ∧ (Monzo ∨ N26)) uses might be quite small meaning that it is not worthwhile for instance.

simonmichael:
> On Feb 19, 2023, at 06:06, Olivia Mackintosh ***@***.***> wrote:
> I've been using hledger for a few years to run my personal and company accounts and I think it is superb. But one shortcoming vs commercial solutions has been that there's no native bank integrations beyond downloading CSV/OFX files and translating them.
> 
> Being a bit lazy, I wrote a couple of API libraries for N26 (popular EU internet bank) and Monzo (popular UK internet bank) to retrieve transactions and translate them into the native Transaction type and it works rather well and want to float the idea of contributing them as optional packages.
> 
> Does anybody have any thoughts as to if this is a good and welcome idea? I have thought that the set of (hedger ∧ (Monzo ∨ N26)) uses might be quite small meaning that it is not worthwhile for instance.
> 

Absolutely! 

https://github.com/simonmichael/hledger/tree/master/examples <https://github.com/simonmichael/hledger/tree/master/examples> ,
especially https://github.com/simonmichael/hledger/tree/master/examples/csv <https://github.com/simonmichael/hledger/tree/master/examples/csv> ,
and https://hledger.org/scripts.html <https://hledger.org/scripts.html>

are where we gather such things currently. As we gather more it could become more organised.

simonmichael:
Hi @livmackintosh .. are you still using hledger ? Any CSV rules or tools or how-tos you'd like to share ?

PR #2735, opened 2026-09-17
translate report texts to user language using haskell-gettext

thielema:
This is a very rough draft.
I translate some report texts using `haskell-gettext`.
`haskell-gettext` is a pure Haskell implementation of `gettext` that does not rely on a global variable for the text dictionary.

I define an enumeration of all translatable texts in `Hledger.Cli.Message`.
I do not use the original English strings as keys,
because sometimes the same English string needs different treatment.
E.g. I have `Average` and `RightAverage`, see below.

You can run it like so:
~~~~
> msgfmt hledger/po/en.po -o hledger/po/en.mo
> msgfmt hledger/po/de.po -o hledger/po/de.mo
> hledger ... --catalog=hledger/po/en.mo
~~~~

Currently the tests will fail because of missing English Catalog.

Problems I found:
* There are small differences in the original strings, like "Commodity" vs. "commodity" and "Total" vs. "Total:". It simplifies translations if we can get rid of such variations.
* Modules like `Hledger.Cli.Commands.Incomestatement` define a `CompoundBalanceCommandSpec` but at this stage we have no `Catalog` and cannot translate. Thus I turned `Text` type into `Message`. However, `CompoundBalanceCommandSpec` is defined in `hledger-lib`, where the `Message` type is not yet available. Thus I added a type parameter `msg`. I think the other two already existing type parameters serve the same purpose.
* The user does not choose a language but instead passes a gettext file using the new option `--catalog`. Currently I load `Catalog` in `compoundBalanceCommandWithCatalog` and `balance`. Maybe there is a more central place.
* This draft omits translation of intervals and report attributes. Their translation is tricky. In English it is "Monthly Income" and "Monthy Income Statement" but in German it is "Monatliches Einkommen", but "Monatliche Einkommenüberschussrechnung" (note the missing "s"), because "Einkommen" has neutral gender and "Rechnung" has female gender.

ToDo:
* convert `*.po` files to `*.mo` files using `msgfmt` in Shake.hs
* Currently I have defined `ReportOpts.catalog_ :: Maybe Catalog`. I prefer to embed the English catalog in the Hledger executable and use this as default, such that we can turn `Maybe Catalog` into `Catalog`.


acinader:
got it. I'm trying to finish up a translation pr too. I'm taking a different, more invasive approach. I'll try to get it polished off today or tomorrow so I can post it and we can compare. 

https://groups.google.com/g/hledger/c/PmcY3J8WiDc

acinader:
i don't think we need to bother with .mo files.  our translation foot print is very small. I can do some performance testing, but i suspect that it's going to be one to two milis even if i 10x what we think we'll see.

acinader:
How can I help? Would you like me to test? Would you like a formal review? Do you want to discuss?

If you want to help me refine #2736 you could build it locally and review the DE translation. Follow [TRANSLATING.md](https://github.com/acinader/hledger/blob/i18n-poc/doc/TRANSLATING.md), edit a PO file in [Poedit](https://poedit.net/), and see the result locally. 

I think it is ready to go, but I'd like to get a set of eyes on it before we waste Simon's time, partcularly because I am taking a different approach than what he had in mind, so I could use some puch back :).

simonmichael:
> I think it is ready to go, but I'd like to get a set of eyes on it before we waste Simon's time, partcularly because I > am taking a different approach than what he had in mind, so I could use some puch back :).

You two never waste my time. ❤️

But it does take a fair bit of brain space to digest all this and give an opinion, and there's lots to do. I hope to try both/all PRs out soon, certainly before the next preview release.

Slightly related: we are in the last two weeks of the 6 month AI credit provided by Anthropic. After this I plan to tap the brakes on AI assistance, at least for a bit, to reflect and digest. 

So this week and next, I feel a little bit of pressure to spend my hacking time on my list of tasks where AI would be particularly helpful. (More ideas welcome, by the way.) I know one of those tasks could be "help review the open PRs". I did a little of that, which informed my first response. But of course I don't want to be an AI maxxer.

Right now, the more time I spend doing AI-assisted dev, the more features (tested, high quality features) pour in. I'm getting things I wished for for many years. Also, they start to feel a bit less consequential, less valuable..

All this is very weird, isn't it.

acinader:
> Slightly related: we are in the last two weeks of the 6-month AI credit provided by Anthropic. After this, I plan to tap the brakes on AI assistance, at least for a bit, to reflect and digest.

Simon, I will happily cover your Anthropic subscription when the promotion is over. I will petition them to extend it for you, but if they won't, I'll increase my sponsorship to cover it. Well worth the money.

heldger2 is an ideal tool for agents to use for accounting. My codex project uses hledger1.99 to obtain, categorize, and book all of my transactions. I use hledger-web for review and reporting during accounting sessions. It's amazing. I haven't fully fleshed out lot-level tracking, asset allocation, and performance reporting, or what the gap is for hledger-web yet, so there's a lot to do.

Along those lines, my follow-on to translations would be to add skills to help a first-time translator use an agent to prepare a translation.

#2006 would be another super high value add that I'd like to get to.

Working without a coding agent would be like trying to code without online documentation(!), google, or stackoverflow.

You could, but why?

thielema:
I added an attempt to perform inflection of interval names without the need to translate every combination individually. The idea is to query the genus of report names and then to query inflected interval names, accordingly. This will work for some languages, however, in order to account for different word order, I think we also need to change, say, "Income Statemnt" to "{interval} Income Statement" and use text interpolation for constructing localized texts.

acinader:
How my [de.po](https://github.com/acinader/hledger/blob/06970943327228dec7a7265147574b3ecdd8b5cc/hledger-lib/locale/de.po#L46-L49) handles it.


#. Report title templates.
#, python-brace-format
msgid "{interval}{report} {dates}{clarification}{valuation}"
msgstr "{interval}{report} {dates}{clarification}{valuation}"

hledger/Hledger/Cli/Commands/Balancesheet.hs
```diff
@@ -21,7 +21,7 @@ import Hledger.Cli.CompoundBalanceCommand
 
 balancesheetSpec = CompoundBalanceCommandSpec {
   cbcdoc      = $(embedFileRelative "Hledger/Cli/Commands/Balancesheet.txt"),
-  cbctitle    = Msg.BalanceSheet,
+  cbctitle    = Msg.Noun Msg.BalanceSheet,
```
acinader, line 24: Two relevant links from the gettext manual

[Context](https://www.gnu.org/software/gettext/manual/html_node/Entries-with-Context.html)

[Context and ambiguities](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html)

po
```
msgctxt "balance-sheet interval"
msgid "Monthly"
msgstr "Monatliche"
```

hs
```
trc t "balance-sheet interval" "Monthly"
```

acinader: Would you be interested in implementing a context-based approach in #2736 using `cbctitle` and `cbcintervalcontext` or something like that?

For example:

```haskell
cbctitle           = i18n "Balance Sheet"
cbcintervalcontext = "balance-sheet interval"
```

thielema:
On Fri, 18 Sep 2026, Arthur Cinader wrote:

> @acinader commented on this pull request.
> 
> ______________________________________________________________________________________________________________________
> 
> In hledger/Hledger/Cli/Commands/Balancesheet.hs:
> 
> > @@ -21,7 +21,7 @@ import Hledger.Cli.CompoundBalanceCommand
>
>  balancesheetSpec = CompoundBalanceCommandSpec {
>    cbcdoc      = $(embedFileRelative "Hledger/Cli/Commands/Balancesheet.txt"),
> -  cbctitle    = Msg.BalanceSheet,
> +  cbctitle    = Msg.Noun Msg.BalanceSheet,
> 
> Two relevant links from the gettext manual
> 
> Context
> 
> Context and ambiguities
> 
> po
> 
> msgctxt "balance-sheet interval"
> msgid "Monthly"
> msgstr "Monatliche"
> 
> hs
> 
> trc t "balance-sheet interval" "Monthly"


If we have to translate every interval for every report type, we could as 
well translate all combinations of interval and report type.

However, my idea was to reduce the translation work by determining the 
genus of the report name and inflect the interval name accordingly. In 
German it happens that all report names have feminine genus. So we only 
need inflections for one of three genders.

That said, I might turn genders into contexts. This might be more in the 
spirit of gettext.

simonmichael:
Just a drive by comment, but special heuristics that work for one or two languages will really limit us as we add more languages.

simonmichael:
Where I'm coming from: I would rather ship a smaller slower-growing number of translations with perfect grammar & spelling, than a large fast-growing number of sloppy translations. I guess that's my inner perfectionist. Also shipping imperfect unfinished translations, or translations with different capabilities, sounds like it would create more maintenance work.

Of course, fast and cheap is probably the better strategy for growth. But I'd like it least to be *possible* to achieve perfect translations when we want to. I feel pretty sure that in general, when you consider how different languages can be, and given our need to present static and dynamic text seamlessly combined, this is going to require custom logic in many places, ie programming. 

It's possible, but I don't know if worthwhile/manageable in practice, we could combine multiple techniques.

thielema:
On Sat, 19 Sep 2026, Simon Michael wrote:

> [10144?s=20&v=4] simonmichael left a comment (hledgerorg/hledger#2735)
> Message ID: ***@***.***>

> Just a drive by comment, but special heuristics that work for one or two 
> languages will really limit us as we add more languages.

My approach is no limitation, only an optimization for German and, I 
expect, many other languages. It shall prevent the translator from 
translating all combinations of "interval" and "report name" individually. 
If that does not work for your language, you can choose a fake "gender" 
that is specific to each report and then do an individual translation of 
the intervals for every single report.

thielema:
On Sat, 19 Sep 2026, thielema wrote:

> On Sat, 19 Sep 2026, Simon Michael wrote:
>
>> [10144?s=20&v=4] simonmichael left a comment (hledgerorg/hledger#2735)
>> Message ID: ***@***.***>
>
>> Just a drive by comment, but special heuristics that work for one or two 
>> languages will really limit us as we add more languages.
>
> My approach is no limitation, only an optimization for German and, I expect, 
> many other languages. It shall prevent the translator from translating all 
> combinations of "interval" and "report name" individually. If that does not 
> work for your language, you can choose a fake "gender" that is specific to 
> each report and then do an individual translation of the intervals for every 
> single report.

In order to make this work in every case however, we need to define a 
translation of the universal interval, that is, a text for the case that 
there is no interval specification.

acinader:
https://www.gnu.org/software/gettext/manual/html_node/Entire-sentences.html

Here's my concrete suggestion for a plan we should adopt:

For our first translation, let’s defer additional grammar-handling logic and add the Haskell and PO entries needed to make the German translation complete and correct.

The guiding principle should be to prefer more translation entries in the PO file over more grammar-handling logic in Haskell.

I think if we focus our efforts, we could get a correct, shippable German translation very quickly. 

We have i18n scaffolding in #2736, including a tag extraction script, documentation, and test coverage. It works right now and is ready for a translator to iterate on the PO file while reviewing the UI.

The missing piece is the first l10n, which you already have a good start on, but it will require more Haskell work to accommodate a gendered language. Can you do this work?

I'd appreciate it if you could review my PR and share your thoughts from a translator's perspective. I think you'll like it. You could start with the PR README and translation doc. Ask me any questions. In the meantime, I will pull your web-balance-report branch and run it so I can give you feedback.

acinader:
> Where I'm coming from: I would rather ship a smaller slower-growing number of translations with perfect grammar & spelling, than a large fast-growing number of sloppy translations.

Agree. 

Adding a translation to the distribution adds a real comittment that has the potential to slow down any development that touches the ui. In order to add a locale, we'd have to have some level of confidence that there is a contributor that is repared to own it in the face of changes that require additional translations.

acinader:
After thinking about it some more, it is probably best to release only i18n, the extraction script, and documentation. We should not contemplate releasing a localized PO file in the near term. 

A limited release focused on the internationalization (i18n) machinery would let any user copy the template .POT to make their own PO to localize their own hledger. This would be an invitation for potential contributors to open issues where there are translation problems that they can't fix with the current template which would require a code change and a new extraction of the template.

If, after some time, a mature locale has worked through the necessary Haskell changes to support localization (l10n), then it could be part of the distribution. 

How does that sound to you?

Issue #2737, opened 2026-09-20
lib: settle on one HTML renderer (blaze-html), instead of lucid and blaze in parallel


acinader:
## Summary

hledger-lib has two hand-maintained renderers for `Hledger.Write.Spreadsheet` cell grid:

- `Hledger.Write.Html.Lucid` — used by CLI `-O html` report (`balance`, `register`, `aregister`, `print`, `holdings`, `is`/`bs`/`cf`).
- `Hledger.Write.Html.Blaze` — added "for compatibility with hledger-web" - unused

They share `Html.HtmlCommon` (border logic) and `Html.Attribute` (the builtin stylesheet). Changes have to be made twice: Lucid now has amount spans, a `date` class, titles, and readable newlines that Blaze lacks, and the colspan/rowspan fix of 2026-09-11 (b2e2331cf) had to be applied to both.

@simonmichael and @thielema  (#2321, Jan 2025) agree to use blaze.

## Why now

#2242 `thielema:web-balance-report` uses `Write.Html.Blaze`: `/balance` renders the CLI's `balanceReportAsSpreadsheet`

## Proposal

1. Bring `Write.Html.Blaze` up to `Write.Html.Lucid` feature for feature: amount spans (`cellParts`), the `date` class, `titledTableHtml`, the newlines between elements (#2326, #1681).
2. Switch the five CLI modules from `Lucid.renderText` to blaze's `Text.Blaze.Html.Renderer.Text.renderHtml` without any visual change. Confirm with existing e2e tests.
3. Delete `Write.Html.Lucid`, fold `HtmlCommon` into the survivor, and swap the `hledger` package's `lucid` dependency for `blaze-html`/`blaze-markup` (already transitive dependencies of hledger-web via yesod).

Dependency-wise this is a swap, not a removal: `lucid` out of `hledger`, `blaze-html` + `blaze-markup` in. Both are old, small and stable.

## Why blaze rather than lucid

- Yesod's `Html` type *is* blaze markup, so hledger-web can never drop blaze; the only library that can be dropped is lucid.
- Lucid cannot render multi-line or indented output (https://github.com/chrisdone/lucid/discussions/161); the newlines added to hledger's output in 2026-09 (#2326) work around this in hledger rather than in the library.
- Lucid's nicer API was the original reason to prefer it (#2321); Henning: "I do not see the actual advantages of the Lucid API and I am completely ok with Blaze."

## History

PRs, oldest first:

[...]

Issue #2703, opened 2026-08-26
web: send a Content-Security-Policy - The Robots are Coming!


acinader:
<img width="591" height="346" alt="Image" src="https://github.com/user-attachments/assets/09cf976d-4fa6-4fbe-8fdf-73f3484485a2" />

https://x.com/sama/status/2093060670472241368?s=20

Agentic attackers are coming and we should harden our defenses.

hledger-web sends no Content-Security-Policy, so the browser has no second line of defense if journal text ever reaches an HTML sink. Two such bugs turned up recently (GHSA-538p-cvc4-4qjm, and the add-form error message in #2700); a strict policy would have blunted both.

The blocker is that a strict `script-src` blocks inline `<script>`, and hledger-web has three inline blocks left:

1. `templates/default-layout-wrapper.hamlet` — the `no-js`/`js` class swap and `document.hledgerWebBaseurl`.
2. `templates/default-layout-wrapper.hamlet` — the analytics loader, when `extraAnalytics` is configured. It creates a script element at runtime, so it needs the nonce propagated or `'strict-dynamic'`.
3. `templates/default-layout.hamlet` — the sidebar scroll restore. This one has to stay inline: it runs during page parse, which is what #2651 fixed. Moving it to a file brings back the visible jump.

Because of (3), moving everything to external files is not available; the scripts need a per-request nonce.

Sketch:

- `yesodMiddleware` generates a random nonce per request, puts it where templates can read it, and sends `script-src 'self' 'nonce-...'`.
- Each of the three inline blocks carries `nonce=#{...}`.
- `templates/upload-form.hamlet` has an inline `onchange=` handler. Nonces do not apply to inline event handlers, so this should move into `static/hledger.js` as an `addEventListener` rather than be exempted with `'unsafe-hashes'`.
- Report-only first (`Content-Security-Policy-Report-Only`) would show what a policy breaks before enforcing it.

Worth deciding at the same time: whether `style-src` can be strict too, which depends on the remaining `style=` attributes in the templates.

X-Frame-Options and X-Content-Type-Options are already sent as of #2702.

PR #2739, opened 2026-09-21
web: balance report page (#2242)

acinader:
Builds on @thielema's `web-balance-report` branch. 

Fourteen commits on top address the review comments on #2242 and what testing turned up.

`/balance` serves the balance report. Account names and
period headings link into the register (for now).

## Two commits change the CLI's HTML output

- **borders as classes.** hledger-web sends a Content-Security-Policy that
  forbids inline styles. Remove per-cell style attributes. Cells get class
  `border-{top,bottom,left,right}-{single,double}` and the builtin stylesheet
  gets eight new rules. Both writers changed together, so lucid and blaze do
  not drift apart here. Column headers get classes: `account` for accounts, 
  `amount` over figures.  (#2703)
- **a `negative` class** on cells whose amount is negative.

`-O html` output has two changes: the border color now follows
the text color instead of being hardcoded black, and a single-period report's
account column picks up the builtin `th.account, td.account` rule it was
missing, so that column sits flush left as it already does in a multi-period
report.

## Other fixes

- `--depth` at startup and `depth:` in the search box were ignored
- Only the interval of the `period` expression was used, so
  `?period=monthly in 2024` gave monthly columns over the whole journal, and
  `?period=2024` gave a single column of all-time totals. The date span now
  restricts the report as a `date:` term would.
- A `date:` term can set the interval too, as on the command line
  (`date:monthly`), and wins over the `period` parameter.
- A period expression that does not parse is now presented in an error alert.
- The table is styled by `hledger.css` rather than by an inline `<style>`
  block, whose unscoped rules would also have restyled the sidebar. It is a
  bootstrap table now, so it matches the journal and register pages.
- The `period` parameter is kept by the search form on this page only (so far), and by
  its clear button, which used to drop you back to the single-period report.
- Fix tooltips so they match the report for the link.
- Remove the unused `lucid` dependency.
- Apply the journal's commodity display styles.
- `-E` matches account sidebar behavior, which is the opposite of the CLI.
- The startup reporting interval (`-M`, `-Q`, `-p 'monthly in 2025'` …) produced a single column of row totals instead of the periods it asks for. It now opens the multi-period report, and a period parameter still overrides it.
- `--title` was ignored — the heading was always "Balance Report". It is now the command line report's heading ("Balance changes in 2024-01-01..2026-12-31"), or `--title`'s value verbatim when one is given.
- A wide report used to drag the whole main pane sideways, heading and search
  box with it: bootstrap's `.table-responsive` only scrolls below 768px, so the
  wrapper carries a rule of our own, and the table scrolls inside the page.
- Eighteen yesod-test specs for the page, and a manual section.

## The page is standalone for now

One commit removes the `/balance` links from the journal page, so there are no links
to it in hledger-web. I think we should add all of the missing reports first and evaluate
them stand-alone before thinking about how to wire it all together.

Addresses @simonmichael's concern on #2242 about committing hledger-web to new
UI: nothing in the existing UI changes.

## Open question: what should a column heading link to?

A period heading (`2024`) links to `register?q=date:2024`, from
`Hledger.Cli.Anchor`. It should instead go to `/balance?q=date:2024`.

The amount cells should probably become links too 
(`register?q=inacct:<account> date:<period>`), since they are unlinked today
and the heading is currently the only route from this page to the transactions
behind a period. That would give: row → the account's transactions, column →
that period's balances, cell → that account's transactions in that period.

It belongs in `Hledger.Cli.Anchor` but is not included here.

## How to test

Manual test plan: https://gist.github.com/acinader/e58e0445e6d517e153f93a5b685ae6e3. 

It tests:

- `/balance?period=yearly` — the table looks like the journal and register pages
- `/balance?q=depth:1` — as many rows as the sidebar
- `/balance?period=2024` — one column, holding 2024 only
- `/balance?period=monthly%20in%202024` — twelve columns
- `/balance?q=date:monthly` — the interval from the search box
- `/balance?period=bogus` — an alert, not a table
- search, then clear: the period survives both
- `hledger bal -Y -O html | grep -c 'style="'` — zero

## Commits

1. `imp: html:` mark table borders with CSS classes instead of inline styles
2. `dev: cli:` export the balance report spreadsheet parts and title
3. `dev: web:` style the table like the other pages
4. `dev: web:` drop the unused lucid dependency, regenerate the cabal file
5. `dev: web:` honor the depth limit
6. `dev: web:` apply the period expression, report a bad one
7. `imp: lib:` spreadsheet cells for negative amounts get a "negative" class
8. `dev: web:` headings, report links, and the period parameter
9. `test: web:` cover the balance page
10. `doc: web:` describe the balance page and its period parameter
11. `dev: web:` don't link to it from the journal page yet
12. `dev: web:` take the interval from a date: search term
13. `fix: web:` apply the journal's commodity display styles
14. `fix: web:` read -E the way the rest of hledger-web does

AI usage: Claude Opus 5, ~140k output tokens across the branch (the
per-commit estimates in the commit messages sum to ~77k; the rest went to
review, testing, and drafting that produced no commit).


simonmichael:
Should this be marked draft ? This [todo list](https://github.com/hledgerorg/hledger/pull/2242#pullrequestreview-5258710141) seems still in progress

acinader:
> Should this be marked draft ? This [todo list](https://github.com/hledgerorg/hledger/pull/2242#pullrequestreview-5258710141) seems still in progress

This is ready to go. This pr is the result of working through the to-do list.

I do have a question about column, row and cell linking behavior, but I think the current implementation is acceptable and good for an upcoming release.

simonmichael:
It all sounds good to me, and looks good from my testing. And good enough to expose in the UI when you feel ready. I'll merge this soon if there are no objections.

My main feedback: I wonder how to clearly distinguish end balance reports and balance change reports for users, in line with the "easy obvious UX" goal, and hledger's bs/is reports. Do hledger-web users need to see changes in ALE accounts, or end balances in RX accounts ? Should we just provide bs and is reports, or include the more general/advanced behaviour of the `balance` command ?

simonmichael:
As a quick aid, it would be nice to be able to toggle historical mode (-H) on this.

acinader:
today I learned that balance report + -H + type:AL = blance sheet.

I'd like to get this in and then close #2737.

Lots to digest.

I'd like @thielema's feedback too on next steps.

simonmichael:
I'm guessing @thielema won't object to this PR, which continues his earlier work. Merging, thank you both!

simonmichael:
> today I learned that balance report + -H + type:AL = blance sheet.

That's right. And `bal type:RX` = `is`.

`bs` and `is` exist because `balance` is a complicated command with many modes. (And they're standard reports people, at least US people, have heard of.)

